PR #3, opened 2025-01-25
HELP/WIP Create QUICKSTART.md

NyanHelsing:
hey any thoughts? im' getting errors when i build still; 

eg

-  ` src/helpers/BaseChatMesh.cpp:2:10: fatal error: base64.hpp: No such file or directory`
    -  i did `platformio pkg install -l densaugeo/base64` to resolve.
- `examples/simple_repeater/main.cpp:7:12: fatal error: SPIFFS.h: No such file or directory`
    - ???

Also is there an `-e tdeck` i can run?

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -1,0 +1,47 @@
+# Quick Starts for common platforms
+
+## Linux CLI-Only QuickStart
+
+### Starting from the home dir, Clone this repository and `cd` into it
+```
+cd ~
+git clone git@github.com:ripplebiz/MeshCore.git
+cd MeshCore
+```
+
+### Install deps:
+```
+platformio pkg install
+```
+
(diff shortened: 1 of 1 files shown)


ripplebiz:
Thanks for this. Looks ok.
There is no TDeck environment. The TDeck firmware is commercial.

ripplebiz:
The base64 should be handled by the lib_deps addition that you see.
Not sure why that one (densaugeo/base64 @ ~1.4.0) would not resolve(?) 
Seems to be in the PIO registry.

ripplebiz:
Ah, found the problem you mentioned about base64.hpp, etc.
My bad. I've tested ALL the build target/envs now and they all build successfully now.

NyanHelsing:
the tdeck firmware is closed source? seriously??

PR #10, opened 2025-02-03
Use display for msgs  on heltec_v3

fdlamotte:
Some code to show received msgs on the display, found this usefull when testing companion app ...

### DIFF
--- a/examples/simple_secure_chat/main.cpp
+++ b/examples/simple_secure_chat/main.cpp
@@ -53,6 +53,21 @@
   #include <helpers/HeltecV3Board.h>
   #include <helpers/CustomSX1262Wrapper.h>
   static HeltecV3Board board;
+
+  #include <Wire.h>
+  #include <Adafruit_GFX.h>
+  #include <Adafruit_SSD1306.h>
+
+  #define SDA_OLED 17
+  #define SCL_OLED 18
+  #define Vext 36
+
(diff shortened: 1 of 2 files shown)


ripplebiz:
Hi,
Thanks for the samples. Yeah, having some feedback on the OLED would def be nice.
I've had firmwares before that started out with one OLED driver, then another, and it really balloons out quickly.
Keeping these /example firmwares reasonably tidy is already becoming an issue. I have to refactor some of them already, and try to separate concerns a bit more. Otherwise, this all just gets SO hard to maintain.

Having some _abstracted_ UI concepts would be ideal, and to have different display driver helpers. I will hopefully tackle this soon.

fdlamotte:
Totally agree with you on this. And I'm happy to see you want to keep things in control.
I've proposed this PR after having to merge latest changes to the simple_secure_chat, thought Heltecs are very common.

Also, maybe displays are out of scope of the meshcore lib and should be kept apart ... I find it good to have a lib that only cares about the mesh itself

I've added this sample on my github account : https://github.com/fdlamotte/MCChatScreen so it can be built by others. I'll try to keep it in sync it with the base example ...

To use my sample, I just put my projects at the same level as MeshCore, platformio.ini references your repo and there is a symlink to the ed25519 lib (since it is only 240k I should have just made a copy)

I let you close this PR if you agree

Regards,

Florent

PR #23, opened 2025-02-14
T1000e code

fdlamotte:
These files have not been taken from other projects (and I removed references from them to symbols defined in variant.h) ;)
Liam is investigating for the bsp part ...

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -76,6 +76,10 @@
   #include <helpers/nrf52/RAK4631Board.h>
   #include <helpers/CustomSX1262Wrapper.h>
   static RAK4631Board board;
+#elif defined(T1000_E)
+  #include <helpers/nrf52/T1000eBoard.h>
+  #include <helpers/CustomLR1110Wrapper.h>
+  static T1000eBoard board;
 #else
   #error "need to provide a 'board' object"
 #endif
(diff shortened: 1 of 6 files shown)


ripplebiz:
Hi,
I don't see the platformio.ini changes, or board.json, variant.ccp (if needed?)

fdlamotte:
they are not ...
the idea was to put in the main repo the files that have no issues ...
I'll rewrite bsp and then send the platform.io at the same time

also there is a need for the example files to support t1000e

fdlamotte:
Sorry for the quick response ... I was driving to the alps were we'll spend a week in family (12h is a little too much ;))

I only sent the "ok" file (and the most important I think) because I think I won't be able to tackle with the BSP this week, but liam told he would eventually investigate. 

If the helper files for the T1000e and LR1110 are in the main repo, he'll be able to deal with the bsp "just by" cloning your repo, adding his variant/board files and modify the platform.ini

Also I have prepared the MeshCoreT1000e repo to remove the helper files and only provide the platform.ini in the way I build things now (I tested with my fork, and it's ok). The MeshCoreT1000e repo will then be my base for a custom firmware dealing with the peripherals of the T1000e and including the companion radio

Whatever you do with this PR, things are going great ! (And I'm ok with current situation concerning the two repos, I find it works good)

regards,

Florent

PR #32, opened 2025-02-20
add new commands to companion firmware to import and export identity

liamcottle:
This PR adds two new commands to import and export the main identity private key in the Companion Radio firmware.

`CMD_EXPORT_PRIVATE_KEY`: `23`
`CMD_IMPORT_PRIVATE_KEY`: `24`
`RESP_CODE_PRIVATE_KEY` : `14`

- `CMD_EXPORT_PRIVATE_KEY` replies with `RESP_CODE_PRIVATE_KEY` containing the 64 byte main identity key.
- `CMD_IMPORT_PRIVATE_KEY` accepts a 64 byte private key which is loaded and saved as the main identity key.

Could possibly be renamed to `CMD_EXPORT_IDENTITY_PRIVATE_KEY`, but naming is internal only...

I opted to leave out the public key from the export, as the firmware can derive it from the private key on import.

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -121,6 +121,8 @@ static uint32_t _atoi(const char* sp) {
 #define CMD_GET_BATTERY_VOLTAGE   20
 #define CMD_SET_TUNING_PARAMS     21
 #define CMD_DEVICE_QEURY          22
+#define CMD_EXPORT_PRIVATE_KEY    23
+#define CMD_IMPORT_PRIVATE_KEY    24
 
 #define RESP_CODE_OK                0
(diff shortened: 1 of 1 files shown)


ripplebiz:
Am a bit worried about this one. Could open an exploit where private keys could just silently get lifted off devices without the user knowing.

liamcottle:
Yep, fair enough!

I've added the below build flags to enable/disable this feature as discussed elsewhere:

- `-D ENABLE_PRIVATE_KEY_EXPORT 1`
- `-D ENABLE_PRIVATE_KEY_IMPORT 1`

By default, these are both set to `0`, in which case the radio will reply with `RESP_CODE_DISABLED`.

PR #46, opened 2025-02-24
Base class for companion radios

fdlamotte:
Did some refactoring to get a base class for companion radios ... applied it on the example

I think it'll help in providing custom firmware for our devices.

Tested on T1000e and Heltec v3


fdlamotte:
I have only tested with ble ...

edit : serial works as well, on heltec v3

PR #50, opened 2025-02-26
Fixes BLE advertising after pairing failure

liamcottle:
When pairing with an ESP32 device, if you cancel the pairing process, or enter the wrong pin, most of the time the device never starts advertising itself again. This means the user has to restart the board to be able to pair again.

It seems like it's just a timing issue, as adding in this short sleep before checking connection count, I have not been able to get it stuck in the non-advertising state anymore.

The nRF52 boards, more specifically the RAK4631, doesn't appear to have this issue. So I haven't added the delay for that board.

Hopefully this is the last of the BLE patches, for now ;)


liamcottle:
Fixed by 54af8ad2a97ca4238dece18781eb5049c1174bb9

PR #52, opened 2025-02-26
Add nRF52 S140 SoftDevice Header Files

liamcottle:
This PR adds the nRF52 S140 SoftDevice header files, so we can compile the T1000-E firmware from a fresh clone.

The header files have been sourced directly from the Nordic Semiconductor website.
https://www.nordicsemi.com/Products/Development-software/S140/Download

It's probably the easiest approach to keep these files in the repo, otherwise we'd need an extra script to pull them in from another source at build time.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -482,6 +482,8 @@ extends = nrf52_base
 platform_packages = framework-arduinoadafruitnrf52
 build_flags = ${nrf52_base.build_flags} 
   -I src/helpers/nrf52
+  -I variants/t1000-e/s140_nrf52_7.3.0_API/include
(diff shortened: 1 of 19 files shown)


fdlamotte:
Thanks for this liam,
could reproduce the issue after purging pio cache ... it must have gathered this one when I compiled MT for the T1000
I applied your PR and works like a charm

I would have applied the build flags only to the nrf52840_t1000e target though

good job

ripplebiz:
So, is this a case of some missing/old/poor PlatformIO support? Just curious. 
eg. is the library manager just not automatically pulling these in because they haven't been set up properly by Seeed, or Nordic?

PR #61, opened 2025-03-03
new Heltec_v3_companion_radio_wifi target

fdlamotte:
config must be hardcoded in platformio.ini and you'll need to figure out your IP but it seems to work ;)


ripplebiz:
Thanks for submitting this. 
I'll merge it in, but would like it to be considered experimental (for now). People can do builds, and their own testing, etc, but prob won't feature in the Github releases section until it's had a fair old shake.

PR #62, opened 2025-03-04
Add LILYGO T-LoRa V2.1-1.6 firmware

awolden:
* Adds support for LILYGO T-LoRa V2.1-1.6
* Brings in a lighterweight crypto library for ed25519 because the existing implementation was causing stack crashes on older esp32 boards.
  * This CPU has limited stack size and crashed out when signing packets. This brings in `libsodium` which doesn't crash on signing. I put it behind a build flag even to be safe.
  * [Discord context
](https://discord.com/channels/1343693475589263471/1344882100007862283/1346578498944241674)
* Fully tested and working


ripplebiz:
Thanks for your contribution.
I'll try to review this today. 
In the meantime, can you change the base branch of the PR to the 'dev' branch?

awolden:
@ripplebiz done. I cleaned it up quite a bit as well.

ripplebiz:
There is a problem with having the different private key formats, as there will be some firmwares with the ENABLE_PRIVATE_KEY_IMPORT/EXPORT enabled, and this would create an incompatibility.
It's a _very_ low-level change you're proposing, with the alt crypto lib, and it would require a LOT of testing to make sure it is completely interoperable with other firmwares, other boards, etc.

ripplebiz:
Also, if stack size is the main issue, a simpler option is you can set custom 'loop task' stack size for the ESP32 targets with:   -D ARDUINO_LOOP_STACK_SIZE= ...

awolden:
@ripplebiz That's fair, and the arduino flag was the first thing I tried at 8kb and it was still failing, which I assumed was double the stacksize, but it's possible the default was already 8kb 🤦 .I just tried at 32kb and 16kb and it worked fine. I pushed the simplification.

Also I see https://github.com/ripplebiz/MeshCore/issues/63 linked to this PR, but I do not believe it would be compatible with this since it's an ESP32-S3 board and this is ESP32. This board config won't flash on to that variation.

There are multiple variations from Lilly go on this:

ESP32 + 1276 (this)
ESP32-S3 + 1276
ESP32-S3 + 1262

PR #74, opened 2025-03-07
add: BLE_NAME_PREFIX flag

ddmunhoz:
This adds BLE_NAME_PREFIX flag to allow customization of the bluetooth device name prefix.

**BLE_NAME_PREFIX behaviour:**

**if unset:** defaults to "MeshCore-"
**If set:** uses the value defined in platformio.ini - iex.: -D BLE_NAME_PREFIX=\"Test\"

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -1272,7 +1272,13 @@ void setup() {
 
 #ifdef BLE_PIN_CODE
   char dev_name[32+10];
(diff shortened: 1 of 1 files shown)


ripplebiz:
Thanks for the addition. Just wondering why the .gitignore change?

ddmunhoz:
My mistake, it was there on my local branch to ignore my MakeFile used to simplify the build/test process. Dropped it, should be good to go!

PR #78, opened 2025-03-07
Enable SSD1306 display class for LilyGo_TLora_V2_1_1_6

awolden:
The LilyGo_TLora_V2_1_1_6_Repeater board has a SSD1306Display so enabling it is pretty simple. This should also apply to T3S3 boards, but I don't have on hand to test.

Screenshot 2025-03-07 at 2 36 13 PM


jquatier:
working well on my LilyGo_TLora_V2_1_1_6

ripplebiz:
Thanks for that, and verifying.

PR #83, opened 2025-03-09
Add an ESP32_CPU_FREQ build flag to lower freq

fdlamotte:
It seems we can gain some uptime by lowering freq (80MHz seems sufficient and one can even try 40MHz ...)

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,6 +1,9 @@
 # MeshCore-FAQ
 A list of frequently-asked questions and answers for MeshCore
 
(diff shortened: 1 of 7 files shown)


ripplebiz:
Hi. Could you move the change to the ESP32Board::begin() ?
... and remove the .ini change. (or maybe have it commented out, to show that this option exists?)

fdlamotte:
> Hi. Could you move the change to the ESP32Board::begin() ? ... and remove the .ini change. (or maybe have it commented out, to show that this option exists?)

done, you're right, it's even better this way ;)

PR #87, opened 2025-03-10
enhancement on the UI and support for power off and status_led

fdlamotte:
In the end this is more generic than t1000e support ...

Added a powerOff() method for boards (defaults to infinite loop ~, deepSleep on heltecs and was already here for t1000e) => long press on user_btn = poweroff (on the heltec you can poweron via reset)

the idea was to support a status led, that could be used for other boards as well 
 - off time shows bat status (over 3.8mV => 2s off, under => 4s off)
 - on time tells if there are msgs pending (500ms msg, 2s no msg)
this is a start ... and there are already some ideas by Mike/N2QDK on discord

there was a need for specifying the polarity of poweroff (because on t1000 button pressed is high level ...)

I've added msg count ...


mikewren:
Adding some ideas here regarding LED status blink strategy:

Four blink "positions" that are cycled through every n seconds:

1) BLE Connection to App status (connected or disconnected)
2) GPS fix/lock (fix or no fix)
3) Battery State of Charge (SoC) (greater than or less than a specific threshold, sell app section below)
4) New Message(s) (a new message has been received since the since the app was last viewed - this may be challenging to determine)

The theory here is that a short blink is normal/good - while a longer blink is “hey something needs to be addressed." I'm thinking that maybe a 100ms (as short as reasonably possible) for short blinks, 500ms for long blinks, with a 250ms interval between positions. A short blink should be subtle and unobtrusive.

Charging would be 0-100% LED ramp up/down every 3 seconds, with the four blink codes doing their thing every  every third charging indicator ramp up/down cycle.

When fully charged (100% State of Charge) and connected to power, LED is on continuous at 50% brightness.

In the app(s), if companion BLE device is a t1000e:

1) Display (un-dim/gray out) a toggle for mute buzzer.
2) Display (un-dim/gray out) toggle for LED indicator status lights.
3) Display cheat sheet for blink codes somewhere in the app.
(MAYBE) 4) Display a slider for battery threshold the status LED will display a "long" blink for to indicate an issue.  I think 30% SoC is a reasonable default, but some people may want notice sooner (50% SoC) or later (10% SoC) that the battery is going flat.

Some examples:

A "short long short short" would indicate the T1000e is connected to an app via BLE, DOESN'T have a gps fix, battery is OK and no new messages.

A " short short long short" would indicate an app is connected via BLE, has a GPS fix, battery is BELOW the "good" threshold, and there are no new messages.

ripplebiz:
Finally getting some time to review.
Mostly looks good. Could you change base branch to 'dev', then I can merge

fdlamotte:
hope everything is fine ... 
just changed the base branch, better policy ...

PR #89, opened 2025-03-11
Add ESP8285 ELRS receiver as board

xznhj8129:
This pull adds ESP8285-based ExpressLRS receivers as supported boards with, for now, USB companion and Repeater modes. It was tested to function with WSL3 nodes. There's several caveats:
* Due to limited RAM, only 25 contacts are supported. More contacts fill the RAM to cause instability or just cause a build failure.
* The boards.json is taken straight from ELRS and is missing a few things, i could use some help there
* I'm not sure if there's gonna be room for the Wifi companion with that RAM level.
* There is no USB (serial adapter required) or BLE on them

Obviously this is a very limited application but opening the door to using ELRS receivers as nodes will open up a lot of exciting possibilities as most of them cost only a fraction of the cost of usual Lora modules, and the newer ones use ESP32s wich will allow a more fully featured build. Meshtastic does offer ELRS _transmitters_ as board options but receivers can be far more flexible in custom builds due to their smaller footprint.
Corrections welcome, this is the first time i develop using PlatformIO.


liamcottle:
Hey there, I'm going to close this out since there's been no activity since it was opened 6 months ago and there's now merge conflicts. Quite a bit has changed in the firmware since this was opened.

PR #90, opened 2025-03-11
Faketec support

oltaco:
Support for Faketec DIY devices, with either HT-RA62 or RA01sh radios.

-che


ripplebiz:
So, this is still in progress? Have reviewed and can merge if you like.

oltaco:
It should be OK, works for HT-RA62 radios and feedback is that it works for RA01sh.

I just wanted to test with an RA01sh radio for myself, but haven't had a chance to build one yet. Plan to get that done this weekend, and then I'll make it as ready.

oltaco:
Ok, confirmed working with RA01sh radio. Tested room server as well, seems to work.

One last question remains, should we rename this to promicro_diy to make it more inline with MT for people who are looking to make the switch?

ripplebiz:
I don't really know which name most people would recognise. I don't mind it going by the Faketec_ prefixes. I assume most people would know what that means?

oltaco:
I think so too, it seems to be the most popular and well known board and I'm sure if people are building their own they'll have no trouble realising this is what they want. 👍

PR #91, opened 2025-03-11
Basic gps support for Heltec_v3 and T1000e

fdlamotte:
this is a basic implementation, tried to be as generic as possible ...

works on t1000e and HeltecV3, disabled if HAS_GPS is not defined


MikesAllotment:
Can the same enhancement also be applied to the Heltec V2 variant please?  (Please don't forget about the V2s - there are a lot out there ;-) )

Issue #77, opened 2025-03-07
[Enhancement] Interlinking nodes through WiFi


tbb98:
Recently, Meshtastic 2.6 has introduced Meshtastic over LAN.

> Meshtastic 2.6 adds support for meshing over a local network (LAN) using UDP, currently available for ESP32 devices on WiFi. This feature allows nodes to communicate over a standard network connection, extending your mesh without relying solely on RF signals. This can be especially useful in locations where RF coverage is limited or when you want to bridge multiple Meshtastic networks over existing infrastructure. Once enabled, nodes automatically discover and connect to each other over the local network with minimal setup.

I believe implementing a similar feature in MeshCore could be highly beneficial for several reasons:  

- **Multiple repeaters in a building** – This would allow multiple repeaters to be placed in different corners of a building (e.g., when roof access is unavailable) without congesting the LoRa network.  
- **Internet interlinking** – A simple way to bridge networks using existing infrastructure, with VPN tunneling handled by the upstream router. (I've read on Discord that the app automatically uploads the nodes it sees—perhaps this could be extended to support displaying IP links.)  
- **Not the default behavior** – The complexity of setup would discourage this from being the default behavior, which is ideal for communities that prefer LoRa-only communication.  
- **Remote management of repeaters** – This would enable managing repeaters even when the user is outside the mesh.  
- **General-purpose chat with LoRa fallback** – This could serve as a broader chat application while retaining LoRa as a backup communication method.  
- **ESP32 auto-discovery via ESP-NOW** – Since ESP32 supports ESP-NOW (its makeshift nRF24 mode), it could be used for automatic discovery of reachable repeaters and clients, even by default when WiFi is not configured.  
- **OTA updates over WiFi** – Since ESP32 nodes are already connected via WiFi, firmware updates could be pushed over IP.  

What are your thoughts on this? 😄

tbb98:
I'm sorry, I've just realized, it's technically the duplicate of #37

liamcottle:
Closing. As mentioned above, duplicate of https://github.com/meshcore-dev/MeshCore/issues/37

Issue #37, opened 2025-02-22
[Enhancement] Non Lora Meshing (e.g. via ESPNow, TCP/UDP via WiFi)


Noki:
I would really love to see additional options for letting devices mesh in addition to using Lora e.g. via ESPNow, or TCP/UDP via WiFi. Here are some use cases:

- Make it easy to bridge 433 Lora and 868 Lora networks
- Reduce Lora airtime and avoid duty cycle limits for small distances (e.g. roof repeater / indoor node) maybe even making Lora optional for an indoor node
- Make it easy to bridge long distances via existing infastructure (e.g. via a WiFi mesh network or the internet in general)

I have no idea if these are viable options, but I definitely wanted to share them here and have them documented.

ripplebiz:
Yes, I have been musing over adding these capabilities. Definitely do-able.
I definitely want to explore stuff like this.

yNosGR:
Having a KISS/TNC interface would be nice as well. For example, being able to bridge/mesh over a serial TNC of over a tcp port like FreeDV-TNC2.

kurtrenauer:
MQTT-Bridge would be welcome very much as well, for my use cases and existing infrastructur.

Atorcha:
MQTT usefull

sevenbitbyte:
I'd like to see Serial and possibly CANBus on the list.

xpinguinx:
Integrating MQTT protocol would be awesome!

ZilvinasJonaitis:
I wish I could connect two LoRa nodes by wire (Serial). This could be especially useful for outdoor repeaters comprising of two repeater nodes in one box. One repeater dedicated to local communications within a range of omnidirectional antenna, another repeater node dedicated for long-haul link to remote repeater(s) via directional antenna.
Of course, there should be some interference management protocol implemented as both repeaters will operate on the same LoRa frequency. Some minimal serial link delay with buffering could a solution.

jbrazio:
Just my two cents, I would really hope Meshcore don't go into the direction of non RF bridges just like the other we all know.

TyIsI:
@jbrazio Honestly, I would argue that bridging to/over different media is not bad, as long as it doesn't become a default backhaul over the internet solution. Creating seamless extension/integration over distance/frequencies is what can really make it resilient.

Or even running local links on different frequencies than backhaul links.

kryptonut:
[...]

PR #96, opened 2025-03-12
wifi bridge for repeaters

### LINKED ISSUE #117: Design/Documentation - Bridge Packet Data Structure
The community seems excited about building several types of bridges #37. I'm working on the wifi+udp bridge ( #96 ) and I thought it might be nice to get some feedback on the bridged packet data structure I'm thinking about using. I expect that bridging metadata should be a feature bridge operators are able to opt-in to, so imo bridges should work with or without this data. When enabled the bridging node adds helpful debug data about packet reception. A signature is also included enabling future bridging tooling to optionally verify uplinked data was provided by trusted nodes.

| Bytes | Field | Type | Description |
|-------|-------|------|-------------|
| 1     | Version | `uint8` | Version of the bridging packet
| 4     | Frequency | `float` | LoRa Frequency the `Mesh::Packet` was heard on by the bridging node
| 4     | rssi | `float` | RSSI value of the packet as a float
| 4     | SNR  | `float` |  SNR value of the packet as a float
| 4     | Timestamp | `uint32` | Time the packet was received
| 32    | Node | `uint8[32]` |  Public key of the node bridging packets
| 3-255 | Packet | `Mesh::Packet` |     |
| 64    | Signature | `uint8[32]` | Signature of all preceeding fields

sevenbitbyte:
Allows repeater nodes to route packets over local wifi networks using UDP broadcasts.

This enables node operators to compose complex repeater solutions. Users may wish to experiment with different LoRa settings like higher data rates, using multiple frequencies for different purposes or even using a mixture of directional antennas.

## Issues

- [ ] partial #37 
- [ ] partial #77
- [ ] closes #117

## How it works

When a packet would normally be repeated we also echo it onto the wifi network as a broadcast UDP packet on the specified port. Bridged nodes who hear the UDP network broadcast then process the packet as if they heard if over LoRa and route the packet as usual.

Right now the UDP packets have no metadata and are just exact same data as the underlying meshcore packet.

Screenshot from 2025-03-12 18-56-03

For end users this all works seamlessly. As long as the users do a flooded advertisement or copy each other's contact info out of band they will eventually discover the bridged route the same as any other meshcore multihop path. 

## New build flags

These flags are only available on builds for boards with WiFi:

- `WIFI_ENABLE`
- `WIFI_BRIDGE_HOST` - bool - When true node will host access point
- `WIFI_BRIDGE_SSID`
- `WIFI_BRIDGE_PASSWORD`
- `UDP_SERVER_ENABLE`
- `UDP_SERVER_PORT` - int16 - What port to listen for UDP packets. Defaults to 8000
- `UDP_PACKET_BUFFER_SIZE` - Defaults to 10 - How many network packets to buffer

## New Commands

 - `get wifi.enable`
 - `get wifi.ap_enable`
 - `get wifi.ssid`
 - `get wifi.password`
 - `get udp.enable`
 - `get udp.port`
 - `set wifi.enable [on|off]`
 - `set wifi.ap_enable [on|off]`
 - `set wifi.ssid [ssid]`
 - `set wifi.password [password]`
 - `set udp.enable [on|off]`
 - `set udp.port [port]`
 
## Todo:

- [x] Allow users to change Wifi & UDP settings
- [x] Verify bridge is sending all expected packets
- [ ] Use debug methods `MESH_DEBUG_PRINTLN` instead of `Serial.print*`

## Questions:
- Should there be more metadata in the UDP packets? Should scoring data like snr/rssi be included?
- Should this be a new build type or is it ok included with all wifi capable repeater nodes?
- How much serial output should there be?


examples/simple_repeater/main.cpp
```diff
@@ -229,6 +286,148 @@ class MyMesh : public mesh::Mesh, public CommonCLICallbacks {
...
+
+      }
+
+      _mgr->queueOutbound(pkt, priority, futureMillis(_delay));
+    }
+  }
+
+  public:
+
+  void handleUdpPackets(){
```
sevenbitbyte: Indention

sevenbitbyte:
Oh yep, I'm missing packets that enter the outbound sendqueue in other ways (like advertisements).

Should probably change how I hook into Dispatcher. Instead of hooking  `Dispatcher::processRecvPacket` I should hook into `Dispatcher::checkSend` it looks like

https://github.com/ripplebiz/MeshCore/blob/main/src/Dispatcher.cpp#L191-L192

Noki:
I would love to see a small addition to this. Please add the ability to manually specify a list of UDP peers by ips/hosts in addition to broadcasting. Broadcasting works within the same location, but not if you want to bride locations e.g. via the internet or a routed wireless network.

Here in Berlin we have a huge wireless community network (WLAN) that spans the city (https://hopglass.berlin.freifunk.net/) and we would like to use it to bridge MeshCore repeaters in different areas of the city until the Lora based mesh is dense enough to close the gaps. I think this is a use case many communities have.

sevenbitbyte:
I was expressly not trying to solve the issue of packet routing beyond a local network or the physical LoRa range of MeshCore networks. so I'm not sure this PR can/should scale in this way, it would get quite a bit more complex to do all of that.

Currently with UDP broadcast we don't have to maintain any additional network state, repeater mappings or more network config that might break if host IPs change. My use case is specifically repeater operators who wish to run a mixture of directional antennas or bridge multiple frequencies. UDP broadcasts makes it VERY easy to add and remove nodes as needed and all the nodes will heal as new nodes come and go. I'm also focused on applications where there's not additional wifi infrastructure at all.

But, I think a simple compromise solution would be to support a single UDP target address to be specified OR to use a local broadcast. It would also make a LOT more sense for complex peering solutions like Berlin to build networking applications that simply plugin to the UDP data streams and shares them between operators in a way that can transit your network.

Noki:
I totally agree that UDP broadcast is a very good solution for bridges at the same location and like the idea of keeping it simple and as close to a mesh as possible. We have multiple locations that have 433 and 868 MHz nodes and might get more directional nodes at those locations in the future and I would use UDP broadcast for those setups. I really like that solution and think it should be enabled by default.

Adding UDP target addresses/hosts would be only for bridging remote locations via existing networks and I would not worry too much about things breaking here due to host or IP changes. In my opinion is the responsibility of the operator to ensure connectivity and most people are quite capable of handling this e.g. by using a dyndns service and port forwarding.

For our use case UDP peers by host/ip should work in addition to local broadcast and it would be really good to make it a reasonable small number of peers (e.g. 3 to 5) so you can actually build a network from it. One address would only allow us to bridge two locations which is not sufficient.

btw.: here in Berlin we can simply use static IP addresses and also hostnames. UDP packets can transit the network just fine, but broadcast traffic can not. It is a routed network and broadcast domains are limited to the individual locations, a network wide broadcast domain would not work, simply due to the size of the network.

src/helpers/CommonCLI.cpp
```diff
@@ -275,6 +301,31 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
...
           strcpy(reply, "Error, max 64");
         }
+      } else if (memcmp(config, "wifi.enable ", 12) == 0) {
+        _prefs->wifi_enable = memcmp(&config[12], "on", 2) == 0;
+        savePrefs();
+        strcpy(reply, _prefs->wifi_enable ? "OK - Reboot to apply wifi ON" : "OK - Reboot to apply wifi OFF");
+      } else if (memcmp(config, "wifi.ap_enable ", 15) == 0) {
+        _prefs->wifi_ap_enable = memcmp(&config[15], "on", 2) == 0;
+        savePrefs();
+        strcpy(reply, _prefs->wifi_enable ? "OK - Reboot to apply wifi AP ON" : "OK - Reboot to apply wifi AP OFF");
```
sevenbitbyte: should be `_prefs->wifi_ap_enable`

sevenbitbyte:
Hmm found a double send condition. Packets with defined paths which a bridging repeater initially saw on the network are re-broadcast over the air & and over the network.

So here the packet was initially heard over the air from the node at `192.168.1.71`, a node on the network rebroadcast it over the air but also over the network and so it's packet shows up in the network monitor as coming from `192.168.1.80` and with one less path byte.
 
Screenshot from 2025-03-20 17-48-54


Should add a `source` field to the `Mesh::Packet`

sevenbitbyte:
Switching base to dev branch

sevenbitbyte:
Merged changes from dev branch, haven't tested again yet

sevenbitbyte:
Working on a refactor to support multiple types of bridges:

  - `IBridge`
    - `virtual begin()`
    - `virtual loop()`
    - `virtual onPacketRx(Mesh::Packet* pkt)`
    - `virtual onPacketTx(Mesh::Packet* pkt)`
    - `sendToMesh(Mesh::Packet* pkt)`

WifiHero91:
It is very interesting! Good Job!

If you can modify to multicast we can use it on a bigger L3 network. Like hills connected via microwave dishes. :)

src/Packet.h
```diff
@@ -34,6 +34,10 @@ namespace mesh {
 #define PAYLOAD_VER_3       0x02   // FUTURE
 #define PAYLOAD_VER_4       0x03   // FUTURE
 
+#define PACKER_SOURCE_NONE        0x00
```
sevenbitbyte: typo'd

dlakelan:
Not sure if you already are, but Please make this IPv6 native! Ipv6 is already about 50% of internet traffic. There are ipv6 only networks!

I guess the thing would be to send to a v6 multicast address you can configure as a user.

jarrodlombardo:
What about USB or Ethernet bridge? I'd like to have two devices in one enclosure and no wifi involved.

Andrew-a-g:
Nice work!

I tried to test this with little success.  I can get the repeaters to join a wifi network and can ping them etc but they do not do anything to send data to eachother.  I have been through the code but I am sure I am either missing something or misunderstanding.

I set as much as I can in the heltec variant and had to gowith the commands to enable this but could not get it running.  WHen I run the commands to try force enable things it does not seem to work.  See below...

Variant config... (note: I also tried setting one up as an AP which also had the same results - wifi OK but no udp bridging).
[env:Heltec_v3_repeater]
extends = Heltec_lora32_v3
build_flags =
  ${Heltec_lora32_v3.build_flags}
  -D DISPLAY_CLASS=SSD1306Display
  -D ADVERT_NAME='"Heltec Repeater"'
  -D ADVERT_LAT=0.0
  -D ADVERT_LON=0.0
  -D ADMIN_PASSWORD='"password"'
  -D MAX_NEIGHBOURS=8
  -D WIFI_ENABLE=1
  -D WIFI_BRIDGE_SSID='"ssid"'
  -D WIFI_BRIDGE_PWD='"password"'
  -D UDP_SERVER_ENABLE=1
  -D UDP_SERVER_PORT=8000
  -D UDP_PACKET_BUFFER_SIZE=10

Commands...
get udp.network_bridge
  -> > off
set udp.network_bridge 1
  -> OK - udp network_bridge is now OFF
get wifi.info
  -> > esp32s3-E385EC 10.10.3.105 fe80:0000:0000:0000:4a27:e2ff:fee3:85ec 48:27:E2:E3:85:EC
get udp.mode
  -> > broadcast
get udp.address
  -> > 0.0.0.0
get udp.port
  -> > 0
get udp.network_bridge
  -> > off
set udp.network_bridge
  -> unknown config: udp.network_bridge
get udp.tx_bridge
  -> > off
set udp.tx_bridge 1
  -> OK - udp tx_bridge is now OFF

Any ideas?  Would love to test and help out where i can.

recrof:
is this being still worked on @sevenbitbyte? there is now merged esp-now/rs232 bridge: #454 if that helps in any way.

liamcottle:
Hello! Thank you very much for all the time and effort you put into this! Apologies for not having the chance to properly review it.

We now have an AbstractBridge/BridgeBase that was merged in #454. Maybe WiFi bridging could be refactored to build off that?

A lot has changed in the project in the last 6 months since this PR was opened and there's now merge conflicts. Unfortunately I will close this one out, since it's not something we can merge at this time :)

seeseepea:
Sorry to ask a silly question, but this seem to be limited to routing meshnode UDP data using wifi routers as bridges. I was looking for software that would permit wifi enabled non-mesh devices to connect to said wifi router, go to a router hosted page that runs an app allowing the device to send mesh-ages even though it is not mesh network device. That router would then be connected to a meshtastic/LoRa transmitter. The idea is that in an internet dead zone, people with traditional wifi devices would be provided a means to access the mesh network. 
Does such a thing exist or do I have to build it? If I have to build it, where would you suggest I begin.

PR #110, opened 2025-03-16
Add support for Faketec with LLCC68 instead of SX1262

lincomatic:
base class didn't have getIrqStatus() member

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -99,6 +99,7 @@
 #elif defined(FAKETEC)
   #include <helpers/nrf52/faketecBoard.h>
   #include <helpers/CustomSX1262Wrapper.h>
+  #include <helpers/CustomLLCC68Wrapper.h>
   static faketecBoard board;
 #else
   #error "need to provide a 'board' object"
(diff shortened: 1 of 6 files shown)


oltaco (changes requested):
It's probably better if we split the faketec build target into FaketecSX1262 and FaketecLLCC68 so that automated builds can make firmwares for both types.

Maybe in future the radio instantiation could be changed to attempt multiple radio types or something and then we could recombine the targets.

lincomatic:
> It's probably better if we split the faketec build target into FaketecSX1262 and FaketecLLCC68 so that automated builds can make firmwares for both types.
> 
> Maybe in future the radio instantiation could be changed to attempt multiple radio types or something and then we could recombine the targets.

My initial idea was to code it to work on both radios like the current Meshtastic build for the same radio (if SX1262 not found, try LLCC68), but the way the current code is structured makes it difficult to implement w/o making it somewhat ugly, because RADIO/WRAPPER_CLASS are defined in platformio.ini.

I could just make it try LLCC68 if  WRAPPER_CLASS fails, for instance

oltaco:
After thinking about it some more it's probably more appropriate to start a new build target and call it something like ProMicroLLCC68, as the Faketec PCB specifically calls out the sx1262 radio.

lincomatic:
OK, I'll do that. I messed up this PR because I forgot to make new branch, so I'll close it and submit a new PR

PR #111, opened 2025-03-16
added display support for RAK4631

recrof:
removed Wire.begin(int, int) from nrf52 devices(incompatible), cosmetic changes.


ripplebiz:
Some builds might need custom SDA/SCL. 
Can you revert, then use:
   Wire.setPins( ...);
   Wire.begin();

recrof:
> Some builds might need custom SDA/SCL. Can you revert, then use: Wire.setPins( ...); Wire.begin();

done

PR #118, opened 2025-03-18
Update clock validation to fix issue with clock stuck in the future

awolden:
* Standardizes clock validation and removes issue where a node can be improperly configured to be in the future and require a full reset to fix

image

Result:

Screenshot 2025-03-18 at 1 19 41 PM
Screenshot 2025-03-18 at 1 40 52 PM


ripplebiz:
clock integrity is quite important, and this PR effectively removes all guards to incorrect clock reversals.

PR #123, opened 2025-03-19
Faketec display and button support.

oltaco:
Tested working with HT-RA62 and RA01sh, with and without OLED display.

Confirmed button working and that BLE pin is dynamic with display, and falls back to 123456 if no display.

-che


oltaco:
Fixed builds for repeater, room server and companion usb firmwares.

PR #145, opened 2025-03-26
change new default freq to 869.525

recrof:


Noki:
Hope this gets merged soon as people tend to keep the defaults and the previous setting (867.5 / 250 kHz) bandwidth is not even allowed in Germany as it exceeds the allowed 200 kHz for that frequency.

PR #152, opened 2025-03-29
change default frequency as mentioned in FAQ

Noki:
The default frequency should be 869.525. This is the dafault for the builds and many nodes will just run with this because operators will not set proper settings. The old value is not even allowed in Germany as 867.5 is limited to 200 kHz bandwidth.


Noki:
Hmm, just spotted this is a duplicate of https://github.com/ripplebiz/MeshCore/pull/145

PR #162, opened 2025-03-31
Update faq.md

LitBomb:
added the following to FAQ:

public/private flood and repeater set flood.max CLI public key in hex and base64
How paths are created and reuse to minimize utilization T-Deck Diag explanation
Update repeater with a Raspberry Pi
clients don't repeat explanation
other meshcore projects


LitBomb:
need to make a change, will resubmit

PR #178, opened 2025-04-06
make SF=11 default for new installations

recrof:


6Toes-PlanetApe:
+1 I support the idea of SF11 as default. You can always switch to SF10 if you have RX/TX problems when your envirement needs it.

starvald:
agreed

PR #185, opened 2025-04-08
Xiao nrf52

fdlamotte:
Initial support for xiao_nrf52 + wio-sx1262

### DIFF
--- a/boards/seeed-xiao-afruitnrf52-nrf52840.json
+++ b/boards/seeed-xiao-afruitnrf52-nrf52840.json
@@ -1,0 +1,61 @@
+{
+  "build": {
+    "arduino": {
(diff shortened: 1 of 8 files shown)


fdlamotte:
I have a question to ask seeed regarding pinout ...
There is some inconsistency between the variant.h they provide and their schematics

in variant.h pin 17 is hooked to i2c while in schematics it gives charge state !!!

fdlamotte:
No finally it's ok ... the pin map is not linear, so HW pin 17 is D23 in the code ...

fdlamotte:
cleaned the PR ...

fdlamotte:
Could test bat reading ... ok after some adjustments

So this port should be ready to be integrated

PR #193, opened 2025-04-11
T114 display

fdlamotte:
* Added a display class for ST7789 (color TFT screen used on T114)
* Added a target for compiling T114 BLE Companion using the screen (some T114 don't have screen)
* modified ui class so refresh is done only if screen has been modified (ST7789 is slow to refresh and doesn't have double buffer, it will also be necessary for e-ink screens)
* added some colors
* ensure that the new code is not compiled against a SSD1306 based target

Code has been tested on T114, but also on a SSD1306 target (heltec_v3)

### DIFF
--- a/examples/companion_radio/UITask.cpp
+++ b/examples/companion_radio/UITask.cpp
@@ -89,22 +89,29 @@ void UITask::renderCurrScreen() {
     // render message preview
     _display->setCursor(0, 0);
     _display->setTextSize(1);
+    _display->setColor(DisplayDriver::GREEN);
     _display->print(_node_name);
 
     _display->setCursor(0, 12);
+    _display->setColor(DisplayDriver::YELLOW);
     _display->print(_origin);
     _display->setCursor(0, 24);
+    _display->setColor(DisplayDriver::LIGHT);
     _display->print(_msg);
 
     _display->setCursor(100, 9);
     _display->setTextSize(2);
+    _display->setColor(DisplayDriver::ORANGE);
     sprintf(tmp, "%d", _msgcount);
     _display->print(tmp);
   } else {
     // render 'home' screen
+    _display->setColor(DisplayDriver::BLUE);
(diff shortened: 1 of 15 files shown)


liamcottle:
> ST7789 is slow to refresh and doesn't have double buffer

The Adafruit ST7789 driver shouldn't be used for the T114 display.

As you noted it's slow to refresh. This will cause issues with the LoRa TX/RX. I contacted Heltec about this a long time ago when working on display support for RNode Firmware. They said we need to use the Heltec driver for this display, which updates much faster. https://github.com/Bei-Ji-Quan/st7789

@recrof was working on display support for the T114, but I'm unsure if any further progress has been made.

recrof:
I gladly put the work into more capable hands of Florent. thanks for your input @liamcottle

fdlamotte:
will have a look ...

It appears the lib depends on OLEDDisplay.h, which comes from mbed ... still need to figure out the implications of that because if I recall well, mbed is a whole framework

fdlamotte:
I'm thinking about e-ink screen on t-echo ... the issue will be worse with that, and slow refresh is by design 

The good point is that the spi bus for the screen is not the same as the lora one, so we can do things in parallel ... I'm used to use DMA for this kind of transfers on other platforms but I don't know if it is available on arduino framework (quite new to this), could also make a freertos task that receives the screen update order and updates the screen asynchrously

This could be also used on the t114 where the screen also has its own spi bus (fast refresh would be better, of course ...)

ripplebiz:
Hi, this PR has conflicts with base branch now. Can you resolve these please?

fdlamotte:
Conflict has been resolved

ripplebiz:
I'll merge this, but would be good to potentially use double-buffering in the (near) future, where endFrame() blits to actual display. The LovyanGfx lib is usually pretty good for this.

fdlamotte:
I'll take a look at it, double buffering is what I was aiming to

will address t-echo first ;)

PR #197, opened 2025-04-13
Update faq to reflect new default spread factor

neutralinsomniac:


LitBomb:
looks good!  thanks for making the change.

ripplebiz:
There isn't _global_ consensus on SF... it's still very much up in the air.

neutralinsomniac:
Shouldn't the FAQ match what's currently running on the default firmware though?

PR #204, opened 2025-04-15
Improved battery voltage estimate for HeltecV3 boards.

gw1urf:
This PR changes the HeltecV3 board's battery measurement function. It uses the analogReadMilliVolts function, rather than plain analogRead. The former incorporates the ESP32's saved analog calibration, which improves the linearity of ADC readings.


446564:
Are we just waiting to test this? 

Also @gw1urf this should be targeting the dev branch not main 😎

MikesAllotment:
Should the same change also be applied to V2 board - not just V3?

recrof:
closing due of lack of activity

PR #216, opened 2025-04-21
Support for Pi PicoW

AeroXuk:
This PR adds support for the Pi PicoW in the following modes:
- Companion Radio over USB Serial
- Repeater
- Room Server
- Terminal Chat

It should be possible to add at a future time:
- Companion Radio over BLE
- Companion Radio over WiFi


AeroXuk:
`ADC_MULTIPLIER` value needs correcting, I wasn't sure how it was calculated.

PR #217, opened 2025-04-21
Support for Pi PicoW

AeroXuk:
This PR adds support for the Pi PicoW in the following modes:
- Companion Radio over USB Serial
- Repeater
- Room Server
- Terminal Chat

It should be possible to add at a future time:
- Companion Radio over BLE
- Companion Radio over WiFi

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -3,6 +3,8 @@
 
 #if defined(NRF52_PLATFORM)
   #include <InternalFileSystem.h>
+#elif defined(RP2040_PLATFORM)
+  #include <LittleFS.h>
 #elif defined(ESP32)
   #include <SPIFFS.h>
 #endif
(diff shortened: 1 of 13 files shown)


AeroXuk:
`ADC_MULTIPLIER` value needs correcting, I wasn't sure how it was calculated.

Update: I think `3.1 * 3.3 * 1000` might be the correct value, but don't have enough time to test now.

ripplebiz:
Thanks for this. All looks fine.
I will merge, but have noted there might be a problem with the IdentityStore. I made a mistake with the nRF52 variant a while back, not realising it needs a fs.mkdir(), unlike SPIFFS which doesn't really have directories.

AeroXuk:
> Thanks for this. All looks fine.
> 
> I will merge, but have noted there might be a problem with the IdentityStore. I made a mistake with the nRF52 variant a while back, not realising it needs a fs.mkdir(), unlike SPIFFS which doesn't really have directories.

Thanks for the heads up. Firmware appeared to be working and retrieved identity after a reboot or power outage. I will double check when I can but I think the Pico-Arduino LittleFS implementation handles this, at least according to the docs.

https://arduino-pico.readthedocs.io/en/latest/fs.html#littlefs-file-system-limitations

PR #221, opened 2025-04-23
Added  Pin A0 for PIN_VBAT_READ to xiao_s3_wio Board

Emmpunkt:
For measuring the Battery Voltage.
Voltage divider with 2 Resistors of same size to Pin A0 (D0)


ripplebiz:
Because the Xiao doesn't support this as standard, and that this is something you have to wire yourself, it only makes sense to have this config change done _manually_ .

PR #225, opened 2025-04-25
companion_radio: greatly reduce the status LED usage

### LINKED ISSUE #200: Change T1000-E LED pulse rate.
The current pulse rate is causing premature battery drain, the LED is on for 3 seconds and off for 3 seconds, even when its not on charge.  this needs to be changed to a 150ms pulse every 10 seconds.
Thank you.

alesgenova:
If the companion device has a status LED it typically stays on for 2s and then off for 2s.

This PR greatly reduces the LED activity, by implementing the following pattern:

```
// No messages
|++|--------|++|--------|++|--------|...
0s 0.02s    4s 4.02s    8s 8.02s    12s
```

```
// With messages
|++++|------|++++|------|++++|------|...
0s   0.2s   4s   4.2s   8s   8.2s   12s
```

## Advantages over current implementation:
- The LED is typically ON 0.5% of the time instead of 50% of the time
- The LED cycle is always exactly 4 seconds, while previously it was variable at 4 seconds (no messages) or 2.5 seconds (messages)
- Since the cycle is slower, and the LED is ON for a very short time, we avoid polling the battery status each cycle

Tested on T1000-E

Fixes #200

### DIFF
--- a/examples/companion_radio/UITask.cpp
+++ b/examples/companion_radio/UITask.cpp
@@ -4,6 +4,12 @@
 
 #define AUTO_OFF_MILLIS   15000   // 15 seconds
 
+#ifdef PIN_STATUS_LED
+#define LED_ON_MILLIS     20
+#define LED_ON_MSG_MILLIS 200
+#define LED_CYCLE_MILLIS  4000
+#endif
+
 #ifndef USER_BTN_PRESSED
 #define USER_BTN_PRESSED LOW
 #endif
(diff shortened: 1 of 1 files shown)


446564:
I personally don't use the led except to check if the device is on or not. So anything to reduce battery use is great imo 😎

akrabu:
I'd personally love to be able to turn off the status part of the LED altogether (maybe just a single blip on startup), and only use the LED for new message alerts.

IoTThinks:
> I'd personally love to be able to turn off the status part of the LED altogether (maybe just a single blip on startup), and only use the LED for new message alerts.

Me too.

PR #228, opened 2025-04-26
remove pin mode setup from uitask

### LINKED ISSUE #215: Rak user button
I flashed the latest firmware in Ble version for my rak wisblock board. Before meshcore I had flashed meshtastic and the button worked(as per meshtastic guide I used PIN 31 aka AIN1). I flashed the file in meshcore releases but upon pressing the button nothing happens, the display stays off. I also tried changing the button parameter by changing the PIN from 9 to 31, recompiled and loaded the new firmware but still nothing.

liamcottle:
This PR removes the `pinMode` call from `UITask.cpp`. Fixes #215, if `INPUT` is needed instead of `INPUT_PULLUP` for a different board type, this should be set in the board file instead of `UITask.cpp`.

### DIFF
--- a/examples/companion_radio/UITask.cpp
+++ b/examples/companion_radio/UITask.cpp
@@ -43,10 +43,6 @@ void UITask::begin(DisplayDriver* display, const char* node_name, const char* bu
     *dash = 0;
   }
 
-  #ifdef PIN_USER_BTN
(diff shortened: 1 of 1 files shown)


fdlamotte:
I think this was necessary for t114, but I'll handle the setup in variant, where it should be ...

PR #230, opened 2025-04-28
Add a way to easily define blinking patterns based on device state

alesgenova:
This PR adds an helper class to easily create blinking patterns.
Depending on the state of the companion device, we swap out the active blinking pattern.

4 patterns implemented so far:
- Idle (blink once for 10ms every 4 seconds)
- Pending messages (blink once for 400ms every 4 seconds)
- Connected to app (quickly blink twice for 10ms every 4 seconds)
- Pairing mode (not used yet, stay on for 1.5s every 2 seconds)

Tested on the t1000e

### DIFF
--- a/examples/companion_radio/BlinkPattern.cpp
+++ b/examples/companion_radio/BlinkPattern.cpp
@@ -1,0 +1,39 @@
+#include "BlinkPattern.h"
+
+BlinkPattern::BlinkPattern(bool initial_state, const unsigned long* pattern, size_t size)
+  :_initial_state(initial_state)
+{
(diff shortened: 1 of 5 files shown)


alesgenova:
@recrof This PR is now up to date with the latest upstream changes to `UITask`

From the user point of view the significant change is that the device blinks twice if there is an active BLE connection

liamcottle:
Hey there! Thanks for the PR :) Unfortunately this got stale, a lot has changed in the last 5 months, especially with the UI task rewrite. I'll close this out as there are now merge conflicts.

PR #231, opened 2025-04-28
companion radio: manually advertise for 15s instead of always automaticallly advertising

alesgenova:
I don't expect this PR to be merged as is, but I would like to start a discussion.

In my opinion there is a security concern in the way display-less companion radios are currently implemented.
This is the existing workflow:

- For display-less devices the bluetooth pin is static and set at build-time (default: "123456" in the provided builds)
- The device boots and automatically advertises itself indefinitely (think pairing mode)
- Any phone in range can connect to it using the "secret" pin
- If the BLE connection to your phone stops for any reason, the companion radio goes back to advertising indefinitely

While this workflow is convenient for the user, I think it's pretty unsafe, since the companion radio is always in pairing mode whenever not connected to a phone, and the default pin is publicly known. Anybody in bluetooth range to your companion radio can connect to it, if you are not already connected to it.

In this PR I implement a different workflow, which while not as convenient, I still think is bearable:
- The device boots and automatically advertises itself for 15 seconds
- If no phone connects to it within 15 seconds it will stop advertising
- If the BLE connection to your phone stops for any reason, the companion radio will not automatically advertise.
- When not advertising, the user can long-click the user button to enable advertising for 15 seconds, allowing you to connect your phone with the radio again

Thanks to the changes proposed in #230 , the user will be able to tell by looking at the LED if the radio is connected/advertising/idle

As a bonus, I refactored a little the `SerialBLEInterface` implementation so it uses event callbacks provided by the ble library.

What do you think about this approach? Can we come up with something as convenient as possible without the security concerns?

### DIFF
--- a/examples/companion_radio/BlinkPattern.cpp
+++ b/examples/companion_radio/BlinkPattern.cpp
@@ -1,0 +1,39 @@
+#include "BlinkPattern.h"
+
+BlinkPattern::BlinkPattern(bool initial_state, const unsigned long* pattern, size_t size)
+  :_initial_state(initial_state)
+{
+  _pattern_size = size > 8 ? 8 : size;
+
+  for (size_t i = 0; i < _pattern_size; ++i) {
+    _pattern[i] = pattern[i];
+  }
+}
+
+void BlinkPattern::start(unsigned long t0) {
+  _pattern_idx = _pattern_size - 1;
+  _state = !_initial_state;
+  _next_change = t0;
+}
+
+bool BlinkPattern::loop(unsigned long t) {
+  if (t >= _next_change) {
+    size_t next_idx = (_pattern_idx + 1) % _pattern_size;
+    _next_change += _pattern[next_idx];
+
+    if (next_idx != 0) {
+      _next_change -= _pattern[_pattern_idx];
+    }
+
+    _state = !_state;
+    _pattern_idx = next_idx;
+
+    return true;
+  }
+
+  return false;
+}
+
+bool BlinkPattern::state() const {
+  return _state;
+}
--- a/examples/companion_radio/BlinkPattern.h
+++ b/examples/companion_radio/BlinkPattern.h
@@ -1,0 +1,21 @@
+#pragma once
+
+#include <stddef.h>
+
+class BlinkPattern {
+  bool _initial_state = false; // state of the led at the beginning of the periodic pattern
+  bool _state = false; // current state of the led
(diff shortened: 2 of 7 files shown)


fdlamotte:
I have this feature on the custom fw for t1000 

* It works well as long as you have your node with you ;)
* The energy gains are not great
* security wise, I think it's also possible to change the pin in the app after first cx

In the end its configurable and I have it always disabled ...

fdlamotte:
Some more comments ;)

First I think the discussion is very interesting, please forgive me if my response sounded too negative ...

* Refactoring of SerialBLEInterface is very welcome ;)
* 15s is a little short I think (on the t1000 the default was 3min and the counter restarts at ble disconnection)
* I prefer when advertising is done on short press ... and I think it should start a train of blinking at the same time

alesgenova:
@fdlamotte 

> I have this feature on the custom fw for t1000
> 
Not sure I understand, is the custom firmware based on this PR or is it something else?

>    *  security wise, I think it's also possible to change the pin in the app after first cx
> 
That's the main issue, as far as I can tell, I couldn't find a way to customize the PIN trough the app settings, that's why I was proposing this not-ideal firmware only change.

If the PIN could be changed after connecting it for the first time, then most of the security concerns would go away, and it would be ok to automatically be advertising for longer (3 min as you say?) after disconnecting

fdlamotte:
> @fdlamotte
> 
> > I have this feature on the custom fw for t1000
> 
> Not sure I understand, is the custom firmware based on this PR or is it something else?
> 

No I maintain a custom firmware for T1000 with extra features (gps, cli ...) here : https://github.com/fdlamotte/MeshCoreT1000e

> That's the main issue, as far as I can tell, I couldn't find a way to customize the PIN trough the app settings, that's why I was proposing this not-ideal firmware only change.

well I confirm you that the firmware lets you change the pin, it's been available in meshcore-cli via `set pin` command for a long time. Didn't ask for it so if it does not exist yet in the app, it should land quickly ;)

I've just tested it, you need to reboot but it works ;)

alesgenova:
Closing this for now, because having the ability to change the PIN from the app satisfies most of my security concerns.

For future reference, this branch has the following changes:

- nrf52 companion only: automatically advertises for 60 seconds on startup or whenever the connected device disconnects
- while the companion is advertising a specific LED pattern is displayed to the user
- the user can start another 60s advertisement cycle by long pressing on the user button
- `SerialBLEInterface` for nrf52 is refactored to use the provided connect/disconnect/advertisement callbacks instead of polling the state

PR #237, opened 2025-04-30
T114 monochrome

fdlamotte:
Here is the new screen driver proposal for t114

It uses @liamcottle proposed code for faster refresh, the drawback is that the display is now monochrome

I've tried to set screen color according to context 
 - connected/not connected
 - messages
 - pin value

The library used is a little dirty but works ...

Screen is now default for t114 targets. There is no way to poll for screen (spi is only MOSI) but this should not hurt normal behavior of t114 with no  screen. Only drawback is that the user will have to set its pin to use as ble device (meaning he should flash a companion radio usb, set the pin and then flash a companion radio ble). I don't think there are many t114 with no screen and people who buy it this way know what they are doing ...

made some fixes to t114 port also ...


bobricius:
What should I do to make the display colorful again????
THANKS !

PR #240, opened 2025-05-01
WIP: Run multiple mesh nodes on a single device

alesgenova:
So I decided to do a little experiment the last couple of nights and it ended up working pretty good, but before I spend more time on this I want to check if it's something that the community is interested in.

So this PR is a proof of concept that allows us to run multiple nodes on a single device. This allows us to do the following things for example:
- Run multiple chat servers on a single device (this case is implemented here)
- Run both a client node and a repeater
- Run two or more repeaters with different settings of what will be repeated

At a very high level this is what's happening here:

- I add a new abstraction called `MeshNode`. 
- Each instance of `MeshNode` has its own unique identity. 
- Instead of subclassing `Mesh` itself, we will have specific implementations of `MeshNode` for server room, repeater, companion, etc. (provided here is an example for the `RoomNode`).
- The `Mesh` keeps a reference to an array of nodes, and when a packet is received, it gets forwarded to the various nodes, which will determine if it's meant for them or not, and will either process it or ignore it.

So using this architecture, I was able to modify the room server example so that 3 room servers are run on the same hardware, and it actually works, I was able to chat on each instance sharing messages with two clients.

```
ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
...
ROOM  name: Room_0  identity_key: _main_0  ID: 30F3524B6721DC13954BDD9435268C8D16656C42F83E1C2C865115D61C8A8E8C
ROOM  name: Room_1  identity_key: _main_1  ID: BC0CCF60443669AE0DB87E0F7904486A487043359D53D3753AEBF3FC8770D5ED
ROOM  name: Room_2  identity_key: _main_2  ID: C206756E4AF93AF646E6ED87C09520820CD817B0F80ED11BAA1E5D564709BA7E
```

Screenshot_20250501_000401_MeshCore

Anyway, this is a proof of concept and is in not anywhere ready to be merged, but I wanted to gauge interest before I sink more time into it.


antijingoist:
Gauging interest?

OMG, yes

mrlmuk:
This seems like a brilliant idea which will avoid a lot of waste where people need multiple devices for each role. I currently have a Heltec v3 which really should be a repeater but is set up as a client with WiFi so Home Assistant can connect to it.

mattiaskagstrom:
having a repeater node also running as a room server would be awesome

andyspoo:
I have a V4 as a repeater. Most of it's memory is just not being used. Being able to also use it as a room server at the same time would be awesome.

LitBomb:
> I have a V4 as a repeater. Most of it's memory is just not being used. Being able to also use it as a room server at the same time would be awesome.

you can already enable repeat on room server.  although room server with repeat on doesn't have some of the more important repeater features, so it is not recommended.  It is best to run room server as room server only.

liamcottle:
Looks interesting, might be something to look into in the future. Will close this out due to old age, no activity and lots of conflicts.

swaits:
You could easily build something like this on top of DongLoRa: https://github.com/swaits/donglora

This turns your Heltecs (etc) into raw LoRa dongles. And it ships with a multiplexer so you can access it from as many things as you want over Unix and TCP sockets.

Disclaimer: I am the author.

PR #247, opened 2025-05-05
UI Text Width - minor improvement

jquatier:
Small PR to clean up some magic numbers in the UITask. All our display drivers support a native way to get text width based off the current font size. This helps when centering text in the UI (we do this for the boot screen). This should be a lot cleaner.

- Added getTextWidth method to DisplayDriver interface
- Implemented getTextWidth in all display classes
- Updated examples to use getTextWidth directly to center text
- Tested on Heltec V3 and T114

IMG_5592 2


ripplebiz:
Nice one!

PR #255, opened 2025-05-07
fixing scaling of bitmaps for 7789 display

jquatier:
Added some basic scaling to the drawXbm method in the ST7789 display code which impacts the T114. All of the other methods in this class account for scaling but bitmaps need to be scaled accordingly. Right now in the dev branch the logo is tucked in the top left corner and smaller after the recent scaling changes. This seems to work well. If I had a T-echo to test on I'd add this to that display class too but might need someone else's help with that.

IMG_5613

### DIFF
--- a/src/helpers/ui/ST7789Display.cpp
+++ b/src/helpers/ui/ST7789Display.cpp
@@ -111,7 +111,35 @@ void ST7789Display::drawRect(int x, int y, int w, int h) {
 }
 
 void ST7789Display::drawXbm(int x, int y, const uint8_t* bits, int w, int h) {
-  display.drawBitmap(x*SCALE_X + X_OFFSET, y*SCALE_Y + Y_OFFSET, w, h, bits);
+  // Calculate the base position in display coordinates
+  uint16_t startX = x * SCALE_X + X_OFFSET;
+  uint16_t startY = y * SCALE_Y + Y_OFFSET;
+  
+  // Width in bytes for bitmap processing
+  uint16_t widthInBytes = (w + 7) / 8;
+  
+  // Process the bitmap row by row
+  for (uint16_t by = 0; by < h; by++) {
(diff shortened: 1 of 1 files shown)


ripplebiz:
I think that approach will sorta 'appear' to work, but looks like there'll be overlapping logical pixels, and also the aspect ratio won't necessarily be 1:1

ripplebiz:
Wonder if it could be done with some sort of cumulative integer error, so some logical pixel are, say 2 wide, some 3 wide, etc. eg.   for (int x = 0; x < w; x++) {  int x1 = (int) (x * SCALE_X); int x2 = (int) ((x + 1) * SCALE_X);  int block_w = x2 - x1;  ...  }

liamcottle:
For reference, this was my implementation for scaling when working on T114 display for RNode firmware.

Unfortunately it had the issue where if the scale was not 1x or 2x, for example scale was 1.5x or 1.75x, it looked bad...

https://github.com/markqvist/RNode_Firmware/blob/fff96fb5c0f6e14f948ec2d92c408ae2cc1ddb65/Display.h#L482

jquatier:
Cool let me try a few things out tonight!

jquatier:
Tried out another way, basically starting from each logical pixel in the bitmap:
    - Calculates the exact starting and ending coordinates (x1,y1) to (x2,y2)
    - Determines the actual width and height of each pixel block (block_w, block_h)
    - Uses fillRect to draw precisely-sized blocks for each bitmap pixel
    
Avoids overlapping pixels that occurred in the previous implementation and distributes pixels evenly based on scale factors (some pixels will be 1px wide, others 2px wide, etc.) and uses drawRect to draw them.

jquatier:
Current build:
Screenshot 2025-05-07 at 6 31 50 PM

PR #258, opened 2025-05-07
Heltec Wireless Tracker support

seagull9000:
Initial support for Heltec Wireless Tracker companion radio ble based on the V3 
- Display is functional
- Battery Voltage is functional

Possible issues remaining:
- TX LED does not light on transmission - this may be a bug in the V3 code
- Battery voltage level did not reach 100% on LiPo 18650 - may require further tweaking of values my battery may not be entirely healthy.

Note:  I don't have a V3 so I can't regression test for those devices.

ref: https://github.com/ripplebiz/MeshCore/issues/239

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -59,7 +59,9 @@
 
 #ifdef DISPLAY_CLASS
   #include "UITask.h"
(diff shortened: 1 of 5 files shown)


ripplebiz:
Nice one

PR #261, opened 2025-05-08
Some defines to use Adafruit_littlefs with stm32

fdlamotte:
I still work out of tree on the e5, but need this to compile ;)


ripplebiz:
Awesome stuff.

Just wondering if we can organise the LittleFS stuff into the /arch/stm32 directory, like I did with the AsyncElegantOTA lib? It's just a static library that cannot be pulled from pio registry, so is put there instead.  And, then it is added as libdep like:
lib_deps =
  file://arch/esp32/AsyncElegantOTA

Can you try this with stm32/littlefs?

ripplebiz:
Also, noticed there's a missing 'return false;'  in:
bool startOTAUpdate(const char* id, char reply[]) override {};

fdlamotte:
> Awesome stuff.

Great it pleases you, but it mostly shows how great meshcore is !

> 
> Just wondering if we can organise the LittleFS stuff into the /arch/stm32 directory, like I did with the AsyncElegantOTA lib? It's just a static library that cannot be pulled from pio registry, so is put there instead. And, then it is added as libdep like: lib_deps = file://arch/esp32/AsyncElegantOTA
> 
> Can you try this with stm32/littlefs?

That was exactly what I was looking for, thanks for orienting me, this seems to work like a charm !

fdlamotte:
made a cleaner PR ;)

PR #266, opened 2025-05-10
T114 GPS Support

jquatier:
Adding support for T114 GPS via new SensorManager. Screenshot is with debug turned on. 
Tested via the app too and lat/long is updated.

Screenshot 2025-05-09 at 8 36 43 PM


ripplebiz:
Would be good if there was a way to detect at runtime if there is a GPS connected, and only expose the "gps" custom var if detected. Then the app can know as well.

jquatier:
Added some code for that and tested it on my device with gps connected and disconnected, let me know if you were thinking about it differently. I think this approach works well since we might not want GPS powered on all the time. This keeps it off until the user sets the setting, but it still initially powers it on to check if it's there.

PR #267, opened 2025-05-10
Loads of adds for T-Beam Supreme

cod3doomy:
T-Beam Supreme:

- Increased BLE companion channels to 8
- Added SH1106 template files
- Restructured display include selection in the example main.cpp files
- Uncommented line 99 in ST7789Spi.h that was the cause of compiler failure
- Created power_init() for setting up the PMU
- Created sensor and GPS structure and verified GPS is working with coordinates read from meshcli interface


ripplebiz:
Hi,
Could you please remove the .vscode files? (they shouldn't be included, and would have thought the gitignore would exclude those)

ripplebiz:
It looks like the Wire1.begin() is missing now (?)

ripplebiz:
There are other build targets that use the 'ST7735' and 'ST7789' #define's, so be careful when changing the main modules, and only your own platformio.ini file. I'm pretty sure some other targets are now broken (like Heltec Tracker)

PR #270, opened 2025-05-11
big cleanup, refactoring, radiolib.setCurrentLimit fixes for all radios

recrof:
cleanup: moved all variant code to variant folders, removed duplicate definitions and unified variant.h files; pin definitions should now only be in platformio.ini or variant.h
cleanup: moved all radiolib wrappers to dedicated folder
fixes: introduced SX127X_CURRENT_LIMIT and set it to 120mA (radiolib default is 60), raised SX126X_CURRENT_LIMIT to 140mA (radio maximum from datasheet)


marrold:
Thanks for this, just a note - In the ideal world a refactor / cleanup and something like changing the current limit would be in two separate PRs

recrof:
> Thanks for this, just a note - In the ideal world a refactor / cleanup and something like changing the current limit would be in two separate PRs

of course you're right. it unfortunately fused into one during yesterday's debugging session. I'll try to not make that mistake again.

recrof:
allright.. will close this one, as it's too big to be reviewed and add it as series of smaller PRs.

PR #278, opened 2025-05-12
Word wrap node name on home screen, slightly better battery icon and implement user button on RAK4631 using analogue pin 31 (same as MT) 

adam2872:


adam2872:
will re-pr separate changes

PR #279, opened 2025-05-12
Word wrap node name on home screen and slightly better battery icon

adam2872:


ripplebiz:
I generally don't like to use the std::string class, and prefer static C strings.
Also, can you change the base branch to 'dev' (not main)

adam2872:
need to made changes

PR #281, opened 2025-05-13
t-beam supreme: added GPS functionality

cod3doomy:
Enabled GPS and verified with meshcli.
All supreme envs build w/o issues.

image


ripplebiz:
Looks good. I'll merge this, but note I prefer not to use the String class, or any dynamic memory allocation, as a general rule. I'll refactor this after merging.

Issue #283, opened 2025-05-13
Add support for INA3221 current sensor to the Repeater firmware


ngavars:
**My Use Case**

A solar repeater node that is using non-lithium battery (Lead Acid, LTO, Na+) often would use a DC-DC buck/boost converter to power the node. This means that the node would not be able to use the built-in ADC to measure battery voltage. Hence the requirement for current/voltage sensor like INA3221.
INA3221 is particularly useful, because it has 3 channels and would be able to measure also solar panel output and, possibly, other voltages.

It would be great if it would be possible to current/voltage/power values from all 3 INA3221 channels somewhere in telemetry section of the app (or via meshcore-cli).

ngavars:
I have added support for INA3221 to the Promicro variant. I'll be creating a PR shortly.

oltaco:
This can be closed, can you do the honours @liamcottle

PR #284, opened 2025-05-13
Telemetry: INA3221 current sensor support for Promicro

ngavars:
INA3221 current sensor support, see Issue #283


ripplebiz:
As a general rule I prefer not to use the String class, or any dynamic memory allocations. For embedded work this is makes for a more determinate execution environment, and helps keep things simple.
In that same vein, the SettingsManager doesn't really add value IMO. 
Happy to merge this with a refactor with those two points.

ngavars:
Ah, I had a feeling that the string class and my poor little settings manager will be frowned upon 😀
The added value of settings manager, at least in my mind, was easier track of different settings as more and more sensors get added, each with their own settings.
But I am happy to refactor those.
Thanks for your feedback.

ripplebiz:
Also, can you change the base branch to 'dev' please

ngavars:
@ripplebiz I removed the SettingsManager and the String is gone. Also changed the INA3221 address to 0x42 so that it can be used with one or several INA219 in the future (and it matches MT). I also added a new TELEM_PERM_ENVIRONMENT permission for access to env sensors.

PR #286, opened 2025-05-14
esp32: make 80Mhz default

recrof:
for better battery life and cooler devices


446564:
do we need some testing?

ripplebiz:
> do we need some testing?

Yes please!

446564:
> > do we need some testing?
> 
> Yes please!

I've been meaning to get to testing this but don't have a busy mesh to test with. Starting to wonder about ways to test with actual radios programmatically. simulate high traffic etc

446564:
performance seems the same as before, but power draw drops from 150 mAh to 110 mAh.

PR #288, opened 2025-05-14
Word wrap node name on home screen

adam2872:


ripplebiz:
I think this kind of problem should be solved at the DisplayDriver level.
I will be working in a refactor of the UITask soon, and can have some kind of word wrap introduced then. 
BTW, some of the underlying low-level graphics libs do stuff like this, so better to just have a high-level method that maps to whichever impl.

PR #289, opened 2025-05-14
Implement user button on RAK4631 using analogue pin 31 (same as MT)

adam2872:


ripplebiz:
If think this line:
    #ifdef PIN_USER_BTN || PIN_USER_BTN_ANA

should be:
  #if defined(PIN_USER_BTN) || defined(PIN_USER_BTN_ANA)

ripplebiz:
This line:
   pinMode(PIN_USER_BTN_ANA, INPUT_PULLUP);

should be:
  pinMode(PIN_USER_BTN_ANA, INPUT);

adam2872:
> This line: pinMode(PIN_USER_BTN_ANA, INPUT_PULLUP);
> 
> should be: pinMode(PIN_USER_BTN_ANA, INPUT);

It needs the pullup or else the value read floats around and can lead to false triggers

adam2872:
> If think this line: #ifdef PIN_USER_BTN || PIN_USER_BTN_ANA
> 
> should be: #if defined(PIN_USER_BTN) || defined(PIN_USER_BTN_ANA)

I’ll do some testing with this.

ripplebiz:
> > This line: pinMode(PIN_USER_BTN_ANA, INPUT_PULLUP);
> > should be: pinMode(PIN_USER_BTN_ANA, INPUT);
> 
> It needs the pullup or else the value read floats around and can lead to false triggers

If this (the _ANA) pin is for analogRead(), then there shouldn't be any pullups.

adam2872:
> > > This line: pinMode(PIN_USER_BTN_ANA, INPUT_PULLUP);
> > > should be: pinMode(PIN_USER_BTN_ANA, INPUT);
> > 
> > 
> > It needs the pullup or else the value read floats around and can lead to false triggers
> 
> If this (the _ANA) pin is for analogRead(), then there shouldn't be any pullups.

You would have thought so, but there is and it does work.

adam2872:
> If think this line: #ifdef PIN_USER_BTN || PIN_USER_BTN_ANA
> 
> should be: #if defined(PIN_USER_BTN) || defined(PIN_USER_BTN_ANA)

Commit created to reflect these changes

PR #290, opened 2025-05-14
Slightly better battery icon

adam2872:


446564:
Looks nice to me. 

The original icon: 

PXL_20250519_030237187

The updated icon:

PXL_20250519_030322737 NIGHT

PR #291, opened 2025-05-15
Build fixes for ProMicro and ProMicroLLCC68

oltaco:
Moved the INA3221 dependency up to the top level for ProMicro and added it to ProMicroLLCC68 as well.


ripplebiz:
This has been remedied in PR #287

PR #294, opened 2025-05-16
Sh1106 display

446564:
Adds support for monochrome 128x64 OLED display


446564:
let me know if there is somewhere to add documentation for things like includes or using this in a new variant. I thought of using comments but maybe a README is better in the directory?

PR #300, opened 2025-05-18
t-beam supreme: PMU and i2c fixes

cod3doomy:
Fixed i2c (Wire) init issue by defining pins in platformio 
Added an i2c scanning function for debug
Corrected the pmu power up sequence


ripplebiz:
Could you change the  Serial.println() to MESH_DEBUG_PRINTLN()

cod3doomy:
I moved the function into the ifdef so it won't run unless debugging. Will that work or do you still need me to remove the serial.print calls?

PR #305, opened 2025-05-19
Telemetry: Create sensor classes that can be shared across variants

ngavars:
The purpose of this PR is to move some of the sensor related code to /src/helpers/sensors so that the code can be re-used in different variants.

I also created the EnvironmentSensorManager that is a SensorManager implementation implementing currently available environment sensors. This is meant as a drop-in replacement to vanilla SensorManager for boards that currently do not have any sensors at all (or as a base class with basic sensor functionality).

### DIFF
--- a/src/helpers/sensors/EnvironmentSensorManager.h
+++ b/src/helpers/sensors/EnvironmentSensorManager.h
@@ -7,8 +7,6 @@
 #include <Adafruit_AHTX0.h>
 #include <Adafruit_BME280.h>
 
-#define NUM_SENSOR_SETTINGS 3
-
 #define TELEM_AHTX_ADDRESS      0x38      // AHT10, AHT20 temperature and humidity sensor I2C address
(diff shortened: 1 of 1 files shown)


ripplebiz:
Am ok about there being this shared class for collection of common sensors, but the 4 'wrapper' classes don't offer much. The methods map almost 1-to-1 to the underlying Adafruit ones, which are already nice and easy to use.

ngavars:
Ok. I'll remove the wrappers.

ngavars:
@ripplebiz I have updated the PR.

The following changes were made:
1. Removed library wrapper classes
2. Switched to Adafruit libraries
3. I removed ability to disable channels for INA3221, because it has no real-life use case. Adafruit library does not even have a function to disable a channel. I still left the check for enabled channels when the data is being read, because a channel can be disabled by the sensor itself due to some reasons (malfunction?).
4. Which means that there are no configurable settings at the moment for the sensors that we have.

Tested with all the sensors on Promicro.

PR #307, opened 2025-05-19
Telemetry: Implement environment sensors for Heltec V3 variant

ngavars:
This PR depends on #305 

This PR adds environment sensor support to Helltec V3. The same set of sensors that were available on Promicro until now.


ngavars:
Well, I guess I should not be creating two parallel PRs. This one looks a lot bigger than it actually is because of changes that went into #305

PR #308, opened 2025-05-19
t-beam supreme: display fix, BME add, user btn fix

cod3doomy:
-Fixed build issues after display recent refactor
-Added BME280 support and updated SensorManager.h to include this data 
-Fixed user button define and verified it turns the display on

### DIFF
--- a/src/helpers/SensorManager.h
+++ b/src/helpers/SensorManager.h
@@ -10,10 +10,10 @@
 
 class SensorManager {
 public:
-  double node_lat, node_lon;  // modify these, if you want to affect Advert location
+  double node_lat, node_lon, node_temp, node_hum, node_pres;  // modify these, if you want to affect Advert location
   double node_altitude;       // altitude in meters
 
-  SensorManager() { node_lat = 0; node_lon = 0; node_altitude = 0; }
+  SensorManager() { node_lat = 0; node_lon = 0; node_altitude = 0; node_temp = 0; node_hum = 0; node_pres = 0;}
   virtual bool begin() { return false; }
   virtual bool querySensors(uint8_t requester_permissions, CayenneLPP& telemetry) { return false; }
   virtual void loop() { }
--- a/variants/lilygo_tbeam_supreme_SX1262/target.cpp
+++ b/variants/lilygo_tbeam_supreme_SX1262/target.cpp
@@ -1,10 +1,16 @@
 #include <Arduino.h>
 #include "target.h"
 #include <helpers/sensors/MicroNMEALocationProvider.h>
+#include <Adafruit_BME280.h>
 
 TBeamS3SupremeBoard board;
 
+#ifdef DISPLAY_CLASS
+  DISPLAY_CLASS display;
+#endif
+
 bool pmuIntFlag;
+//#define SEALEVELPRESSURE_HPA (1013.25)
 
 #ifndef LORA_CR
   #define LORA_CR      5
@@ -23,6 +29,7 @@ ESP32RTCClock fallback_clock;
 AutoDiscoverRTCClock rtc_clock(fallback_clock);
 MicroNMEALocationProvider nmea = MicroNMEALocationProvider(Serial1);
 TbeamSupSensorManager sensors = TbeamSupSensorManager(nmea);
+Adafruit_BME280 bme;
 
 static void setPMUIntFlag(){
   pmuIntFlag = true;
@@ -46,7 +53,7 @@ void scanDevices(TwoWire *w)
             switch (addr) {
             case 0x77:
             case 0x76:
-                Serial.println("\tFound BMX280 Sensor");
+                Serial.println("\tFound BME280 Sensor");
                 deviceOnline |= BME280_ONLINE;
                 break;
             case 0x34:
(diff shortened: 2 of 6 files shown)


ripplebiz:
You shouldn't need to modify the SensorManager class.
Move the _temp, _hum, _pres data members to the TbeamSupSensorManager class.
(SensorManager only has the _lat, _lon members because it's something the advert code needs to access).

Also, I'd probably make the 'bme' object a member of TbeamSupSensorManager, for neater encapsulation.

cod3doomy:
Thanks for the advice. I cleaned up the code and reverted SensorManager

variants/lilygo_tbeam_supreme_SX1262/target.cpp
```diff
@@ -340,55 +368,104 @@ void TbeamSupSensorManager::sleep_gps() {
...
-int TbeamSupSensorManager::getNumSettings() const { return 1; }  // just one supported: "gps" (power switch)
+int TbeamSupSensorManager::getNumSettings() const {
+  return sensorNum; 
+}
 
 const char* TbeamSupSensorManager::getSettingName(int i) const {
-  return i == 0 ? "gps" : NULL;
+  switch(i){
+    case 0: return "gps";
+    case 1: return "bme280";
```
ripplebiz: The bme280 isn't able to be controlled via the app (eg. turned on/off), so shouldn't be exposed as a custom var

variants/lilygo_tbeam_supreme_SX1262/target.cpp
```diff
@@ -340,55 +368,104 @@ void TbeamSupSensorManager::sleep_gps() {
...
+    default: NULL;
+  }
 }
 
 const char* TbeamSupSensorManager::getSettingValue(int i) const {
-  if (i == 0) {
-    return gps_active ? "1" : "0";
+  switch(i){
+    case 0: return gps_active == true ? "1" : "0";
+    case 1: return bme_active == true ? "1" : "0";
```
ripplebiz: Same here, it shouldn't be a custom var

variants/lilygo_tbeam_supreme_SX1262/target.h
```diff
@@ -8,29 +8,46 @@
...
+#include <Adafruit_BME280.h>
 
 class TbeamSupSensorManager: public SensorManager {
     bool gps_active = false;
+    bool bme_active = false;
     LocationProvider * _nmea;
-  
+    Adafruit_BME280 bme;
+    double node_temp, node_hum, node_pres;
+    int sensorNum = 2;
```
ripplebiz: this should just be 1

variants/lilygo_tbeam_supreme_SX1262/target.h
```diff
@@ -48,7 +65,8 @@ enum {
   OSC32768_ONLINE      = _BV(13),
 };
 
-bool power_init();
+void scanDevices(TwoWire *w);
+static bool l76kProbe();
```
ripplebiz: Can get rid of these declarations... not needing visibility outside of the target.cpp module

variants/lilygo_tbeam_supreme_SX1262/target.cpp
```diff
@@ -340,55 +368,104 @@ void TbeamSupSensorManager::sleep_gps() {
...
+    gps_active = gps_alive;
+    MESH_DEBUG_PRINTLN("GPS init failed and GPS is not active");
+  return gps_alive;
 }
 
 bool TbeamSupSensorManager::querySensors(uint8_t requester_permissions, CayenneLPP& telemetry) {
   if (requester_permissions & TELEM_PERM_LOCATION) {   // does requester have permission?
     telemetry.addGPS(TELEM_CHANNEL_SELF, node_lat, node_lon, node_altitude);
   }
+  if (requester_permissions & TELEM_PERM_ENVIRONMENT) {   // does requester have permission?
```
ripplebiz: should also be a 
   && bme_active

variants/lilygo_tbeam_supreme_SX1262/target.cpp
```diff
@@ -340,55 +368,104 @@ void TbeamSupSensorManager::sleep_gps() {
...
       node_lat = ((double)_nmea->getLatitude())/1000000.;
       node_lon = ((double)_nmea->getLongitude())/1000000.;
       node_altitude = ((double)_nmea->getAltitude()) / 1000.0;
       //Serial.printf("lat %f lon %f\r\n", _lat, _lon);
     }
-    next_gps_update = millis() + 1000;
+
+    //read BME280 values
+    //node_alt = bme.readAltitude(SEALEVELPRESSURE_HPA);
+    node_temp = bme.readTemperature();
```
ripplebiz: add:
   if (bme_active) {

src/helpers/SensorManager.h
```diff
@@ -13,7 +13,7 @@ class SensorManager {
   double node_lat, node_lon;  // modify these, if you want to affect Advert location
   double node_altitude;       // altitude in meters
 
-  SensorManager() { node_lat = 0; node_lon = 0; node_altitude = 0; }
+  SensorManager() { node_lat = 0; node_lon = 0; node_altitude = 0;}
```
ripplebiz: it's minor I know, but shouldn't be diff here.
(tracking changes to the core/common files is important, and don't want lots of minor commits coming in blame trail when trying to find issues)

ripplebiz (changes requested):

PR #312, opened 2025-05-20
initial support for generic RTTTL (Ring Tone Transfer Language) notifier

seagull9000:
New PR for RTTTL notifier to replace https://github.com/ripplebiz/MeshCore/pull/272

Made against dev.


ripplebiz:
OK, will merge, but will refactor a few things.

PR #314, opened 2025-05-20
RTTTL on message types

seagull9000:
Add enums for message event types

Implemented on contactMessage with hard coded melody.

All other messages types suppressed.  This is just:
newContactMessage
channelMessage


ripplebiz:
The enum definition belongs in UITask.h
Also, would name it just something generic like UIEventType

ripplebiz:
Also, make sure you don't leave any Serial.print()s , or put them in MESH_DEBUG_PRINT's  if they are useful for debugging (and only optionally compiled in)

PR #318, opened 2025-05-21
ProMicro GPS support

oltaco:
Support for NMEA based GPS on ProMicro. Tested working with ublox neo-6m.

@ngavars Maybe you could take a look at this and make sure it doesn't break your other sensors? We should probably sort out the GetNumSettings function so that it reports correctly depending on what sensors were detected as well?


ngavars:
@oltaco I have updated my sensors PR - most of the sensor stuff has moved to EnvironmentSensorManager in helpers/sensors. This simplifies sensor manager implementation for each variant. Also - the sensors that we currently support do not expose any writable settings, so it simplifies things.
My PR has not yet been accepted.

I did not have time to try your code with GPS + sensors yet. It's next on my list.

oltaco:
No problems, your new PR looks great. I will close this PR and rebase GPS support from your PR.

PR #321, opened 2025-05-22
GPS support added to EnvironmentSensorClass

oltaco:
Added GPS support to EnvironmentSensorClass. Tested with ublox neo-6m and promicro.

@ngavars could you take a look and make sure it doesn't break the other sensors when you get a chance?

### DIFF
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -1,5 +1,8 @@
 #include "EnvironmentSensorManager.h"
 
+
+#include <helpers/sensors/LocationProvider.h>
+
 static Adafruit_AHTX0 AHTX0;
 static Adafruit_BME280 BME280;
 static Adafruit_INA3221 INA3221;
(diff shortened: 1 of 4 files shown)


ngavars:
I can confirm that GPS and sensors are working on Promicro.

ripplebiz:
We're going to need to be able to conditionally compile/support the various sensors, especially GPS, since there won't be defines for the PIN_GPS_* pins for a number of variants who pull this class in.

recommend something like this in the variant .ini's
   -D ENV_INCLUDE_GPS=1
   -D ENV_INCLUDE_INA3321=1
   -D ENV_INCLUDE_BME280=1

etc.

and have things like:

#if ENV_INCLUDE_BME280
  if (BME280.begin(TELEM_BME280_ADDRESS, &Wire)) {
      MESH_DEBUG_PRINTLN("Found BME280 at address: %02X", TELEM_BME280_ADDRESS);
      MESH_DEBUG_PRINTLN("BME sensor ID: %02X", BME280.sensorID);
      BME280_initialized = true;
  } else {
      BME280_initialized = false;
      MESH_DEBUG_PRINTLN("BME280 was not found at I2C address %02X", TELEM_BME280_ADDRESS);
  }
#endif

oltaco:
No problems, easy done 👍

oltaco:
~~Ok I've wrapped all the sensors in conditionals now, however it does still require the MicroNMEA dependency because it's beyond my abilities to refactor that out.~~

New commit coming soon.

oltaco:
Now with even more conditionals! It's ugly but it does work. I have tested with all of the ENV_INCLUDE* defines and their dependencies commented out, still compiles and telemetry returns battery level and lat/lon as manually set.

PR #322, opened 2025-05-22
rak3x72 : first commit

fdlamotte:
That's first commit for RAK3172, based on STM32WL platform

Seems to work fine, but I need to test range

updated wio-e5 to be on par (specific board class for both to get good vendor name, and RX_BOOSTED_GAIN toggle)


ripplebiz:
Good stuff

PR #326, opened 2025-05-22
Adding sensor classes support to Heltec v3

memo-567:
Adding sensor classes support to Heltec v3

Works with pin 17 and 18 only. Using 41 and 42 causes display problems.

### DIFF
--- a/variants/heltec_v3/platformio.ini
+++ b/variants/heltec_v3/platformio.ini
@@ -17,6 +17,10 @@ build_flags =
   -D SX126X_DIO3_TCXO_VOLTAGE=1.8
   -D SX126X_CURRENT_LIMIT=140
   -D SX126X_RX_BOOSTED_GAIN=1
+  -D ENV_INCLUDE_AHTX0=1
+  -D ENV_INCLUDE_BME280=1
+  -D ENV_INCLUDE_INA3221=1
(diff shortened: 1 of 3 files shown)


ngavars:
You can add libraries to Heltec_lora32_v3 config only. All others extend from it.

ghost:
> You can add libraries to Heltec_lora32_v3 config only. All others extend from it.

Ah okay - thanks  :)

ngavars:
This also applies to the source file filter, so you can specify   +<helpers/sensors>  only in the base configuration for Heltec V3.

ghost:
> This also applies to the source file filter, so you can specify +<helpers/sensors> only in the base configuration for Heltec V3.

Thanks again :)

ngavars:
With recent changes you need to add the following flags to pio ini:

  -D ENV_INCLUDE_AHTX0=1
  -D ENV_INCLUDE_BME280=1
  -D ENV_INCLUDE_INA3221=1
  -D ENV_INCLUDE_INA219=1

These define which sensors will be enabled for the variant. There's also -D ENV_INCLUDE_GPS=1 if you feel adventurous and want to enable GPS (but it needs RX/TX pins defined too).

ghost:
> With recent changes you need to add the following flags to pio ini:

Thanks for the tip :)

PR #329, opened 2025-05-23
Add sensor support to Xiao Nrf

ngavars:
This PR adds sensor support to Xiao Nrf. GPS is not enabled because it needs 3 additional pins.


ngavars:
Tested with Xiao NRF basic.

PR #332, opened 2025-05-26
RAK4631: Add GPS support

cod3doomy:
Added RAK12500 GPS support
Added socket scan capability to determine which socket the GPS is connected to

T-beam supreme: added conditional to displaying location data to only when gps is active


ripplebiz:
Would be good to have the GPS stuff conditionally compiled in with the same  -D ENV_INCLUDE_GPS=1  like the EnvironmentSensorManager class.

cod3doomy:
I'll work on integrating the EnvironmentSensorManager into this code and then update this PR.

cod3doomy:
So, to use the RAK12500 universally across the RAK base boards you have to use I2C. This is what I have in my code and it requires a U-Blox library from SparkFun. As of now the EnvironmentSensorManager would not work with the RAK12500 I2C GPS unless the library reference is added and a separate initBasicGPS() I2C function is created. I could do this, but I don't want the EnvironmentSensorManager to get messy and for now we could at least get the GPS functionality out for the RAK users in the meantime.

PR #340, opened 2025-05-27
ESP-NOW Heltec V3

memo-567:
Basic adaptation of the Generic_ESPNOW_* targets for ESP-NOW Heltec V3.
The display works so far. The sensors should work as well, but I don't have a way to test it at the moment.


ghost:
Display support ✅
Sensor support ✅
RTC support ✅

I still have to clean up and remove LoRa and BLE code that are not necessary for ESP-NOW.

PR #342, opened 2025-05-28
ThinkNode m1 GPS support

jquatier:
- Adds GPS support to ThinkNode M1

Tested with success on my device. No screenshot this time =)

These same changes can be made to the T-echo too, but I don't own one of those devices to test on. Happy to do a PR for that as long as someone else can test it.

### DIFF
--- a/variants/thinknode_m1/platformio.ini
+++ b/variants/thinknode_m1/platformio.ini
@@ -25,6 +25,9 @@ build_src_filter = ${nrf52840_thinknode_m1.build_src_filter}
   +<helpers/*.cpp>
   +<helpers/nrf52/ThinkNodeM1Board.cpp>
   +<../variants/thinknode_m1>
+lib_deps =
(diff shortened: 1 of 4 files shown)


jquatier:
Great feedback from @liamcottle, we need to account for the hardware GPS switch being off (which appears to kill power). I have a few ideas. Will finish this up tomorrow. Moving the PR to draft for now.

liamcottle:
Just tested with the new commit that handles the GPS switch.

This is working really nicely! Awesome work. I've tested turning the board on and off with different switch states while watching the live telemetry updates from another companion.

It seems to be correctly updating the GPS when the hardware switch is set to ON position, and no longer updates when the switch is set to OFF position.

Enabling and disabling GPS setting from inside the app is always working as intended, and when using the hardware switch I see the option update in the app when navigating back to the settings screen.

Unless you can think of anything else that needs doing, I think this is good! :)

Issue #333, opened 2025-05-26
Button class with callbacks proposal


seagull9000:
This is a proposal to add a button class to the UI to aid development of new features and to maintain consistency across devices.

It also helps map out a button strategy / user experience.

The class should also support multiple buttons (e.g T-beam) and possibly menu control for devices with a acreen.

This is a starter for feedback on the idea.

So far I can see the following can be used for the user button:

- Long-Long (LL) press (> 4s): Shutdown (or deep sleep where shutdown not supported)  - available for all firmwares
- Long (L1) press x1 (> 1.5s < 3s): 
    - If a message is displayed on screen, dismiss it (companion firmware)
    - If a menu item is displayed, sends OK (all firmwares)
- Short press x1 (S1) (<= 1s)
    - if screen is off, turn it on (all firmwares)
    - if device sleeping, wake it up (all firmwares)
    - If screen is on, and message is displayed, advance to next (companion firmware)
    - If menu item is displayed move to next menu item
- Short press x2 (S2) - double press
    - Send flood advert (all firmwares)
    - if menu item is displayed, toggle setting (all firmwares), then L1 to OK
- Short press x3 (S3) - triple press
    - Toggle repeat on/off (repeater/room server firmwares)

These are just suggestions and other combinations could potentially be used (e.g. SL (short-long).

A call back function can be registered for each type of press (LL, L1, S1, S2, S3, etc).

EDIT: Just saw the Thinknode M1 with rotary switch.  This could be supported in a button class too.

hank:
Perhaps we could put a new set of virtuals in mesh::MainBoard to implement these, then stub them out in ESPBoard and the other child classes with either empty implementations or some standard behavior. That way we can just override them in the board classes. For the TBeam Supreme at least, this is implemented with XPowersLib at least in part:

PMU.setPowerKeyPressOffTime(XPOWERS_POWEROFF_4S);

The other parts seem to be handled by UITask::buttonHandler. Perhaps the buttonHandler could call the callbacks.

The only thing I would worry about is the consistency of the UI across devices and keeping that unified through changes.

seagull9000:
See jquatier's PR for potential implementation ... https://github.com/ripplebiz/MeshCore/pull/343

seagull9000:
Will close this as the above PR has implimented a button class.

PR #226, opened 2025-04-25
feat(ui): implement shutdown functionality with enhanced button handling

rightup:
- Added long-press shutdown logic to button handler
- Introduced display countdown and final prompt messages
- Added build flag `UI_CAN_SHUTDOWN` to enable shutdown feature
- Modularised shutdown logic into helper methods (`performShutdown`, `showCountdown`, etc.)

fix(ui): correct variable type for time tracking

- Changed `cur_time` from `int` to `unsigned long` to properly store `millis()` value

PR #343, opened 2025-05-29
Improved Button Management

jquatier:
Tossing this up for comments mostly, I am perfectly ok if this isn't merged right away or even at all, but wanted to get at least some eyes on it. I've seen some comments on wanting to refactor UITask too, so if its better to hold off on this, no worries!

The button handler in UITask was getting a little crazy with all the logic, and would get more crazy if we keep adding to it. This PR encapsulates the button into its own class that can fire multiple and different events for the UI to "listen" to. 

For now I've added:

- **Any Press**: turn on display / update it (this ends up firing before the events below too)
- **Single Press**: dismiss messages if there are any (can be used for pagination later)
- **Double Press**: nothing. (idea: send advert?)
- **Triple Press**: put buzzer in quiet mode. (Is this better as a nodepref so it can be persisted?)
- **Long Press**: call existing `shutdown()`

No new functionality was added, aside from the buzzer quiet mode. I like the cleanness of the actions being in different functions now. I also like this still being tied with UITask so that we can make sure all boards act the same. We can still implement secondary buttons for boards that have those, but at least we will know they all behave the same. 

Curious what folks think of this, again just tossing the idea up for now. Tested on Heltec V3, T114, T1000E, and ThinkNode M1.

#333 #226 are related

### DIFF
--- a/examples/companion_radio/Button.cpp
+++ b/examples/companion_radio/Button.cpp
@@ -60,10 +60,10 @@ void Button::update() {
         _state = IDLE;
     }
     
-    // Handle long press
+    // Handle long press while button is held
     if (_state == PRESSED && (now - _pressTime) > BUTTON_LONG_PRESS_TIME_MS) {
         triggerEvent(LONG_PRESS);
-        _state = IDLE;  // Prevent multiple long press events
+        _state = IDLE;  // Prevent multiple press events
         _clickCount = 0;
     }
 }
@@ -97,7 +97,7 @@ void Button::handleStateChange() {
             } else {
                 // Long press already handled in update()
                 _state = IDLE;
-                _clickCount = 0;  // Reset click count after long press
+                _clickCount = 0;
             }
         }
     }
--- a/examples/companion_radio/Button.cpp
+++ b/examples/companion_radio/Button.cpp
@@ -13,9 +13,6 @@ Button::Button(uint8_t pin, bool activeState, bool isAnalog, uint16_t analogThre
 }
 
 void Button::begin() {
-    if (!_isAnalog) {
-        pinMode(_pin, INPUT_PULLUP);
-    }
     _currentState = readButton();
     _lastState = _currentState;
 }
--- a/examples/companion_radio/Button.cpp
+++ b/examples/companion_radio/Button.cpp
@@ -1,0 +1,127 @@
+#include "Button.h"
+
+Button::Button(uint8_t pin, bool activeState) 
+    : _pin(pin), _activeState(activeState), _isAnalog(false), _analogThreshold(20) {
+    _currentState = !_activeState;
+    _lastState = _currentState;
+}
+
+Button::Button(uint8_t pin, bool activeState, bool isAnalog, uint16_t analogThreshold)
+    : _pin(pin), _activeState(activeState), _isAnalog(isAnalog), _analogThreshold(analogThreshold) {
+    _currentState = !_activeState;
+    _lastState = _currentState;
+}
+
+void Button::begin() {
+    if (!_isAnalog) {
+        pinMode(_pin, INPUT_PULLUP);
+    }
+    _currentState = readButton();
+    _lastState = _currentState;
+}
+
+void Button::update() {
+    uint32_t now = millis();
+    
+    // Read button at specified interval
+    if (now - _lastReadTime < BUTTON_READ_INTERVAL_MS) {
+        return;
+    }
+    _lastReadTime = now;
+    
+    bool newState = readButton();
+    
+    // Check if state has changed
+    if (newState != _lastState) {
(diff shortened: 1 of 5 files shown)


seagull9000:
I think you can distinguish between Long press and Long-Long press.  The latter would be used for shutdown/deep sleep depenidning on the architecture.    The former (Long) can be used for an 'OK/execute'  (see https://github.com/ripplebiz/MeshCore/issues/333).

Also, Long-Long should not wait for button to be lifted.  If its down for more than 4-5 seconds, then execute the curresponding call back for Long-Long.  The current button handler doesn't do this and waits for the button to be lifted before it will play the shutdown melody.  By using a 'timeout', the user knows they have pressed long enough instead of guessting.

Perhaps also allow a mechanism for assigning callbacks by passing the button sequence and function as a method.  E.g.
   addCallback(L1,myLongPressFunction);  // Where L1 is an enum of the press type

hank:
Double press for advert is great. I can try this on a device or two soon.

hank:
Had to add
```  +<../examples/companion_radio/Button.cpp>```
To my platformio.ini to get the t114 companion to compile without a linker error.

debug messages seem to work great! Now just to add an advert on doubleclick.

seagull9000:
Should the button class be across all devices rather than just companion radio?  That way it can be used with repeater etc.  Using callbacks bound to enumerated press types is then what you would use in the individual firmwares.  I thik it keeps the code cleaner and avoids having to update multile firmwares when changing the button logic.

jquatier:
> avoids having to update multile firmwares when changing the button logic.

Like I mentioned, There’s been some convos about refactoring UI Task and I think 1 goal would be sharing that across all, which would include buttons. The UITasks are mostly duplicate code already with minor differences. The companion one gets touched most often since the others don’t generally need all the same features.

jquatier:
Going to work on this some more tonight. I’ll add the timeout in for long hold, it’s a good idea to make the experience smoother.

jquatier:
> Had to add ` +<../examples/companion_radio/Button.cpp>` To my platformio.ini to get the t114 companion to compile without a linker error.
> 
> debug messages seem to work great! Now just to add an advert on doubleclick.

Looks like the t114 platformio doesn’t include the whole companion folder like others do (see Heltec v3, m1, etc.) easy fix though thanks for the callout! Might end up moving that file anyway.

hank:
Just finished a big refactor of the companion folder here

[companion-mymesh-refactor](https://github.com/hank/MeshCore/compare/companion-mymesh-refactor)

02b94a94c2c9fbd3ad78269825be57e9392bede1

Pull that in - it extracts MyMesh into a class and header, externs the proper things, and has your button code send an advert, which seems successful on the T114. I also cleaned up all the companion ini files so they should all build!

hank:
ok well, I got the debug print that it sent the advert, but it doesn't seem like it's actually sending even though I believe I'm doing what main.cpp does to send it. So that needs a look. But you can call anything in the MyMesh class through the global `the_mesh`.

jquatier:
Few fixes for issues I noticed:
- refreshing an e-paper display on any press was causing missed clicks, this moves the refresh to the other actions. any press is still used to wake the display quickly
- t114 seems to be the only variant with the UI code included that needed the build src filter updated after all

Timeout for a long press works well, great for the shutdown action. Buzzer silence works well too. I'd like to add some feedback for that one, so I'll add that to my list, either for this PR or the next one.

Decided to not change any current behavior in this PR for what clicks do, just adding a buzzer silence option on triple click. Goal is to keep the PR smaller since this already introduces enough new stuff. Hoping to get @ripplebiz eyes on it before I go too much further and we can chat with him on the self-advert stuff.

seagull9000:
That cleans things up in UITask nicely with those callbacks - looking good to me.

If you want me to test with the T1000e, are you able to post a zip on discord?

hank:
Cool I can re-do the cleanup I did as a separate pull request once this one is merged. That way we get the button behavior in, then fix up main.cpp and do the restructuring afterward as a separate action.

hank:
This is built from c445bbeaf250792cbdce943c9f76afb08d3da32a

[t1000e_companion_radio_ble-1.6.3.1-c445bbe-all.zip](https://github.com/user-attachments/files/20531803/t1000e_companion_radio_ble-1.6.3.1-c445bbe-all.zip)

ripplebiz:
Good stuff.

LitBomb:
Please take a look at issue 414 

https://github.com/ripplebiz/MeshCore/issues/414

Rak 19007 user button on AIN1 Worked on 1.6.2 to turn the screen back on.  On 1.7.0 the button doen't do anything.  

In contrast, rak19026's built in user button works on both 1.6.2 and 1.7.0.

PR #344, opened 2025-05-29
Add T-Echo GPS

jquatier:
This is untested since I don't have the device, but I said I would toss something up. I'm 80% sure this will work, and if it doesn't its likely the pins that are off. Hopefully someone else can help me test it.


jquatier:
Got a successful test in discord. Seems to work well.

PR #348, opened 2025-05-31
gps : sync time on fix

fdlamotte:
Some changes in LocationProvider

* will sync time at first fix (_time_sync_needed variable)
* syncTime method will force _time_sync_needed to true
* "sync" custom variable on T1000 let the user force update

Validated change in companion radio to use rtc_clock defined in target (was still creating a new rtc_clock).

### DIFF
--- a/src/helpers/sensors/MicroNMEALocationProvider.h
+++ b/src/helpers/sensors/MicroNMEALocationProvider.h
@@ -19,13 +19,16 @@
 class MicroNMEALocationProvider : public LocationProvider {
     char _nmeaBuffer[100];
     MicroNMEA nmea;
+    mesh::RTCClock* _clock;
     Stream* _gps_serial;
     int _pin_reset;
     int _pin_en;
+    long next_check = 0;
+    long time_valid = 0;
 
 public :
-    MicroNMEALocationProvider(Stream& ser, int pin_reset = GPS_RESET, int pin_en = GPS_EN) :
-    _gps_serial(&ser), nmea(_nmeaBuffer, sizeof(_nmeaBuffer)), _pin_reset(pin_reset), _pin_en(pin_en) {
+    MicroNMEALocationProvider(Stream& ser, mesh::RTCClock* clock = &rtc_clock, int pin_reset = GPS_RESET, int pin_en = GPS_EN) :
+    _gps_serial(&ser), nmea(_nmeaBuffer, sizeof(_nmeaBuffer)), _pin_reset(pin_reset), _pin_en(pin_en), _clock(clock) {
         if (_pin_reset != -1) {
             pinMode(_pin_reset, OUTPUT);
             digitalWrite(_pin_reset, GPS_RESET_FORCE);
@@ -59,6 +62,7 @@ public :
         }        
     }
 
+    void syncTime() override { nmea.clear(); LocationProvider::syncTime(); }
     long getLatitude() override { return nmea.getLatitude(); }
     long getLongitude() override { return nmea.getLongitude(); }
     long getAltitude() override { 
@@ -66,6 +70,7 @@ public :
         nmea.getAltitude(alt);
         return alt;
     }
+    long satellitesCount() override { return nmea.getNumSatellites(); }
     bool isValid() override { return nmea.isValid(); }
 
     long getTimestamp() override { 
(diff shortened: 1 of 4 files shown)


variants/t1000-e/target.cpp
```diff
@@ -179,14 +179,26 @@ void T1000SensorManager::loop() {
...
+int T1000SensorManager::getNumSettings() const { return 2; }  // just one supported: "gps" (power switch)
 
 const char* T1000SensorManager::getSettingName(int i) const {
-  return i == 0 ? "gps" : NULL;
+  switch (i) {
+    case 0:
+      return "gps";
+      break;
+    case 1:
+      return "sync";
```
ripplebiz: The "sync" custom var doesn't make much sense. These custom vars are just a kind of bridge between the mobile app and the companion radios, for additional features that need some control.
The app already syncs the companion clock from the handset's time, so I don't know what this var is for(?)

fdlamotte: This was just a cmd var, to trigger a sync from the gps (value returns to 0 when synced)

I agree it does not make must sense for a node connected to a mobile phone (as gps in the node in that same case).

The use case for this would more be on an SBC that might even not have access to the internet to sync time

src/helpers/sensors/MicroNMEALocationProvider.h
```diff
@@ -59,27 +62,49 @@ public :
...
             nmea.process(c);
         }
+
+        if (!isValid()) time_valid = 0;
+
+        if (millis() > next_check) {
+            next_check = millis() + 1000;
+            if (_time_sync_needed && time_valid > 2) {
+                if (_clock != NULL) {
+                    rtc_clock.setCurrentTime(getTimestamp());
```
ripplebiz: this should be _clock->setCurrentTime( )

variants/t1000-e/target.cpp
ripplebiz: I still don't really follow what the scenario would be where/how triggering the syncTime() would be?

fdlamotte: I thinks it mostly gives the possibility of doing it, because rtcs tend to drift and some more than others (my rak3372 repeater is now 3h behind). So I thought if someone finds its clock is too far from real time, it could trigger a sync. 

I admit I never did it past the first sync ... even on the custom firmware for t1000 where it was possible. Or only once yesterday, because there was a glitch and date got set to 2080 ! the gps must have had reported the fix, while the time had not yet been sent ...

Maybe it would be better to have gps time (as timestamp) in the telemetry ? There is an lcc type for that. Then people could sync time on their device from node's gps and then sync node time ...

liamcottle: My opinion is the custom vars are more intended for custom board specific settings, not so much to be used as custom commands.

I believe these will need to be persisted to flash in a future firmware update, as things like GPS being auto turned on at boot would be really useful.

However something like a "sync time from gps right now" should probably be a `CMD_SYNC_TIME_FROM_GPS` since this isn't really a setting that can be toggled, nor is it something that can be persisted :)

I guess it could be useful for debugging, as well as if you had a companion node that was controlled via serial only, or some other program, and not used with the app (since that auto syncs the time upon connect).

src/helpers/sensors/MicroNMEALocationProvider.h
```diff
@@ -59,27 +62,49 @@ public :
...
             nmea.process(c);
         }
+
+        if (!isValid()) time_valid = 0;
+
+        if (millis() > next_check) {
+            next_check = millis() + 1000;
+            if (_time_sync_needed && time_valid > 2) {
+                if (_clock != NULL) {
+                    _clock.setCurrentTime(getTimestamp());
```
ripplebiz: should be _clock->

variants/t1000-e/target.cpp
ripplebiz: Yeah, I'd prefer there to be a command of some sort.
Rest of the PR is fine, so if you can just remove the "sync" custom var, should be fine.

ripplebiz:
NOTE: I've changed the MicroNMEA... constructor so the clock param is NULL by default.
This way, all the existing targets will (hopefully) behave the same. If each target wants to enable this feature, they should first test, then explictly pass &rtc_clock like the T1000e target is doing.

fdlamotte:
That's how I had made it at first ;) 
Seems totally fine, and can be toggled only for repeaters for instance where it probably makes more sense ;)

PR #351, opened 2025-05-31
wio_e5 : bme280 support

fdlamotte:
Some sensors for wio-e5 ... that's necessary for a sensor node ;)

Could not use Adafruit lib ... some strange things with spi defs, while I don't use spi :(

But found an alternative

### DIFF
--- a/variants/wio-e5-mini/target.h
+++ b/variants/wio-e5-mini/target.h
@@ -25,11 +25,11 @@ public:
 };
 
 class WIOE5SensorManager : public SensorManager {
-    BME280I2C* _bme;
+    BME280I2C bme;
     bool has_bme = false;   
 
 public:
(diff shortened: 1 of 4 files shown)


variants/wio-e5-mini/target.h
```diff
@@ -21,10 +24,20 @@ class WIOE5Board : public STM32Board {
     }
 };
 
+class WIOE5SensorManager : public SensorManager {
+    BME280I2C* _bme;
```
ripplebiz: Prob simpler to just have the bme instance a member, rather than passed in via constructor. ie. it's wholly encapsulated by the sensorManager class

fdlamotte: right, done

I also did a slight change to stm32 internal FS to force formatting from debugger

PR #353, opened 2025-06-01
MyMesh Refactor and Advert Button

hank:
This branch does a few things:

* Pulls the companion MyMesh class into a header and CPP file and makes the right externed globals for the Button, MyMesh, main and UITask files to actually integrate with each other
* Adds a double press action to button behavior on all devices that sends an advert
* Fixes all the companion ini files to include all files in the companion directory so there shouldn't be linker errors
* Ports a self advert func from repeater since that was missing in the MyMesh for companion

FIRMWARE_VERSION ends up in MyMesh.h, which can be moved if desired. main.cpp just isn't a great place for it. This brings up a potential future task of cleaning up the MyMesh classes in the different example codebases to make a proper class structure that would support Repeater, Room Server and Companion (and other) implementations.

I just tested the doublepress advert on a tbeam sx1276 and it works great, shows up immediately on another node's discover list.

```
DEBUG: UITask: any press triggered
DEBUG: UITask: any press triggered
DEBUG: UITask: double press triggered, sending advert
DEBUG: Advert sent!
```


446564:
I'll test this on the g2 nano tonight

examples/companion_radio/UITask.h
```diff
@@ -1,4 +1,5 @@
-#pragma once
+#ifndef UI_TASK_H
+#define UI_TASK_H
```
ripplebiz: '#pragma once'  does this, and is very widely supported now.

variants/t114/platformio.ini
```diff
@@ -75,7 +75,7 @@ build_flags =
 ;  -D ENABLE_PRIVATE_KEY_IMPORT=1
 ;  -D ENABLE_PRIVATE_KEY_EXPORT=1
 ;  -D MESH_PACKET_LOGGING=1
-;  -D MESH_DEBUG=1
+  -D MESH_DEBUG=1
```
ripplebiz: try not to leave these uncommented/enabled when committing

examples/companion_radio/MyMesh.h
```diff
@@ -0,0 +1,283 @@
...
+
+#define SEND_TIMEOUT_BASE_MILLIS          500
+#define FLOOD_SEND_TIMEOUT_FACTOR         16.0f
+#define DIRECT_SEND_PERHOP_FACTOR         6.0f
+#define DIRECT_SEND_PERHOP_EXTRA_MILLIS   250
+#define LAZY_CONTACTS_WRITE_DELAY        5000
+
+#define  PUBLIC_GROUP_PSK  "izOH6cXN6mrJ5e26oRXNcg=="
+
+#define CMD_APP_START              1
```
ripplebiz: the CMD_ and RESP_ codes don't need to be known by other modules, so can be moved to the .cpp file

ripplebiz:
The project doesn't have a linter or code format def yet, but it's customary to use the predominant brace/indenting style of the repo.

hank:
* pragma onced
* Moved CMD/RESP macros as requested
* Added clang-format to the repo which vscode will use to format code, and set it up to match what I saw in the repo with the braces and spacing. Tweak to your heart's content!
* Reformatted all my changes
* Removed debug macro from ini
* Merged in current dev, which required porting a change from main.cpp into MyMesh.cpp because of refactor.

Please review! Thanks!

ripplebiz:
Thanks for the changes.
There's a lot at stake with changes like this. It's very hard to tell if you've made a mistake somewhere with this refactor, and it has the potential to break ALL companion firmwares. Still, the MyMesh class could really do with being moved out of main, so I'm trying to really weigh this one up properly.

Still, I'd like as many unnecessary changes as possible to be reverted. Like the simple_secure_chat/main.cpp. Doesn't look like there are any _actual_ changes there. Same with the Button.* files. If you could revert these files please. (I know, the author of Button didn't follow code formatting, but I'd rather leave as-is, and stick to the 'no unnecessary diffs pls' principle.

Also, with the clang formatter, there are still some differences, like else and else if being on same line. And function opening brace on same line.

hank:
* Reverted changes to Button files as requested
* Reverted changes to simple_secure_chat
* Switched AllowShortIfStatementsOnASingleLine to true to clang format rules
* Switched BreakBeforeBraces to Attach to make all opening braces occur on the same line
* Reformatted MyMesh.cpp as an example of clang-format changes. Can update more if needed.
* Did my best to minimize changes to the repo
* Built many companion firmwares and tested one to ensure things aren't completely broken, more testing recommended.

Issue #354, opened 2025-06-01
[Feature request]: Companion radio remote DFU trigger


seagull9000:
I know it sounds counterintuitive to have this feature for a companion node, however there will be times when a convert from meshtastic flashes their rooftop node with the companion radio firmware.  

It's difficult to foresee that this will  inevitably result in a climb to the roof with laptop in tow when it comes time to make that rooftop node a repeater node.

I'd like to say it's a trap for young players, but it can happen to someone more experienced who is encouraging young players to try meshcore :/  

Suggest feature be built into the Meshcore App.

This could possibly be addressed in documentation, but reading documentation is usually reserved as a last resort activity and by then the damage is done.

EDIT:  from Liam:
App would need a new CMD_START_OTA command code, and it would need to reply with the same response that includes the Mac address etc
with a RESP_CODE_OTA_STARTED

lucidnx:
I want this too. My node is fixed in backpack on powerbank and it's kinda annoying if I know it can be done without touching it . Work smart, not hard (lazy people have best ideas)

mannkind:
+1 to remote updating companion nodes; the convenience factor is huge and possibly the _only_ thing I miss from that other mesh network.

Even if the node is in my living room, I want to be able to start a remote DFU without physically interacting with it.

liamcottle:
Duplicate of #160. I do plan to get this supported by the firmware.

PR #357, opened 2025-06-01
faq.md: update OTA firmware instructions

LitBomb:
added ESP32 OTA firmware update instructions
added nRF OTA firmware update instructions to use the new nRF DFU app on android and iOS

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -65,7 +65,8 @@ author: https://github.com/LitBomb<!-- omit from toc -->
   - [6.4. Q: I can't connect via Bluetooth, what is the Bluetooth pairing code?](#64-q-i-cant-connect-via-bluetooth-what-is-the-bluetooth-pairing-code)
   - [6.5. Q: My Heltec V3 keeps disconnecting from my smartphone.  It can't hold a solid Bluetooth connection.](#65-q-my-heltec-v3-keeps-disconnecting-from-my-smartphone--it-cant-hold-a-solid-bluetooth-connection)
 - [7. Other Questions:](#7-other-questions)
-  - [7.1. Q: How to  Update repeater and room server firmware over the air?](#71-q-how-to--update-repeater-and-room-server-firmware-over-the-air)
+  - [7.2 Q: How to update ESP32-based devices over the air?](#72-q-how-to-update-esp32-based-devices-over-the-air)
+  - [7.1 Q: How to update nRF (RAK, T114, Seed XIAO) repeater and room server firmware over the air using the new simpler DFU app?](#71-q-how-to-update-nrf-rak-t114-seed-xiao-repeater-and-room-server-firmware-over-the-air-using-the-new-simpler-dfu-app)
 
 ## 1. Introduction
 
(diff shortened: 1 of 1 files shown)


seagull9000:
This isn't currently anywhere, but is there a spot to mention the issue this feature request is trying to solve?  It's related to the firmware flashing but by the time the new entrant gets to this its too late...

https://github.com/ripplebiz/MeshCore/issues/354

People coming from meshtastic could easily be caught out thinking the companion node will repeat.

It's a bit of a gotcha moment when you realise.

LitBomb:
> This isn't currently anywhere, but is there a spot to mention the issue this feature request is trying to solve? It's related to the firmware flashing but by the time the new entrant gets to this its too late...
> 
> #354
> 
> People coming from meshtastic could easily be caught out thinking the companion node will repeat.
> 
> It's a bit of a gotcha moment when you realise.

see 5.2 in FAQ: https://github.com/ripplebiz/MeshCore/blob/main/docs/faq.md#52-q-do-meshcore-clients-repeat

This seems like an edge case: installing companion firmware on a node thinking it would repeat and then put it up on some high places that is hard to reach and hard to take down.

If you have only one device, you should install companion.  If you have more than one device and want to setup one device as a repeater, and somehow without first knowing what a repeater does or doesn't do in meshcore seems like an edge case.  When you install firmware to a device for the very first time, I would imagine you would test it for a bunch to times to check that it works.  

I can't think of a case that a first time meshcore user installing firmware to a device for the first time, install companion firmware on to a node, and them set it up at some very hard to reach location, and then test it and then realize "oh it doesn't work, I should have installed the repeater firmware instead, even though this is my one an only device".

If I misunderstood you, I apologize, and please let me know what user scenario you are thinking about.  Thanks.

jon-reid-nz:
> > This isn't currently anywhere, but is there a spot to mention the issue this feature request is trying to solve? It's related to the firmware flashing but by the time the new entrant gets to this its too late...
> > #354
> > People coming from meshtastic could easily be caught out thinking the companion node will repeat.
> > It's a bit of a gotcha moment when you realise.
> 
> see 5.2 in FAQ: https://github.com/ripplebiz/MeshCore/blob/main/docs/faq.md#52-q-do-meshcore-clients-repeat
> 
> This seems like an edge case: installing companion firmware on a node thinking it would repeat and then put it up on some high places that is hard to reach and hard to take down.
> 
> If you have only one device, you should install companion. If you have more than one device and want to setup one device as a repeater, and somehow without first knowing what a repeater does or doesn't do in meshcore seems like an edge case. When you install firmware to a device for the very first time, I would imagine you would test it for a bunch to times to check that it works.
> 
> I can't think of a case that a first time meshcore user installing firmware to a device for the first time, install companion firmware on to a node, and them set it up at some very hard to reach location, and then test it and then realize "oh it doesn't work, I should have installed the repeater firmware instead, even though this is my one an only device".
> 
> If I misunderstood you, I apologize, and please let me know what user scenario you are thinking about. Thanks.

I did this very thing, put companion FW on my roof node using OTA - not realising that I would lose the ability to OTA - I guess I should have done more reading before flashing !

All that is required is a small warning on the flasher page when downloading the ZIP file for companion radio USB update is required subsequently.

PR #359, opened 2025-06-02
Scan devices in board

hank:
Moving scanDevices (renamed to scanI2CDevices for clarity) up into the ESP32Board level. Considered higher but figured I'd leave that for another day - we can always implement it just for ESP32 boards for now and then move it up as a virtual later. This will allow us to deduplicate code between the tbeam and tbeamsupreme which is nice.

In addition, ran into a major compilation issue in the latest HEAD of dev with ESPNow. Fixed it by adding an ESPNow INI define and adding preprocessor guards around that code so it's not compiled into tbeam code for no reason. Even when I made it compile through code changes, it overran the flash size.

### DIFF
--- a/src/helpers/ESP32Board.cpp
+++ b/src/helpers/ESP32Board.cpp
@@ -43,4 +43,64 @@ bool ESP32Board::startOTAUpdate(const char* id, char reply[]) {
 }
 #endif
 
+uint8_t ESP32Board::scanI2CDevices(TwoWire *w)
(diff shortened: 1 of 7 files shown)


liamcottle:
Hello, thanks for the contribution. It looks like this hasn't had any traction since it was opened 6 months ago. A lot of changes have been made to the code base since then, so I'll close this one out for now :)

PR #367, opened 2025-06-04
add basic alert system to companion UI

446564:
Adds `_alert[80]` which can be set along with `_needs_refresh` to trigger a simple on screen alert that lasts 1s at this time.

Implements POC with double press to advert action

### DIFF
--- a/examples/companion_radio/UITask.cpp
+++ b/examples/companion_radio/UITask.cpp
@@ -162,7 +162,16 @@ void UITask::renderCurrScreen() {
   if (_display == NULL) return;  // assert() ??
 
   char tmp[80];
-  if (_origin[0] && _msg[0]) { // message preview
+  if (_alert[0]) {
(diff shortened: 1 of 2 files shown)


446564:
Let me know if this is unwanted or I'm doing something odd that needs adjustment.

I also added the buzzer ack sound when an advert is sent via double press for some haptics.

liamcottle:
Nice and simple, I like it. Could probably add an alert for the triple press buzzer toggle as well. To show "Buzzer: on" and "Buzzer: off".

446564:
> Nice and simple, I like it. Could probably add an alert for the triple press buzzer toggle as well. To show "Buzzer: on" and "Buzzer: off".

Ya that would be nice. I can add that maybe later tonight or tomorrow the latest. Or in a little follow up. Whatever is easier.

PR #378, opened 2025-06-07
add mising config for `openocd_target`

recrof:


446564:
I'll add it to my list to test

PR #379, opened 2025-06-07
Minor companion ui fixes

jquatier:
1. If the boot took longer than a few seconds we would completely miss the boot screen. This isn't too critical but it's nice to see the current version number and logo. using the `ui_started_at` timestamp to make sure we always show it no matter how long boot took. This happened a lot on e-paper displays too due to the display taking some extra time to fire up.
2. Display wouldnt refresh after being turned on with a single button press. Needed an extra `else` to make sure the render happens after display turns on. This was causing a blank screen occasionally due to the `renderCurrScreen()` not being called in that scenario.
3. Added an alert message for quiet mode on/off
4. Fixed alignment of alert text

Tested on T114, Heltec V3, and ThinkNode M1

### DIFF
--- a/examples/companion_radio/UITask.cpp
+++ b/examples/companion_radio/UITask.cpp
@@ -163,9 +163,9 @@ void UITask::renderCurrScreen() {
 
   char tmp[80];
   if (_alert[0]) {
+    _display->setTextSize(1.4);
     uint16_t textWidth = _display->getTextWidth(_alert);
     _display->setCursor((_display->width() - textWidth) / 2, 22);
-    _display->setTextSize(1.4);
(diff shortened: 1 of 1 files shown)


liamcottle:
Awesome, looks good to me! Just a couple of notes:

- Can you please change the alert text to `Buzzer: ON` and `Buzzer: OFF`, or if you prefer `Sound: ON` / `Sound: OFF`.
- Can you please remove the buzzer sound that is played when buzzer is put into "off" mode.

jquatier:
Yeah thats way better 👍 should be all set!

liamcottle:
Great, thanks! Should be good to get merged in now :) @ripplebiz

PR #395, opened 2025-06-14
Native port

ajvpot:


dreirund:
Ahoj,

can we habe a comment on why this is closed, what is missing, ...?

Regards!

PR #399, opened 2025-06-17
print BLE pairing pin to serial

recrof:
Users that use boards with damaged or no display on targets that don't have display detection(or not possible) need a way to read bluetooth dynamic pin.


ripplebiz:
This will mess up the _USB companions.
Maybe make this a thing in the CLI Rescue mode?
Could print the current pin as part of the initial 'welcome' output?

recrof:
> This will mess up the _USB companions. Maybe make this a thing in the CLI Rescue mode? Could print the current pin as part of the initial 'welcome' output?

I tought about that - USB companions don't use this function at all.

PR #403, opened 2025-06-17
add board battery percentage

446564:
defaults to lipo with no changes to existing variants

- adds two chemistry types to start BAT_LIPO and BATT_LIFEPO
- adds getBattPercent to MainBoard
- adds getBattType to MainBoard
- companion UI now uses getBattPercent in renderBatteryIndicator

### DIFF
--- a/src/MeshCore.h
+++ b/src/MeshCore.h
@@ -33,9 +33,30 @@ namespace mesh {
 #define  BD_STARTUP_NORMAL     0  // getStartupReason() codes
 #define  BD_STARTUP_RX_PACKET  1
 
+#define BATT_LIPO     0
+#define BATT_LIFEPO   1
+
 class MainBoard {
 public:
   virtual uint16_t getBattMilliVolts() = 0;
+  uint8_t getBattPercent() {
+    uint32_t mv = getBattMilliVolts();
+    // Convert millivolts to percentage
+    int minMilliVolts;
+    int maxMilliVolts;
+    switch (getBattType()) {
+    case BATT_LIPO:
+      minMilliVolts = 3000;
+      maxMilliVolts = 4200;
+      break;
+    case BATT_LIFEPO:
+      minMilliVolts = 2500;
+      maxMilliVolts = 3400;
+      break;
+    }
+    return ((mv - minMilliVolts) * 100) / (maxMilliVolts - minMilliVolts);
+  }
+  uint8_t getBattType() { return BATT_LIPO; };
   virtual const char* getManufacturerName() const = 0;
   virtual void onBeforeTransmit() { }
   virtual void onAfterTransmit() { }
--- a/examples/companion_radio/UITask.cpp
+++ b/examples/companion_radio/UITask.cpp
@@ -132,11 +132,7 @@ void UITask::newMsg(uint8_t path_len, const char* from_name, const char* text, i
   }
 }
 
-void UITask::renderBatteryIndicator(uint16_t batteryMilliVolts) {
-  // Convert millivolts to percentage
-  const int minMilliVolts = 3000; // Minimum voltage (e.g., 3.0V)
-  const int maxMilliVolts = 4200; // Maximum voltage (e.g., 4.2V)
-  int batteryPercentage = ((batteryMilliVolts - minMilliVolts) * 100) / (maxMilliVolts - minMilliVolts);
+void UITask::renderBatteryIndicator(uint16_t batteryPercentage) {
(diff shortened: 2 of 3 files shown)


jbrazio:
How about the mobile apps ? This is where most people interact with the devices and the batt% is calculated by the app itself based on the telemetry.

Today all batt% are wrong because they are not reflecting the non-linear nature of LiPo when trying to translate voltage to "capacity".

I would suggest to update your implementation of `getBattPercent()` to better estimate the "capacity" of a LiPo.
Then it would be great for the mobile apps to start using this instead. :-P

src/MeshCore.h
```diff
@@ -33,9 +33,30 @@ namespace mesh {
...
 public:
   virtual uint16_t getBattMilliVolts() = 0;
+  uint8_t getBattPercent() {
+    uint32_t mv = getBattMilliVolts();
+    // Convert millivolts to percentage
+    int minMilliVolts;
+    int maxMilliVolts;
+    switch (getBattType()) {
+    case BATT_LIPO:
+      minMilliVolts = 3000;
```
jbrazio: The cut off value for a LiPo around 3.2V minimises wear and should be considered the 0%.

446564: the firmware currently just goes with 3V so I kept that, but lots of room for improvement with battery readings here :smile:

446564:
> I would suggest to update your implementation of `getBattPercent()` to better estimate the "capacity" of a LiPo. Then it would be great for the mobile apps to start using this instead. :-P

I'm working on some ideas to approach battery curves, thinking of maybe a map. Just dividing the usable range by 100 and mapping to the closest. rounding down seems better than rounding up.

Then from there mapping the values to match the actual average discharge curves.

446564:
This last change is giving me trouble, it should work fine but when I flash it on a companion with a display it hangs before rendering the screen. Something with the way I'm grabbing the percentage from the map but no compiler errors so I'm not sure what is wrong.

Any help is appreciated. It does not hang when flashing on a companion without a display, i.e. T1000-e.

446564:
turns out arduino can't do maps, at least without a 3rd party implementation. I switched to an array of struct but the jump in storage was a bit much, from 48 bytes to 240 or so for one battery curve. I wonder if two arrays of ints is less overhead, just use the index to reference each half of the pair..

jbrazio:
Keep it simple.. lookup tables.

```C
#define TABLE_SIZE 21

const uint16_t millivoltTable[TABLE_SIZE] = {
    4200, 4150, 4120, 4100, 4050,
    4000, 3850, 3800, 3750, 3700,
    3650, 3600, 3500, 3450, 3400,
    3300, 3200, 3100, 3000, 2800,
    2700
};

const uint8_t percentTable[TABLE_SIZE] = {
    100, 95, 90, 85, 80,
    75, 70, 65, 60, 55,
    50, 45, 40, 35, 30,
    25, 20, 15, 10, 5,
    0
};


uint8_t getBestFittingValue(uint16_t mV) {
    uint8_t ret = 0;
    uint16_t min = (uint16_t)-1; // use a large initial value

    for (uint8_t i = 0; i < TABLE_SIZE; i++) {
        uint16_t diff = abs(millivoltTable[i] - mV);
        if (diff < min) {
            min = diff;
            ret = percentTable[i];
        }
    }

    return ret;
}
```

You can have multiple tables like `millivoltTableLiPo`, `millivoltTableLiFe`, etc.
Update the `getBestFittingValue` to have an additional argument [pointer] to the table.. and use it to look up the final value.

446564:
this is working except for the part using classes as I have no idea what i'm doing.

I went with one multidimensional array as it was significantly less memory than multiple arrays.

jbrazio:
> one multidimensional array as it was significantly less memory than multiple arrays

This is not true.. let's have a look:
1. Your array is uint16_t * 3 * 100 = 600 bytes.
2. Those two arrays are `uint16_t * 21 (42 bytes)` + `uint8_t * 21 (21 bytes)` which is 63 bytes total.

Even if you add another 42 bytes for the LiFe array.. we are far away from your implementation memory footprint.

446564:
I'm going for 1% resolution, so 3 x 100 elements for two chemistries and the charge scale.

I suppose we could get it a bit slimmer by having only the chemistry curves and infer the SoC percentage but the index we found it during the scan

446564:
either way, I have it all sorted out and removed the class idea. just finishing up adding the lifepo4 values and will squash and resubmit a fresh PR

PR #404, opened 2025-06-18
Gps toggle on 4 clicks

fdlamotte:
Really should be done on 3 clicks or less but would change behaviour of current installs ...

The simplest way was to add a virtual function in the board class ... by default does nothing but if implemented calls the good hw toggle ...

Also put board class in variant dir for t1000e thinks its where it should be after doing it for the stm32 variants ;)

### DIFF
--- a/examples/companion_radio/Button.cpp
+++ b/examples/companion_radio/Button.cpp
@@ -50,9 +50,12 @@ void Button::update() {
             triggerEvent(SHORT_PRESS);
         } else if (_clickCount == 2) {
             triggerEvent(DOUBLE_PRESS);
-        } else if (_clickCount >= 3) {
+        } else if (_clickCount == 3) {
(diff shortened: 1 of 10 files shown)


ripplebiz:
Mostly looks good, but I don't think the Board abstract class should know anything about peripherals like GPS.
I get that this was the easiest way to achieve the gps toggle, but I think the better way is for companion main to inject the SensorManager into UITask:

  UITask ui_task(&board, &sensors);

Then do a device-agnostic GPS toggle, using just the abstract methods. A bit more work, granted, but keeps the interfaces cleaner.

fdlamotte:
I agree

fdlamotte:
I've passed the SensorManager class to UITask at begin, seems interesting as it also opens displaying of sensor data 

now UI can toggle gps if the toggle method has been overloaded in sensors manager

PR #406, opened 2025-06-19
Techo rx1262 refactor

fdlamotte:
Same for techo

### DIFF
--- a/variants/techo/target.cpp
+++ b/variants/techo/target.cpp
@@ -25,34 +25,7 @@ TechoSensorManager sensors = TechoSensorManager(nmea);
 bool radio_init() {
   rtc_clock.begin(Wire);
   
-#ifdef SX126X_DIO3_TCXO_VOLTAGE
-  float tcxo = SX126X_DIO3_TCXO_VOLTAGE;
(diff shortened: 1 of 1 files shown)


variants/techo/target.cpp
```diff
@@ -25,34 +25,7 @@ TechoSensorManager sensors = TechoSensorManager(nmea);
...
-#endif
-#ifdef SX126X_DIO2_AS_RF_SWITCH
-  radio.setDio2AsRfSwitch(SX126X_DIO2_AS_RF_SWITCH);
-#endif
-#ifdef SX126X_RX_BOOSTED_GAIN
-  radio.setRxBoostedGainMode(SX126X_RX_BOOSTED_GAIN);
-#endif
-
-  return true;  // success
+  return radio.std_init(&SPI);
```
jquatier, line 28: I think the LORA_CR defined above can go too.

fdlamotte: right, the case where LORA_CR is not set is handled in std_init

I'll remove them from t_echo, xiao_nrf52 and xiao_s3 then

PR #413, opened 2025-06-19
add board battery percentage

446564:
defaults to lipo with no changes to existing variants

- adds two chemistry types to start lipo and lifepo4
- adds getBattPercent to MainBoard
- adds getBattType to MainBboard
- companion UI now uses getBattPercent in renderBatteryIndicator

### DIFF
--- a/src/helpers/Battery.cpp
+++ b/src/helpers/Battery.cpp
@@ -4,7 +4,8 @@
 uint16_t batt_percent(uint8_t batt, uint16_t mv) {
     uint16_t last = mv; // set last to mv so it can never start greater than itself
     uint16_t p = 0;
-    for (int i = 0; i < 102; i++) { // always 101 elements, 100 percent and 0
+    // 101 elements including 0 percentile
+    for (int i = 0; i < 101; i++) {
         // if we match the current value or are between the current and last value
         // we return the lower of the the two, giving accuracy +/- 1% soc
         if (mv <= last && mv >= batt_curve[i][batt]) {


src/helpers/Battery.cpp
```diff
@@ -0,0 +1,17 @@
+#include <Arduino.h>
+#include "Battery.h"
+
+uint16_t batt_percent(uint8_t batt, uint16_t mv) {
+    uint16_t last = mv; // set last to mv so it can never start greater than itself
+    uint16_t p = 0;
+    for (int i = 0; i < 102; i++) { // always 101 elements, 100 percent and 0
```
446564: pretty sure I have a small bug here, 0-99 == 100 so should be `i < 101`. will fix shortly

446564 (commented):
can't approve my own review after changes...

446564:
Any reason we aren't moving forward with some version of this? 

Example: my companion is currently reporting 53% in app @3.65V but that's about 10% maybe.

We can of course offload this to the mobile app but it's not future proof and doesn't allow custom power setups. As well the UI will show incorrect battery status.

cc @ripplebiz

taedryn:
I would like to see this move forward as well, reporting (accurate and realistic) battery percentage is a very valuable feature to have.

446564:
> I would like to see this move forward as well, reporting (accurate and realistic) battery percentage is a very valuable feature to have.

And should be done at the lowest level possible so different clients not only do not have to implement their own but the implementations do not differ. Only the hardware knows what SoC a given voltage is equal to.

shultsb:
This feature would be very helpful!

khudson:
i was just about to submit a bug for this so i'm glad there's a PR to fix it already

446564:
needs rebasing and moving to new UI

jbrazio:
Needs rebasing.. and much more:
1. lookup tables for multiple batt chemistries
2. cli/app command to set the batt chemistry (we will not build one target per batt type)
3. less data points (5% resolution should be more than enough)

https://github.com/meshcore-dev/MeshCore/pull/403#issuecomment-2988694086 :-)

446564:
> Needs rebasing.. and much more:
> 
>     1. lookup tables for multiple batt chemistries

How do you propose a lookup table for a variable input value? Also this is a lookup table, it's a concept not a specific structure. I thought map but we dont know the value being looked up, which is further complicated by number 3 below.

As well, I designed the data structure to be easy to maintain and expand. It's easier to read a table like this. And I doubt the small storage savings is worth the extra complexity.

>     2. cli/app command to set the batt chemistry (we will not build one target per batt type)

Absolutely but not in this PR. Smaller changes are easier to manage and right now everything is assumed to be lipo anyway so we can just leave that as default. Anyone getting battery voltage from an alternative chemistry is already building custom firmware.


>     3. less data points (5% resolution should be more than enough)
>
I disagree.

jbrazio:
> How do you propose a lookup table for a variable input value? Also this is a lookup table, it's a concept not a specific structure. I thought map but we dont know the value being looked up, which is further complicated by number 3 below.

https://github.com/meshcore-dev/MeshCore/pull/403#issuecomment-2988694086
That's a basic C design pattern..

> As well, I designed the data structure to be easy to maintain and expand. It's easier to read a table like this. And I doubt the small storage savings is worth the extra complexity.

No.. you added bloat and complexity.

> Absolutely but not in this PR. 

Then start there.

> now everything is assumed to be lipo anyway

Wrong again.. the current model assumes a theoretical battery with a linear relationship between voltage and capacity.

> I disagree.

If you need more than 5% accuracy, look at the volts.

446564:
Do it yourself then

LitBomb:
@jbrazio  no offense, but unless I misread it, the tone in your reply is not very collaborative.  Please tell me I read it wrong.

jbrazio:
No offence taken. I would not call it "not very collaborative" but it's indeed blunt. This thread has a long history, actually it's the second PR about the same topic with exactly the same fundamental issues.

LitBomb:
@jbrazio 

blunt with constructive feedback is good.  blunt and blunt alone is bad.  It is easy for anyone to criticize.  Constructive and collaborative means pointing out where and what can be improved and also provide support and recommendation and even code.  not simply replying with "bloat and complexity" and "wrong again".  

Scott does this very well, if you read his feedback on PRs.  If he has something to say, he tells you why he likes it or not, and he tells what the submitter can do to improve and sometimes provide a line or two of code to demonstrate.  

This is a feature that we sorely need.  If it is not good enough, you can be the change that you wish to see, like @446564.  Please don't reply bluntly for the sake of being blunt.  We don't want to be like the other mesh project.

jbrazio:
I always suggested what would need to be done in order to have a better approach to the problem.
More than that.. only "if I do it myself". ;-)

PR #416, opened 2025-06-19
Button handling fix 

### LINKED ISSUE #414: User button on RAK 19007 AIN1 works on 1.6.2, doesn't work on 1.7.0
I have a RAK19007 with a user button on pin AIN1.  with 1.6.2 companion firmware, the user button works.  It can wake up the display.  

On 1.7.0, with the new button function change, the user button on the 19007 doesn't do anything.  It doesn't even wake up the screen.

I am guessing the new button feature changed to use a different pin for the user button? 

on my RAK19026, the button works with 1.6.2 and 1.7.0. 

A full erase on the 19007 and then flash 1.7.0 doesn't fix the issue.

jquatier:
Fix button handling to allow both a digital user button and an analog user button simultaneously. This largely impacts the rak4631 variant since it [has both defined](https://github.com/ripplebiz/MeshCore/blob/main/variants/rak4631/platformio.ini#L9). Some rak boards have a built-in button and some do not.

Fixes #414


ripplebiz:
might need to update your branch, as there's now a quad press

jquatier:
should be current now.

PR #420, opened 2025-06-20
return range with advert interval setting error

### LINKED ISSUE #142: flood.advert.interval accepts input in hours, but advert.interval accepts input in minutes
The new flood.advert.interval accepts input in hours.  But advert.interval accepts input in minutes.  Would be good they both accept the same time unit.

Also, while trying to set flood advert interval to 60, thinking it is in minutes, it tells me "Error: max is 48 hours".  When I try to set it to 1 (hour), it says "Error: min is 3 hours".  Would be good to print both acceptable min and max values every time, whether it exceeds min or max, so users don't have to hit the errors twice trying to guess what the min and max are.  Thanks.

446564:
updates both local and flood advert preference
setting errors to include the allowable range and
time units

closes #142

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -245,10 +245,8 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
         strcpy(reply, "OK");
       } else if (memcmp(config, "flood.advert.interval ", 22) == 0) {
         int hours = _atoi(&config[22]);
-        if (hours > 0 && hours < 3) {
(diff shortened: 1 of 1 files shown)


446564:
this **does not** change the time units, but if that is desired can also include that change. Just trying to decide what makes the most sense, probably both in minutes.

liamcottle:
> this **does not** change the time units, but if that is desired can also include that change. Just trying to decide what makes the most sense, probably both in minutes.

We won't be able to change the time units in the CLI command unless it's a new command.

Since the app uses those CLI commands for the UI. Older clients would be sending invalid values to newer repeaters.

PR #421, opened 2025-06-20
fix Heltec v2 getBattMilliVolts

### LINKED ISSUE #161: Heltec V2 - Battery level shows as 45% when powered via USB or full battery
Heltec V2 - Battery level shows as 45% when powered via USB or full battery.  Not sure if this is a bug in the app or whether the battery calculation formula is incorrect in the CompanionRadio variant code for Heltec V2.

446564:
closes #161

### DIFF
--- a/src/helpers/HeltecV2Board.h
+++ b/src/helpers/HeltecV2Board.h
@@ -68,7 +68,8 @@ public:
     }
     raw = raw / 8;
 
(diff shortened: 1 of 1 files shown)


446564:
needs testers, firmware in issue comments #161

446564:
worse than before

PR #424, opened 2025-06-20
fix Heltec v2 getBattMilliVolts ADC multiplier

446564:


446564:
preliminary results seem good, waiting for long term results

PR #426, opened 2025-06-20
allows manual BLE pin even when device has display

### LINKED ISSUE #109: Fixed BLE pin code gets ignored
In the companion_radio_ble build, the a fixed BLE pin code gets ignored if the device previously was flashed with a dynamic pin code. I flashed a build with display support into a device by accident. Subsequently, flashed a build with BLE_PIN_CODE defined to be nonzero, and it was still generating dynamic pin codes (which I couldn't see, because the device had no display).

The offending code is in main.cpp:

```
  #ifdef BLE_PIN_CODE
    if (_prefs.ble_pin == 0) {
    #ifdef DISPLAY_CLASS
      _active_ble_pin = trng.nextInt(100000, 999999);  // random pin each session
    #else
      _active_ble_pin = BLE_PIN_CODE;  // otherwise static pin
    #endif
    } else {
      _active_ble_pin = _prefs.ble_pin;
    }
  #else
    _active_ble_pin = 0;
  #endif` 
```

Note than if _prefs.ble == 0, it ignores BLE_PIN_CODE.

446564:
also updates remaining variants to use default of
123456 instead of 0 for random pin

closes #109


446564:
tested on t1000e and nano g2 ultra, more encouraged.

446564:
tested on T114 as well

DRiKE:
Thanks, works for me on a Heltec v3 wireless stick (the one where the pairing code doesn't fit on the display)!

Issue #328, opened 2025-05-23
Feature Request - GPS on Heltec v3


mrlmuk:
Now that GPS is available on the Heltec Tracker would it be a lot of work to also enable it on a Heltec v3?

Looking at the variants/.../platformio.ini they both appear to be based around the same chip and board design so perhaps most of the work is already done. They differ in the display and the GPIO pins would need to be set correctly for GPS. I will have a go at a custom build / variant.

mrlmuk:
Found this will read more ...
http://github.com/ripplebiz/MeshCore/pull/91
Basic gps support for Heltec_v3 and T1000e

oltaco:
I can implement this, are there standard pins to use for the GPS on heltec?

oltaco:
Submitted a PR for GPS support on Heltec V3. Tested with Ublox Neo-6M. Connect RX on GPS module to pin 45 and TX to pin 46.

Note that while you can enable and disable the GPS there is no way to actually power down the GPS without using a mosfet and adjusting PIN_GPS_EN.

mrlmuk:
Thanks for having a look at this. 

Those were the pins which seem convenient and others have used such but I imagine this would always be a custom firmware someone would configure and build anyway. I'm aware of the battery drain issues and inability to turn it off I can figure out the hardware side and add a switch or mosfet easily as shown in these links.

https://www.thingiverse.com/thing:6967414. While this case is nice it has some frustrating design quirks which waste space.
https://www.reddit.com/r/meshtastic/comments/1i2obfg/has_anyone_added_a_gps_to_their_heltec_v3_i_have/
https://meshtastic.org/docs/hardware/devices/heltec-automation/lora32/peripherals/ 

I will see if I can make sense of what you have done to test it later to try and give some helpful feedback.

Update, what I tried was very close to what you have done but it would not build. I didn't have the knowledge to get it working but at least I learnt quite a bit along the way.

DaBosh:
i have one with this pinout, can offer in app to select gpio pins for tx nd rx

<img width="332" alt="Image" src="https://github.com/user-attachments/assets/2f5eb6bd-831d-4ea5-8737-de49dcc23b36" />

oltaco:
> i have one with this pinout, can offer in app to select gpio pins for tx nd rx
> 

I don't know of any plans to support changing the pins in app.

mrlmuk:
[...]

PR #433, opened 2025-06-21
Heltec v3 GPS support

oltaco:
GPS support using EnvironmentSensorManager.

Tested with Ublox Neo-6M. Connect RX on GPS module to pin 45 and TX to pin 46.

Note that while you can enable and disable the GPS there is no way to actually power down the GPS module without using a mosfet and adjusting PIN_GPS_EN.

Closes: #328

### DIFF
--- a/variants/heltec_v3/platformio.ini
+++ b/variants/heltec_v3/platformio.ini
@@ -21,6 +21,10 @@ build_flags =
   -D ENV_INCLUDE_BME280=1
   -D ENV_INCLUDE_INA3221=1
   -D ENV_INCLUDE_INA219=1  
+  -D ENV_INCLUDE_GPS=1
+  -D PIN_GPS_RX=45
+  -D PIN_GPS_TX=46
+  -D PIN_GPS_EN=-1
 build_src_filter = ${esp32_base.build_src_filter}
   +<../variants/heltec_v3>
   +<helpers/sensors>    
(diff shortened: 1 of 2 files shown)


ripplebiz:
Could you do a timing test for me? I'm curious, for setups that DON'T have GPS wired up, how long does each EnvironmentSensorManager::loop() invocation take?  Is probably fine, but just want to make sure there isn't any latency being added, eg to repeaters, with having that processing in every loop()

oltaco:
Good call, will investigate and get back to you.

oltaco:
So with ENV_INCLUDE_GPS=1 and GPS **unwired** EnvironmentSensorManager::loop() is taking 8-10 micros, with the occasional single loop spike to ~15-17us. The attached log has 20 seconds worth of measurement for this setup.

I also checked with GPS wired just to see what the hit would be. In that setup each second when a new NMEA message comes in every second or third loop takes ~215-240us, probably every time a new character is coming in over serial I guess?

[envsenseman loop gps unwired.txt](https://github.com/user-attachments/files/20845248/envsenseman.loop.gps.unwired.txt)

fdlamotte:
Thanks for the patch, my heltec with gps now gives me position with the dev branch (gps wired on 33/34/47, it also has an RST pin but its wiring doesn't seem necessary)

Maybe a configurable baudrate would be interesting (a GPS_BAUDRATE symbol), by chance my GPS was at 9600 (well that's the one with which MicroNMEALocationProvider was written, so the defaults might be my fault ...)

PR #434, opened 2025-06-21
MACRO cleanup

jbrazio:
This PR ~~cleans up the variants folder and~~ moves duplicated MACRO flags to the `arduino_base` configuration. It also adds (disabled by default) more of the available debug flags.


jbrazio:
@ripplebiz should I keep resolving the conflicts with `dev` for this PR ?

ripplebiz:
I'm not really a fan of PR's like this, as the merge conflicts go both ways, and I prefer that the various 'champions' of the variants, do refactors of the variants they are actively working on.  There is currently a fairly big push to try to get sensor/GPS/UI parity among all the variants, and that needs to take precedence.

jbrazio:
OK np, just to understand the effort as this was mainly housekeeping removing dupe flags.

It could also be possible to reduce scope for this PR, just keep the central part at `platformio.ini` and then each "champion" would clean up their variant flags when refactoring.

PR #439, opened 2025-06-23
T-Beam refactor

cod3doomy:
There is a lot to this PR, so if there are any questions let me know.

The idea here is to merge T-Beam fw so that there is less redundant code. Most versions (except the 0.7) share PMU code, init sequence, and most IO pin definitions.

- Merged all T-Beam board.h files into one TBeamBoard.
- Added PMU code to identify different AXP chips.
- Modified "lilygo tbeam SX1276" variant to cover all T-Beam SX1276 versions
- Modified "lilygo tbeam SX1262" variant to cover all T-Beam SX1262 versions
- Enabled GPS on all T-Beam versions/models
- Enabled BME280 on the Supreme

I am also updating EnvironmentSensorManager to allow for boards that do or don't have GPS enable pins, as well as a PERSISTANT_GPS define check for boards that want GPS to stay active after boot.


src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -180,13 +182,25 @@ bool EnvironmentSensorManager::setSettingValue(const char* name, const char* val
 
 #if ENV_INCLUDE_GPS
 void EnvironmentSensorManager::initBasicGPS() {
-  Serial1.setPins(PIN_GPS_TX, PIN_GPS_RX);
+
+  Serial1.setPins(PIN_GPS_RX, PIN_GPS_TX);
```
ripplebiz, line 186: Is this RX/TX swap intended? If this is only for T-Beam, will need to put it in a #if conditional

cod3doomy: Looking at the setPins arguments it shows rx first and then tx, so I followed that logic. If we want to stick with tx first and then rx second, I can swap mine without issues. Which way would you recommend we go with?

oltaco: @cod3doomy It's all about how you wire up the GPS. We are using PIN_GPS_TX to refer to the pin marked TX on the GPS module, so that pin should be connected to the RX pin for the serial port.

cod3doomy: Figured as much. Just wanted to check. I can swap them back.

ripplebiz:
BTW, there is still a /variants/lilygo_tbeam variant. Should that one be deleted now?

cod3doomy:
Yes, you can remove it. The lilygo_tbeam is now the t-beam sx1276 variant. 
There should be three variants now:
Tbeam sx1262
Tbeam sx1276
Tbeam supreme sx1262

PR #444, opened 2025-06-24
Fix repeater/room server uptime rollover issue.

oltaco:
Currently total_up_time_secs is based on millis() so the maximum is ~50 day before it rolls over to 0.

This PR adds a bootTime var to CommonCLI which gets set at first time sync, after which we can use that to compute uptime in seconds.


oltaco:
I just realised that this likely needs changes to work properly in the case of nodes with RTC installed. Any other issues with it @ripplebiz ?

liamcottle:
Hey, closing this one out since it now has merge conflicts and it doesn't look like it was going to be merged at this time.

PR #446, opened 2025-06-25
RAK4631: BME680 add and GPS cleanup

cod3doomy:
- Added the BME680 environment sensor functionality
- Added the GPS Repeater env for those wanting it (GPS is force alive, since the app has no control yet)
- Cleaned up the GPS and other RAK4631SensorManager code

Verified build and functionality on normal and GPS repeater and companion envs.

IAQ readout is still a work in progress, but a placeholder can be seen on Channel 2 of the app telemetry.

### DIFF
--- a/variants/rak4631/target.cpp
+++ b/variants/rak4631/target.cpp
@@ -19,6 +19,21 @@ RAK4631SensorManager sensors = RAK4631SensorManager(nmea);
 RAK4631SensorManager sensors;
 #endif
 
+#if ENV_INCLUDE_BME680
+#ifndef TELEM_BME680_ADDRESS
+#define TELEM_BME680_ADDRESS    0x76      // BME680 environmental sensor I2C address
+#endif
+#include <bsec2.h>
+static Bsec2 BME680;
+float rawPressure = 0;
+float rawTemperature = 0;
+float compTemperature = 0;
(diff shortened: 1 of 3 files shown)


variants/rak4631/target.cpp
```diff
@@ -147,83 +166,266 @@ bool RAK4631SensorManager::gpsIsAwake(uint32_t ioPin){
...
+    }
+
+    MESH_DEBUG_PRINTLN("BSEC outputs:\n\tTime stamp = %f", String((int) (outputs.output[0].time_stamp / INT64_C(1000000))));
+    for (uint8_t i = 0; i < outputs.nOutputs; i++)
+    {
+        const bsecData output  = outputs.output[i];
+        switch (output.sensor_id)
+        {
+          case BSEC_OUTPUT_IAQ:
+              MESH_DEBUG_PRINTLN("\tIAQ = %f", String(output.signal));
```
ripplebiz: Please do not use the String class. Also, here you can just use the appropriate formatter, like %f, and pass the float. You don't need to Stringify

cod3doomy: Sorry, I'm still learning. I'll keep this in mind for the future. I removed all string type casting and changed a few lines of the BME680 IAQ data output.

PR #469, opened 2025-07-02
Fixing radio include and BT pin for Heltec Paper

jquatier:
The code formatter reordered some very important includes before I committed my code which is causing radio init to fail on this board. We should probably turn off "SortIncludes" in clang format for now. I didn't notice it sadly.

Also, fixing the BLE pin to be like the other variants since the code checks for 123456 and then decides to generate a pin.


jquatier:
@ripplebiz this would be great to get in for a 1.7.3 (or soonest), currently the HT Paper on the web flasher will not work properly without this. Thanks!

PR #475, opened 2025-07-02
Remove ScanResponse.addName() to fix re-advertising after disconnect

rightup:
Removed the call to Bluefruit.ScanResponse.addName() in startAdv(), as it was preventing BLE from reliably restarting advertising after a disconnect.

Hypothesis: adding the device name to the scan response exceeds internal buffer limits or causes a conflict with advertising timing, leading to the BLE stack silently failing to re-advertise.

Tested successfully (on T-1000) without this line, advertising now resumes correctly after disconnection (on iPhone)

### DIFF
--- a/src/helpers/nrf52/SerialBLEInterface.cpp
+++ b/src/helpers/nrf52/SerialBLEInterface.cpp
@@ -18,6 +18,10 @@ void SerialBLEInterface::begin(const char* device_name, uint32_t pin_code) {
 }
 
 void SerialBLEInterface::startAdv() {
+  Bluefruit.Advertising.stop(); // always clean restart
+  Bluefruit.Advertising.clearData(); // clear advertising data
+  Bluefruit.ScanResponse.clearData(); // clear scan response data
+  
   // Advertising packet
   Bluefruit.Advertising.addFlags(BLE_GAP_ADV_FLAGS_LE_ONLY_GENERAL_DISC_MODE);
   Bluefruit.Advertising.addTxPower();
@@ -27,7 +31,7 @@ void SerialBLEInterface::startAdv() {
 
   // Secondary Scan Response packet (optional)
   // Since there is no room for 'Name' in Advertising packet
-  // Bluefruit.ScanResponse.addName();
+  Bluefruit.ScanResponse.addName();
 
   /* Start Advertising
    * - Enable auto advertising if disconnected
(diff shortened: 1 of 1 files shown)


fdlamotte:
thanks @rightup this one seems good

if addName is optional and it still works without it I'm ok to remove ;)

liamcottle:
Howdy! I did a quick test on the web, and without a call to `addName()`, it showed as an unknown device, however after adding that line back, the name would show up.

Screenshot 2025-07-04 at 8 16 35 PM

Screenshot 2025-07-04 at 8 19 03 PM


If multiple calls to `Bluefruit.ScanResponse.addName()` is indeed the issue, it might also be relevant to move this entire block from `SerialBLEInterface::startAdv` into `SerialBLEInterface::begin`:

```
// Advertising packet
Bluefruit.Advertising.addFlags(BLE_GAP_ADV_FLAGS_LE_ONLY_GENERAL_DISC_MODE);
Bluefruit.Advertising.addTxPower();
  
// Include the BLE UART (AKA 'NUS') 128-bit UUID
Bluefruit.Advertising.addService(bleuart);

// Secondary Scan Response packet (optional)
// Since there is no room for 'Name' in Advertising packet
Bluefruit.ScanResponse.addName();
```
  
This is not something I've tested, and have no idea if there are side effects to calling it once on startup, but looking at the source of the `addFlags` and `addTxPower` etc, they are also appending to the advertising data on each call. So maybe these should be called _once_, in `begin` instead...

Unfortunately I don't have the capacity to deep dive all of those changes right now, but thought I best drop this info in here while I'm thinking about it.

rightup:
Hi there,

You still need to include Bluefruit.setName(device_name); within the Bluefruit.begin() call.

The issue seems to occur when startAdv() is called on reconnect and Bluefruit.ScanResponse.addName(); is applied again. (not re advertising)

I had only tested on iPhones (where the issue was noticeable). I’ll re-test on other devices once I'm home, but I suspect the phone might be caching the name if you have noticed the above.

rightup:
I’ve had some time to test this across multiple platforms:
	•	iPhone, Android, and native Windows all correctly display the device name.
	•	However, Chrome and Edge (on desktop) show the device as “Unknown” — it seems they don’t consistently perform scan requests and instead rely on the name being included in the main advertising payload.

While not ideal, the underlying issue I’m addressing is a Bluetooth stack freeze during re-advertising, which appears to be triggered by the device name — and possibly other data — being appended or duplicated across the advertising and scan response payloads.

I’ll continue investigating a more robust way to handle this without sacrificing stability or cross-platform visibility.

rightup:
Manually restart BLE advertising after disconnect to prevent stack freeze

Replaced use of `restartOnDisconnect(true)` with explicit (existing) manual re-advertising logic.
This avoids Bluetooth stack instability caused by overlapping advertising state,
Changes:

- Added explicit `Bluefruit.Advertising.stop()` and data clears in `startAdv()`
- Disabled automatic restart with `restartOnDisconnect(false)`
- Re-advertising now fully handled in `checkRecvFrame()` loop


Tested on: iPhone, Android, Windows, and Chrome – confirmed stable reconnects and name visibility.

@liamcottle  do you mind having another look when you get a moment, thanks

liamcottle:
Flashed your latest commits on a brand new device that I've never paired with, nor had MeshCore on, and it looks to correctly show the default device name when scanning on web now :)

PR #478, opened 2025-07-03
xiao_nrf: rxen was not defined with alt pinout

fdlamotte:
This variation of pinout did not work anymore :(


fdlamotte:
this will be refactored in other PR

PR #479, opened 2025-07-03
cleanup xiao nrf52

446564:
move variant specific code out of src/helpers
redefine RXEN for alternate radio pinout (S3)


446564:
cherry pick #478

PR #484, opened 2025-07-04
LR's (corrected) calculation instead of SX and minor changes

liquidraver:
Radiolib LR gettimeonair calculation wrongly assumes preamble is not in symbols already (in Lora, it is) so it's trying to make symbols from symbols.
if we are using radiolib's code, plugging in preamble lengths shows the error:
preamble 8 = 8
preamble 16 = 0
preamble 24 = 32

That's why it was working with 8 or 24 but not with 16.

override code copied from radiolib, stripped down to lora only and commented/corrected

also CRC increased and explicit header added (info via radiolib's wiki)


fdlamotte:
Tested on my side with good results (talking to sx1268 and stm32wl radios)

446564:
~~amazing, I guess I will wait to update repeater firmware :)~~

nice work

Issue #497, opened 2025-07-06
Double Press generates Zero Hop Advert - should be Flood Advert


MikesAllotment:
The current double press button handler in Companion Radio generates a Zero Hop Advert - should be Flood Advert.

bool MyMesh::advert() {
  mesh::Packet* pkt;
  if (_prefs.advert_loc_policy == ADVERT_LOC_NONE) {
    pkt = createSelfAdvert(_prefs.node_name);
  } else {
    pkt = createSelfAdvert(_prefs.node_name, sensors.node_lat, sensors.node_lon);
  }
  if (pkt) {
    sendZeroHop(pkt);   // **<<----**  Wrong
    return true;
  } else {
    return false;
  }
}

Change to ....

bool MyMesh::advert() {
  mesh::Packet* pkt;
  if (_prefs.advert_loc_policy == ADVERT_LOC_NONE) {
    pkt = createSelfAdvert(_prefs.node_name);
  } else {
    pkt = createSelfAdvert(_prefs.node_name, sensors.node_lat, sensors.node_lon);
  }
  if (pkt) {
    sendFlood(pkt);   // **<<----**  Should be
    return true;
  } else {
    return false;
  }
}

recrof:
could you please open a pull request?

MikesAllotment:
I've never raised a PR - wouldn't know where to start.

recrof:
> I've never raised a PR - wouldn't know where to start.

it's pretty easy.. go to https://github.com/ripplebiz/MeshCore/, switch the branch to dev, find the file that needs to be changed, use the pencil to edit the file, make the changes and then submit it - PR creation will be one of the options.

seagull9000:
Why should it be flood? 

There must have been some logic for it to be zero advert.  Does this need some discussion?

MikesAllotment:
I was supposed to be flood ... it's a bug.

446564:
Can be closed? @ripplebiz

webmonkey:
> Does this need some discussion?

I think so. It's clear that there are different opinions.

MikesAllotment:
I have posted in the Development channel on Discord for discussion.

PR #512, opened 2025-07-10
Double-click advert-type fix

webmonkey:
Send flood ad instead of zero hop

Raised in #497


ripplebiz:
This was never meant to be a flood advert short-cut. If you need to do that, then the app is there.
This is just a convenience for when you are swapping details with someone.

Issue #515, opened 2025-07-12
Add Wifi to Station G2


zjs81:

liamcottle:
Feel free to open a PR to add this variant. It won't be available in the web flasher as users still need to compile it themselves to add the WiFi SSID and password. Please see https://github.com/meshcore-dev/MeshCore/issues/524 for reference.

PR #516, opened 2025-07-12
#515 add wifi suppot to station g2

zjs81:
Introduce WIfi support for station G2. This also allows clients to connect using bluetooth but if they do they will be disconnected via tcp and vice versa


recrof:
I'm affraid you're also merging commits that are in main into dev with this PR. you need close this PR, checkout dev on your repo, make changes there and then open PR again - it should only have your commit(s).

zjs81:
Thanks for the info

Issue #440, opened 2025-06-23
NTP over LoRa


mikecarper:
https://github-wiki-see.page/m/hardwario/lora-modem/wiki/Time-Synchronization

Something simple like this. Only serial or gps devices can send these. Probably limit to 1 hop so time drift doesn't get too high 
{
  "type": "time_sync",
  "epoch": 1719082800, 
}

seagull9000:
This may be a silly question but it has been nagging at me for some time ...

.. why can't repeaters use the timestamps in passing messages for time sync?

If we assume that the App device has good time, then I don't see why this wouldn't work.   Granted, there may be a few seconds delay when a significant number of multiple hops are involved from when the App timestamps the message and when the repeater receives it.  

As in the above post, you could limit the drift by limiting the hops from which a repeater will take the time.  This might then cause some remote/intermediate repeaters in a path to never update, though.  Maybe make the number of hops for a valid timestamp configurable in those repeater but default to 1 otherwise.

 It would avoid the need to flood time sync signals around (which could have the same multi-hop issue anyway).

In case of bad actors, the repeaters could refuse to sync outside a certain range or ensure it gets a number of consistent time stamps before acting on them.  They would need to have a base time for an unset clock so this can be used as an exception to this rule.

lucidnx:
yeah, I'v had same idea. RadioLib can do TX time calculation so FW can be compensating delays. probably it's already happening if you sync time manually.

zjs81:
Added a pull request using @seagull9000 's suggestion

n2aws:
@zjs81 great work!


I'd actually like to see this go a step further, probably as another "issue". but,  since repeaters will typically be in decent locations,  and GPS hardware is so inexpensive now,  it'd be awesome if repeater firmwares allowed the device to get time directly from GPS,  and then became "stratum 1 timeservers" to the rest of the mesh.  This would help alleviate the problem mentioned above about nodes that are many hops away not being synced, because a nearby repeater could just have a GPS module installed...  Also,  with companion firmware, I know the GPS needs to be re-enabled on each node restart but for repeaters, I think  it should default to being on if hardware is found..   

Thoughts?   Should I open this as a separate issue?

mikecarper:
[...]

PR #518, opened 2025-07-12
#440 add time sync using rec packets

zjs81:


zjs81:
Need to find a way to prevent bad actors. Rejecting PR for now.

PR #522, opened 2025-07-13
Heltec Wireless Paper fix: radio init failed: -2

recrof:


recrof:
fixes https://github.com/ripplebiz/MeshCore/issues/519

PR #529, opened 2025-07-16
Fixed Barometric Pressure Reading for BMP280

marcelverdult:
Measurement has to be divided by 100 same as BME280


jappyjan:
Congrats to your first PR ❤️😀🎉

PR #530, opened 2025-07-16
added set/get voltage_multiplier to CLI

marcelverdult:
added a possibility to set/get voltage multiplier via cli so you could calibrate battery voltage without the need to recompile and flash

ToDo: use the saved multiplier in battery voltage calculation

### DIFF
--- a/variants/techo/variant.h
+++ b/variants/techo/variant.h
@@ -61,9 +61,9 @@
 ////////////////////////////////////////////////////////////////////////////////
 // Builtin LEDs
 
-#define LED_RED                 (-1)
-#define LED_GREEN               (-1)
-#define LED_BLUE                (-1)
+#define LED_RED                 (34)
+#define LED_GREEN               (33)
+#define LED_BLUE                (14)
 
 #define PIN_STATUS_LED          LED_GREEN
 #define LED_BUILTIN             LED_GREEN


src/helpers/CommonCLI.cpp
```diff
@@ -342,6 +348,15 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
         _prefs->freq = atof(&config[5]);
         savePrefs();
         strcpy(reply, "OK - reboot to apply");
+      } else if (memcmp(config, "voltage_multiplier ", 19) == 0) {
```
ripplebiz: for consistency, prob should be "voltage.multiplier "

src/helpers/CommonCLI.cpp
```diff
@@ -218,6 +222,8 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
         mesh::Utils::toHex(&reply[2], _callbacks->getSelfIdPubKey(), PUB_KEY_SIZE);
       } else if (memcmp(config, "role", 4) == 0) {
         sprintf(reply, "> %s", _callbacks->getRole());
+      } else if (memcmp(config, "voltage_multiplier", 18) == 0) {
```
ripplebiz: for consistency, prob should be "voltage.multiplier"

src/helpers/CommonCLI.cpp
```diff
@@ -69,6 +71,7 @@ void CommonCLI::loadPrefsInt(FILESYSTEM* fs, const char* filename) {
     _prefs->sf = constrain(_prefs->sf, 7, 12);
     _prefs->cr = constrain(_prefs->cr, 5, 8);
     _prefs->tx_power_dbm = constrain(_prefs->tx_power_dbm, 1, 30);
+    _prefs->voltage_multiplier = constrain(_prefs->voltage_multiplier, 1, 10);
```
ripplebiz: range should include 0.
Also, does the upper maximum make sense? Just wonder if this should just check for negative values?

ripplebiz:
Please change base branch to 'dev'

PR #537, opened 2025-07-19
RAK4631: Add BME680 and RAK18001 Buzzer

cod3doomy:
Large add to ESM for the BME680. Confirmed working on all RAK sockets. Non-RAK boards untested. 
-Temp, Humidity, and Pressure are sent to Channel 1 in Liam's app. 
-IAQ and CO2 are sent to Channel 2 as Generic Sensor and Concentration

RAK18001 buzzer added for use on Socket D on both the 19007 and 19003. Verified working with startup sound as well as message received.

### DIFF
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -21,6 +21,22 @@ static Adafruit_AHTX0 AHTX0;
 static Adafruit_BME280 BME280;
 #endif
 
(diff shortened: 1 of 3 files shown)


recrof:
I have potentially conflicting PR: https://github.com/ripplebiz/MeshCore/pull/541
if you can merge it to this one, I will close it. I have tested both BMP280 and BME280 and they work without problem. I can see no reason why not include other i2c sensors as well.

oliverl-21:
will we see this merged?

PR #547, opened 2025-07-22
sensors: gpio command

fdlamotte:
Simple implementation to control gpio pins

Pins are defined in `get/setGpio` in board class
A `gpio` cli command has been defined, with no arguments it gives the value of the gpio, if an argument is given it will be parsed as an hex value and sent to the `setGpio` method

Poc uses rak3372 to control the two leds ...

read only pins should be possible


fdlamotte:
Some modifications to my initial proposal 
* gpio -> io (shorter and same signification)
* reset/set/toggle bit masks (prefix with r/s/t) (`io t3` or `io t 3` will toggle the two LSB

PR #553, opened 2025-07-25
Disable LED lights on t-echo device

ch4r13:
Sorry for the less-than-ideal change in the code, but is it possible to disable the LED lights on the T-Echo device for now, until someone adds an option in the UI to disable them?
They unnecessarily drain the battery, especially since the device cannot enter deep sleep by user control for now.

I hope there will be plenty of users who don’t like a device flashing like a Christmas tree. 

Edit: sorry for Unverified commit, I used accidentally old one of my ssh keys ...

### DIFF
--- a/variants/techo/variant.h
+++ b/variants/techo/variant.h
@@ -61,9 +61,9 @@
 ////////////////////////////////////////////////////////////////////////////////
 // Builtin LEDs
 
-#define LED_RED                 (34)
-#define LED_GREEN               (33)
-#define LED_BLUE                (14)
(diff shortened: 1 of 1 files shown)


recrof:
This change is very crude, with only one board affected.. we should not encourage changes that disable functionality without actual fixes. this should never be open as a PR.

ch4r13:
You are absolutely right, sorry for this PR. I tried to modify the behavior of flashing LED, but it's called deeper in the code and my programming knowledge is very fringe. I will try to find some more time to make it better. @recrof

Issue #560, opened 2025-07-30
T-1000e - long press on H/W button to power down will power back up if held too long


va3jme:
When you press and hold the hardware button to turn the device off, if you hold it for too long, the device detects the continued press as a power on request. You have to time your release of the button "just so", so that the device doesn't restart when you really intended to shut it down. 

I'm not really sure how the code is structured, but one possibility would be to have the device beep a single time after you've held it long enough to shut it off, but not actually complete the shut off until you've released the button. That way once the device has shut down, it's not detecting a key press because, well, the key is not pressed and so it will not start right back up again.

seagull9000:
What version is the t1000e on?

The shutdown Melody should play when it is shutting down.  If you release the button before the melody ends, then it shouldn't start up again.

If the buzzer has been muted, then you may not hear the melody.  That can be fixed easily enough in the code

liamcottle:
Looks like this was fixed in #636

va3jme:
Thanks everyone for finding the duplicate for this one. I should have looked harder before submitting a report.

PR #583, opened 2025-08-06
fix: T1000 Fix for power off - long hold button. Slow triple press for mute.

ventz:
Fixes:
* Fix for power off - long press and wait. If it's connected to power, it won't work.

Feature:
* Add for mute - 3 slow presses.


fdlamotte:
Hi, I think I can handle this one ;)

What issue are you trying to solve ? Because last time I used it long press and triple press worked ok ... did you have issues with the long press duration ? (It was initially set to 5s ...)

I'm ok for the usb detection ...

There is another issue you can look at since you're at it, when powering off, if the button stays pressed the device will just reboot ... its ok with the sound because we know when to release button but not without sound.

It would be great if you could handle the btn release ... my idea was simply to light the led in power off and wait for the button to be released before actually calling `sd_power_off` (edit: this would solve https://github.com/meshcore-dev/MeshCore/issues/560)

edit : also we usually do the commits to the dev branch instead of main

fdlamotte:
I'll close this, as the issue with poweroff got a fix with 1.8 (and there were some changes on button handling)

Not sure increasing long press duration was really an answer

PR #584, opened 2025-08-06
nano g2 gps fixes

446564:
various fixes for nano g2 gps

- baud rate changed to default
- switch state detected at boot to enable position
- small edge cases fixed
- more mesh debugging output added

Co-authored-by: LitBomb
Co-authored-by: Nick from BOSTON


LitBomb:
This is a necessary fix to get GPS working on the Nano G2.  with 1.7.4 released built, the Nano G2's GPS doesn't appear to work.  The PPS LED still flashes hinting at the GPS does get satellite lock.  This is very similar to adding a random GPS to the OG T-Deck and requiring user to set the GPS baud rate.  

Ded's fix is to set the GPS baud rate in code to 9600 baud.  with 9600 baud rate, the 3 of us have tested 4 different Nano G2s and now they all have working GPS with live position.

enabling and disabling GPS from the phone app still works as expected. 

Thanks @446564 for spending the time to swat this bug down!

446564:
the intial cause was that some MT firmware was setting a different baud rate and saving that in the gps eeprom. So my device was using 115200 while most were 9600. This changes to the expected default and we can consider a probe to detect odd baud rates if/when more users pop up with non working GPS units.

variants/nano_g2_ultra/target.cpp
```diff
@@ -42,45 +42,44 @@ void radio_set_tx_power(uint8_t dbm)
...
-  else
-  {
-    MESH_DEBUG_PRINTLN("No GPS detected");
+  Serial1.setPins(PIN_GPS_TX, PIN_GPS_RX);
+  Serial1.begin(9600);
+  MESH_DEBUG_PRINTLN("Checking GPS switch state");
+  delay(1000);
+
+  // Check initial switch state to determine if GPS should be active
+  if (gps_active = Serial1.available() > 0) {
```
446564: Missed a little bug here, probably should just be `Serial1.available() > 0` as if true, `start_gps` will set `gps_active` to true.

ripplebiz: Rest looks good, but can you fix the above?

PR #585, opened 2025-08-06
Fix: Reset ESP32 RTC clock on software reset

rightup:
Updated `ESP32RTCClock::begin()` to treat software reset (`ESP_RST_SW`) the same as power-on reset (`ESP_RST_POWERON`), ensuring the clock resets to the default time (15 May 2024, 8:50pm) when triggered via the CLI reboot command or any other software-initiated reset.

No personal issues with a good old-fashioned power cycle, but some have reported their clock getting accidentally set far into the future. Probably less funny when the device is zip-tied 8m up a tree ... not exactly reset-button friendly!

### DIFF
--- a/src/helpers/ESP32Board.h
+++ b/src/helpers/ESP32Board.h
@@ -94,7 +94,7 @@ public:
   ESP32RTCClock() { }
   void begin() {
     esp_reset_reason_t reason = esp_reset_reason();
-    if (reason == ESP_RST_POWERON) {
+    if (reason == ESP_RST_POWERON || reason == ESP_RST_SW) {
       // start with some date/time in the recent past
       struct timeval tv;
       tv.tv_sec = 1715770351;  // 15 May 2024, 8:50pm


rightup:
Hey, just a note: I used Fix: in the commit message, but it’s possible the original behavior was by design,  so feel free to treat this as a merge suggestion rather than a bug fix, depending on intent.

Either way, resetting the clock on soft reboot feels like the safer option in the field,  especially when devices are zip-tied to trees 😅

ripplebiz:
Yeah, this one is a little tricky. It was intentional to keep the ESP system time across soft resets, and I seem to recall it being handy in various contexts. But am struggling to rummage through my memory and figure out if this really should be changed...

rightup:
Totally fair, I figured this might’ve been intentional, which is why I hesitated on calling it a “fix”.

I haven’t personally seen the time set incorrectly, but I’ve noticed a few user reports where the clock jumped a few hours into the future. If that does happen, the only way to reset it is via a power cycle or physically pressing the reset button — which isn’t ideal when the device is mounted somewhere creative.

No problem at all — happy to drop this if it risks interfering with other use cases. I know timing in embedded is always tricky… there’s a reason even Doc Brown had issues with it. 🤣

recrof:
if this is not the best solution, can we introduce another command that will reboot with clock reset?

ripplebiz:
Generally, most want clock preserved across a reboot...  for 'bad' clocks, need some other solution.
Maybe new CLI command  "clock reset" ?

PR #590, opened 2025-08-08
update git ignore

446564:
exclude lsp dirs for clangd and ccls
exclude pio compile commands and flags files


446564:
I don't know why I have two lines removed. I'll resubmit later

PR #592, opened 2025-08-09
Changed radio_init to use radio.std_init

wel97459:
The waveshare_rp2040_lora variant is not following the other variants by using radio.std_init to set up the radio.

The original code looks to have been pulled from generic-e22 code which also has the same preamble of 8 hard code into the function call to radio.begin. 
[Line 36](https://github.com/wel97459/MeshCore/blob/c48ec40ae680647d483b9eff8029f23bcebf3718/variants/generic-e22/target.cpp#L36)

The hardcode preamble in the radio.std_init function is 16.


ripplebiz:
Could you please change the base branch to 'dev'.

PR #593, opened 2025-08-09
Added a default header for settings that are repeated in multiple files.

wel97459:
Pretty straightforward.

Also has defines to help with region configurations of the radio.


liamcottle:
Hey there! Thanks for the PR :) Looks like this one has become stale and now has merge conflicts. I don't think this is something we were looking to merge at this time. So I'll close this one out.

PR #597, opened 2025-08-10
add lsp ignores

446564:
adds cache and ccls dirs and clangd files to git ignore


446564:
this is for those using clangd or ccls with platformio

PR #598, opened 2025-08-10
add new UI to nano g2

446564:


446564:
sorry for the format change, that just the included clang formatting that was added a while back

PR #603, opened 2025-08-11
Add deep sleep with USER button wake-up for Heltec V3

d6e:
## Problem
Currently meshcore running on heltec v3 boards can't be turned off so the battery drains very quickly and stays dead which is bad for the device. 

## Summary
- Implements deep sleep functionality for Heltec V3 boards when the USER button is held for 5 seconds
- Reset button is required to start again (this is because the USER button has a mode conflict issue which prevents it from being a way to keep the device asleep *and* wake the device reliably)
- Shows power-off confirmation message on display before sleep

## Test Plan
- [x] Verify device enters deep sleep when holding the power button for 5 seconds
- [x] Confirm RESET button press wakes device from sleep
- [x] Validate display shows power-off message before sleep

### DIFF
--- a/examples/companion_radio/UITask.cpp
+++ b/examples/companion_radio/UITask.cpp
@@ -293,10 +293,24 @@ void UITask::shutdown(bool restart){
 
   #endif // PIN_BUZZER
 
-  if (restart)
+  if (restart) {
     _board->reboot();
-  else
+  } else {
+    // Show power-off message on display
+    if (_display != NULL) {
+      _display->startFrame();
+      _display->setCursor(0, 0);
+      _display->setTextSize(2);
+      _display->print("Power Off");
+      _display->setCursor(0, 20);
+      _display->setTextSize(1);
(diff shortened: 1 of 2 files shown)


fvanson:
Thank you for quick fix & reaction!
Would it be possible to also introduce for the Heltec T114?

d6e:
> Would it be possible to also introduce for the Heltec T114?

I don't have one of those to test with, I could copy over the changes if you'd like though. Make a separate issue for it

fvanson:
> > Would it be possible to also introduce for the Heltec T114?
> 
> I don't have one of those to test with, I could copy over the changes if you'd like though. Make a separate issue for it

Thank you for your support! I understand. I will make a separate issue for it, as I think this is a very good to have functionality.

ripplebiz:
The new UITask structure in the 'dev' branch has now defined this, with the Board::powerOff() method. HeltecV3 and T114 have been implemented.

PR #608, opened 2025-08-13
Add heltec_vision_master_e213 board.

Quency-D:
1. Add heltec_vision_master_e213 board.
2. Add new screen recognition function to heltec_wireless_paper.

### DIFF
--- a/src/helpers/ui/E213Display.cpp
+++ b/src/helpers/ui/E213Display.cpp
@@ -2,7 +2,7 @@
 
 #include "../../MeshCore.h"
 
-EInkDetectionResult E213Display::detectEInk()
+BaseDisplay* E213Display::detectEInk()
 {
     // Test 1: Logic of BUSY pin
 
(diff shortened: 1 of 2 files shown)


src/helpers/ui/E213Display.cpp
```diff
@@ -2,36 +2,83 @@
...
+
+    // Test complete. Release pin
+    pinMode(DISP_RST, INPUT);
+
+    if (busyLogic == LOW)
+        return V_LCMEN213EFC1;
+    else // busy HIGH
+        return V_E0213A367;
+}
+
```
ripplebiz, line 40: Based on which display is detected, can you just store a pointer to display or display1, but as base class, which seems to be: BaseDisplay
So have a member var:  BaseDisplay* _detected; 

And instead of al of the if (_version) ...   just called _detected->setRotation() for example.

Quency-D: Thanks for your suggestion, it has been modified.

PR #621, opened 2025-08-15
Add target that enables MESH_PACKET_LOGGING on select devices

Cisien:
Adds a variant to the repeater targets for rak 4631, xaio s3, and station g2 boards.

Primary motivation is to help others on my local mesh run our data collector, and so I don't have to manually edit the platformio.ini each time i build.


haylinmoore:
This would be nice to have

ripplebiz:
I'm wondering if a better way to go about this is to enhance the existing "log ..." CLI commands? Currently "log start" writes to a temp file, and "log" just dumps that to Serial. We could extend this same mechanism, but have a new preference to say _where_ to write log to. eg. "log output {serial|file}", and there could be an implicit "log start" at startup if pref = "serial" ?

Cisien:
I like that idea, saves having to spin custom builds completely. I'll work on that when I get a chance.

PR #628, opened 2025-08-16
[Feature]: Xiao ESP32-C6 (Meshimi) Add MAX17261 fuel gauge support with voltage + temperature telemetry

alexbegoon:
-  Introduces MAX17261 fuel gauge support for the Xiao ESP32-C6 Meshimi variant, providing accurate battery voltage and die temperature readings, and wiring temperature into telemetry.
- New board class: variants/xiao_c6/MeshimiBoard.{h,cpp}
- Telemetry integration: variants/xiao_c6/MeshimiSensors.{h,cpp}

### DIFF
--- a/variants/xiao_c6/MeshimiBoard.cpp
+++ b/variants/xiao_c6/MeshimiBoard.cpp
@@ -1,0 +1,156 @@
+// Build only for Meshimi board-enabled targets
+#ifdef USE_MESHIMI_BOARD
+#include "MeshimiBoard.h"
+
+// MAX17261 driver
+#include <max17261.h>
+#include <Arduino.h>
+#include <Wire.h>
+#include <math.h>
+#include <string.h>
+
+static struct max17261_conf g_max17261_conf;
+static bool g_max17261_inited = false;
+static bool g_max17261_conf_inited = false;
+static uint32_t g_last_mv_read_ms = 0;
+static uint16_t g_last_mv_cached = 0;
+static uint32_t g_last_temp_read_ms = 0;
+static float g_last_temp_cached = NAN;
+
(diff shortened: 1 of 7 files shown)


fdlamotte:
Interesting ...

But why not adding MAX1261 to  EnvironmentalSensorManager and use EnvironmentalSensorManager in xiaoC6 port ? This would benefit every variant and let XiaoC6 use any sensor.

alexbegoon:
> Interesting ...
> 
> But why not adding MAX1261 to EnvironmentalSensorManager and use EnvironmentalSensorManager in xiaoC6 port ? This would benefit every variant and let XiaoC6 use any sensor.

Not sure. Need to check the design.
May be having GaugeSensorManager will be better then.

fdlamotte:
I thought you would implement your fuel gauge as a sensor in the telemetry framework ?

Instead you build a totally new framework for power monitoring ... this is a huge work (and this does not seem badly written), but I feel it makes things really overcomplicated ! I'd like to have @ripplebiz impressions on that one ...

Why not adding the gauge as a sensor, and then, if you want to monitor your battery with this sensor (instead of the std ADC) see how the value can be obtained from the sensor in the existing framework ?

I'm sorry maybe I misleaded you with my previous comment ?

alexbegoon:
@fdlamotte This is inspired by the conversation that happened at a discord here: https://discord.com/channels/1343693475589263471/1391673743453192242/1407413402925793461

From my POV it make sense. Thanks for a feedback!

liamcottle:
Hello, thanks for the PR. I'll close this one out due to old age and no traction gained to push for merging.

PR #629, opened 2025-08-16
techo_ui: some tweaks

fdlamotte:
* fix on text width calculation (some text was written in next line, added 1 to calc size)
* display turns on even if there is no backlight defined
* shows 9 lines in recent list

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -14,6 +14,10 @@
 
 #define LONG_PRESS_MILLIS   1200
 
+#ifndef UI_RECENT_LIST_SIZE
+  #define UI_RECENT_LIST_SIZE 4
+#endif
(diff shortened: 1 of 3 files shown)


examples/companion_radio/ui-new/UITask.cpp
```diff
@@ -156,10 +160,10 @@ class HomeScreen : public UIScreen {
         display.drawTextCentered(display.width() / 2, 43, tmp);
       }
     } else if (_page == HomePage::RECENT) {
-      the_mesh.getRecentlyHeard(recent, 4);
+      the_mesh.getRecentlyHeard(recent, UI_RECENT_LIST_SIZE);
```
ripplebiz, line 163: The recent[] array will also need to be declared with this size

fdlamotte:
Ok, fixed the commit and rebased

PR #635, opened 2025-08-17
Update CommonCLI.cpp

yg-ht:
I keep wanting a help menu in the CLI but it isn't there.  Had a look and realised a small refactor would help make this maintainable.  I have:

- made a single source of truth for the output strings
- added a number of formatter functions to enable small help responses (essentially grouped things)
- added the calling routines for the above into the main logic

CLI message responses for use over the mesh need to be <= 160 chars which I *think* I have achieved here.  Needed to group things so that it didn't just try to dump everything in one hit and go way over that 160 limit.  Was also conscious of future expansion so I hope these messages are still small enough for additional options in the future.

I believe this is a direct drop-in with no other files / code affected.  This is first commit after development though and has not yet been tested on a device.  Please can I ask for help on this as it might take me a bit to actually test it (I am away from devices right now).

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -3,6 +3,176 @@
 #include "TxtDataHelpers.h"
 #include <RTClib.h>
 
+// ---------- Command and key registry for zero-maintenance `help` ----------
+// Define command strings once, use everywhere, and list them for help output.
+static constexpr const char* CMD_REBOOT        = "reboot";
+static constexpr const char* CMD_ADVERT        = "advert";
+static constexpr const char* CMD_CLOCK         = "clock";
+static constexpr const char* CMD_CLOCK_SYNC    = "clock sync";
+static constexpr const char* CMD_START_OTA     = "start ota";
+static constexpr const char* CMD_NEIGHBORS     = "neighbors";
+static constexpr const char* CMD_TEMPRADIO     = "tempradio";
+static constexpr const char* CMD_PASSWORD      = "password";
+static constexpr const char* CMD_CLEAR_STATS   = "clear stats";
+static constexpr const char* CMD_GET           = "get";
+static constexpr const char* CMD_SET           = "set";
+static constexpr const char* CMD_ERASE         = "erase";
+static constexpr const char* CMD_VER           = "ver";
+static constexpr const char* CMD_LOG_START     = "log start";
+static constexpr const char* CMD_LOG_STOP      = "log stop";
+static constexpr const char* CMD_LOG_ERASE     = "log erase";
+static constexpr const char* CMD_LOG_DUMP      = "log";
+static constexpr const char* CMD_HELP          = "help";
+
+// Config keys supported by get/set. Listed once for help, used in comparisons.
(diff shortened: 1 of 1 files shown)


src/helpers/CommonCLI.cpp
```diff
@@ -129,12 +299,28 @@ void CommonCLI::savePrefs() {
 }
 
 void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, char* reply) {
-    if (memcmp(command, "reboot", 6) == 0) {
+    // Split help to keep messages short
+    // device, radio, mesh, auth, debug
+    if (memcmp(command, CMD_HELP, 4) == 0) {
+      if (memcmp(command, "help device cmds", 16) == 0) { format_help_device_cmds(reply); return; }
+      if (memcmp(command, "help device keys", 16) == 0) { format_help_device_keys(reply); return; }
+      if (memcmp(command, "help radio cmds", 15)  == 0) { format_help_radio_cmds(reply);  return; }
```
4np, line 307: It would probably make sense to use the constants here as well, something like:

```suggestion
      if (memcmp(command, CMD_HELP + SPACE + KEY_RADIO + "cmds", 15)  == 0) { format_help_radio_cmds(reply);  return; }
```

4np (commented):
Good to have help! I was constantly looking for this as well :)

src/helpers/CommonCLI.cpp
```diff
@@ -180,88 +366,88 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
       } else {
         strcpy(reply, "Error, invalid params");
       }
-    } else if (memcmp(command, "password ", 9) == 0) {
+    } else if (memcmp(command, CMD_PASSWORD, 9) == 0) {
```
4np, line 369: The original string is nine characters `"password "` (note the space at the end), whereas the constants is `"password"` (no space, 8 characters). So it looks like this if statement will be false.

4np (commented):
IMHO it would be better to tokenize commands (e.g. `start ota` > cmd: 'start' & arg: 'ota']) and use those rather than string comparisons on command + space (e.g. `"password"` vs `"password "`). 

Something along these lines:

```c
#include <string.h>
#include <stdio.h>

...

void handleCommand(char *input) {
  // split by spaces
  char *cmd = strtok(input, " ");
  char *arg = strtok(NULL, " ");  // first argument
  char *arg2 = strtok(NULL, " "); // optional second, etc.

  if (cmd == NULL) return;

  if (strcmp(cmd, "start") == 0) {
    if (arg && strcmp(arg, "ota") == 0) {
      Serial.println("Starting OTA...");
      // startOTA();
    }  else {
      Serial.println("Unknown start option");
    }
  } else if (strcmp(cmd, "stop") == 0) {
    Serial.println("Stopping...");
  } else if (strcmp(cmd, "status") == 0) {
    Serial.println("Status is OK");
  } else {
    Serial.println("Unknown command");
  }
}
```

liamcottle:
Hello, thanks for the PR! I'll close this one out due to old age and merge conflicts. I'm not sure this is the approach we want to take for a help menu at this time.

PR #644, opened 2025-08-20
Control cooldown and priority

ViezeVingertjes:
**Before**
Same cooldown for all - control could lag under load.
Priorities were ad‑hoc and scattered:
Flood: PATH=2, ADVERT=3, DATA=1
Direct: PATH=1, DATA=0
Zero‑hop: 0
TRACE: 5

**After**
Control uses shorter cooldown - snappier joins/ACKs.
Priorities centralized via getPriorityFor() and named constants:
PRI_CONTROL_HIGH=0 (ACK/PATH, zero‑hop)
PRI_DATA=1 (REQ/RESPONSE/TXT)
PRI_BEACON=3 (ADVERT)
PRI_TRACE=5 (TRACE)

### DIFF
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -65,7 +65,11 @@ void Dispatcher::loop() {
       //Serial.print("  airtime="); Serial.println(t);
 
       // will need radio silence up to next_tx_time
-      next_tx_time = futureMillis(t * getAirtimeBudgetFactor());
+      if (outbound && outbound->isControlPayload()) {
+        next_tx_time = futureMillis(t * getControlAirtimeBudgetFactor());
+      } else {
+        next_tx_time = futureMillis(t * getDataAirtimeBudgetFactor());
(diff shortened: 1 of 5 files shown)


ripplebiz:
This is a pretty low-level change, and has impact on EVERY node out there, so we'll need some kind of 'buy in' for changes like this. ie. see if you can get people in the Development channel to run this branch, and get some feedback.
We really need to know if this has benefits or risks, etc.

ViezeVingertjes:
> This is a pretty low-level change, and has impact on EVERY node out there, so we'll need some kind of 'buy in' for changes like this. ie. see if you can get people in the Development channel to run this branch, and get some feedback. We really need to know if this has benefits or risks, etc.

I totally agree! I'm just sharing some improvements from my personal build in case they're of interest. If they don't fit the project's direction or gain some support, it's nothing lost really. (no pressure)

PR #653, opened 2025-08-22
Extra filesystem support for NRF52 (CustomLFS)

oltaco:
This is probably going to need a heap of changes before it is ready to be merged but let's get the ball rolling!


oltaco:
Ok, readme and versioning have been reverted to align with dev 👍

examples/companion_radio/DataStore.cpp
```diff
@@ -203,11 +261,20 @@ void DataStore::savePrefs(const NodePrefs& _prefs, double node_lat, double node_
 }
 
 void DataStore::loadContacts(DataStoreHost* host) {
+#if defined(NRF52_PLATFORM) || defined(STM32_PLATFORM)
```
ripplebiz, line 252: Could probably reduce a lot of the #if conditionals with a private helper method, something like:
   FILESYSTEM* getContactsFS() { ... }

examples/companion_radio/DataStore.cpp
```diff
@@ -269,11 +340,20 @@ void DataStore::saveContacts(DataStoreHost* host) {
 }
 
 void DataStore::loadChannels(DataStoreHost* host) {
```
ripplebiz, line 321: Similar here.... the helper method could maybe also incorporate this, ie. return FS pointer for contacts/channels?
   FILESYSTEM* getContactsChannelsFS() ... ?

examples/companion_radio/DataStore.cpp
oltaco: Totally agree, I wanted to do something to tidy up all the #ifs but I got a bit stuck when I got to here. Will take your suggestion and implement a private helper that returns the correct pointer.

oltaco:
Added the private helper method _ getContactsChannelsFS() and tidied up most of the ifdefs. The single FS constructor now explicitly sets _fsExtra to nullptr. DataStore::formatFileSystem() now checks for _fsExtra == nullptr to decide whether to format both filesystems.

I also made some changes to the rescue CLI to improve it for dual filesystems. File listings are prefixed with either UserData or ExtraFS make it more clear where the files are stored. ls/rm/cat commands all work as expected across both filesystems.

Did a lot of testing of migration from UserData to ExtraFS, both onchip and external flash. Also tested to make sure everything still works on esp32 devices (nothing should have changed there but better safe than sorry!).

I ran into one weird issue, one of my devices was locking up at startup when using ExtraFS (onchip) mode, but it was fine with standard UserData and QSPI mode. It turned out there must have been some corruption in the ExtraFS area beacuse I made a UF2 file which writes zeroes to the 100kb ExtraFS area and then it booted fine. Just something to be aware of.

edit: Also had some discussion with @recrof about creating some nrf_erase type firmware that can format the external QSPI flash. Unfortunately it will have to be a per-device firmware because of pin differences etc. I've got something that does this already, it just needs a bit of a tidy before release.

PR #657, opened 2025-08-24
[companion] Emit error when radio init fails

adrienthebo:
# Problem

When the example companion app inits, it will print `Please wait...`, attempt radio init, and then silently halt on radio init failure. The absence of error logging means that users can't determine if they've flashed the wrong firmware or if the radio has gotten hung during initialization.

# Solution

Print a failure message on radio init failure.

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -104,7 +104,20 @@ void setup() {
   }
 #endif
 
-  if (!radio_init()) { halt(); }
+  if (!radio_init()) {
+#ifdef DISPLAY_CLASS
+  if (disp != NULL) {
+    disp->startFrame();
+    disp->clear();
+    disp->setCursor(0, 0);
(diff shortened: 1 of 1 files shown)


ripplebiz:
Thanks for this, but we always target the 'dev' branch for PR's, plus the UI code in dev is currently quite different, with two different ui's in different folders ('ui-new', 'ui-orig').

liamcottle:
Hey, thanks for the contribution. I'll close this out since there's been no activity for 3 weeks.

We would be happy to accept a follow up PR that targets the dev branch.
I would probably opt to ensure the error messaging is much more concise.

This is a bit too much information, and won't fit on most of the small screens.

```
Radio init failure!
Device might be flashed with incompatible firmware.
```

I would probably prefer something along the lines of:

```
Radio Init Failed
Error Code: -123
```

I'm not sure if you can pull out the error code, but would be good to display that if possible

PR #662, opened 2025-08-25
lib_build: Support display and example build

fdlamotte:
builds display code depending on DISPLAY_CLASS


fdlamotte:
Now one can include code from the examples (the whole code but it should be possible to add an option to remove `main.cpp`)

using `-D BUILD_EXAMPLE=example_name` you build example stored in `example_name` dir with your own options and without modifying MC tree ...

PR #663, opened 2025-08-25
Add CLI command to remove neighbour

liamcottle:
This PR adds a new `neighbor.remove <pub_key>` command to the CLI to remove a specific neighbour from the repeater neighbours list. It supports providing a dynamic public key prefix in hex, and will remove all neighbours that match the provided prefix.

### DIFF
--- a/src/helpers/CommonCLI.h
+++ b/src/helpers/CommonCLI.h
@@ -43,7 +43,9 @@ public:
   virtual void dumpLogFile() = 0;
   virtual void setTxPower(uint8_t power_dbm) = 0;
   virtual void formatNeighborsReply(char *reply) = 0;
-  virtual void removeNeighbor(const uint8_t* pubkey, int key_len) = 0;
+  virtual void removeNeighbor(const uint8_t* pubkey, int key_len) {
+    // no op by default
+  };
   virtual mesh::LocalIdentity& getSelfId() = 0;
   virtual void clearStats() = 0;
   virtual void applyTempRadioParams(float freq, float bw, uint8_t sf, uint8_t cr, int timeout_mins) = 0;


examples/simple_repeater/main.cpp
```diff
@@ -719,6 +719,17 @@ class MyMesh : public mesh::Mesh, public CommonCLICallbacks {
     *dp = 0;  // null terminator
   }
 
+  void removeNeighbor(const uint8_t* pubkey, int key_len) override {
+#if MAX_NEIGHBOURS
+    for (int i = 0; i < MAX_NEIGHBOURS; i++) {
+      NeighbourInfo* neighbour = &neighbours[i];
+      if(memcmp(neighbour->id.pub_key, pubkey, key_len) == 0){
+        neighbours[i] = {}; // clear neighbour entry
```
ripplebiz: Apparently, = {} is a C++ 11 thing. Not sure how portable this is(?)

examples/simple_repeater/main.cpp
```diff
@@ -719,6 +719,17 @@ class MyMesh : public mesh::Mesh, public CommonCLICallbacks {
...
+      }
+    }
+#endif
+  }
+
```
ripplebiz, line 732: Also, you'll need to add some empty override of removeNeighbor() for room server and sensor sketches too.

liamcottle: Is it not possible to provide a default (no-op) implementation at the CommonCLI.h level so we don't have to copy paste empty implementations of unused functions to all the other firmware types?

examples/simple_repeater/main.cpp
liamcottle: Not too sure on that one, can convert to `memset` to set to all zeros, but as far as I know, that's not the ideal approach depending on what sort of public variables are inside of the class? I do see that's used in the default initialiser, so probably is fine.

examples/simple_repeater/main.cpp
ripplebiz: I'm probably thinking like a Java/Kotlin dev, and it being an 'interface'. But, you can probably have the abstract 'interface' have a default impl of empty body, instead of '= 0'

examples/simple_repeater/main.cpp
liamcottle: Have switched to using new instance of the struct

examples/simple_repeater/main.cpp
liamcottle: Have added a default no op implementation at the header level, so we don't need to define an empty override for other firmwares. Tested building sensor firmware, and it compiles without complaining about pure virtual function not being overridden. So this should be good to go now too

PR #673, opened 2025-08-27
add remaining roles to t1000e

446564:
usb companion, repeater and room server

### DIFF
--- a/variants/t1000-e/platformio.ini
+++ b/variants/t1000-e/platformio.ini
@@ -47,7 +47,6 @@ build_flags = ${t1000-e.build_flags}
 ;  -D MESH_DEBUG=1
   -D RX_BOOSTED_GAIN=true
   -D RF_SWITCH_TABLE
-  -D DISPLAY_CLASS=NullDisplayDriver
 build_src_filter = ${t1000-e.build_src_filter}
   +<../examples/simple_repeater>
 lib_deps = ${t1000-e.lib_deps}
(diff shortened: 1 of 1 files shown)


variants/t1000-e/platformio.ini
```diff
@@ -34,6 +34,67 @@ build_src_filter = ${nrf52840_t1000e.build_src_filter}
...
+  -D ADVERT_NAME='"t1000-e Repeater"'
+  -D ADVERT_LAT=0.0
+  -D ADVERT_LON=0.0
+  -D ADMIN_PASSWORD='"password"'
+  -D MAX_NEIGHBOURS=8
+;  -D MESH_PACKET_LOGGING=1
+;  -D MESH_DEBUG=1
+  -D RX_BOOSTED_GAIN=true
+  -D RF_SWITCH_TABLE
+  -D DISPLAY_CLASS=NullDisplayDriver
```
fdlamotte: I don't think DISPLAY_CLASS is needed ... or do we have some use for the button in repeater ? or the LED ?

the remaining looks good to me (had made the same PR just after you)

ripplebiz: Yeah, for repeater I don't think this will achieve anything, ie. there is no button press handling.

446564: I can update tomorrow, unless anyone else wants to

446564:
all set

IoTThinks:
So T1000-e can be flashed as repeaters and room servers now.

I guess the Web Flasher does not support this feature yet.

Thanks a lot.

fdlamotte:
> So T1000-e can be flashed as repeaters and room servers now.
> 
> I guess the Web Flasher does not support this feature yet.
> 
> Thanks a lot.

@recrof maybe can do something about this ;)

IoTThinks:
That is so fast.
Thanks a lot.

image

PR #680, opened 2025-08-28
add power off to nano g2

446564:


446564:
sorry not sorry about the formatting, we should format the whole tree using the committed linting rules, change them or remove them

Issue #666, opened 2025-08-25
[Proposed contribution] Implement power savings features on T114


taedryn:
I'm interested in contributing code which will save power, starting on the T114, but eventually applicable to many or all nRF-powered boards.  There are numerous features of the nRF52840 and SX1262 for reducing power consumption, which are not currently being used.  The invocation of these features on the nRF should have no impact on the functionality of the mesh: powering down unused peripherals, switching to a more efficient power source, etc.

The one option I've identified on the SX1262 might impact mesh functionality, and would require careful testing, which is switching to a lower-power mode most of the time, with occasional events to listen for incoming data.

There is a huge opportunity available in modifying the main loop to enable CPU sleep when there isn't pending work.  This is also much riskier, since it does impact the main loop, and pending-work signals aren't currently available in the core code.  For a low-traffic mesh environment, this would result in a massive power reduction (on the order of 1/5 to 1/10th the power consumption, maybe better), and even in a high-traffic environment like the Seattle area mesh, due to the low-speed nature of LoRa communications, the CPU necessarily spends a lot of time doing nothing at full speed, right now.  I suspect this could allow us to use 1/3 the power even in a high-traffic mesh.  An alternative/additional approach would be to enable lower-power clock modes, such as reducing clock frequency, without going into a full Wait For Event sleep, when activity is low.

Eventually I'd like to examine other nRF-based hardware with an eye toward implementing similar improvements.

My first work on this is already underway, to do the zero-risk stuff, like enabling a SoftDevice low power mode, turning off unused peripherals, etc.  I would be doing multiple pull requests, split up by risk level, broadly speaking.

I would appreciate feedback on whether this work would be welcomed, and what conflicts it might have with current development or planned features.

446564:
This is the type of work I would love to do but just have no time recently. There have been a few experiments but not much progress beyond testing it out. There were multiple issues with those approaches like slow to very slow reconnect over BT for example.

So I think this would be very welcome. I am happy to test ideas on my t1000e as well :)

taedryn:
Well, certainly the zero-risk stuff like turning off peripherals seems like low-hanging fruit.  This is stuff that's literally not connected to anything on the board.  Leaving it on is just warming the board for no reason.

I was looking at the BT stuff, but given that we already have complaints about nodes being hard to connect with, I'm not inclined to do anything there, at least until I understand the code a bit better.

fdlamotte:
I'll be happy to follow your work @taedryn 

[...]

PR #683, opened 2025-08-29
Improving T114 power consumption

taedryn:
This change addresses the low-hanging-fruit of power optimization: disabling unused hardware and peripherals in the CPU, and enabling the nRF's `sd_power_mode_set(NRF_POWER_MODE_LOWPWR);` mode.

This code has been tested for multiple days on a T114 as a Bluetooth companion, and found to perform identically compared to before these changes were made.  It has not yet been tested as a repeater or room server, though there should be no difference in performance under those roles.

This change takes my test hardware's standby (screen off, GPS off, receiving normally; what happens when you leave the node alone until the screen times out) power consumption from about 22 mA to about 17 mA.  This represents about a 20% reduction in power consumption, which is significant, given how much of the time nodes spend in this state.  Power consumption is reduced by a similar 2-3 mA in other states (screen on, or transmitting an advert at 0 dBm).

The risk of this change is very low.  The unused peripherals aren't in use, so powering them is just heating the board for no benefit.  The LOWPWR mode is used in a few other boards, including the T1000e.

These changes could easily be made to other nRF boards.  The LOWPWR mode is probably providing most of the benefit, and could be applied across all nRF-based devices with little risk, though obviously this change should be tested in the real world on the impacted hardware before it's rolled out across the codebase.

### DIFF
--- a/src/helpers/nrf52/T114Board.cpp
+++ b/src/helpers/nrf52/T114Board.cpp
@@ -26,6 +26,45 @@ void T114Board::begin() {
 
   pinMode(PIN_VBAT_READ, INPUT);
 
+  // Enable SoftDevice low-power mode
+  sd_power_mode_set(NRF_POWER_MODE_LOWPWR);
+
+  // Enable DC/DC converter for better efficiency (REG1 stage)
+  NRF_POWER->DCDCEN = 1;
+
+  // Power down unused communication peripherals
+  // UART1 - Not used on T114
+  NRF_UARTE1->ENABLE = 0;
+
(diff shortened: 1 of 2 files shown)


taedryn:
This PR addresses https://github.com/meshcore-dev/MeshCore/issues/666

ripplebiz:
Good stuff.

IoTThinks:
This code is good.
RAK4631 repeaters can reduce the power consumption from 12.5 mA down to 8.5 mA.

PR #692, opened 2025-08-31
ThinkNode M1: low battery auto shutdown

liamcottle:
This PR enables automatic shutdown when the ThinkNode M1 battery hits 3.3v or less.
We want to avoid writing to flash when voltage is low, as it can lead to file system corruption.

It also shows a `Low Battery. Shutting Down!` message on screen before it shuts down.

Since this is an eink display, it will persist after power loss. This is gated behind an `#ifdef THINKNODE_M1` as I don't have other eink displays to test the aesthetics of the font size and spacing.

I noticed the blue status LED was still on when powering down, so this is manually disabled as well.


recrof:
this shoud be also added to lilygo t-echo, as it uses exactly the same display.

PR #706, opened 2025-09-01
fixed max_contacts to 300 for heltec v3 companion_ble

recrof:


recrof:
@ripplebiz this one is regression from 1.8.0 to 1.8.1 due my accidental change. should be safe to merge

PR #707, opened 2025-09-01
✨ feat: add virtual keyboard

csrutil:
- Implement VirtualKeyboard class with full keyboard functionality
- Support cursor navigation, text input, and special keys (DEL, SPACE, CAPS, OK)
- Compatible with both MESHCORE and MESHTASTIC display systems
- Include multi-language layout support (EN/CN keyboards)
- Add key press callback system for extensibility
- Handle caps lock toggle and text buffer management

### DIFF
--- a/src/helpers/VirtualKeyboard.cpp
+++ b/src/helpers/VirtualKeyboard.cpp
@@ -1,0 +1,285 @@
+/*
+ * VirtualKeyboard.cpp
+ * Author TSAO (hey@tsao.dev) 2025
+ */
+
+#include "VirtualKeyboard.h"
+
+#ifdef MESHCORE
+#include "MeshCore.h"
+#elif defined(MESHTASTIC)
+#include "configuration.h"
+#include "graphics/ScreenFonts.h"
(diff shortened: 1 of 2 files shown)


ripplebiz:
PR's should have 'dev' branch as base

fdlamotte:
I find the proof of concept great and it would be a great addition !

I quickly looked into this, and I think its far from ready to be integrated though. Main showstopper I see is that it does not integrate with uitask and instead directly targets the low level libs for ssd1306 display

Still love the demo, and hope we'll be able to integrate this functionality !

I'm going to repurpose an heltec_v3 just for trying it ;)

ripplebiz:
Agree that this would instead need to be, say, a UIScreen class that defines the behaviours and renderings, etc. And the #if MESHTASTIC code really doesn't belong here. In a general-purpose lib it would tho.

liamcottle:
Thanks for the PR! The demo videos I saw on the GitHub were really cool. For now, due to lack of activity, I'll close this one out.

PR #708, opened 2025-09-01
✨ feat: add vibration feedback system

csrutil:
- Add genericVibration class for haptic feedback control
- Implement 1-second vibration pulses with 5-second cooldown
- Support non-blocking operation with loop-based timing
- Enable vibration feedback for events like new messages and contacts

### DIFF
--- a/src/helpers/ui/vibration.cpp
+++ b/src/helpers/ui/vibration.cpp
@@ -1,0 +1,43 @@
+#ifdef PIN_VIBRATION
+#include "vibration.h"
+
+void genericVibration::begin()
+{
+    pinMode(PIN_VIBRATION, OUTPUT);
+    digitalWrite(PIN_VIBRATION, LOW);
+    duration = 0;
+}
+
+void genericVibration::trigger()
+{
+    duration = millis();
+    digitalWrite(PIN_VIBRATION, HIGH);
+}
+
+void genericVibration::loop()
+{
+    if (isVibrating()) {
+        if ((millis() / 1000) % 2 == 0) {
+            digitalWrite(PIN_VIBRATION, LOW);
+        } else {
+            digitalWrite(PIN_VIBRATION, HIGH);
+        }
+
+        if (millis() - duration > 5000) {
+            stop();
+        }
+    }
+}
+
+bool genericVibration::isVibrating()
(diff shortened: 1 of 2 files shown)


ripplebiz:
PR's should have 'dev' branch as base

csrutil:
Well noted, I just changed the base branch to dev!

ripplebiz:
Thanks. Is generally structured well, however, I'd suggest we rename the soundBuzzer() method to a more generic concept, maybe alertUser(UIEventType t) ... 

And then the UITask impl can do potentially both, or one, of the current PIN_BUZZER conditional code, or this vibrate code.

4np:
Perhaps accepting a collection of event types would be useful rather than a single type. Then you could send an alert to `[.buzzer, .tactile, .speaker, .display]`.

Pseudo code in Swift:

```swift
enum UIEventType: Int {
  case buzzer
  case tactile
  case speaker
  case display

  static let all: [Self] = [.buzzer, .tactile, .speaker, .display]
}
```

ripplebiz:
Thanks for that. I have merged, but have also pushed a refactor/cleanup.

csrutil:
thanks @ripplebiz, it looks better after the cleanup. Hey @4np, it's a good idea, lets see if we need it in the future, thank you!

orsib78:
Hi! Just a quick question — I’m planning to build a BT Companion using a Heltec T114. I also have a speaker and a vibration motor for the build. Are these features already implemented? And if so, which pins on the board should I use to connect the speaker and/or the vibration motor?

RobsonMi:
> Hi! Just a quick question — I’m planning to build a BT Companion using a Heltec T114. I also have a speaker and a vibration motor for the build. Are these features already implemented? And if so, which pins on the board should I use to connect the speaker and/or the vibration motor?

As far as I can see the pins are PIN_VIBRATION and PIN_BUZZER. Logic is included in most/all boards but the pins are not defined in any board config. You'd have to define them for your board and build the firmware yourself at this time. Also default vibration timeout is 5 seconds. So every time you receive any user event other than "none" it will vibrate for 5 seconds. You can define VIBRATION_TIMEOUT in ms to overwrite this.

PR #713, opened 2025-09-02
add heltec meshpocket board.

Quency-D:
I have tested that my proposed changes behave as described.
 I have tested that my proposed changes do not cause any obvious regressions on the following devices:
 - [x] heltec-meshpocket
 - [x] heltec-t114


variants/mesh_pocket/target.h
```diff
@@ -0,0 +1,45 @@
...
+public:
+  MeshPocketSensorManager() {};
+  bool begin() override;
+  bool querySensors(uint8_t requester_permissions, CayenneLPP& telemetry);
+  void loop();
+  int getNumSettings() const override;
+  const char* getSettingName(int i) const override;
+  const char* getSettingValue(int i) const override;
+  bool setSettingValue(const char* name, const char* value) override;
+};
```
ripplebiz: Are there plans to customise SensorManager? If not, you can just declare:
  extern SensorManager sensors;

src/helpers/ui/GxEPDDisplay.cpp
```diff
@@ -77,41 +74,41 @@ void GxEPDDisplay::setColor(Color c) {
...
 void GxEPDDisplay::setCursor(int x, int y) {
-  display.setCursor(x*SCALE_X, (y+10)*SCALE_Y);
+  display.setCursor((x+offset_x)*scale_x, (y+offset_y)*scale_y);
 }
 
 void GxEPDDisplay::print(const char* str) {
   display.print(str);
 }
 
 void GxEPDDisplay::fillRect(int x, int y, int w, int h) {
```
ripplebiz: Just wondering why offset_x and offset_y are only applied in setCursor()?

Quency-D: offset_x and offset_y only affect the text cursor position, making it easy to uniformly adjust the text area, while graphics drawing maintains the original coordinates for precise control.

variants/mesh_pocket/target.h
Quency-D: The sensor section has been removed.

Quency-D:
It has been re-adapted according to commit #722.

fdlamotte:
Still works good on TEcho, no regressions from the changes ... good work !

@Quency-D you might consider adding the following flags for meshpocket
*  `-D AUTO_OFF_MILLIS=0` to get an always-on display (refresh is done only when needed now)
*  `-D DISABLE_DIAGNOSTIC_OUTPUT` this one will make GxEPD less verbose ;)

Quency-D:
> Still works good on TEcho, no regressions from the changes ... good work !
> 
> @Quency-D you might consider adding the following flags for meshpocket
> 
> * `-D AUTO_OFF_MILLIS=0` to get an always-on display (refresh is done only when needed now)
> * `-D DISABLE_DIAGNOSTIC_OUTPUT` this one will make GxEPD less verbose ;)

Thank you very much for your suggestions, that's great. I've modified both.

PR #718, opened 2025-09-02
Techo display improvements

fdlamotte:
Better management of the backlight (using the touch button)

Track changes in a screen
 * _changed (queried with has_changed) is set when a change was found with previous refresh while updating
 * _entered is set when entering (and will force a change)

for the moment, conditional refresh is only enabled when EINK_DISPLAY is defined

when refresh is not performed, the display receives a `startFrame` and no `endFrame`, but that is not really an issue as the buffer is normally cleared with next `startFrame`, maybe some double buffering setups could suffer (hence EINK_DISPLAY)

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -519,6 +519,14 @@ void UITask::loop() {
     c = handleLongPress(KEY_ENTER);
   }
 #endif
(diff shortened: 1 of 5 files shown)


recrof:
@ripplebiz this is a tremendous update for eink displays - it "cured" badly washed screens of few people already. I'm very much for merging this into dev, together with #722

ripplebiz:
OK, looks like #722 replaces this one, so I'll close this one

PR #719, opened 2025-09-02
renamend and refactored vision master, added usb roles

recrof:
initial version vision master family companion envs did not get picked up by build system, so they were missing in 1.8.0 and 1.8.1 releases.


variants/heltec_e213/platformio.ini
```diff
@@ -1,14 +1,22 @@
...
+[Heltec_E213_base]
 extends = esp32_base
-board = heltec_vision_master_e213
+board = heltec_e213
 build_flags =
   ${esp32_base.build_flags}
-  -I variants/heltec_vision_master_e213
-  -D HELTEC_VISION_MASTER_E213
+  -I variants/heltec_e213
+  -D Vision_Master_E213
```
ripplebiz, line 7: Need to be very careful changing or removing -D (defines). There don't seem to be any references to this one, however. 
Also, if introducing a new one, UPPER CASE please.

Same with the others below for E290, T190

recrof: I would use capitals, but it's part of the display library. ref: https://github.com/todd-herbert/heltec-eink-modules/blob/9207eb6ab2b96f66298e0488740218c17b006af7/src/Displays/BaseDisplay/enums.h#L16
my hands are tied.

PR #720, opened 2025-09-02
new-ui: add double/triple clicks, buzzer and gps toggle functions

oltaco:
This adds double and triple click functionality to user button, as well as buzzer and gps toggle functions, and hooks triple click to buzzer toggle for now.

I left out the old quad-press for GPS for now because it's available in the app anyway.


oltaco:
A little more work needed.

oltaco:
Much nicer with sliding window for multi click detection, and long press during multiclick detection cancels pending clicks.

PR #724, opened 2025-09-03
tidy up thinknode_m1 variant

recrof:
- moved board class files to `variants/thinknode_m1` folder
- simplified platfomio.ini definitions (as we did on other nrf52 boards already)
- moved LoRa pins from board files into platformio.ini

### DIFF
--- a/variants/thinknode_m1/ThinkNodeM1Board.cpp
+++ b/variants/thinknode_m1/ThinkNodeM1Board.cpp
@@ -1,10 +1,10 @@
-#include <Arduino.h>
 #include "ThinkNodeM1Board.h"
+#include <Arduino.h>
 
 #ifdef THINKNODE_M1
 
-#include <bluefruit.h>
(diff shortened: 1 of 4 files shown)


variants/thinknode_m1/ThinkNodeM1Board.cpp
```diff
@@ -1,10 +1,10 @@
-#include <Arduino.h>
 #include "ThinkNodeM1Board.h"
+#include <Arduino.h>
```
ripplebiz, line 2: Just wondering why you move the order of this include? Sometimes there are things like uint8_t's and other stdint types, which Arduino.h pulls in which the Board header might be depending on(?)

recrof: I can return it back.. seems it's made by formatter I use.. no idea why. I always test if all roles compile on all cleanups I do.

PR #730, opened 2025-09-04
Adds name filtering and text truncation for display in HomeScreen

michaelhart:
Adds methods to replace non-ASCII characters and truncate text with ellipsis based on available width in Recent adverts. For devices like the Nano G2 Ultra, this addresses the multiple glitched characters that a single emoji would render as and instead shows "`█`".

Lilygo T-Echo Before and After:

image
image

Nano G2 Ultra Before and After:

image
image

Note: This doesn't address text truncation of the companion node name in the top, only in the list of recently heard adverts.

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -90,6 +90,7 @@ class HomeScreen : public UIScreen {
   bool _shutdown_init;
   AdvertPath recent[UI_RECENT_LIST_SIZE];
 
+
   void renderBatteryIndicator(DisplayDriver& display, uint16_t batteryMilliVolts) {
     // Convert millivolts to percentage
     const int minMilliVolts = 3000; // Minimum voltage (e.g., 3.0V)
@@ -157,10 +158,12 @@ public:
   int render(DisplayDriver& display) override {
     char tmp[80];
     // node name
-    display.setCursor(0, 0);
     display.setTextSize(1);
     display.setColor(DisplayDriver::GREEN);
-    display.print(_node_prefs->node_name);
+    char filtered_name[sizeof(_node_prefs->node_name)];
+    display.translateUTF8ToBlocks(filtered_name, _node_prefs->node_name, sizeof(filtered_name));
+    display.setCursor(0, 0);
+    display.print(filtered_name);
 
     // battery voltage
     renderBatteryIndicator(display, _task->getBattMilliVolts());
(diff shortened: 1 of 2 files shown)


liamcottle:
Tested on ThinkNode M1, looks much better. Thanks for contributing! :) I'll leave the code review for Scott.

examples/companion_radio/ui-new/UITask.cpp
```diff
@@ -87,6 +87,47 @@ class HomeScreen : public UIScreen {
...
+      if (c >= 32 && c <= 126) {
+        dest[j++] = c;  // ASCII printable
+      } else if (c >= 0x80) {
+        dest[j++] = '\xDB';  // CP437 full block █
+        while (src[i+1] && (src[i+1] & 0xC0) == 0x80) 
+          i++;  // skip UTF-8 continuation bytes
+      }
+    }
+    dest[j] = 0;
+  }
```
ripplebiz: The method name is a bit confusing. Maybe something like 'translateUTF8toBlocks()' ?

examples/companion_radio/ui-new/UITask.cpp
```diff
@@ -87,6 +87,47 @@ class HomeScreen : public UIScreen {
...
+    }
+    
+    int ellipsis_width = display.getTextWidth(ellipsis);
+    int len = strlen(text);
+    
+    while (len > 0 && display.getTextWidth(text) > max_width - ellipsis_width) {
+      text[--len] = 0;
+    }
+    strcat(text, ellipsis);
+  }
```
ripplebiz: I have actually been meaning to introduce something like this, but at the DisplayDriver level.
If you look at drawTextCentered() for instance, there is a dumb, generic impl there, but it can be overridden by sub-classes if the particular display has a better way of doing it.
Would be good to have a new DisplayDriver method, say, drawTextEllipsized(x, y, w, str) with a basic imp like above, but where display-specific could have some optimised impl

liamcottle:
Could this be refactored into a display driver helper such as `drawTextEllipsized` as mentioned by Scott?
Would allow for using in other parts of the UI without much fuss :)

Just noticed a bunch of other commits get merged/forced in.
I'm assuming this was main into dev, or dev into main.
Maybe this PR should be rebased onto current dev branch so we don't have the mess of the other commits.

Definitely keen to see this UI improvement make it into firmware.

liamcottle:
Oh, it looks like you did that! Nice. Missed it within all those other commits :D I would definitely recommend a rebase of your branch to clean up the commit diff.

michaelhart:
Sorry I had to go on a choose your own adventure with git to get the PR clean again. I think it should be ready for review again now 😅

examples/companion_radio/ui-new/UITask.cpp
michaelhart: Done ✅

examples/companion_radio/ui-new/UITask.cpp
michaelhart: Relocated; let me know if that isn't how you had intended

PR #736, opened 2025-09-05
Introduce BLE low-power mode and enable DC/DC converter

ViezeVingertjes:
This pull request introduces several optimizations to significantly reduce power consumption for nRF52-based devices, specifically enabling them for the `t1000-e` variant.

**Changes:**

*   **Conditional BLE Low-Power Configuration:**
    *   Introduces a `BLE_LOW_POWER` compile-time flag.
    *   When enabled, this flag adjusts BLE parameters for lower power usage by:
        *   Reducing the TX power.
        *   Increasing the advertising and connection intervals.
    *   This mode is now enabled by default for the `t1000-e` variant via `platformio.ini`.

*   **Enable DC/DC Converter:**
    *   The DC/DC converter on the nRF52 is now enabled for the `t1000-e`, leading to improved power efficiency over the default LDO regulator.


fdlamotte:
Interesting ...

Had already played with ble settings (there even was a cli cmd for that in custom firmware) ... but I didn't know about enabling DC-DC converter, and it should be great

Testing that on my t1000 before merging it .. thanks !

fdlamotte:
there is a setting I'm not sure about in BLE: bandwidth

I've set it to max because it was the only way I had to get a correct MTU ... haven't you seen any problem with this setting ? For example when trying to send long messages ...

edit: that is what I thought, can't even get contacts from the device :(

ViezeVingertjes:
> Interesting ...
> 
> Had already played with ble settings (there even was a cli cmd for that in custom firmware) ... but I didn't know about enabling DC-DC converter, and it should be great
> 
> Testing that on my t1000 before merging it .. thanks !

No problem! it was in there somewhere already, just not for the T1000-E. Have been running these changes for a bit now and seems to be running stable (for me that is; so please do test).

ViezeVingertjes:
> there is a setting I'm not sure about in BLE: bandwidth
> 
> I've set it to max because it was the only way I had to get a correct MTU ... haven't you seen any problem with this setting ? For example when trying to send long messages ...
> 
> edit: that is what I thought, can't even get contacts from the device :(

hmm that's odd... i do not have many contacts though...
Can you test it by tweaking that one? without changing the others perhaps? Maybe there is a sweet spot then.

fdlamotte:
> hmm that's odd... i do not have many contacts though... Can you test it by tweaking that one? without changing the others perhaps? Maybe there is a sweet spot then.

I run a T1000e just with the DC-DC settings to test battery life (will take time), let me try on the other one 

If you try with the apps, I think it keeps the contacts from previous fetches ... you may try with mccli, it does a fresh fetch at the start ;)

Also, I'd prefer not have function parameters behind `#ifdefs` it would be more readable to place the whole functions 

I'd rather have something like 
```
#ifdef BLE_LOW_POWER
  Bluefruit.configPrphBandwidth(BANDWIDTH_NORMAL);
  Bluefruit.configPrphConn(400, BLE_GAP_EVENT_LENGTH_MIN, 8, 8)
#else
  Bluefruit.configPrphBandwidth(BANDWIDTH_MAX);
  Bluefruit.configPrphConn(250, BLE_GAP_EVENT_LENGTH_MIN, 16, 16)  // increase MTU
#endif
```
where I can clearly see what the alternatives do ;)

fdlamotte:
by reverting the bandwidth change (and gap_event_length) contacts are correctly fetched

I think this must be kept with original values ... or we don't have the good MTU (and even then MTU is low for the transfers we do)

Don't think there is much to save here, power wise ;)

ViezeVingertjes:
Maybe just minimize to this? 7363a4f67dbeb95cbecffde5f6e6ddd2ae6fcfa8 @fdlamotte 
Should be the safe route, although now BLE_LOW_POWER is only used for the tx power...

fdlamotte:
I'm ok with this one ;)

You could go lower for BLE_POWER, I've done -16 I think, maybe an int BLE_TX_POWER parameter instead of a boolean ? then we set it to 0 by default in platformio.ini (but it might upset people trying to reach their T1000 from the bedroom ...)

for that I would put a 
```
#ifndef BLE_TX_POWER
#define BLE_TX_POWER 4
#endif
```
in SerialBLEInterface.h

If you're ok, I let you do that and I'll merge the PR

ViezeVingertjes:
> I'm ok with this one ;)
> 
> You could go lower for BLE_POWER, I've done -16 I think, maybe an int BLE_TX_POWER parameter instead of a boolean ? then we set it to 0 by default in platformio.ini (but it might upset people trying to reach their T1000 from the bedroom ...)
> 
> for that I would put a
> 
> ```
> #ifndef BLE_TX_POWER
> #define BLE_TX_POWER 4
> #endif
> ```
> 
> in SerialBLEInterface.h
> 
> If you're ok, I let you do that and I'll merge the PR


5370667bd85b9c3c24d7f430a318e975dc9ab5d4

fdlamotte:
the PR looks great now, just as lean as we like them !

good job !

IoTThinks:
This PR is not in v1.8.1 yet I believe.
Looking forward to see if T1000e can have longer battery life (>1 day).

Have you tested how long T1000e can last with this PR?

ViezeVingertjes:
> This PR is not in v1.8.1 yet I believe. Looking forward to see if T1000e can have longer battery life (>1 day).
> 
> Have you tested how long T1000e can last with this PR?

Thats correct, its only in dev at this time. 
Mine has been on for 3 days now and could probably go one more easily. (i always forget to charge it...) 
Note: It probably also depends on activity so milage may vary, but 1% per hour is about where its at now, which makes sense for a 700mAh battery. Also; that 0.6-1% /h also goes for when in hibernation, as its not off completely.

IoTThinks:
Great. I will test the code from the dev branch then.
I guess this improvement will be greatly loved.
Looking forward to have it in the next release.

IoTThinks:
I'm testing the dev firmware on dev branch.
I did full charge for both T1000-e.

After 12 hours
- On Android Phone, stock v1.8.1, battery is 73%.
- On iPhone, dev firmware, battery is 84%. Expected discharge rate is 1% from your comment.

Let me see if dev firmware can survive two days.

ViezeVingertjes:
> I'm testing the dev firmware on dev branch. I did full charge for both T1000-e.
> 
> After 12 hours
> 
> * On Android Phone, stock v1.8.1, battery is 73%.
> * On iPhone, dev firmware, battery is 84%. Expected discharge rate is 1% from your comment.
> 
> Let me see if dev firmware can survive two days.

Should last two days, but the % will be off at some point. MC does not take in account that the battery will drop faster when the charge get to a lower point.

IoTThinks:
72% for dev firmware.
66% for stock v1.8.1.

I guess marginal improvement only.

The dev firmware does not have BLE disconnection on iphone anymore. May be from other PRs.

Usually, the offgrid trip may last a day. So it is only meaningful if T1000-e can last two days or more.

No idea if we can turn off the Led blinking to save more power.

ViezeVingertjes:
> 72% for dev firmware. 66% for stock v1.8.1.
> 
> I guess marginal improvement only.
> 
> The dev firmware does not have BLE disconnection on iphone anymore. May be from other PRs.
> 
> Usually, the offgrid trip may last a day. So it is only meaningful if T1000-e can last two days or more.
> 
> No idea if we can turn off the Led blinking to save more power.

Could it be skewed by the amount of activity and possibly even iOS vs Android? T1000e easily does 3 days here on dev, which was not possible before, but that might not be related to this specific PR, there are multiple PR's targeting the T1000e in dev. but any improvement is welcome i guess. ^^

trakais:
For what it's worth, I flashed Dev onto my T1000e after this change, it seems to be doing a little better, but died before 48 hours were over. And I used it only a few times to send a test message, not a lot of activity. So mileage may vary

IoTThinks:
Some chats in the morning only.
Less than 20 chat msgs.

Exactly 24 hours now.
Android with stock v1.8.1: 63%.
Iphone with dev firmware: 65%.

Just reporting only.
They may both reach another half day.

IoTThinks:
After 33h.
Stock firmware 1.8.1 with Android is out of battery. Dead. Need to charge to boot.
This is expected.

Dev firmware with iphone. Still alive with 60% at 3.72v.
I guess the lower BLE TX helps when battery level is lower.
Great.
Let's if dev node can survive 48 hours.

src/helpers/nrf52/SerialBLEInterface.cpp
```diff
@@ -80,7 +80,7 @@ void SerialBLEInterface::startAdv() {
    * https://developer.apple.com/library/content/qa/qa1931/_index.html   
    */
   Bluefruit.Advertising.restartOnDisconnect(false); // don't restart automatically as we handle it in onDisconnect
-  Bluefruit.Advertising.setInterval(32, 244);    // in unit of 0.625 ms
+  Bluefruit.Advertising.setInterval(32, 1600);
```
liamcottle, line 83: This one slipped through code review, and should be rolled back to `244` which is `0.625 * 244 = 152.5 ms`.

The value of `1600` which is `0.625 * 1600 = 1000 ms` isn't listed in the Apple recommended discovery intervals:
https://developer.apple.com/library/archive/qa/qa1931/_index.html so I'm not too sure why this was changed to 1600.

I've noticed connecting to devices (after being disconnected for 30 seconds, which is the fast interval timeout) is significantly slower due to this change. Sometimes waiting up to 20 seconds for the Android app to connect. Reverting this change, the app is able to reconnect within 2 seconds.

> The recommended advertising pattern and advertising intervals are:
> First, advertise at 20 ms intervals for at least 30 seconds
> If not discovered after 30 seconds, you may change to one of the following longer intervals: 152.5 ms, 211.25 ms, 318.75 ms, 417.5 ms, 546.25 ms, 760 ms, 852.5 ms, 1022.5 ms, 1285 ms
> **Important:** These specific intervals are the ones to be used exactly! Even the slightest deviation from these actual intervals may dramatically increase your time to discovery.

Our esp32 SerialBLEInterface implementation is using the defaults of 32/244, so reverting this back ensures a consistent setup for our nrf52 implementation.

```
# BLEAdvertising.h
#define BLE_ADV_INTERVAL_FAST_DFLT       32  // 20    ms (in 0.625 ms unit)
#define BLE_ADV_INTERVAL_SLOW_DFLT       244 // 152.5 ms (in 0.625 ms unit)
#define BLE_ADV_FAST_TIMEOUT_DFLT        30  // in seconds
```

Any changes to the advertising intervals should probably be put behind a power saving flag, with a note documenting that things will take far longer to connect in this mode.

It should be noted that the advertising interval change here only saves power when you're not connected. If you're connected to the device, it won't be advertising anyway.

ViezeVingertjes: Ah did not know about that one. Did not see any issues like that personally though. Perhaps we can add it somewhere as a reminder, comment or whatever, for the next person?

Also, can the other changes stay? As dev did extend the lifetime by a lot in combination with the other prs for a lot of us from hours to days. Don't want to accidental revert that i think.

fdlamotte: Thanks for reverting this @liamcottle I should have known as I was there when we set the value ;)

IoTThinks:
Realized just now.
4 hours before reaching 48 hours.
The node with dev firmware and iPhone was out of battery.

So less than 44 hours.
From 24+ hours to 44 hours.
Quite good actually.

I realized I could see the battery of dev firmware on T1000-e came down to 44% or even 34%.

Last time, a few hours after 60%, the node would be out of battery.

So yes, the improvement works.

src/helpers/nrf52/SerialBLEInterface.cpp
liamcottle: > Also, can the other changes stay? As dev did extend the lifetime by a lot in combination with the other prs for a lot of us from hours to days. Don't want to accidental revert that i think.

Yep, allgood. `BLE_TX_POWER` is still the same value for all other devices, it's only been dropped from `4` to `0` for the T1000E via the platformio build flags. So I won't lose BLE range (due to lower power) on my other nRF devices ;)

PR #784 was merged to revert the advertising interval change back to `244`. Other commits are probably fine to stay :)

PR #740, opened 2025-09-05
feat: Added EByte EoRa-S3-XXXTB Support

sschueller:
The EoRa-S3-400TB and EoRa-S3-900TB are the same except for the frequency. They have an E22-900-MM22S which is an sx1262.

The board appears to be almost identical to the LilyGo_T3S3_sx1262.

The following works:

- Display
- Button
- Battery percentage
- Green TX lights on Lora Send
- BLe connection
- Lora RX

The following I could not get to work yet:

- Lora TX (light flashes but adverts are never received)
- SD Card (not sure where to configure that or test it, I documented the GPIOs in the code)

Product Info: https://www.cdebyte.com/products/EoRa-S3-400TB

Schematic: https://www.cdebyte.com/pdf-down.aspx?id=2991

Test device purchased here: https://de.aliexpress.com/item/1005006593195102.html (Ad states SX1268 but received was an E22-900-MM22S which is a sx1262)

If anyone has an idea why TX may not be working please let me know. The SX1262 pins appear to be correctly defined.


ripplebiz:
PR's should target the 'dev' branch as base

PR #745, opened 2025-09-06
fix: only hide pin after successful authentication in SerialBLEInterface

ViezeVingertjes:


liamcottle:
Looks good thanks! We will look to migrate to the connection state callbacks to mark it as disconnected as well.

PR #761, opened 2025-09-09
Set correct max upload size for NRF52 companion targets, add QSPIFLASH to USB targets

oltaco:
This PR corrects the max upload size for NRF companion targets so we can't accidentally overwrite it with a big firmware in future.
It also adds QSPIFLASH for a couple of variants which were missing it for their companion USB targets.

I did consider whether the variants with QSPIFLASH enabled should keep the full max upload size, but given that EXTRAFS is still defined at the root platformio.ini level it is probably safer to reflect that.

For future reference this is how to calculate max upload size, when ExtraFS starts at D4000:
SoftDevice S140 v7.3.0 - D4000 - 27000 = AD000 / 708608 bytes in decimal
SoftDevice S140 v6.1.1 - D4000 - 26000 = AE000 / 712704 bytes in decimal

### DIFF
--- a/variants/heltec_mesh_solar/platformio.ini
+++ b/variants/heltec_mesh_solar/platformio.ini
@@ -57,6 +57,7 @@ build_flags =
 
 [env:Heltec_mesh_solar_companion_radio_ble]
 extends = Heltec_mesh_solar
+board_upload.maximum_size = 712704
 build_flags =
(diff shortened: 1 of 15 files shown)


oltaco:
Closing this PR, will resubmit a new one with updated linker script for extraFS targets.

PR #765, opened 2025-09-09
Fix incorrect maximum RAM and flash size in NRF board.jsons

oltaco:
Every NRF52 board json file has maximum_ram_size set incorrectly, and a couple of them have maximum_size (flash size) set incorrectly. It's not super important, as it's only used by the platformio size-checker to create the output for the build log and to stop you *uploading* a firmware above what you have set in maximum_size.

This PR solves that, and brings the RAK board.json in tree so that we can fix it too.

Note that neither maximum_ram_size/maximum_size settings will stop the firmware actually getting built. The linker will throw an error and stop if you exceed the limits set in the linker scripts, but those limits are not the same as what is set in the board.json files.

nrf52840_s140_v*.ld defines maximum RAM as ``LENGTH = 0x20040000 - 0x20006000`` in nrf52840_s140_v*.ld, minus another 2kb for the stack as seen in nrf_common.ld:
```
__StackSize = 1024*2;

/* Set stack top to end of RAM, and stack limit move down by size of stack_dummy section */
PROVIDE(__StackTop   = ORIGIN(RAM) + LENGTH(RAM));
PROVIDE(__StackLimit = __StackTop - __StackSize);
PROVIDE(__stack      = __StackTop);
```

This leaves us with 235520 bytes of RAM allowed by the linker.

You can confirm this by forcing a RAM usage increase with a big array, and once RAM usage gets past 235520 bytes the linker will choke: ``arm-none-eabi/bin/ld.exe: region `RAM'
overflowed by 537128960 bytes``

### DIFF
--- a/boards/wiscore_rak4631.json
+++ b/boards/wiscore_rak4631.json
@@ -1,0 +1,76 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "nrf52840_s140_v6.ld"
+    },
+    "core": "nRF5",
+    "cpu": "cortex-m4",
+    "extra_flags": "-DARDUINO_NRF52840_FEATHER -DNRF52840_XXAA",
+    "f_cpu": "64000000L",
+    "hwids": [
+      [
+        "0x239A",
+        "0x8029"
+      ],
+      [
+        "0x239A",
+        "0x0029"
+      ],
+      [
+        "0x239A",
+        "0x002A"
+      ],
+      [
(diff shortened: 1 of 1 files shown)


weebl2000:
@oltaco is this PR still needed? I'm having a look at older PRs and see if they're still relevant.

weebl2000:
@oltaco please see https://github.com/oltaco/MeshCore/pull/5

PR #767, opened 2025-09-10
Fix: WisMeshTag power off and wake up

liamcottle:
This PR fixes the issue where holding the user button to power off, will not fully power off until the user releases the button, and then presses it once more.

The power off logic was expecting HIGH to indicate button press, however this board uses LOW to indicate button press.

Additionally, pressing the button to wake the board when it was in powered off state was not working. This has also been adjusted to use a pull up, and sense for low state to trigger power on.

Requesting review/test by @recrof as I believe you have this board :)


recrof (approved):
tested and works. you have my blessing!

PR #770, opened 2025-09-10
Normalize repeater target names

jbrazio:
Fix a typo with `variants/lilygo_tlora_v2_1` and take the opportunity to normalize all targets to lower case.

### DIFF
--- a/variants/lilygo_t3s3/platformio.ini
+++ b/variants/lilygo_t3s3/platformio.ini
@@ -33,7 +33,7 @@ lib_deps =
   adafruit/Adafruit SSD1306 @ ^2.5.13
 
 ; === LilyGo T3S3 with SX1262 environments ===
-[env:LilyGo_T3S3_sx1262_Repeater]
+[env:LilyGo_T3S3_sx1262_repeater]
 extends = LilyGo_T3S3_sx1262
 build_flags =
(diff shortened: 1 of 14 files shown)


jbrazio:
@ripplebiz can you review this ? Thanks.

jbrazio:
@liamcottle @andymux @fdlamotte 15 days to review a basic PR.. do you guys need a hand ? I can help. 🥸

liamcottle:
Hey there,

Thanks for the tag. This might be a "basic PR", but it's also low priority...

We've had a lot of changes being merged in, and the last few releases have resulted in broken firmwares.

We don't really want to be rushed to merge in every change that presents itself.

Things like the web flasher also need to be discussed, as these changes would likely result in the .bin and .uf2 filenames changing case, which may or may not cause issues for the web flasher if it's expecting the old filename casing.

We appreciate the offer, but we will review changes as we can :)

recrof:
@jbrazio I'm strongly recommending you to be patient.. this attitude does not make good blood in the team.

jbrazio:
Please don’t overreact; that wasn’t my intention. The last part of my comment was meant to be a playful joke. However, the point I raised is still valid: "review" simply refers to providing feedback, and letting something sit for two weeks without any communication isn’t beneficial for anyone involved.

@liamcottle @recrof

PR #771, opened 2025-09-11
T114 CPU sleep

taedryn:
This change adds two new features: the Mesh object gains the ability to report when it has pending activity (both in the near-term future, and in the immediate future), and the companion radio example gains the ability to send the CPU into a Wait For Event sleep when there's no pending activity detected.  This latter feature is limited to only the T114 in this change, since that's the only hardware I have.  The change should be applicable to any companion with an nRF processor, but it needs to be tested on specific hardware targets before it can be rolled out for general release.

I've real-world tested this code for over a week, and a TFT-equipped T114 with a 3000 mAh battery lasted about 8 days without any external power input, and was just as stable as without this change.  I consider this to be a low-risk modification to the companion software.  I'll explore making these changes to the repeater software in a separate PR.

These changes appear to drop standby power consumption (GPS off, screen off, no mesh activity) to about 13 mA, from around 17 with my previous "disable unused peripherals" change, and around 20 mA from the code as I found it before disabling peripherals.

I also commented the elaborate #ifdef/else/endif statements in the `setup()` function of the companion's `main.cpp` file so I could keep them clearer in my head.

### DIFF
--- a/variants/heltec_t114/T114Board.cpp
+++ b/variants/heltec_t114/T114Board.cpp
@@ -28,7 +28,7 @@ void T114Board::begin() {
   sd_power_mode_set(NRF_POWER_MODE_LOWPWR);
 
   // Enable DC/DC converter for better efficiency (REG1 stage)
-  sd_power_dcdc_mode_set(NRF_POWER_DCDC_ENABLE)
+  sd_power_dcdc_mode_set(NRF_POWER_DCDC_ENABLE);
 
   // Power down unused communication peripherals
   // UART1 - Not used on T114


jbrazio:
This is a step into the right direction, the overall implementation is what concerns me in the long term. I feel we need to move into the direction of a centralised `PowerManager`.

ripplebiz:
Appreciate the efforts here, but agree that a new PowerManager class is prob better strategy, where at start of loop() it clears some flag/state, and all the various sub-systems like UI or Mesh, etc, can set some indicator of if they are 'busy' or maybe set some estimate of milliseconds when they next need to do 'work'. Then at end of loop(), PowerManager can do some minimum calc and determine what is a 'safe' amount of time that CPU can sleep for

taedryn:
I'm not familiar enough with the majority of the code base to make changes like that yet. I agree that's a better approach, but I would like this change to move forward, perhaps under an experimental #ifdef flag for others to test with, so we can get more experience with these techniques across more devices.

Ultimately, the MeshCore algorithms need to move away from a loop mentality and into an event/interrupt driven mentality to make real headway on power reduction. The work I'm doing will always be a bandaid compared to actually refactoring the code to be power friendly. A power manager is a good interim step between my piecemeal code and a full event-driven overhaul, until that work can take place.

I would also note that we don't need to calculate sleep times at all, at least for the nRF processors. The WFE sleep specifically sets the CPU to wake on any event, so there's no risk of missing incoming stimulus, as I understand it.

Vasili-Sk:
i agree that event/interrupt way of operation is the most preferable. then on idle task cpu will go to sleep with WFE and after receiving any data - wake up and go doing tasks

fschrempf:
> Ultimately, the MeshCore algorithms need to move away from a loop mentality and into an event/interrupt driven mentality to make real headway on power reduction. The work I'm doing will always be a bandaid compared to actually refactoring the code to be power friendly

That is very true! Unfortunately it won't be easy to do this in full scale with the current firmware codebase as in some cases the existing Arduino core and libraries prevents using interrupt-driven approaches or makes using them unnecessarily complicated.

That's why I hope that one day there will be a fully event-driven firmware (probably in Rust) that makes power-saving the default from the ground up and not something that needs hours of additional development and headaches.

liamcottle:
Hey, thanks for the PR! Lots has changed in the firmware regarding power saving over the last 5 months since this PR was opened. Based on other comments here, a different approach has been looked at. I'll close this one out for now.

PR #775, opened 2025-09-12
Add battery voltage within battery icon in companion ui 

JamieKyriacou:
- Add battery voltage within battery icon in companion ui for the new ui.
- Reduce the size of the battery full bar so that it is below the text.
- I have tested this on my Heltec t114, but since I don't have a device that uses the orig-ui I haven't changed that code. I have the code prepared if it was to be included as well.
- Since it seems like #413  hasn't be merged, this may be a good interim solution. Especially since some may argue the voltage is more important to see than the percentage.

IMG_3944


liamcottle:
Hey, thanks for the PR!

Lots has been happening in the project, so we apologise for the slow follow up.

For now, I'm closing out older PRs that haven't had any traction.

Issue #411, opened 2025-06-19
Pairing PIN disappears to quickluy


SkyperTHC:
The bluetooth pairing pin is shown for perhaps 5 seconds. THat's not long enough for many users to complete the entire pairing process.

It should be shown longer, or forever until a key is pressed.

446564:
I agree, I think longer but not forever.

The reason being is that if you have paired to a phone and disconnect and reconnect the radio, you don't need the code again. So staying lit up would not serve a purpose there, the previous pairing remains valid but the device provides a new code in case you need to pair to another device etc.

recrof:
i think he's referencing behaviour of current implementation we have - if you connect to the board and the phone/computer shows pairing dialog with pin input, pin disappears from the board screen. I seen this happen on nRF OLED devices. seems that board thinks it's already connected and thuss UITask removes the pin from display.

liamcottle:
Fixed for nRF52 devices in #776

PR #776, opened 2025-09-12
Fix: BLE pin disappearing too quickly on nrf52 devices

liamcottle:
This PR fixes the issue where the BLE pin displayed on screen disappears very quickly after connecting, even if you haven't typed the pin in yet. The UI shows `< Connected >` and the user has a bad time unless they memorised the pin quick enough before it disappears.

I've switched to using the `onSecured` callback, instead of the `onConnected` callback, to mark the device as connected. This callback is fired when pairing has been successful, or when an app reconnects when already previously paired.

The BLE pin now stays on screen until the user has successfully paired.

This has been tested, and working on the following devices:
- ThinkNode M1
- Heltec T114
- RAK4631 WisMesh Pocket

The PR only makes the change on nrf52, as the esp32 `SerialBLEInterface` appears to mark devices as connected in the `onMtuChanged` callback, which seems to only happen after pairing is complete.

I do note that the esp32 class is still using the `checkRecvFrame` to update device connection state, and this should really be refactored to use the callbacks as well, however it has been left as is to keep this PR simple.

### DIFF
--- a/src/helpers/nrf52/SerialBLEInterface.cpp
+++ b/src/helpers/nrf52/SerialBLEInterface.cpp
@@ -4,10 +4,7 @@ static SerialBLEInterface* instance;
 
 void SerialBLEInterface::onConnect(uint16_t connection_handle) {
   BLE_DEBUG_PRINTLN("SerialBLEInterface: connected");
-  if(instance){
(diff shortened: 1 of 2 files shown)


liamcottle:
Fixes #411

PR #777, opened 2025-09-12
[Docs] Add companion not showing up over Bluetooth to FAQ

bryantkelley:
Adds an FAQ to make sure people flashed the correct firmware when their companion isn't showing up on bluetooth.


liamcottle:
Thanks!

PR #778, opened 2025-09-12
Add GPS duty cycling - GPS Refactor

awolden:
Refactors GPS system in the name of better handling duty cycling. Kind of pulled on the thread here and took some feedback from @fdlamotte to refactor the gps system.

* Pulls all major GPS logic into its own manager
* Moves all board specific logic into its own providers
* Updates GPS manager to read from available prefs
* Adds in validation huerestics for determining valid gps signal
* Syncs clock when a valid GPS reading happens
* will support gps interval pref when fully wired

**Testing:**
* Haven't tested NMEA 
* Tested with rak and 12500

**Todo:**

- [ ] Test with NMEA
- [ ] Update all variant files as needed

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -145,3 +145,4 @@ lib_deps =
   adafruit/Adafruit_VL53L0X @ ^1.2.4
   stevemarple/MicroNMEA @ ^2.0.6
   adafruit/Adafruit BME680 Library @ ^2.0.4
+  sparkfun/SparkFun u-blox GNSS Arduino Library@^2.2.27
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -95,13 +95,6 @@ static Adafruit_MLX90614 MLX90614;
 static Adafruit_VL53L0X VL53L0X;
 #endif
 
-#if ENV_INCLUDE_GPS && defined(RAK_BOARD) && !defined(RAK_WISMESH_TAG)
-#define RAK_WISBLOCK_GPS
-#endif
-
-#ifdef RAK_WISBLOCK_GPS
-#include "location/UbloxI2CGpsDriver.h"
-#endif
 
 bool EnvironmentSensorManager::begin() {
   #if ENV_INCLUDE_GPS
--- a/src/helpers/sensors/location/LocationProvider.cpp
+++ b/src/helpers/sensors/location/LocationProvider.cpp
@@ -1,7 +1,9 @@
 #include "LocationProvider.h"
 #include <Arduino.h>
 
-#include "UbloxI2CGpsDriver.h"
+#ifdef GPS_UBLOX_I2C
+  #include "UbloxI2CGpsDriver.h"
+#endif
 #include "MicroNMEAGpsDriver.h"
 
 const long LOOP_THROTTLE_MS = 1000;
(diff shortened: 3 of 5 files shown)


4np:
It would be nice if the polling interval could be variable, based on accelerometer values (for example 3-axis [`RAK1904`](https://store.rakwireless.com/products/rak1904-lis3dh-3-axis-acceleration-sensor?srsltid=AfmBOorSudrsmaYObTS5btm2wtjN2nuhW_FaQ7bcM8QtmLv7PgDDARe4) or 9-axis [`RAK12034`](https://store.rakwireless.com/products/rak12034-9-axis-accelerometer?srsltid=AfmBOopJ9IRRCRxkeXmfVkMqwFbcLObrSNCImuXtAwSS32xEj32pczl2&variant=41843038814406)).

If there's no accelerometer, default to 5 mins, and if there's an accelerometer and acceleration is detected then adjust the GPS update interval based on the acceleration rate.

awolden:
@4np ya I considered adding some more sophistication to it, we could wait for a minimum accuracy, take an average over 3 readings, waiting for accelerometer values, I didn't want to try add too much too fast when some of this can be layered on. (plus I don't have an accelerometer to test with)

4np:
Yeah makes sense, it probably deviates too far from the scope of the PR. Still something that would be nice to have on the roadmap at some point.

fdlamotte:
As I've played a little with gps lately I think I'll handle this one ;) I think idea and implementation are good

@awolden have you considered implementing pwm in the LocationProvider instead of sensor manager ? I think it might make sense ...

For debugging, the new ui page I propose in #820 will be good, it shows sat count, fix status and coordinates as they come from nmea ...

I'll wait for #820 before going further, as it removes unnecessary duplicate code between sensor manager and location provider ;)

awolden:
I can move it, will have time to do that soon

fdlamotte:
@ripplebiz you were saying a rewrite of the gps part would probably be necessary ...

We should probably take the opportunity of this PR to try rationalizing things a bit around GPS ... because, here we have gps -> locationProvider -> gpsmanager -> SensorManager ... and everyone has his own loop ;)

awolden:
Ya something can probably be improved with that. `LocationProvider` no longer has a loop in this case though. It was replaced with a `sync` method. In this usage the location providers only handle impl specific IO.  `GpsManager` has the only loop.

fdlamotte:
This is getting really interesting ;)

I wanted to try on T1000 (my techo is at home right now), I can't get it compiled without rewriting some code because T1000 does not use EnvironmentSensorManager (t1000SensorManager was there before and t1000 does not have any sensor from ESM but has its own sensors hooked to adcs, so there is no reason to use ESM ...)

I would prefer sticking with LocationProvider handling all the gps logic. I agree there is a fundamental issue with behaviour inheritance (don't know if thats the term in english) ... at least, your proposal does a very good cleaning both in ESM and LocationProvider, which is good ...

In the semantics, if we were to use composition, I would rather have LocationProvider delegating to a GpsDriver than GpsManager delegating to a LocationProvider. Or we could have GpsManager and GpsDriver (but LocationProvider sounds a little too high level in my ears).

I'll try to make your implementation work with T1000, it will help me understand it better ... (T1000 was the first to get gps, LocationManager has been made for it in the old custom fw)

awolden:
@fdlamotte The inheritance route gets a little wonky with the detection logic and hard to parse with the way the individual providers inherit from the super class. Would have to put the detection logic back into the SensorManager. The around keeping LocationProvider and considering the implementation specific classes 'drivers' is a great one. I refactored it around that idea. 

LocationProvider now handles the GPS management logic
GpsDriver now handles the specific IO logic

liamcottle:
Closing out due to lack of activity and merge conflicts. I believe several PRs have been merged semi recently regarding gps interval config via cli.

PR #786, opened 2025-09-15
tidy up xiao nrf52 variant

recrof:
- add all sensors supported by env sensor manager(apart from GPS - rx/tx pins are used by i2c)
- remove unnecessary nrf52840_xiao pio def
- remove alternative pinout envs (consulted with @fdlamotte)
- make source formatting in line with `.clang-format`


fdlamotte:
Could test it as companion firmware with bme280 ... went smooth (from compiling to getting sensor value ;))

For the alternative pinouts I'm totally in-line. We should only propose default pinouts in tree (for baking the officially supported boards for the flasher). 
For alternative pinouts and custom boards, the out-of-tree builds offers greater flexibility.

PR #787, opened 2025-09-15
fix building errors for wismesh tag companion ble

recrof:


recrof:
here is the build log: https://files.brazio.org/meshcore/nightly/report.txt

PR #795, opened 2025-09-16
Use python3 not python in build.sh

tahnok:
Since the bin/uf2conv/uf2conf.py script uses python3, use python3 as the command instead of python. On my ubuntu 24.04 machine, I don't have a python command in my path by default


liamcottle:
Thanks!

PR #796, opened 2025-09-16
make offline queue channel messages mutable

446564:
older channel messages can be overwritten, keeping other mssagage types

this allows a user to be away for a long time and still get the most recent channel messages without losing any direct messages for exampe

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -177,13 +177,29 @@ void MyMesh::updateContactFromFrame(ContactInfo &contact, uint32_t& last_mod, co
 
 void MyMesh::addToOfflineQueue(const uint8_t frame[], int len) {
   if (offline_queue_len >= OFFLINE_QUEUE_SIZE) {
-    MESH_DEBUG_PRINTLN("ERROR: offline_queue is full!");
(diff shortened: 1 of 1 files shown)


ripplebiz:
OK, have merged, but I'll do a follow-up commit, as I think the delete logic should have i < offline_queue_len - 1

446564:
> OK, have merged, but I'll do a follow-up commit, as I think the delete logic should have i < offline_queue_len - 1

Wouldn't that mean from idx 0-254? Because it's less than and not less than or equal to.

ripplebiz:
> > OK, have merged, but I'll do a follow-up commit, as I think the delete logic should have i < offline_queue_len - 1
> 
> Wouldn't that mean from idx 0-254? Because it's less than and not less than or equal to.

If you look in the delete loop, you'd be _reading_ off the end of the array, and writing that into the last element.

446564:
> > > OK, have merged, but I'll do a follow-up commit, as I think the delete logic should have i < offline_queue_len - 1
> > 
> > 
> > Wouldn't that mean from idx 0-254? Because it's less than and not less than or equal to.
> 
> If you look in the delete loop, you'd be _reading_ off the end of the array, and writing that into the last element.

Ah ya, I get it now. Thanks for catching.

PR #806, opened 2025-09-20
Safer _countLfsBlock / _getLfsUsedBlockCount

oltaco:
Adds a bounds check to _countLfsBlock / _getLfsUsedBlockCount and passes the error back up to lfs_traverse so it can exit gracefully in the case of lfs_traverse following a bad skip list instead of attempting to read out of bounds blocks and crashing.

This stops the device from insta-hang/bootloop but it doesn't address the corruption. Maybe it would be better to call format() instead of just printing a MESH_DEBUG log?

### DIFF
--- a/examples/companion_radio/DataStore.cpp
+++ b/examples/companion_radio/DataStore.cpp
@@ -42,12 +42,17 @@ static File openWrite(FILESYSTEM* fs, const char* filename) {
 #endif
 }
 
+#if defined(NRF52_PLATFORM) || defined(STM32_PLATFORM)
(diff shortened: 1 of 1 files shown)


liamcottle:
Nice find, I added the block counting code a while back so we could check how much storage was used.

Also fun fact, the reason this is probably happening when connecting to BLE, is because the mobile app fetches the battery voltage after connect.

The firmware also returns the storage usage in that same `CMD_GET_BATT_AND_STORAGE` request/response...

PR #811, opened 2025-09-21
Tracker l1: environment sensors

fdlamotte:
This PR should be added once wio_tracker_eink is in

* I had to make some modifications to the variant.h, i2c pins on the grove connector were inverted
* also a modification to EnvironmentSensorManager was necessary because on nrf52 there is no begin function that let the user set the pins and frequency, this has to be done before
* only BME280 and GPS have been added ... when all sensors are in the node hangs at boot, so I've only included tested sensors
* MicroNMEA is set-up to sync rtc on first gps fix


ripplebiz:
Looks like you'll need to update your PR branch with current dev

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -108,7 +108,13 @@ bool EnvironmentSensorManager::begin() {
   #endif
 
   #if ENV_PIN_SDA && ENV_PIN_SCL
+    #ifdef NRF52_PLATFORM
+  Wire1.setPins(ENV_PIN_SDA, ENV_PIN_SCL);
+  Wire1.setClock(100000);
+  Wire1.begin();
```
ripplebiz, line 114: SO, this is for ALL nRF targets. Can you verify this is correct for all nRF variants?

fdlamotte:
I'll do so, but nevertheless, this PR has power draw issue ...

I'll try to check where it comes from, but battery doesn't hold ... (I left the node at more than 4V an it had dropped to 3.75V this morning)

given the previous runs I did not expect less than 3.9

oltaco:
@fdlamotte It's probably because the GPS isn't being put into standby like it was when the GPS code was in target.cpp.

``  digitalWrite(PIN_GPS_STANDBY, LOW); // Put GPS back into standby mode``

Tbh It feels like too much board specific code has made it into EnvironmentSensorManager now, mainly due to GPS stuff. Perhaps GPS stuff could be kept in boards and accessed via callbacks or something?

src/helpers/sensors/EnvironmentSensorManager.cpp
fdlamotte: The thing is that only Heltec_v3 uses `ENV_PIN_SDA`, and when I defined it, it did not compile on tracker_l1

You just can't do `Wire1.begin(ENV_PIN_SDA, ENV_PIN_SCL, 100000);` on nrf52, seems to be an esp32 only thing

```
[florent@fl380:~/src/meshcore/MeshCore] (tracker_l1_environment_sensors)
$ grep -r ENV_PIN_SDA
variants/heltec_v3/platformio.ini:  -D ENV_PIN_SDA=33
variants/wio-tracker-l1/platformio.ini:  -D ENV_PIN_SDA=PIN_WIRE1_SDA
variants/wio-tracker-l1-eink/platformio.ini:  -D ENV_PIN_SDA=PIN_WIRE1_SDA
src/helpers/sensors/EnvironmentSensorManager.cpp:#if ENV_PIN_SDA && ENV_PIN_SCL
src/helpers/sensors/EnvironmentSensorManager.cpp:  #if ENV_PIN_SDA && ENV_PIN_SCL
src/helpers/sensors/EnvironmentSensorManager.cpp:  Wire1.setPins(ENV_PIN_SDA, ENV_PIN_SCL);
src/helpers/sensors/EnvironmentSensorManager.cpp:  Wire1.begin(ENV_PIN_SDA, ENV_PIN_SCL, 100000);
src/helpers/sensors/EnvironmentSensorManager.cpp:  MESH_DEBUG_PRINTLN("Second I2C initialized on pins SDA: %d SCL: %d", ENV_PIN_SDA, ENV_PIN_SCL);```

fdlamotte:
> @fdlamotte It's probably because the GPS isn't being put into standby like it was when the GPS code was in target.cpp.
> 
> ` digitalWrite(PIN_GPS_STANDBY, LOW); // Put GPS back into standby mode`

I've mapped PIN_GPS_STANDBY to PIN_GPS_EN thinking it would be enough (they both power down gps when low) I'll check with debugger ;)

oltaco:
Take a look at the variant.h, PIN_GPS_EN is defined as D18 already (this came from seeed's variant.h but it was commented out there, and obviously wrong like so much else).

I think if you remove PIN_GPS_EN or just change it to (0) then the power draw will probably be gone.

fdlamotte:
the variant files from seeed are sooo buggy !

edit: And its even worse that I thought because 18 is an I2C pin !!!

PR #813, opened 2025-09-22
add heltec v4 board.

Quency-D:
This board has PA, so to comply with regulations, the maximum output of sx1262 is 10.

### DIFF
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -433,8 +433,19 @@ void EnvironmentSensorManager::initBasicGPS() {
   // Try to detect if GPS is physically connected to determine if we should expose the setting
   #ifdef PIN_GPS_EN
     pinMode(PIN_GPS_EN, OUTPUT);
+  #ifdef PIN_GPS_EN_ACTIVE
+    digitalWrite(PIN_GPS_EN, PIN_GPS_EN_ACTIVE);   // Power on GPS
+  #else
     digitalWrite(PIN_GPS_EN, HIGH);   // Power on GPS
   #endif
+  #endif
+
+#ifdef PIN_GPS_RESET
+    pinMode(PIN_GPS_RESET, OUTPUT);
+    digitalWrite(PIN_GPS_RESET, PIN_GPS_RESET_ACTIVE); // assert for 10ms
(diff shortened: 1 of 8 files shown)


ripplebiz:
I just started experimenting with the V4 board I was sent,  then .. your PR came in!

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -433,8 +433,19 @@ void EnvironmentSensorManager::initBasicGPS() {
...
+  #ifdef PIN_GPS_EN_ACTIVE
+    digitalWrite(PIN_GPS_EN, PIN_GPS_EN_ACTIVE);   // Power on GPS
+  #else
     digitalWrite(PIN_GPS_EN, HIGH);   // Power on GPS
   #endif
+  #endif
+
+#ifdef PIN_GPS_RESET
+    pinMode(PIN_GPS_RESET, OUTPUT);
+    digitalWrite(PIN_GPS_RESET, PIN_GPS_RESET_ACTIVE); // assert for 10ms
```
fdlamotte, line 445: It would have been nice to give a default value to PIN_GPS_RESET_ACTIVE ... ;)
Because now compilation breaks on TEcho (at least)

I think reset is handled in MicroNMEALocationProvider ...

I'll see how I can clean this mess

fdlamotte: this is where gps reset is done

https://github.com/meshcore-dev/MeshCore/blob/ea13fa899ee1df2bcd5fee82fd47e20d2ec160b9/src/helpers/sensors/MicroNMEALocationProvider.h#L51-L57

I agree the `GPS_RESET_FORCE` symbol could be changed to `PIN_GPS_RESET_ACTIVE` but it has been here for a long time now ...

fdlamotte: the big issue is that EnvironmentSensorManager is a big mess regarding GPS management ...

I won't be able to fix everything ... will just try to make things work, but there is a need for some rewriting here ...

fdlamotte: See #820 

I've tried to use the symbols introduced in this PR ... so it should work on heltec_v4, but would need to be verified

PR #820, opened 2025-09-23
ESM: delegate gps management to LocationProvider

fdlamotte:
There was duplicate code (and duplicate symbols) that led to lots of confusion

Tried to use gps init routines from LocationProvider in ESM, instead of having the same code (with different symbol names)

### DIFF
--- a/src/helpers/sensors/MicroNMEALocationProvider.h
+++ b/src/helpers/sensors/MicroNMEALocationProvider.h
@@ -5,15 +5,31 @@
 #include <RTClib.h>
 
 #ifndef GPS_EN
-#define GPS_EN (-1)
+    #ifdef PIN_GPS_EN
+        #define GPS_EN PIN_GPS_EN
+    #else
+        #define GPS_EN (-1)
+    #endif
+#endif
+
+#ifndef PIN_GPS_EN_ACTIVE
+    #define PIN_GPS_EN_ACTIVE HIGH
 #endif
 
 #ifndef GPS_RESET
-#define GPS_RESET (-1)
+    #ifdef PIN_GPS_RESET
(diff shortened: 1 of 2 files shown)


recrof:
this one is also important for fixing failing t-echo build: https://files.brazio.org/meshcore/nightly/report.txt

Quency-D:
Great, this looks cleaner.

fdlamotte:
> Great, this looks cleaner.

Hi @Quency-D, really can't blame you because, there was (and it's still not totally fixed) a mess here. In the end we owe you one because it forced me to act ;)  

I'm seeing some interest in #778 and think this could be handed to location provider too ... I've let a message there

src/helpers/sensors/MicroNMEALocationProvider.h
```diff
@@ -40,12 +56,12 @@ public :
     }
 
     void begin() override {
+        if (_pin_en != -1) {
+            digitalWrite(_pin_en, PIN_GPS_EN_ACTIVE);
+        }
         if (_pin_reset != -1) {
             digitalWrite(_pin_reset, !GPS_RESET_FORCE);
```
ripplebiz, line 63: Just wondering if the begin() also needs to do the same reset logic as before... with the delay(10) etc?

fdlamotte: I'll be awk most of the day ...

But the idea here was power on through begin() and reset() to reset (I even wait 100ms here which is a lot ...)

fdlamotte: I rethought about it and called reset after begin to force reset (and keep exact same behaviour of ESM as it has been used for months ;) )

should be ok now

PR #821, opened 2025-09-23
ui_task: initial gps page

fdlamotte:
only shows if selected at build time ;)

added a getLocationProvider function in SensorManager (defaults to NULL)


examples/companion_radio/ui-new/UITask.cpp
```diff
@@ -75,6 +75,9 @@ class HomeScreen : public UIScreen {
     RADIO,
     BLUETOOTH,
     ADVERT,
+#if UI_GPS_PAGE == 1
```
ripplebiz: Could have this enabled by   #if ENV_INCLUDE_GPS   ?

fdlamotte: If this adds value for every user, why not

info on gps page is raw info from location provider ... coords will get back to 0 if there is no fix. this is mostly a debug interface

fdlamotte: let's use ENV_INCLUDE_GPS

as a side note, I remarked the top dots were not centered after adding pages ;)

fdlamotte: there is an issue with using ENV_INCLUDE_GPS, it indicates GPS support is built in, but gps can be absent ...

examples/companion_radio/ui-new/UITask.cpp
```diff
@@ -250,6 +253,47 @@ class HomeScreen : public UIScreen {
...
+          display.width()-display.getTextWidth(buf)-1, y);
+        display.print(buf);
+        y = y + 12;
+        display.setCursor(0,y);
+        display.print("pos");
+        sprintf(buf, "%.4f %.4f", 
+          nmea->getLatitude()/1000000., nmea->getLongitude()/1000000.);
+        display.setCursor(
+          display.width()-display.getTextWidth(buf)-1, y);
+        display.print(buf);
```
ripplebiz: I've been meaning to introduce a new DisplayDriver::drawTextRightAlgin(), like the drawTextCentered().
Helps slim-down the UI code, but also introduces ways specific sub-classes (of DisplayDriver) can implement the right-aligning. Some GFX like Lovyan can do this for you (and highly optimised)

fdlamotte: I'll propose something then, seems better

src/helpers/ui/DisplayDriver.h
```diff
@@ -32,6 +32,15 @@ class DisplayDriver {
     setCursor(mid_x - w/2, y);
     print(str);
   }
+  virtual void drawTextRightAlign(int x_margin, int y, const char* str) {
```
ripplebiz: The X param is usually the _anchor_, ie. the end coordinate. eg. you might want to right align, but with the edge in the middle of the display. So, caller passes the x coord of the Right edge.

fdlamotte: I see ... maybe we could say that negative values of x are anchored to the right ? We would have best of both world ?

fdlamotte:
find it practical with negative values ... and much more consistent ;)

ripplebiz:
> find it practical with negative values ... and much more consistent ;)

Interesting, but that could backfire quite easily. I have some scrolling code, for instance, which uses equivalent of drawLeftAligned() but passes x value that is off the left side of screen, ie. to scroll text horizontally.  You wouldn't be able to do this with this scheme

fdlamotte:
ok, now no side-effect or hidden behavior ;)

PR #825, opened 2025-09-24
Enhance Station G2 configuration to support environment sensors and G…

zjs81:
…PS. Updated platformio.ini to include sensor_base dependencies and modified target files to initialize EnvironmentSensorManager with GPS support conditionally.

### DIFF
--- a/variants/station_g2/target.cpp
+++ b/variants/station_g2/target.cpp
@@ -14,7 +14,13 @@ WRAPPER_CLASS radio_driver(radio, board);
 
 ESP32RTCClock fallback_clock;
 AutoDiscoverRTCClock rtc_clock(fallback_clock);
-SensorManager sensors;
+
+#if ENV_INCLUDE_GPS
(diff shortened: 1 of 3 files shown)


recrof:
looks good

variants/station_g2/target.cpp
```diff
@@ -25,6 +31,9 @@ bool radio_init() {
   fallback_clock.begin();
   rtc_clock.begin(Wire);
 
+  // Initialize sensors
+  sensors.begin();
```
ripplebiz, line 35: sensors.begin() is called by the various firmwares already, ie. should Not be in target.cpp

liquidraver:
If I attach a GPS to the "GROVE GPS" port, my setup works with this:
-D PIN_GPS_RX 15
-D PIN_GPS_TX 7

variants/station_g2/target.cpp
zjs81: Thanks I'll make that change this weekend.

recrof:
this is already implemented in https://github.com/meshcore-dev/MeshCore/pull/922

PR #829, opened 2025-09-24
 Feat: add `T-Echo-Lite` Variant to MeshCore

SoulOfNoob:
https://github.com/meshcore-dev/MeshCore/issues/824


SoulOfNoob:
wrong source branch..

PR #830, opened 2025-09-24
 Feat: add `T-Echo-Lite` Variant to MeshCore

SoulOfNoob:
https://github.com/meshcore-dev/MeshCore/issues/824


fdlamotte:
seems ok for base branch 
will have a look later, at work atm ...

SoulOfNoob:
I still have to fix the GPS but probably won't have time until thursday. 
I think it's a new module and will take a bit more effort so i would open another PR if i get it working locally right? or should i cram it into this one as well even if it takes a while?

fdlamotte:
Nice to see you've managed to handle display class ;)

For the gps it would be better if you have to wait a little .. there are some pending changes on #820 so a little wait and then we'll introduce it ;)

fdlamotte:
In the end there is not even a change in the display class (some welcome cleaning) ;)

I could validate there is no issue on techo and on tracker_l1_eink, the two other variants relying on gxepd2 are very similar to those two so I expect it should work.

As discussed we should take a fork of gxepd2 under meshcore-dev umbrella (@recrof) 

I 'll let the new device in congrats ! Pleas keep an eye on it, as I'm not sure anyone else has it yet ...

SoulOfNoob:
Coool, thanks for the effort. 

Since i plan on using it as my EDC device i will most certanly keep an eye on it ^^

PR #831, opened 2025-09-24
Add bridge management CLI

jbrazio:
This pull request adds support for new CLI bridge management commands.

Full list of commands added:
- get/set bridge.enabled [on|off]
- get/set bridge.delay [0-10000]
- get/set bridge.source [tx|rx]
- get/set bridge.baud [9600-115200]
- get/set bridge.secret [shared-secret]


jbrazio:
@fdlamotte ready for your usual test suite ? ;-D

fdlamotte:
wanted to test it. First tried with wio-e5 ... got some error related to undefined setBridgeState, added the WITH_BRIDGE symbol in platformio (was thinking Mesh.h was not included ok ...)

As I was not sure of my environment for stm32 I give a shot at heltec v3, got the same error (in te middle of warnings)

```
src/helpers/CommonCLI.cpp: In member function 'void CommonCLI::handleCommand(uint32_t, const char*, char*)':
src/helpers/CommonCLI.cpp:277:21: error: 'class CommonCLICallbacks' has no member named 'setBridgeState'; did you mean 'getBuildDate'?
         _callbacks->setBridgeState(_prefs->bridge_enabled);
                     ^~~~~~~~~~~~~~
```

Seems that when building RS232 bridge the setBridgeState method is not used ... (I think you told me about this not being tested)


Now I wonder about saving the bridge state in CommonCli, as it is used by repeaters (that can have bridges) but also sensors ... @ripplebiz do you think this can be a problem ? when changing role for instance ... I have the impression that possible implication can go further than simply the repeater code ...

Also I was wondering if setBridgeState could be alwas compiled (with a version that does nothing if there is no bridge ... instead of not being implemented). It would also simplify conditional compiling ;)

src/helpers/CommonCLI.cpp
```diff
@@ -32,32 +32,34 @@ void CommonCLI::loadPrefsInt(FILESYSTEM* fs, const char* filename) {
...
-    file.read((uint8_t *) &_prefs->cr, sizeof(_prefs->cr));  // 113
-    file.read((uint8_t *) &_prefs->allow_read_only, sizeof(_prefs->allow_read_only));  // 114
-    file.read((uint8_t *) &_prefs->multi_acks, sizeof(_prefs->multi_acks));  // 115
-    file.read((uint8_t *) &_prefs->bw, sizeof(_prefs->bw));  // 116
-    file.read((uint8_t *) &_prefs->agc_reset_interval, sizeof(_prefs->agc_reset_interval));  // 120
-    file.read(pad, 3);   // 121
-    file.read((uint8_t *) &_prefs->flood_max, sizeof(_prefs->flood_max));   // 124
-    file.read((uint8_t *) &_prefs->flood_advert_interval, sizeof(_prefs->flood_advert_interval));  // 125
-    file.read((uint8_t *) &_prefs->interference_threshold, sizeof(_prefs->interference_threshold));  // 126
+    file.read((uint8_t *)&_prefs->airtime_factor, sizeof(_prefs->airtime_factor));    // 0
```
fdlamotte, line 35: it's interesting to rewrite lines for better readability ... but it complexifies the review process as here for instance all the lines were modified and it becomes difficult to see the two added lines

I prefer having separate PRs for code refactoring (and only if necessary)

just a global remark ;)

jbrazio:
Please advise how you want me to proceed on this part (the configuration struct).

src/helpers/CommonCLI.cpp
jbrazio: This is not my first rodeo.. 🥸

I would agree with you if this was totally unrelated with the main subject of the PR.. but it's not, it's the same domain.

fdlamotte:
> Please advise how you want me to proceed on this part (the configuration struct).

For the configuration struct it seems ok to me ... but we'd probably need validation from the ones who wrote it (@ripplebiz at least)

For the PR in general, I'm still waiting to be able to compile a simple rs232 bridge to review it ... which I can't ;)

jbrazio:
> For the PR in general, I'm still waiting to be able to compile a simple rs232 bridge to review it ... which I can't ;)

lol true true :sweat: I was waiting for a feedback on the best way to handle the config struct.
I will patch that later on then.

fdlamotte:
looks like we were caught in a deadlock ;)

jbrazio:
Nothing that an ISR cannot save. :rofl:

fdlamotte:
did quick tests on wio-e5 repeater with rs232 bridge
seems to work as intended ... 
will go deeper ;)

jbrazio:
more stuff for testing

fdlamotte:
commands to manage bridge are interesting

I'm thinking about another one, to query the repeater about its bridge capability could be `get bridge.type` and would return `none`, `rs232`, `espnow` or watever

fdlamotte:
Seems close to be ready ;) 

I would suggest adding some conditional to the debug printfs ... maybe using a BRIDGE_DEBUG_PRINTLN macro ?

fdlamotte:
I've been playing a little with `bridge.source` ... that's interesting ;)

I  definitely prefer setting it on tx as I prefer having a symetrical path ...

jbrazio:
On the inbound repeater set it as source rx and repeater off, the inbound repeater will not repeat but still fwd the packet to the bridge so its repeated by the other side. 🚀

examples/simple_repeater/MyMesh.h
```diff
@@ -182,4 +182,17 @@ class MyMesh : public mesh::Mesh, public CommonCLICallbacks {
   void clearStats() override;
   void handleCommand(uint32_t sender_timestamp, char* command, char* reply);
   void loop();
+
+#if defined(WITH_BRIDGE)
+  void setBridgeState(bool enable) override {
+    if (enable == bridge.getState()) return;
+    enable ? bridge.begin() : bridge.end();
```
ripplebiz: I may be wrong, but I have a feeling the ternary operator here can be stripped out by some C++ compilers. ie. it can be seen as a nil expression. Safer to use good ol' if/else

src/helpers/AbstractBridge.h
```diff
@@ -11,6 +11,18 @@ class AbstractBridge {
...
+   * @brief Stops the bridge.
+   */
+  virtual void end() = 0;
+
+  /**
+   * @brief Gets the current state of the bridge.
+   *
+   * @return true if the bridge is initialized and running, false otherwise.
+   */
+  virtual bool getState() const = 0;
```
ripplebiz: prob more readable if method name is something like isRunning()

src/helpers/CommonCLI.cpp
```diff
@@ -32,32 +32,38 @@ void CommonCLI::loadPrefsInt(FILESYSTEM* fs, const char* filename) {
   if (file) {
     uint8_t pad[8];
 
-    file.read((uint8_t *) &_prefs->airtime_factor, sizeof(_prefs->airtime_factor));  // 0
```
ripplebiz, line 35: Please do not reformat existing code blocks, just for neatness sake. It adds significant cognitive load on reviewers, and makes tracking diffs (over time) much harder.

src/helpers/bridges/RS232Bridge.cpp
```diff
@@ -4,10 +4,11 @@
 
 #ifdef WITH_RS232_BRIDGE
 
-RS232Bridge::RS232Bridge(Stream &serial, mesh::PacketManager *mgr, mesh::RTCClock *rtc)
-    : BridgeBase(mgr, rtc), _serial(&serial) {}
+RS232Bridge::RS232Bridge(NodePrefs *prefs, Stream &serial, mesh::PacketManager *mgr, mesh::RTCClock *rtc)
+    : BridgeBase(prefs, mgr, rtc), _serial(&serial) {}
 
 void RS232Bridge::begin() {
+  Serial.printf("%s: RS232 BRIDGE: Initializing at %d baud...\n", getLogDateTime(), _prefs->bridge_baud);
```
ripplebiz: Need to put these behind some conditional macro

src/helpers/bridges/RS232Bridge.cpp
```diff
@@ -25,53 +26,26 @@ void RS232Bridge::begin() {
...
-    buffer[4 + len] = (checksum >> 8) & 0xFF; // Checksum high byte
-    buffer[5 + len] = checksum & 0xFF;        // Checksum low byte
+  // Update bridge state
+  _initialized = true;
+}
 
-    // Send complete packet
-    _serial->write(buffer, len + SERIAL_OVERHEAD);
+void RS232Bridge::end() {
+  Serial.printf("%s: RS232 BRIDGE: Stopping...\n", getLogDateTime());
```
ripplebiz: conditional macro...

ripplebiz (commented):
Overall, good work.

src/helpers/bridges/RS232Bridge.cpp
jbrazio: @fdlamotte suggested the same, will actually create BRIDGE_DEBUG_PRINTLN macro and have a define to enable the output.

fdlamotte:
now everything has been addressed ... good job !
let's merge

Andrew-a-g:
> On the inbound repeater set it as source rx and repeater off, the inbound repeater will not repeat but still fwd the packet to the bridge so its repeated by the other side. 🚀

Testing this scenario ATM. I have two repeaters - one with no source config to behave as a normal repeater, the other (Rx node) is configured with repeat off and source Rx.

Question is: is there a way to see packets that came in through the source Rx node without looking at the debug logs on the repeater? Is the bridge totally transparent?

jbrazio:
The one with source set as RX will be transparent. Basically the other side of the bridge will "think" it received the packet directly from RF, the only caveat is that the SNR will not reflect the original one.

Andrew-a-g:
Thanks. Am seeing it working here.. was just very confused by the behaviour. I can't help but think in this scenario it would be better (less confusing) to disable repeat in LoRa but still count as a repeat over bridge. 

I'll test further but nice work. Very good use case for a v3 or similar node to pair with an LNA node that's deaf due to 4g interference etc.

PR #835, opened 2025-09-25
fix: revert ENV_PIN_SDA for Wio Tracker L1 non-eink

oltaco:
With ENV_PIN_SDA defined on the on the OLED Wio Tracker L1 the device gets stuck on "Loading...", BLE is advertising and you can "connect" but the device never sends any data.


fdlamotte:
I was afraid of this one ... 
I would probably search the fault in ESM ... remove all sensors and add them one by one to find the culprit

fdlamotte:
I'm letting this in, as current state of tracker is bad whith this regard ... and it will at least boot ;)

But if `PIN_ENV_SDA` and `PIN_ENV_SCL` are not defined, the main i2c bus gets scanned for any sensor (as they're all in), this might have been better to disable all sensors except GPS (and maybe BME280 and INA219 that I have been tested on eink but I'm sure they are correctly init)

PR #836, opened 2025-09-25
[BUG] Wio Tracker L1 freezes during startup

csrutil:
The changes from [commit `4aa58ad`](https://github.com/meshcore-dev/MeshCore/commit/4aa58ade8a32f9c5a3e65d400b64610fb702a03d) cause the L1 variant to fail at boot.

To fix the issue, comment out the `ENV_PIN_SDA` and `ENV_PIN_SCL` definitions in `platformio.ini`.


csrutil:
I need to check why ;)

recrof:
thanks for prompt fix.. I will close this duplicate, as there is earlier one already.

PR #845, opened 2025-09-27
Fix T-Echo Build

oliverl-21:
Fixes T-Echo Build on Dev


fdlamotte:
we know the issue but have a more "global" fix in the pipe ... thanks anyway ;)

PR #858, opened 2025-09-28
Fix: add missing build flags for tiny_relay

csrutil:
- Add ADVERT_LAT and ADVERT_LON definitions for both repeater and sensor variants
- Set MAX_NEIGHBOURS to 50 for improved network capacity


liamcottle:
Thanks!

PR #865, opened 2025-09-29
fix(node): validate & sanitize node name

### LINKED ISSUE #798: Name input check missing
Hi,

the name field of node has input check missing.

I tried to create a repeater with a name: `xxx - Soon™` which rendered the node completly unusable and also the web configurator showed the following.

Image

only a factory reset helped in this case.

spacepc-de:
Fixes #798


liamcottle:
Hi there, thanks for the contribution.

It looks like this doesn't allow emojis, other language characters, and limits to ASCII only and not UTF-8?

We most certainly want to allow for emojis, but we haven't decided on what validation rules we want to impose on the node names yet.

spacepc-de:
I can change it to UTF8 if you want.

liamcottle:
Closing out as this is now stale. Firmware now has some basic validation on node names.
https://github.com/meshcore-dev/MeshCore/blob/246d681b2bb9201952bf93d7970bce054ceb3a00/src/helpers/CommonCLI.cpp#L17

PR #872, opened 2025-10-01
Fix/serialble overflow guard

spacepc-de:
- Fixes a potential buffer overflow when reading from the Bluefruit BLE UART on nRF52 by
  clamping reads to MAX_FRAME_SIZE and draining any surplus bytes.
  - No protocol or API changes; normal behavior remains unchanged for valid frames.

  Motivation

  - SerialBLEInterface::checkRecvFrame() previously did:
      - len = bleuart.available()
      - bleuart.readBytes(dest, len)
  - If len > MAX_FRAME_SIZE (172), this could write past the caller’s buffer (e.g.,
  cmd_frame), causing memory corruption or crashes. This situation is possible if the peer
  sends multiple frames quickly or if the BLE stack coalesces data.

  Changes

  - In src/helpers/nrf52/SerialBLEInterface.cpp:
      - Limit read size to MAX_FRAME_SIZE.
      - If more data is pending than MAX_FRAME_SIZE, drain the remaining bytes in small
  chunks.
      - Log a warning with BLE_DEBUG_PRINTLN when an overflow scenario is detected.
  - The function still returns the number of bytes copied into the destination buffer,
  preserving existing semantics.

  Behavior and Compatibility

  - Normal operation (each frame ≤ MAX_FRAME_SIZE) is unaffected.
  - If more than MAX_FRAME_SIZE bytes are pending:
      - The first MAX_FRAME_SIZE bytes are delivered to the caller.
      - The surplus is discarded to avoid mixing multiple frames into a single return or
  leaving partial data that would corrupt subsequent parsing.
  - No changes required in client applications; in rare overflow cases, the sender may need to
  retry per its existing timeout logic.

  Testing

  - Verified that single frames ≤ MAX_FRAME_SIZE continue to flow unchanged.
  - Simulated overflow by sending multiple back-to-back writes that exceed 172 bytes:
      - Confirmed no crashes or memory corruption.
      - Observed “WARN: BLE RX overflow” log and that the first frame is processed, with
  excess drained.
  - Sanity-checked that send path and timing logic remain the same.

Impact
  - In overflow scenarios, some extra incoming bytes may be dropped; this is preferable to
  undefined behavior and aligns with the current single-frame-per-read assumption.

### DIFF
--- a/src/helpers/nrf52/SerialBLEInterface.cpp
+++ b/src/helpers/nrf52/SerialBLEInterface.cpp
@@ -178,11 +178,29 @@ size_t SerialBLEInterface::checkRecvFrame(uint8_t dest[]) {
       send_queue[i] = send_queue[i + 1];
     }
   } else {
-    int len = bleuart.available();
-    if (len > 0) {
-      bleuart.readBytes(dest, len);
-      BLE_DEBUG_PRINTLN("readBytes: sz=%d, hdr=%d", len, (uint32_t) dest[0]);
-      return len;
+    int avail = bleuart.available();
+    if (avail > 0) {
+      // Cap read to avoid overflowing caller's buffer (MAX_FRAME_SIZE)
+      int to_read = avail > MAX_FRAME_SIZE ? MAX_FRAME_SIZE : avail;
+      int got = bleuart.readBytes(dest, to_read);
+
+      // If more data is pending than we can safely store, drain the excess
+      if (avail > MAX_FRAME_SIZE) {
+        int remaining = avail - to_read;
+        uint8_t discard[32];
+        int drained_total = 0;
+        while (remaining > 0) {
+          int chunk = remaining > (int)sizeof(discard) ? (int)sizeof(discard) : remaining;
+          int drained = bleuart.readBytes(discard, chunk);
+          if (drained <= 0) break;
+          drained_total += drained;
+          remaining -= drained;
+        }
+        BLE_DEBUG_PRINTLN("WARN: BLE RX overflow: avail=%d, read=%d, drained=%d", avail, got, drained_total);
+      }
+
+      BLE_DEBUG_PRINTLN("readBytes: sz=%d, hdr=%d", got, (uint32_t) dest[0]);
+      return got;
     }
   }
   return 0;


olanwe:
Hi, we have received reports from the southern German mesh indicating that this PR resolves the freeze issue caused by BT connection problems on the T1000-e.
It might be a good idea to consider merging it into the dev branch soon.

spacepc-de:
> Hi, we have received reports from the southern German mesh indicating that this PR resolves the freeze issue caused by BT connection problems on the T1000-e. It might be a good idea to consider merging it into the dev branch soon.

olanwe, yes, it solved the problem for me. I can send you the firmware as long as there is no new firmware available. I wrote to you via MeshCore but couldn't get through. However, I can see all your nodes :)
Greetings from JOST in Konstanz.

olanwe:
Hi Jost!
Sorry, I didn’t have your node in my contact list, which I assume is why the message didn’t get through. I tried replying to you but received a “failed” error, so I’m not sure if the transmission was successful.
Sending an e-mail to your info... address didn’t work either (I got a 550 error from the mail system).
Anyway, thanks a lot for the fix! I’ve already built the firmware with your changes and uploaded it to weyes.de/yztzc so our fellow mesh colleagues can use and test it as well.
I hope your PR will be merged soon.
Greetings from Achstetten,
Oliver aka Olanwe

Michael-Frank:
Stable for me (Android Google Pixel -> T1000e) with a latested dev branch + this PR.
Survived multiple disconnects, syncling large contact lists an so on.
Related or not:  local communit resports, that the  same custom firware does not solve "T1000e crashing" when using it with an IPhone.
@spacepc-de maybe you have an idea?
Update: 
- this pr is stable for me with android (@olanwe )
- iphone issue is tracked and worked on seperatly in #1067

Michael-Frank:
Short update:
 i sucessfuyll ran custom firmware with this PR merged into it now for 2 Weeks stable and it fixed BT connection stability.
 - With latested dev prior to 1.10 release + this pr
 - with 1.10 release + this pr

 Same for mulitple members in the local mesh, like @olanwe  -  all running patched firmwares with this PR and reporting stability improvements.

@spacepc-de : if you need more testing or feedack please let me know.  
 Have you spoken to @oltaco and or @recrof ? They have done a lot of BT related fixes recently.

liquidraver:
You guys seems to have solved the core problem, thanks!
I'll try to migrate this change to my fixing branch and test :)
I tweaked TX side for a week thinking it's us that do something wrong, and it seems it's iOS :)

fschrempf:
>  i sucessfuyll ran custom firmware with this PR merged into it now for 2 Weeks stable and it fixed BT connection stability.

Same here. We have tested this for almost two weeks on a WisMesh Tag and several days on a T1000-E. Both devices haven't shown any of the BT issues anymore that were frequently happening before.

liquidraver:
we could freeze the device with this code when rapidly connecting/disconnecting with timeout (eg you walk on the edge of BT range, or slamming the phone in the microwave and opening-closing the door every 2-5 sec.)

We are working on other stuff on serialbleinterface to avoid freezes, I'll add this method to it for extra safeguard to test.

IoTThinks:
True, I have some problem with T1000-e after some time too.
May need to reboot it.

liamcottle:
Howdy! Thanks for the contribution :)

After extensive testing of various changes, we have merged in #1177 which overhauled the BLE connection logic on nRF based devices.

I'll close this one out, as I don't believe this is needed now.

PR #885, opened 2025-10-04
Add MAX_LORA_TX_POWER build flag for Heltec V4 configuration

ViezeVingertjes:


trisweb:
Should LORA_TX_POWER also be defaulted to 22 for this board? Or is a default less than the board's designed power desirable, or a pattern used with other boards?

ViezeVingertjes:
> Should LORA_TX_POWER also be defaulted to 22 for this board? Or is a default less than the board's designed power desirable, or a pattern used with other boards?

Probably that one should be on the right amount to stay within 'legal' limits when amplification is applied, 22 (before PA) would be way over for sure, but this at least ensures that one can adjust it through the cli/app and it is persisted after a reboot.

```
  -D LORA_TX_POWER=10 ;If it is configured as 10 here, the final output will be 22 dbm.
```

The comment behind it makes believe that it will result in 22dbm, which is acceptable for most countries imo.

trisweb:
That makes sense. As long as we are able to raise it and it is saved, that will work fine.

LitBomb:
for new high output tx devices, it should be consistent with existing high output devices like Station G2 and Ikoka E22 30dbm and 33dbm.  Currently both only set `-D LORA_TX_POWER=` to their max specification output, no additional MAX build flag.  For example, Station G2's LORA_TX_POWER is set to 19 per https://wiki.uniteng.com/en/meshtastic/station-g2

Either way, please have consistent build flags across devices.

ViezeVingertjes:
> Currently both only set `-D LORA_TX_POWER=` to their max specification output, no additional MAX build flag.

Ah, i see. I agree in that case. I do not know what value would be correct for the V4 in that case though.
If anyone knows; i don't mind pushing a commit  to make it consistent.

trisweb:
@ViezeVingertjes I quite like the explicit setting of a MAX power. Only thing that makes it complicated is ensuring other boards have the same setup.

However if the convention currently is to just use the one power level and set it to the max (users may then turn it down if necessary for their jurisdiction) then I suggest we do that for now and refactor a max in the future if necessary.

The power for the Heltec V4 should be 22, to set the 1262 to full power. The board has a 15dB attenuator, and the power would not exceed its design max of 28dBm when set to the maximum for the 1262, and I believe this is what it expects.

LitBomb:
whatever we do, do not do what meshtastic does.  last time used it, meshtastic's client shows 30dbm in the phone UI, which is a lie, and the firmware sets the actual tx power and it is hidden from the app/users.  

also for reference, repeater firmware defaults to LORA_TX_POWER, and repeater administrators can use the command `get tx` and `set tx <number>` or the repeater web config page (on flasher.meshcore.co.uk) to set the repeater's tx power.  as we know, companion radio tx output can be configured with phone app.

ViezeVingertjes:
> @ViezeVingertjes I quite like the explicit setting of a MAX power. Only thing that makes it complicated is ensuring other boards have the same setup.

After a bit of sleep; i also think we should properly set the MAX so its explicitly set. So that would mean that for this V4 its now correctly set. We probably could create an issue, to consolidate the other boards later. Then this one can be merged and the V4 users can at least set their TX power persistently after its merged.

ViezeVingertjes:
> whatever we do, do not do what meshtastic does. last time used it, meshtastic's client shows 30dbm in the phone UI, which is a lie, and the firmware sets the actual tx power and it is hidden from the app/users.

Haha no! in MC, what you see is what you get. The problem without this fix is that you can set it to 22dB for example, it goes in effect, but on reboot it reverts to that 10.

trisweb:
There is an argument to be made that the actual setting of say 22dBm is a lie, if the radio then amplifies it and the real output power is 27dBm.

I know we have to field questions frequently on discord etc about "why can't I set my fancy new high power radio to its full power??" So it's obviously not 100% clear to users.

But trying to be too smart about it is also potentially confusing.

My recommendation is just that we document the power settings and their impact for each radio, if it's not 1:1. Especially with popular radios like the V4 that have an LNA, users aren't gonna know, and the question will keep coming up unless it's well handled. But I think that will be enough.

ViezeVingertjes:
Totally agree, but would be out of scope for this one, i think. Would be a waste to not merge this soon (or a different fix), as people currently are getting their V4's in and their setting is not persisted.

trisweb:
Yes, should be fixed first.

Any progress? How can we get this merged?

ViezeVingertjes:
> Any progress? How can we get this merged?

Perhaps @recrof or @fdlamotte could review if they have some time.

trisweb:
Some specs on how these power settings actually impact output power:

image

So as I suspected, the correct maximum is 22. And a good default is indeed 10. But the maximum should be easily available for users to set (to 22).

PR #889, opened 2025-10-05
Sensecap indicator

fdlamotte:
Initial support for SenseCap Indicator on espnow ...

The best part is probably the LVGFX display class, with touch support ;)

### DIFF
--- a/variants/sensecap_indicator-espnow/SCIndicatorDisplay.h
+++ b/variants/sensecap_indicator-espnow/SCIndicatorDisplay.h
@@ -124,6 +124,5 @@ public:
 class SCIndicatorDisplay : public LGFXDisplay {
   LGFX disp;
 public:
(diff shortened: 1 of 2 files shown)


variants/sensecap_indicator-espnow/SCIndicatorDisplay.h
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    setPanel(&_panel_instance);
+  }
+};
+
+class SCIndicatorDisplay : public LGFXDisplay {
+  LGFX disp;
+public:
+  SCIndicatorDisplay() : LGFXDisplay(480, 480)
```
ripplebiz: The concrete LGFX display instance prob should be passed as constructor param, as it is a dependency required by base class

PR #890, opened 2025-10-05
CommonCLI: gps management commands

fdlamotte:
Some commands to manage GPS (on repeaters for the moment)

* gps : prints gps status (on/off, fix/no fix, sat nb)
* gps on
* gps off
* gps sync : will sync time at next fix

This will be enhanced with parameters when #831 is in (because it adds parameters ;))
* get/set gps.onboot

and possibly parameters for pwm functionality #778 that is currently under review 

Tested on Techo repeater

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -401,6 +401,19 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
       sprintf(reply, "%s (Build: %s)", _callbacks->getFirmwareVer(), _callbacks->getBuildDate());
     } else if (memcmp(command, "board", 5) == 0) {
       sprintf(reply, "%s", _board->getManufacturerName());
+#if ENV_INCLUDE_GPS == 1
+    } else if (memcmp(command, "gps on", 6) == 0) {
+      _callbacks->gpsStart();
+      strcpy(reply, "ok");
+    } else if (memcmp(command, "gps off", 7) == 0) {
+      _callbacks->gpsStop();
+      strcpy(reply, "ok");
+    } else if (memcmp(command, "gps sync", 8) == 0) {
+      _callbacks->gpsSyncTime();
+      strcpy(reply, "Waiting fix ...");
+    } else if (memcmp(command, "gps", 3) == 0) {
+      _callbacks->gpsGetStatus(reply);
+#endif
     } else if (memcmp(command, "log start", 9) == 0) {
       _callbacks->setLoggingOn(true);
       strcpy(reply, "   logging on");
--- a/src/helpers/CommonCLI.h
+++ b/src/helpers/CommonCLI.h
@@ -50,6 +50,10 @@ public:
   virtual void saveIdentity(const mesh::LocalIdentity& new_id) = 0;
   virtual void clearStats() = 0;
   virtual void applyTempRadioParams(float freq, float bw, uint8_t sf, uint8_t cr, int timeout_mins) = 0;
+  virtual void gpsGetStatus(char * reply) {}
+  virtual void gpsStart() {}
+  virtual void gpsStop() {}
+  virtual void gpsSyncTime() {}
 };
 
 class CommonCLI {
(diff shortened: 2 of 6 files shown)


mattzzw:
Nice! How about adding a `gps info` for things like "time since last fix" or "sats in view"?

fdlamotte:
> Nice! How about adding a `gps info` for things like "time since last fix" or "sats in view"?

gps command already does that:

```
gps
  -> off
gps on
  -> ok
gps
  -> on, active, no fix, 0 sats
gps
  -> on, active, fix, 9 sats
```

I note the idea for time since last fix but it's currently not available in MicroNMEALocationManager ;)

src/helpers/sensors/MicroNMEALocationProvider.h
```diff
@@ -78,6 +78,16 @@ public :
         }        
     }
 
+    bool isEnabled() override {
+        // directly read the enable pin if present as gps can be
+        // activated/deactivated outside of here ...
+        if (_pin_en != -1) {
+            return digitalRead(_pin_en) == PIN_GPS_EN_ACTIVE;
```
ripplebiz, line 85: Do _all_ variants define PIN_GPS_EN_ACTIVE ?

src/helpers/CommonCLI.h
```diff
@@ -50,6 +50,10 @@ class CommonCLICallbacks {
   virtual void saveIdentity(const mesh::LocalIdentity& new_id) = 0;
   virtual void clearStats() = 0;
   virtual void applyTempRadioParams(float freq, float bw, uint8_t sf, uint8_t cr, int timeout_mins) = 0;
+  virtual void gpsGetStatus(char * reply) {}
```
ripplebiz: the default impl here would need to set reply to something, either empty string, or something like "gps not found"

src/helpers/sensors/MicroNMEALocationProvider.h
fdlamotte: no, but there is a safeguard 

https://github.com/meshcore-dev/MeshCore/blob/f5f58863276f22c857a64ad7d923b18dd1d1f701/src/helpers/sensors/MicroNMEALocationProvider.h#L15-L17

fdlamotte:
Removed the callbacks, so now everything is done in common_cli (now works with sensors too)

I also added generic `sensor get/set` commands

src/helpers/CommonCLI.cpp
```diff
@@ -504,6 +514,90 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
       sprintf(reply, "%s (Build: %s)", _callbacks->getFirmwareVer(), _callbacks->getBuildDate());
     } else if (memcmp(command, "board", 5) == 0) {
       sprintf(reply, "%s", _board->getManufacturerName());
+    } else if (memcmp(command, "sensor get ", 11) == 0) {
+      const char* key = command + 11;
+      const char* val = sensors.getSettingByKey(key);
+      if (val != NULL) {
```
ripplebiz, line 525: probably need some prefix here so automated parsers can distinguish. ie. they just see "can't find custom var" as a value. Maybe "> {val}"  for non-null, and "null"  ?

src/helpers/CommonCLI.cpp
```diff
@@ -504,6 +514,90 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
...
+        strcpy(reply, val);
+      } else {
+        strcpy(reply, "can't find custom var");
+      }
+    } else if (memcmp(command, "sensor set ", 11) == 0) {
+      const char* args = &command[11];
```
ripplebiz: Have a look at the 'tempradio' case, and how to use the helper mesh::Utils::parseTextParts()
the strchr() below can return NULL with a badly formatted command

fdlamotte:
Currently on repeaters and sensors gps values do not get it automatically into adverts (and it should not always)

I've just added a command `gps setloc` that copies the current coordinates into prefs and saves prefs. This way, once we get a fix, we can configure the node position manually

I think I'll next implement `advert_loc_policy` for repeaters, rooms and sensors, but on these nodes it should be off by default

PR #892, opened 2025-10-06
BLE_LONG_PRESS macro to allow hiding of the PIN on home screen

andyshinn:
A feature to hide the BLE PIN behind a long press on the home screen. I wasn't keen on showing the pin out in the open automatically when the screen was on so I thought this might be a good compromise. If it makes more sense to add this to a setting that can be switched on and off at runtime then I could try to expand it around that instead of a macro.

See example below:

https://github.com/user-attachments/assets/d4d90924-2dcf-4891-94b5-90f7e0c1fae8

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -91,6 +91,7 @@ class HomeScreen : public UIScreen {
   NodePrefs* _node_prefs;
   uint8_t _page;
   bool _shutdown_init;
+  bool _show_ble_pin;
   AdvertPath recent[UI_RECENT_LIST_SIZE];
 
 
@@ -149,8 +150,8 @@ class HomeScreen : public UIScreen {
 
 public:
   HomeScreen(UITask* task, mesh::RTCClock* rtc, SensorManager* sensors, NodePrefs* node_prefs)
-     : _task(task), _rtc(rtc), _sensors(sensors), _node_prefs(node_prefs), _page(0), 
-       _shutdown_init(false), sensors_lpp(200) {  }
+     : _task(task), _rtc(rtc), _sensors(sensors), _node_prefs(node_prefs), _page(0),
+       _shutdown_init(false), sensors_lpp(200), _show_ble_pin(false) {  }
 
   void poll() override {
     if (_shutdown_init && !_task->isButtonPressed()) {  // must wait for USR button to be released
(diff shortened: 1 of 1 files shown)


tekstrand:
This would be great to have.

bplein:
I've wanted this for a while!

examples/companion_radio/ui-new/UITask.cpp
```diff
@@ -193,10 +194,21 @@ class HomeScreen : public UIScreen {
         display.setTextSize(1);
         display.drawTextCentered(display.width() / 2, 43, "< Connected >");
       } else if (the_mesh.getBLEPin() != 0) { // BT pin
-        display.setColor(DisplayDriver::RED);
-        display.setTextSize(2);
-        sprintf(tmp, "Pin:%d", the_mesh.getBLEPin());
-        display.drawTextCentered(display.width() / 2, 43, tmp);
+#ifdef BLE_LONG_PRESS
```
liamcottle, line 197: Would be happy for the `ifdef` to be removed and have this as standard functionality

examples/companion_radio/ui-new/UITask.cpp
```diff
@@ -149,8 +150,8 @@ class HomeScreen : public UIScreen {
 
 public:
   HomeScreen(UITask* task, mesh::RTCClock* rtc, SensorManager* sensors, NodePrefs* node_prefs)
-     : _task(task), _rtc(rtc), _sensors(sensors), _node_prefs(node_prefs), _page(0), 
-       _shutdown_init(false), sensors_lpp(200) {  }
+     : _task(task), _rtc(rtc), _sensors(sensors), _node_prefs(node_prefs), _page(0),
+       _shutdown_init(false), sensors_lpp(200), _show_ble_pin(false) {  }
```
liamcottle, line 154: would prefer the pin to be shown by default for ease of use for new users

examples/companion_radio/ui-new/UITask.cpp
```diff
@@ -378,6 +390,12 @@ class HomeScreen : public UIScreen {
       }
       return true;
     }
+#ifdef BLE_LONG_PRESS
+    if (c == KEY_ENTER && _page == HomePage::FIRST) {
+      _show_ble_pin = !_show_ble_pin;  // toggle PIN visibility on long press
```
liamcottle, line 395: in a future update, this could possibly become a preference flag that's saved, and loaded on boot

liamcottle (changes requested):
Hey, thanks for the PR! I understand it's taken a very long time to review this, almost 6 months now lol... But if you still want to see it merged in, please see the changes I've requested :)

PR #900, opened 2025-10-07
Add stats to serial CLI

michaelhart:
Implemented formatStatsReply, formatRadioStatsReply, and formatPacketStatsReply methods in MyMesh for both simple_repeater and simple_room_server. Introduced StatsFormatHelper for structured statistics formatting. Updated CommonCLI to handle new stats commands.

The 3 new serial-only commands are stats, stats-radio, and stats-packets.

```
  -> {"battery_mv":0,"uptime_secs":17,"errors":0,"queue_len":2}
  -> {"noise_floor":-110,"last_rssi":-42,"last_snr":49,"tx_air_secs":0,"rx_air_secs":1}
  -> {"recv":6,"sent":3,"flood_tx":3,"direct_tx":0,"flood_rx":6,"direct_rx":0}
```


Cisien:
@ripplebiz @recrof We would love this change. It will let us free up air time by moving repeater stats telemetry to other systems

Cisien:
@ripplebiz Are there any concerns with this change? It would be good to get it merged so we can get these metrics in an automation fashon without having to spend air time.

recrof:
sorry guys, can't merge PRs, I guess @liamcottle and @fdlamotte can review and decide if @ripplebiz is busy

fdlamotte:
I don't have issue with this PR (did not see it before, sorry).

But I would be interested to know why it hasn't been merged yet by @ripplebiz before doing it (sorry, I don't really make things go forward). If it's just for review time, I can do it ... (already overlooked and seems fine)

If I understand well, you want to read the STATUS_REQ from serial cli (it took me time because I usually use cli over lora but I now see why you couldn't access these ;) )

michaelhart:
> If I understand well, you want to read the STATUS_REQ from serial cli (it took me time because I usually use cli over lora but I now see why you couldn't access these ;) )

Yup. A bit more context: It's an increasingly popular idea to use the HA plugin to poll these stats over LoRa/MeshCore, but the downside is that it is imo a very, very poor utilization of precious airtime to have so many people polling so many repeaters to paint a private line on a private chart that they will also probably go months at a time without looking at and potentially even totally forget that it even exists.

From my stats, already these REQ/RES packets at times account for [around 50% of all mesh traffic globally](https://analyzer.letsme.sh/stats) - though it has improved a little recently, and it's my understanding too that this is only a handful of people. The ratio of mesh traffic being used this way is even worse if you look at specific regions, like SEA. Why so many packets can come from as few as 1-2 people's HA is an open and separate issue, but my hope is to provide a substantially better alternative (more granular, publicly available charts on packet analyzer, and even deeper analysis potential in [w0z's Grafana](https://grafana.w0z.is/) when Alex has time to update his ingestor).

My hope with this change is to funnel these stats over the internet, so that we don't kill the usability of our mesh just so a few people can make a personal HA dashboard, and so that we all can still benefit and improve the reliability of the mesh from the insights from this data.

fdlamotte:
There is no big blocker with this PR, seems we all got overwhelmed recently ;) 

There is one small issue whith the `stats` command, as the length is not checked. If other stats commands are added later, there is a risk that an older firmware responds to stats instead of the newer command ...

So the suggestion is either to check the lenth or use another name for `stats`, Scott suggests `stats-core`

michaelhart:
That is a great catch about the fact the stats stuff would have caught future stats-x commands. I've both checked length on all to ensure it always behaves as intended and changed `stats` to `stats-core`.

fdlamotte:
great, I'll just run it on my techo before merging (a last check)

edit: just great, let's merge

thank you for the contribution, and sorry for the delay ... you were right to ping us ;)

michaelhart:
No worries, I completely understand :) Thank you @fdlamotte!

michaelhart:
Followup for where the stats are going: https://analyzer.letsme.sh/node/C79F318AC9063B31C374C7B4476CC32D12F19DF37C2440555DB81FDE34E56442

PR #901, opened 2025-10-08
feat: ignore supplied password for passwordless room servers

muesli:
If the password is intentionally left empty, we can assume the operator wants this room server to be public. In such cases it's less frustrating for users to simply accept any password.

A neater solution might be introducing a separate build flag that explicitly enables password-less public room servers, but the downside is that it requires a separate firmware image or custom build, whereas this solution works for any room server that's already configured without a password. We could think about adding a special case for the "hello" password as well.

Just thinking out loud: we could introduce a separate type indicator (chat, repeater, room, public room) for password-less servers that skips the login procedure altogether to further improve the user experience. Another alternative to that might be a quick status request that indicates if the room server is up and whether it requires authentication.

### DIFF
--- a/examples/simple_room_server/MyMesh.cpp
+++ b/examples/simple_room_server/MyMesh.cpp
@@ -303,7 +303,7 @@ void MyMesh::onAnonDataRecv(mesh::Packet *packet, const uint8_t *secret, const m
       if (strcmp((char *)&data[8], _prefs.password) == 0) { // check for valid admin password
         perm = PERM_ACL_ADMIN;
       } else {
-        if (strcmp((char *)&data[8], _prefs.guest_password) == 0) {   // check the room/public password
+        if (strlen(_prefs.guest_password) == 0 || strcmp((char *)&data[8], _prefs.guest_password) == 0) {   // check the room/public password
           perm = PERM_ACL_READ_WRITE;
         } else if (_prefs.allow_read_only) {
           perm = PERM_ACL_GUEST;


DayleDrinkwater:
This would break `allow_read_only` wouldn't it?

muesli:
> This would break `allow_read_only` wouldn't it?

It would change the behavior under one specific condition, but I don't think I would call it a breaking change.

Let's look at the different read-only scenarios for servers that have...

1. no password & read-only disabled: read-only mode never gets enabled (as before)
2. with password & read-only disabled: read-only mode never gets enabled (as before)
3. with password & read-only enabled: read-only mode gets enabled if you specify an incorrect password (as before)

This leaves us with only the fourth option, which actually changes its behavior. When no password is set & read-only is enabled, the read-only mode now never gets enabled, because any password gets accepted. I'd argue this is still a positive change, as the previous behavior didn't provide any protection or a true "read-only" mode for guests. Specifying an empty password would still result in write access currently.

Appreciate a review! Am I overlooking something?

446564:
review in progress

446564:
>     1. no password & read-only disabled: read-only mode never gets enabled (as before)
> 
>     2. with password & read-only disabled: read-only mode never gets enabled (as before)
> 
>     3. with password & read-only enabled: read-only mode gets enabled if you specify an incorrect password (as before)

All scenarios tested correctly as described. I will move on to code review.

446564 (approved):
I can't approve that it needs to be added, but the PR works as described and does not really change any existing behavior with the exception that now if a guest does not need a password and supplies an incorrect one they still get in.

I think instead of a build flag it might make sense to add a pref and matching command set to allow password less login for guests. this way the app could expose a toggle in management with a description.

PR #911, opened 2025-10-09
Ikoka Nano Variant

bplein:
Created as a fork of the ikoka stick variant.

- Updated for Ikoka Nano legacy pinout
- Removed display support
- Removed user button support
- Retains I2C sensor support

Tested with the ebytes E22 30W module, companion-ble and repeater firmware versions, with an I2C INA3221 power sensor.


bplein:
If one diffs these files with the files in the ikoka stick variant, you'll see most of it is a find and replace of stick->nano, along with some changes in pin definitions, removal of user button and change to the Null display driver.

PR #913, opened 2025-10-11
Feat - Add gat562 meshtiny targets

csrutil:
Add board support for two new nRF52840-based devices with SX1262 LoRa radios:

- GAT562 Mesh Trial Tracker: board definition, variant files, and companion radio targets (USB/BLE)
- GAT562 Mesh Tracker Pro: enhanced variant with joystick support
- Meshtiny: compact board variant with display, buzzer, and joystick controls

Each board includes:
- PlatformIO board definitions with nRF52840 configuration
- Board-specific classes with power management and OTA update support
- Multiple build targets for companion radio applications
- Support for GPS, OLED display (SSD1306), and environmental sensors
- Both USB serial and BLE connectivity options

### DIFF
--- a/boards/gat562_mesh_trial_tracker.json
+++ b/boards/gat562_mesh_trial_tracker.json
@@ -1,0 +1,74 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "nrf52840_s140_v6.ld"
+    },
+    "core": "nRF5",
+    "cpu": "cortex-m4",
+    "extra_flags": "-DARDUINO_NRF52840_FEATHER -DNRF52840_XXAA",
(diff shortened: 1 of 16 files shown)


IoTThinks:
@csrutil: Is this PR working?
Some MeshCore friends ask me for this GAT562 30s Kit – 30dBm Mesh Device too.

So it is best if you can push it as well. I guess it is similar except the TX power.
https://shop.mtoolstec.com/product/gat562-30s-kit-30dbm-mesh-device

Thanks a lot.

liamcottle:
Hey, thanks for the PR!

It looks like this one is very old and we now have support for some of the GAT562 models in recent commits.

I'll close this one out.

PR #918, opened 2025-10-11
xiao c3 cleanup

recrof:
- migrated from SensorManager to EnvironmentSensorManager
- added Room Server and Companion WIFI roles
- moved board file to variants folder
- cleanup

### DIFF
--- a/variants/xiao_c3/platformio.ini
+++ b/variants/xiao_c3/platformio.ini
@@ -5,6 +5,7 @@ build_flags =
   ${esp32_base.build_flags}
   ${sensor_base.build_flags}
   -UENV_INCLUDE_GPS
+  -UENV_INCLUDE_VL53L0X
   -I variants/xiao_c3
   -D ESP32_CPU_FREQ=80
   -D PIN_VBAT_READ=D0


recrof:
@fdlamotte updated the source to exclude VL53L0X sensor code - as it causes bootloops on esp32c3

leon332157:
Hi, I'm still experiencing a bootloop on the latest v1.11.0 BLE firmware with the Xiao esp32c3, even when I use erase flash option. I have to flash v1.9.0 for the device to function normally. Repeater firmware does not seem to have this issue tho. 
image

recrof:
> Hi, I'm still experiencing a bootloop on the latest v1.11.0 BLE firmware with the Xiao esp32c3, even when I use erase flash option. 

yes, we are aware of the problem. if you want to use latest & greatest, try this firmware from @liquidraver, it should not bootloop: https://github.com/liquidraver/MeshCore/releases/tag/unifiedblev1
I'm hoping for it to be merged in upcoming release.

leon332157:
Thank you for the quick response! I did not know that this is a BLE related issue, the firmware you shared is working for me!

recrof:
> Thank you for the quick response! I did not know that this is a BLE related issue, the firmware you shared is working for me!

relation to bluetooth is very distant.. the size of the firmware is key in bootlooping - new `Nimble` bluetooth library is much lighter that espressif one we use currently, so it gives the board some breathing space.

PR #953, opened 2025-10-16
LR1110 packet shift issue workaround

Woodie-07:
This PR provides a workaround for what seems to be a bug with the LR1110 radio itself: when the radio receives a packet corrupted in a certain way, it'll report a 0 length packet with the header error IRQ set, but all subsequent packets will have their data shifted an additional 4 bytes to the right in the RX buffer on top of the radio's reported offset. The shift is reset on standby() which is called on every transmission.

The issue manifests on the companion app as the device seemingly going 'deaf' after a random time with no new messages or advertisements showing up in the app until the user performs an action that results in a transmission. The device has in fact not gone 'deaf' - it is still receiving all packets but they're shifted 4 bytes to the right so are prefixed with zeroes and the end cut off. In debug logs, this will first show as all packets being type=0 route=F when shifted by 4 bytes, then when shift grows to 8 bytes or more it'll show the src -> dst as 00 -> 00.
On repeater firmware, the issue is less apparent but still present. After the bug occurs, the next packet received will be interpreted as type=0 route=F due to the first byte being 0x00. Thus, the repeater will rebroadcast the (corrupted) packet, resulting in a call to standby() due to the transmission, resetting the shift and resolving the issue automatically for subsequent packets.

This patch detects when the shift occurs and corrects for it. It works in the `CustomLR1110` class as so:
- RadioLib's `LR1110::getPacketLength` function called by `RadioLibWrapper::recvRaw` is overridden:
  - It first calls the base function as normal.
  - If the reported length is 0 and `RADIOLIB_LR11X0_IRQ_HEADER_ERR` is set, it'll increase a u8 `shiftCounter` by 4.
- RadioLib's `LR1110::standby` function called internally by RadioLib on transmission is overridden to also reset `shiftCounter` to 0.
- RadioLib's `LR1110::readData` function called by `RadioLibWrapper::recvRaw` is overridden to perform exactly the same operations as the original but add `shiftCounter` to the radio's reported offset in the `readBuffer8` call. Unfortunately, RadioLib's `LR11x0::readBuffer8` is not virtual so I've had to copy and paste the entire `readData` function from RadioLib to change that one function call.

This PR also bumps the minimum RadioLib version from 7.1.2 to 7.3.0. The patch depends on at least 7.2.0 (would need a small change for 7.1.2) but I've increased it to latest as discussed in Discord.

### DIFF
--- a/src/helpers/radiolib/CustomLR1110.h
+++ b/src/helpers/radiolib/CustomLR1110.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <RadioLib.h>
+#include "MeshCore.h"
 
 #define LR1110_IRQ_HAS_PREAMBLE                     0b0000000100  //  4     4     valid LoRa header received
 #define LR1110_IRQ_HEADER_VALID                     0b0000010000  //  4     4     valid LoRa header received
@@ -24,10 +25,10 @@ class CustomLR1110 : public LR1110 {
       if (len == 0) {
         uint32_t irq = getIrqStatus();
         if (irq & RADIOLIB_LR11X0_IRQ_HEADER_ERR) {
-          Serial.println(F("got possible bug packet"));
+          MESH_DEBUG_PRINTLN("LR1110: got header err, assuming shift");
           this->shiftCount += 4; // uint8 will loop around to 0 at 256, perfect as rx buffer is 256 bytes
         } else {
-          Serial.println(F("got zero-length packet without header err irq"));
+          MESH_DEBUG_PRINTLN("LR1110: got zero-length packet without header err irq");
         }
       }
       return len;


fdlamotte:
I've been using this workaround for two days now with no issue.

Only thing that could break other builds is the bump to radiolib 7.3, but it has been used in fresh builds for some time now (at least it's the one used for all my own builds).

big up for @Woodie-07 and all contributors that helped on this one. Big thanks to @jgromes who gave an hand to debug this !

landgenoot:
Is this problem unique to the LR1110? I am experiencing similar behavior on the RAK4631 with SX1262.
My repeater is randomly unreachable (no ping, no admin, no repeats). After advert, it becomes reachable again immediately.

Woodie-07:
> Is this problem unique to the LR1110? I am experiencing similar behavior on the RAK4631 with SX1262.
> 
> My repeater is randomly unreachable (no ping, no admin, no repeats). After advert, it becomes reachable again immediately.

That sounds like something different to me. If it were this issue, repeaters would continue working but would shift a packet randomly and rebroadcast it.

PR #954, opened 2025-10-16
Fix the analogue user button on the new UI by adding a delay to UITask.cpp

adam2872:
The polling of the ADC of the user button is too fast and causes bad reads. The delay is required to allow the ADC to settle between reads.

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -650,6 +650,7 @@ void UITask::loop() {
   }
 #endif
 #if defined(PIN_USER_BTN_ANA)
+  delay(10);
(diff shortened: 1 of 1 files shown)


examples/companion_radio/ui-new/UITask.cpp
```diff
@@ -650,6 +650,7 @@ void UITask::loop() {
   }
 #endif
 #if defined(PIN_USER_BTN_ANA)
+  delay(10);
```
ripplebiz, line 653: Use of delay() is mostly forbidden (except in setup()/begin() code).

adam2872: Ahh, ok, I’ll refactor using another method. Thanks.

PR #961, opened 2025-10-17
Changes retransmit delay to no longer use buckets

Cisien:
This change would give the "listen before send" code a better chance to detect if another nearby repeater is transmitting.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -396,12 +396,27 @@ int MyMesh::calcRxDelay(float score, uint32_t air_time) const {
 }
 
 uint32_t MyMesh::getRetransmitDelay(const mesh::Packet *packet) {
+  if(_prefs.tx_delay_factor == 0) {
+    MESH_DEBUG_PRINTLN("getRetransmitDelay: Total tx delay: %d", _prefs.tx_delay_factor);
+    return 5;
+  }
+  
   uint32_t t = (_radio->getEstAirtimeFor(packet->path_len + packet->payload_len + 2) * _prefs.tx_delay_factor);
-  return getRNG()->nextInt(0, 6) * t;
+  uint32_t total_delay = getRNG()->nextInt(5, t);
+  MESH_DEBUG_PRINTLN("getRetransmitDelay: Total tx delay: %d", total_delay);
+  return total_delay;
 }
+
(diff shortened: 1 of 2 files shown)


Cisien:
One behavior i've noticed, but i'm not sure yet if it's related to this change. For direct path messages, having the `direct.txdelay` value set to 0 makes tracing very unreliable. Setting this to 0.5 brings it to 100%

Should I clamp the minimum delay so it's never 0?

Cisien:
Updated so the minimum value is 5 (arbitrary number), traces are reliable on `direct.txdelay 0` now

examples/simple_repeater/MyMesh.cpp
```diff
@@ -396,12 +396,27 @@ int MyMesh::calcRxDelay(float score, uint32_t air_time) const {
...
 
 uint32_t MyMesh::getRetransmitDelay(const mesh::Packet *packet) {
+  if(_prefs.tx_delay_factor == 0) {
+    MESH_DEBUG_PRINTLN("getRetransmitDelay: Total tx delay: %d", _prefs.tx_delay_factor);
+    return 5;
+  }
+  
   uint32_t t = (_radio->getEstAirtimeFor(packet->path_len + packet->payload_len + 2) * _prefs.tx_delay_factor);
-  return getRNG()->nextInt(0, 6) * t;
+  uint32_t total_delay = getRNG()->nextInt(5, t);
```
ripplebiz, line 405: This would make the maximum delay 't'. Whereas it used to be 5*t was the max.
This would change the interpretation of the _prefs values. Admins can set to some small fraction, like 0.1 (or smaller) if very small delays want to be experimented with. Currently max for the pref is 2.0 (I think), which is probably very wild/unrealistic. 
Also, currently the minimum 'extra' delay is 0, and this introduces min of 5 which could be a fairly 'radical' change. For a first-try, I'd rather introducing JUST the granularity change... ie. no 6 buckets, and just continuous range instead

ripplebiz:
BTW, I've pushed to dev similar because a few of the main team want to try out what more granular tx delays are like.

Cisien:
Excellent, i'll run it on my repeaters and let you know if anything looks off

liamcottle:
Closing out as Scott removed the 6 slots with this commit 8cbcd2271d586997dd3f92011e6c024477c36798 :)

PR #965, opened 2025-10-17
Add basic support for the BME680 sensor

Cisien:


Cisien:
This is a dupe of https://github.com/meshcore-dev/MeshCore/pull/955

PR #966, opened 2025-10-18
Added a delay to the polling of the analogue user button.

adam2872:
Added a delay to the polling of the analogue user button to allow the ADC to settle.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,10 +1,6 @@
 **MeshCore-FAQ**<!-- omit from toc -->
 A list of frequently-asked questions and answers for MeshCore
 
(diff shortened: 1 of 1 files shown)


examples/companion_radio/ui-new/UITask.h
```diff
@@ -40,6 +40,10 @@ class UITask : public AbstractUITask {
   int last_led_increment = 0;
 #endif
 
+#ifdef PIN_USER_BTN_ANA
+  unsigned long _analogue_pin_read_millis;
```
ripplebiz: Should initialise this to zero, either here or in constructor

adam2872: Its now initialised with millis()

ripplebiz:
Just noticed the base branch of the PR is 'main'. Should be 'dev'.

PR #969, opened 2025-10-18
build.sh: Make firmware build date more universal and include clock

jpds:


liamcottle:
Hello, thanks for the contribution!

Personally, I think `07-Dec-2025` is more user friendly than `202512070451`.

So I'll close this one out for now :)

jpds:
> Personally, I think 07-Dec-2025 is more user friendly

Well, that's not true for non-English speakers, nor for distinguishing between multiple builds done in one day. 

Edit: also, I remember originally doing `2025-12-03 15:45`, but the app truncates the string after a certain point.

PR #972, opened 2025-10-18
all nrf52 devices: force framework-arduinoadafruitnrf52 version to 1.10700.0

recrof:
There are multiple fixes in newer framework-arduinoadafruitnrf52 1.10700.0, but the platformio/nordicnrf52 requires version 1.10601.0 - this PR forces usage of newer version.

thanks @oltaco and @MikesAllotment(and all other people on discord) for testing this


oltaco:
We've had nearly a week of positive testing with this by a couple of people on busy meshes, and it looks like it alleviates the filesystem corruption issues.

PR #973, opened 2025-10-18
equalize RAK with all other nrf52 variants and use newer platform with all important fixes

recrof:
thanks @oltaco and @MikesAllotment(and others on discord) for testing this and raising awareness that the RAK is still using old forked, unmainained nrf52 platform.

### DIFF
--- a/boards/rak4631.json
+++ b/boards/rak4631.json
@@ -1,0 +1,72 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "nrf52840_s140_v6.ld"
+    },
+    "core": "nRF5",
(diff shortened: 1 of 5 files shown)


recrof:
extra perk we get from this: it should not be needed to use RAK_PATCH to build rak4631 variant.

fdlamotte:
I think that's the version I've been using in my out-of-tree builds for rak ;)

RAK's version never worked for me ...

edit: and I never had to use rak patches once I got it to work ...

fdlamotte (approved):
ok for me as I will be able to build rak targets in-tree ;)

PR #982, opened 2025-10-20
Add Seeed Wio WM1110 Dev Board variant

ViezeVingertjes:
And its alive...! Works like champ so far, does require a external programmer though, like the J-Link (cheap clone works fine) or a Raspberry PI. Probably not the most beginner friendly kit, but its pretty cheap.

https://www.seeedstudio.com/Wio-WM1110-Dev-Kit-p-5677.html

### DIFF
--- a/variants/wio_wm1110/target.cpp
+++ b/variants/wio_wm1110/target.cpp
@@ -1,23 +1,6 @@
 #include <Arduino.h>
 #include "target.h"
 #include <helpers/ArduinoHelpers.h>
-#include <helpers/sensors/LocationProvider.h>
-
-class WM1110LocationProvider : public LocationProvider {
-public:
-  long getLatitude() override { return 0; }
-  long getLongitude() override { return 0; }
-  long getAltitude() override { return 0; }
-  long satellitesCount() override { return 0; }
-  bool isValid() override { return false; }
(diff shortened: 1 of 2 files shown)


variants/wio_wm1110/target.cpp
```diff
@@ -0,0 +1,110 @@
...
+  long satellitesCount() override { return 0; }
+  bool isValid() override { return false; }
+  long getTimestamp() override { return 0; }
+  void sendSentence(const char* sentence) override {}
+  void reset() override {}
+  void begin() override {}
+  void stop() override {}
+  void loop() override {}
+  bool isEnabled() override { return false; }
+};
```
ripplebiz: If LocationProvider is not supported, you can put  -U ENV_INCLUDE_GPS  in the .ini file

ripplebiz: Or you can set that to zero,  eg.  -D ENV_INCLUDE_GPS=0

ViezeVingertjes: Good one; it does have GPS but i did not find out how to implement it at the time so it was stubbed. But it works just fine without. 5088444f858a3fe7135ed5f77892be69285c76a6

PR #985, opened 2025-10-21
Fix reversed GPS PINs on G2 and enable timesync

liquidraver:


recrof:
this corrects my mistake - I assumed the pins are as-is, but they are reversed.

PR #986, opened 2025-10-21
gps auto timesync and gps autosync CLI command

liquidraver:
My next shot. hopefully in the right place and not too excessive.


fdlamotte:
sorry, we currently have #778 under test ...

then we can add autosync (if it is not already in the other PR ;))

liquidraver:
ah, another refactor! I'm still too early :)

PR #997, opened 2025-10-22
enable quad-tap GPS on wismesh_tag

kallanreed:


recrof:
there are multiple issues with your PR.

1. we never merge to main branch (only to dev)
2. EnvironmentSensorManager.cpp is used by almost every variant, not just wismesh tag - if you don't have means to test your change, it will be rejected.
3. you changed UITask.cpp for every variant, not just wismesh tag, same as 2)

kallanreed:
Sorry, this PR was meant to target my fork.

PR #998, opened 2025-10-22
Add support for bmp085/bmp180 temperature/pressure sensor

tahnok:
Inspired by #955 and the fact that I have an (as yet) unused BMP180 sensor in my repeater node, add support for the BMP05 and BMP180 sensors via Adafruit's library

No way to set the i2c address from the looks of the library, and I'm not really sure which mode is best but 3/ULTRAHIGHRES is the default in the library. Will check out the [datasheet](https://cdn-shop.adafruit.com/datasheets/BST-BMP180-DS000-09.pdf)

Update: it does appear to get a reading, although it's a bit lower (986 hPa reading, vs 1001 hPa from the nearby airport)

### DIFF
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -312,12 +312,9 @@ bool EnvironmentSensorManager::begin() {
   #endif
 
   #if ENV_INCLUDE_BMP085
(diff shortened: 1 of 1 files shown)


tahnok:
Datasheet shows minimal accuracy changes with MODE/oversampling but some increase in current and measurement time, so I switched to none to save power

PR #1001, opened 2025-10-22
Use correct macro for GPS logging test

kallanreed:
The rest of the GPS tests in this file use `#ifdef PIN_GPS_EN` for this logging check. The logged message implies that the GPS can’t be enabled and in the location class, there are two tests. One for EN and one for RESET so they have different meanings.

For devices that have EN but not RESET pin, this log entry is incorrect.


kallanreed:
Hi, it would be nice to get this one. I submitted #1000 because of this bug but the problem was actually the logging.

weebl2000:
@liamcottle I think this is consistent with the rest of the code, one-liner too.

Issue #241, opened 2025-05-01
[Feature Request] Ability to remotely change public key of a repeater


michaeljaddy:
![image](https://github.com/user-attachments/assets/5d205a79-ebdc-4457-a503-2769b18c7ace)

Not sure if this is even technically possible but on our small mesh in Yorkshire we seem to have run into an issue where we have 2 repeaters and they both have the same 2 characters at the start of the public key which appears to be causing some odd routing decisions.

It would be great if there was a way to change the public key without having to physically visit remotely deployed node.

marrold:
Import / Export of the key would also be useful

mrlmuk:
Does this public key match actually cause an issue?  Wont both nodes receive if they are within range and process the message because it matches the key with no knowledge of the other?

The keys can already be imported and exported via a switch in the firmware, you just need to edit and compile a special version. If you want to export the key from something without that option, first install the companion firmware.

michaeljaddy:
> Does this public key match actually cause an issue?  Wont both nodes receive if they are within range and process the message because it matches the key with no knowledge of the other?
> 
> The keys can already be imported and exported via a switch in the firmware, you just need to edit and compile a special version. If you want to export the key from something without that option, first install the companion firmware.
> 
> 

I believe it is causing me issues with being able to log in to a repeater I remotely manage but I could be wrong. At the moment I only have a single path out of my node into the rest of the mesh - via the red circled repeater at the bottom. That repeater has an adjacency to two other repeaters both prefixed with "ec", so I never know if it taking the blue path or the green path. I own the repeater on the left and frequently have difficulty logging in despite good signal strength - so I can only assume the other repeater is replying first saying "invalid password" before I manage to get logged in to mine.

Even if it's not the cause of that particular issue it's still a pain for running traces etc as I can't force it down a particular path.

![image](https://github.com/user-attachments/assets/4967611b-f121-4df1-8ad7-fc93eb5ce1f1)

mrlmuk:
That's a good example which hopefully the devs can test or already know the answer to.

I would have though when logging in the full credentials (pubkey and password) are encrypted with the public key so only the device with the private key could open the message, see its ID and then check the cleartext password. Others would not even see its own public ID so it would not reply to the wrong password.

I will try and keep my second device just to the south east of you as repeater more often as well, if you can reach it.

trisweb:
[...]

PR #1015, opened 2025-10-25
Add CLI command to regenerate key pair

mattzzw:
In a situation where repeaters face zero hop key prefix collisions, keys will have to be regenerated.
Doing this via the `set prv.key` command is only possible via physical connection to the repeater - for good reasons.

Now, in order to be able to solve a prefix collision without having to climb on roof tops with your notebook in your backpack this PR will add the `regeneratekeys` command to enable remote regenerating of the key pair.
By applying the command the repeater will reboot immediately and by doing so will also send an advert so that the device will show up a couple of seconds later in the contacts list with a new public key.

This fixes https://github.com/meshcore-dev/MeshCore/issues/241

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -807,17 +807,16 @@ void MyMesh::clearStats() {
   ((SimpleMeshTables *)getTables())->resetStats();
 }
 
-void MyMesh::regenerateKeys() {
-  MESH_DEBUG_PRINTLN("Generating new keypair");
-  mesh::LocalIdentity new_id = radio_new_identity();
-  
-  int count = 0;
-  while (count < 10 && (new_id.pub_key[0] == 0x00 || new_id.pub_key[0] == 0xFF)) {
-    new_id = radio_new_identity();
-    count++;
+void MyMesh::regenerateKeys(uint8_t byte) { 
+  if (byte >0 && byte < 0xff){
+    MESH_DEBUG_PRINTLN("Generating new keypair");
+    mesh::LocalIdentity new_id = radio_new_identity();
+    
+    while (new_id.pub_key[0] != byte) {
+      new_id = radio_new_identity();
+    }
+    saveIdentity(new_id);
   }
-  
-  saveIdentity(new_id);
 }
 
 void MyMesh::handleCommand(uint32_t sender_timestamp, char *command, char *reply) {
--- a/examples/simple_repeater/MyMesh.h
+++ b/examples/simple_repeater/MyMesh.h
@@ -188,7 +188,7 @@ public:
 
   void saveIdentity(const mesh::LocalIdentity& new_id) override;
   void clearStats() override;
-  void regenerateKeys() override;  
+  void regenerateKeys(uint8_t byte);  
   void handleCommand(uint32_t sender_timestamp, char* command, char* reply);
   void loop();
 
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -245,11 +245,29 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
     } else if (memcmp(command, "clear stats", 11) == 0) {
       _callbacks->clearStats();
       strcpy(reply, "(OK - stats reset)");
-    } else if (memcmp(command, "regeneratekeys", 14) == 0) {
+    } else if (memcmp(command, "regeneratekeys ", 15) == 0) {
+      // Parse first hex digit
+      int value = 0;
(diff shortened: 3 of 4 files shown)


marcelverdult:
yes please!

recrof:
set prv.key might be more useful here, as you need to run this multiple times if the mesh is too crowded with occupied IDs.

mattzzw:
`set prv.key` is currently only available via serial, so it needs physical access to the device. I think it would make a lot of sense to make `set prv.key` available on the cli console. (`get prv.key` is a complete different story, of course.)
However, using `set prv.key` requires knowledge and access to external key generators like https://gessaman.com/mc-keygen or ssh-keygen adding more steps involved.

Therefore I would propose two ways of solving the collision problem:
1. Make `set prv.key`available via cli console (for the use case of very crowded routers with the need to use pre-defined and pre-generated keys)
2. Add the `regeneratekeys` command as a simple and user friendly "one-click solution"

recrof:
> Add the regeneratekeys command as a simple and user friendly "one-click solution"

I don't think it's friendly at all - it just randomly changes the id. if ID utilization is at 80%, you will need to do it 10x and pollute contact lists of whole mesh with bogus entries.

IoTThinks:
> `set prv.key` is currently only available via serial

Yah, most of the time, the repeaters are glued with silicon to protect them from rain.
Serial connection should be avoided at most of the time.

mattzzw:
Another option would be to let the user provide a desired first byte of the public key as a parameter and let the regenerateKeys routine run for ~256 times until we have a match. Basically doing what https://gessaman.com/mc-keygen/ does for the first byte only. Part of that is already implemented in order to prevent 0x00 and 0xff as a first byte.
Don't know about the runtime though. Any thoughts on that?

mattzzw:
Now the user has to define the first byte of the public key to be generated. 
e.g. 

`> regeneratekeys 2f`

On a Heltec v3 it never took longer than 3 minutes to generate a key pair with matching public key, most of the time it took less or about a minute. 
(I tested this using an unused LoRa configuration in my area...)

recrof:
thanks @mattzzw. can you try how long it takes on nrf52 device? those are deployed as solar nodes most of the time - if it takes 3 minutes on esp32s3 device, I fear it might take ~10 minutes on nrf52.

mattzzw:
Yepp, will check on a faketec w/ nrf52.

mattzzw:
It's more anecdotal evidence, but I did 10 runs and the runtimes where acceptable. I would expect the esp32s3 to be 3-5 times faster than the nrf53 in general. ED25519 operations cannot make use of the crypto hardware available in both chips, however the implementation of the software libs can have a huge impact, as usual. 
Here's the times for the NRF52:
4:40 <
2:05
1:57
0:57
0:18
1:15
0:05
0:06
4:23<
0:48

IoTThinks:
Even with 10 repeaters, I happen to see 2 of them have the same first byte.
How deep is the impact for routing if two repeaters have the same first byte? 
The routing may be wrong to reach the destination wholly?

I guess the first byte is only required to be different with nearby neighbors only.
So instead of setting the wanted byte, we can list those bytes we do not want.
Will be faster.

khudson:
It just means both repeaters will repeat the packet if heard by the same previous hop. The worst thing that happens is there is an unnecessary repeat in terms of airtime utilization. Definitively tracking a packet's path by the client app, on the other hand, is something that will be affected as there's no way for the client to discern which repeater with overlapping prefix actually did repeat a packet.

andyspoo:
> `set prv.key` is currently only available via serial, so it needs physical access to the device. I think it would make a lot of sense to make `set prv.key` available on the cli console. (`get prv.key` is a complete different story, of course.) However, using `set prv.key` requires knowledge and access to external key generators like https://gessaman.com/mc-keygen or ssh-keygen adding more steps involved.
> 
> Therefore I would propose two ways of solving the collision problem:
> 
>     1. Make `set prv.key`available via cli console (for the use case of very crowded routers with the need to use pre-defined and pre-generated keys)
> 
>     2. Add the `regeneratekeys` command as a simple and user friendly "one-click solution"

Seeing as we want people to get their repeaters as high as possible preferable on a mast, and the aerial should be as short as possible to reduce noise then the repeater becomes pretty much inaccessible so the ability to do this remotely makes a lot of sense.

recrof:
I already opened PR to allow set prv.key via remote admin: https://github.com/meshcore-dev/MeshCore/pull/1054

liamcottle:
`set prv.key` is now supported via LoRa as of https://github.com/meshcore-dev/MeshCore/pull/1457

The next app update will also have a new "Change Identity Key" section in repeater remote management.
This is already available via the web app. Closing as this is no longer needed. Clients will be able to provide the exact key to use.

PR #1016, opened 2025-10-25
Support for Lilygo T3S3 LR1121

BetaFloof:
I accidentally bought a Lilygo T3S3 with an LR1121 instead of the already supported SX1262, so I tried implementing it myself since radiolib supports it.

I basically copied the configuration for the T3S3 SX1262 and adjusted it to work with the LR1121. It's mostly code adapted from existing variants using the LR1110. 

Using a second node (RPi Pico + SX1262, fw version 1.9.0) I verified the functionality of:
- BLE companion radio
- USB companion radio
- Terminal chat
- Room server
- Repeater (limited test)

I wasn't able to test chatting over the repeater because I currently don't have enough nodes, but serial commands, advertising, connecting as guest & admin and requesting status all worked.

### DIFF
--- a/src/helpers/radiolib/CustomLR1121.h
+++ b/src/helpers/radiolib/CustomLR1121.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <RadioLib.h>
-#include "MeshCore.h"
 
 #define LR1121_IRQ_HAS_PREAMBLE                     0b0000000100  //  4     4     valid LoRa header received
 #define LR1121_IRQ_HEADER_VALID                     0b0000010000  //  4     4     valid LoRa header received
@@ -9,19 +8,6 @@
 class CustomLR1121 : public LR1121 {
   public:
     CustomLR1121(Module *mod) : LR1121(mod) { }
-
-    size_t getPacketLength(bool update) override {
-      size_t len = LR1121::getPacketLength(update);
-      if (len == 0 && getIrqStatus() & RADIOLIB_LR11X0_IRQ_HEADER_ERR) {
-        // we've just recieved a corrupted packet
-        // this may have triggered a bug causing subsequent packets to be shifted
-        // call standby() to return radio to known-good state
-        // recvRaw will call startReceive() to restart rx
-        MESH_DEBUG_PRINTLN("LR1121: got header err, calling standby()");
-        standby();
-      }
-      return len;
-    }
 
     RadioLibTime_t getTimeOnAir(size_t len) override {
   // calculate number of symbols
--- a/src/helpers/radiolib/CustomLR1121.h
+++ b/src/helpers/radiolib/CustomLR1121.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <RadioLib.h>
+#include "MeshCore.h"
 
 #define LR1121_IRQ_HAS_PREAMBLE                     0b0000000100  //  4     4     valid LoRa header received
 #define LR1121_IRQ_HEADER_VALID                     0b0000010000  //  4     4     valid LoRa header received
@@ -8,6 +9,19 @@
 class CustomLR1121 : public LR1121 {
   public:
     CustomLR1121(Module *mod) : LR1121(mod) { }
+
+    size_t getPacketLength(bool update) override {
+      size_t len = LR1121::getPacketLength(update);
+      if (len == 0 && getIrqStatus() & RADIOLIB_LR11X0_IRQ_HEADER_ERR) {
+        // we've just recieved a corrupted packet
+        // this may have triggered a bug causing subsequent packets to be shifted
+        // call standby() to return radio to known-good state
+        // recvRaw will call startReceive() to restart rx
+        MESH_DEBUG_PRINTLN("LR1121: got header err, calling standby()");
+        standby();
+      }
+      return len;
+    }
 
     RadioLibTime_t getTimeOnAir(size_t len) override {
   // calculate number of symbols
--- a/src/helpers/radiolib/CustomLR1121.h
+++ b/src/helpers/radiolib/CustomLR1121.h
@@ -1,0 +1,70 @@
+#pragma once
+
+#include <RadioLib.h>
+
+#define LR1121_IRQ_HAS_PREAMBLE                     0b0000000100  //  4     4     valid LoRa header received
+#define LR1121_IRQ_HEADER_VALID                     0b0000010000  //  4     4     valid LoRa header received
+
+class CustomLR1121 : public LR1121 {
+  public:
+    CustomLR1121(Module *mod) : LR1121(mod) { }
+
+    RadioLibTime_t getTimeOnAir(size_t len) override {
+  // calculate number of symbols
+  float N_symbol = 0;
+  if(this->codingRate <= RADIOLIB_LR11X0_LORA_CR_4_8_SHORT) {
+    // legacy coding rate - nice and simple
+    // get SF coefficients
+    float coeff1 = 0;
+    int16_t coeff2 = 0;
+    int16_t coeff3 = 0;
+    if(this->spreadingFactor < 7) {
+      // SF5, SF6
+      coeff1 = 6.25;
+      coeff2 = 4*this->spreadingFactor;
+      coeff3 = 4*this->spreadingFactor;
+    } else if(this->spreadingFactor < 11) {
+      // SF7. SF8, SF9, SF10
+      coeff1 = 4.25;
+      coeff2 = 4*this->spreadingFactor + 8;
+      coeff3 = 4*this->spreadingFactor;
+    } else {
+      // SF11, SF12
+      coeff1 = 4.25;
+      coeff2 = 4*this->spreadingFactor + 8;
+      coeff3 = 4*(this->spreadingFactor - 2);
+    }
+
+    // get CRC length
+    int16_t N_bitCRC = 16;
+    if(this->crcTypeLoRa == RADIOLIB_LR11X0_LORA_CRC_DISABLED) {
+      N_bitCRC = 0;
+    }
+
+    // get header length
+    int16_t N_symbolHeader = 20;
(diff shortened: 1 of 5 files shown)


recrof:
just few notes: 
1. getPacketLength in CustomLR1121 can be removed - newest radiolib fixed it
2. LR1121 includes 2 radios - 2.4Ghz and sub-ghz one. the firmware/app does not know one radio supports max 20dBm(2.4Ghz) and 22dBm(sub-ghz).. if you switch the frequency to 2.4, you will get weird behaviour because of this.

src/helpers/radiolib/CustomLR1121.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
+
+#include <RadioLib.h>
+
+#define LR1121_IRQ_HAS_PREAMBLE                     0b0000000100  //  4     4     valid LoRa header received
+#define LR1121_IRQ_HEADER_VALID                     0b0000010000  //  4     4     valid LoRa header received
```
ripplebiz, line 6: The RadioLib constants for these are both one bit shifted left. So, just wondering if these IRQ masks are correct?

src/helpers/radiolib/CustomLR1121.h
```diff
@@ -0,0 +1,70 @@
...
+
+class CustomLR1121 : public LR1121 {
+  public:
+    CustomLR1121(Module *mod) : LR1121(mod) { }
+
+    RadioLibTime_t getTimeOnAir(size_t len) override {
```
ripplebiz, line 12: Might also be able to get rid of this. Latest RadioLib (I think) has fixed this timeOnAir calc issue

variants/lilygo_t3s3_lr1121/target.cpp
```diff
@@ -0,0 +1,95 @@
...
+  spi.begin(P_LORA_SCLK, P_LORA_MISO, P_LORA_MOSI, P_LORA_NSS);
+  int status = radio.begin(LORA_FREQ, LORA_BW, LORA_SF, LORA_CR, RADIOLIB_LR11X0_LORA_SYNC_WORD_PRIVATE, LORA_TX_POWER, 16, tcxo);
+  if (status != RADIOLIB_ERR_NONE) {
+    Serial.print("ERROR: radio init failed: ");
+    Serial.println(status);
+    return false;  // fail
+  }
+  
+  radio.setCRC(2);
+  radio.explicitHeader();
```
ripplebiz, line 65: ^^^ don't think this line should be here. Other radio class std_init() impl's don't do this.

src/helpers/radiolib/CustomLR1121.h
oltaco: I think it's more straightforward to just use the macros that RadioLib provides.

RADIOLIB_LR11X0_IRQ_PREAMBLE_DETECTED
RADIOLIB_LR11X0_IRQ_SYNC_WORD_HEADER_VALID

ripplebiz: After doing some separate tests, I think the RadioLib macros for PREAMBE_DETECTED and HEADER_VALID could be wrong.

oltaco: Not good! I checked the datasheet for LR1121 against the RadioLib LR11X0 macros but it wouldn't surprise me if something was wrong because the LR series has been nothing but trouble 😞

kranimantas:
I'm using the LR1121 on T-Lora Pager and it seems that there's a problem with the Pager transmitting the ACK. The device does receive the message, but the sender does not get the ACK. And then the Pager keeps trying (the green dot in the Ultra firmware is on), but never sends that ACK out. While the Pager is doing it, it is not receiving any new messages. A workaround is to try to send out a typed-in message from the Pager, and then it would receive one message again, and would get stuck into that ACK cycle.

Same behaviour on the Companion firmware, so I think this is a radio-level issue.

Works good with Meshtastic though (if we consider the Base UI "good", anyway).

LanWolf:
> I'm using the LR1121 on T-Lora Pager and it seems that there's a problem with the Pager transmitting the ACK. The device does receive the message, but the sender does not get the ACK. And then the Pager keeps trying (the green dot in the Ultra firmware is on), but never sends that ACK out. While the Pager is doing it, it is not receiving any new messages. A workaround is to try to send out a typed-in message from the Pager, and then it would receive one message again, and would get stuck into that ACK cycle.
> 
> Same behaviour on the Companion firmware, so I think this is a radio-level issue.
> 
> Works good with Meshtastic though (if we consider the Base UI "good", anyway).

I am having the same issue, not being able to receive, only one msg after sending one.

liamcottle:
Hey, thanks for the PR!

Unfortunately there's been no activity for a long time and the review comments haven't been resolved.

I'll close this one out for now.

kranimantas:
So if I understand this correctly, LR1121 is abandoned? LR1121 currently seems to be not operational in any Meshcore firmware (it works in Meshtastic though).

liamcottle:
> So if I understand this correctly, LR1121 is abandoned?

It looks like the PR was abandoned.

It was opened in October 2025, had some review comments left, but they were never followed up. So I closed it out.

I don't have any of this hardware, so it's not something I can work on. I have some of the newer LR2021 devices which support multi-SF operation, which I think will be getting some time put into in the coming months, as it would be very interesting to use these features.

Can open the PR again if the author is able to make the changes that were requested. But for now it's closed due to inactivity and no one else looked to take it over.

BetaFloof:
Apologies for my inactivity.
I've had very little free time in the past few months and had honestly forgotten about this.

I appreciate the comments and I'll try my best to address them in the coming days/weeks.

I'm not really familiar with the intricacies of RadioLib or the LR1121 so I'll have to dig deeper than I've done at the start.

BetaFloof:
I've pulled my branch up to latest dev and have added adjustments that have happened in the past.

I have done the same tests as at the start. This time on v1.14.1 and v1.15.0. Again against my second node (RPi Pico + SX1262, fw version 1.14.0).
From what I can tell, all firmwares work as expected.

variants/lilygo_t3s3_lr1121/target.cpp
BetaFloof: I've taken that from the t1000-e variant. I also found it on the Thinknode m3 and wio wm1110 which all use the LR1110.
Though I've tested with it removed and I haven't noticed any issues.

src/helpers/radiolib/CustomLR1121.h
BetaFloof: Removed 👍

src/helpers/radiolib/CustomLR1121.h
BetaFloof: I'm now using the RadioLib macros and haven't noticed any issues

BetaFloof:
> I'm using the LR1121 on T-Lora Pager and it seems that there's a problem with the Pager transmitting the ACK. The device does receive the message, but the sender does not get the ACK. And then the Pager keeps trying (the green dot in the Ultra firmware is on), but never sends that ACK out. While the Pager is doing it, it is not receiving any new messages. A workaround is to try to send out a typed-in message from the Pager, and then it would receive one message again, and would get stuck into that ACK cycle.
> 
> Same behaviour on the Companion firmware, so I think this is a radio-level issue.
> 
> Works good with Meshtastic though (if we consider the Base UI "good", anyway).

@kranimantas @LanWolf

I don't think that issue is related to this PR. My work here is only targeting the LilyGo T3 S3 LR1121 fw. It seems like an issue in the Ultra firmware

BetaFloof:
I believe there is an issue reading the noise floor. I'll have a look into that

jimdigriz:
@BetaFloof if you need any help or do not have the bandwidth to continue this do ask, I have the same hardware and am using it.

Also, I just tested your branch and it worked. Thanks!

kranimantas:
@BetaFloof I just compiled your branch with some tweaks for the T-Pager (only changed the pinout), and BLE companion does indeed work. Then tested with multiple nodes, repeater admin, everything works flawless. The only thing I see, keyboard backlight is always on, but that could be a bug from the previous firmwares I was playing with.

You give us hope! 

@ripplebiz any chance of implementing this in the Ultra firmware? I am also willing to help with testing, if needed. 

image

BetaFloof:
I'm glad this is working for you guys.

@jimdigriz I appreciate your offer to help.
I'd be interested if you have tested it "in the field". In my testing I have received suspiciously few repeater and contact adverts or messages in the public channel from others compared to my Pi Pico node. Not sure if this is an issue with the firmware or my setup.
Also, are you getting a constant noise floor of -120 dB as well?
There's likely an issue in my `getCurrentRSSI()`

jimdigriz:
> In my testing I have received suspiciously few repeater and contact adverts or messages in the public channel from others

@BetaFloof I will try to find the time to do a bit more than an sniff test with the PR. The issue I have is there are only three repeaters (one of which is mine) in the rural area I live. My repeater has been taken down for maintenance and one of the other repeaters is only really accessible via my repeater. I'll see what I can do and if I notice anything I'll fiddle with the firmware to see if I can improve things.

> Also, are you getting a constant noise floor of -120 dB as well?

I'll have a nosey, hopefully night.

oltaco:
> In my testing I have received suspiciously few repeater and contact adverts or messages in the public channel from others compared to my Pi Pico node.

I had a quick look at this old PR and I think the RF switch table might be incorrect? The LilyGO github lists them here
https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/docs/en/t3_s3_lr1121_pa/t3_s3_lr1121_pa_hw.md#lora-antenna-switch-truth-table

BetaFloof:
@oltaco Thanks for that! I've updated the table and I'm now getting pretty much the same adverts and messages as on the Pico.
The noise floor issue also seems to have been caused by this as it's not stuck at -120 dB anymore and updating regularly.

PR #1032, opened 2025-10-30
Allow UIScreen first dibs on button events

kallanreed:
This is part of a larger change to add a new UI screen. I need to use the triple-click gesture so I need the screen to be passed the key before any custom override in the UITask handlers.

In the current implementation triple click will always toggle the buzzer which is an unwanted side effect for my new screen.

This change simplifies UI event handling, making it more consistent and allowing UIScreen implementations to handle keys first.

There should be no functional changes as a result of this PR. I have tested this on my T114.

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -4,7 +4,7 @@
 #include "target.h"
 
 #ifndef AUTO_OFF_MILLIS
-  #define AUTO_OFF_MILLIS    15000  // 15 seconds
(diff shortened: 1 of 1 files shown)


ripplebiz:
Per my comment of #1041, with a new/experimental /test-ui folder, this PR wouldn't be needed.
As long as the UITask in the test-ui conforms to the AbstractUITask, then all is good.

kallanreed:
This is really more code cleanup than a functional change. Seems worth doing regardless of the other PR.

PR #1036, opened 2025-10-31
Refactor DataStore to use openRead() and openWrite()

oltaco:
refactored loadPrefsInt(), loadContacts(), loadChannels(), getBlobByKey() and putBlobByKey() to use openRead() and openWrite()

This gets rid of a few of the architecture ifdefs since that's already handled for us in openRead and openWrite.

Tested on nRF52 and ESP32, but should be fine for STM32 and RP2040 since it was already done this way elsewhere in DataStore (saveContacts for example already used openWrite with no arch ifdefs).


liamcottle:
One thing to note here, is on nrf52, `openWrite` will delete the file before opening it. So I'm wondering if this will nuke all previous adv blobs each time you save a new one, since we don't write all of them back.

oltaco:
Good catch! 🤦‍♂️
I've reverted the change to putBlobByKey() since it's the only place we are still using seek() instead of remove then rewrite.

liamcottle:
Sweet! Looks good. Will leave for @ripplebiz to review in case I missed anything else

ripplebiz:
Nice.  Like seeing the line count reduced! :-)

PR #1041, opened 2025-10-31
Add "Quick Messages" screen to new UI

kallanreed:
This allows users to send a canned message to a channel directly from a companion radio without using the app.
This is useful for when using the radio standalone and when testing with multiple radios.

This PR includes changes the necessary changes from #1032. I'd prefer if #1032 went in separately.
Implementation is hidden behind a build flag that I've enabled only for the Heltec T114 for now.
The set of messages is hard coded.

Usage:
When on the page, long press/Enter to enter the Quick Messages screen.
- double click/Left moves between fields.
- single click/Right cycles forward through the values.
- triple click cycles backward through the values.
- single click on "[send]" will send the message to the selected channel.
- long press will exit the Quick Messages screen.
- If GPS is enabled, one of the message options will be to send your current position.

Tested on T114.

Future ideas:
- allow message set to be configured from the app.

feature in use

### DIFF
--- a/examples/companion_radio/ui-new/QuickMsg.cpp
+++ b/examples/companion_radio/ui-new/QuickMsg.cpp
@@ -8,7 +8,8 @@
 #endif
 
 static constexpr const char* messages[] {
-  "ping", "ack", "yes", "no", "test"
+  "test", "ping", "hello", "ack", "yes", "no", "share location",
+  "come to me", "going to you", "help", "SOS"
 };
 static constexpr size_t messages_count = COUNTOF(messages);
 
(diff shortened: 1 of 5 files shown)


swannman (approved):

ripplebiz:
I'm not sure there really is a consensus on whether having all the double/triple click stuff in the UI like this. 
What I'd suggest is create a new '/test-ui' folder in the companion firmware, and do whatever in that one and see what people's feedback is like, or whether it gains popularity. If so, then the 'official' release variants could switch to that one.

kallanreed:
It is opt-in with a build flag. I set the flag in the T114 build, but I'm happy to remove that from the PR.

liamcottle:
Hey, thanks for the PR! Since this one is quite old now, and didn't get much traction, I'll close it out.

Issue #1044, opened 2025-11-01
[Feature Request] Preserve and use the ed25519 seed to improve compatibility with modern implementations


aqua:
MC's ed25519 library is a 2013 implementation by Orson Peters -- while it generates new keypairs from a 32-byte seed as normal, the representation it uses for the private key is the sha512 of the seed after cofactor clamping.  This works, but few if any modern ed25519 libraries are compatible with that representation.  Most just use the seed, or a few use the right half of the sha512 before clamping, neither of which are possible using the keys stored by the current MC firmware.

This mostly matters to loggers like mctomqtt, which do a `get prv.key` to obtain the secret key so they can decode the raw packets, and are thus very constrained in what crypto libraries they can use (mctomqtt shells out to a Node implementation of a WASM build of Peters' library).

Please consider preserving the original seed, and making it available via the CLI.  I'll send a PR that implements this.  It won't help nodes already deployed with current keys (their seeds are long gone) but loggers will generally be able well connected and able to regen their keys if they need the functionality.

Background:

- https://blog.mozilla.org/warner/2011/11/29/ed25519-keys/ (writeup on some of the various definitions of "private key" used since the original team's reference implementation)
- Example ed25519 APIs in [Python](https://pynacl.readthedocs.io/en/latest/signing/), [Rust](https://docs.rs/ed25519/latest/ed25519/index.html) and [Go](https://pkg.go.dev/golang.org/x/crypto/ed25519#NewKeyFromSeed)

PR #1045, opened 2025-11-01
Preserve and make the ed25519 key accessible via the CLI

aqua:
Addresses issue #1044.

- Stores the original seed during keygen
- Adds `get prv.seed` CLI to obtain it
- Adds `set prv.seed` to generate a new keypair from a seed chosen off-node (could be useful for devices without reliable entropy)
- Reworks Identity::writeTo() to semantically explicit forms since the seed and the pubkey are the same length and can't be disambiguated soley via the size_t.


446564:
@aqua can you please resolve the conflicts on this one?

aqua:
@446564 Done, but PR #1055 would be the better one to review, this one is against main because I suck at github.

446564:
I'll take a look there then, do you want to close this one in favor of that then?

aqua:
I held off because Liam was citing this one in other PRs, but he seems to have popped into #1055 now, so yes.

PR #1047, opened 2025-11-01
Fix the sample RAK repeater build target name

aqua:
The actual target doesn't capitalize the 'r' in repeater.


liamcottle:
Thanks!

PR #1052, opened 2025-11-01
Add per-channel flags; support "no store" channels

swannman:
In the PugetMesh community we have channels like #bot and #testing that are mostly used for "can you hear me?" tests. These are the most heavily-used channels in the region, yet the messages are not useful to store offline - they take up space on the device and are time-consuming to sync when reconnecting a phone to the companion.

This PR introduces a per-channel "no store" flag that tells the companion not to store any messages for the channel unless a phone is connected. This reserves more offline storage for channels that actually matter, and makes reconnecting faster.

This PR also introduces the concept of channel flags. These are persisted in the unused 4-byte header of each channel in the data store. Other per-channel settings such as notifications enabled/disabled can be stored here in the future.

Testing performed on a WisMesh Tag:
### Upgrade
- [x] Upgrade from previous dev branch firmware to this firmware: all channels default to `noStore=false`. 

### Behavior when noStore = false
- [x] Verified that the companion continued to receive and store messages for all channels when disconnected from phone.

### Behavior when noStore = true
- [x] Verified that the companion continued to store messages for these channels when connected to phone
- [x] Verified that the companion did not store messages for these channels when disconnected from phone

### Settings persistence
- [x] Verified that `CMD_GET_CHANNEL_FLAG_NOSTORE` returns the noStore value
- [x] Verified that `CMD_SET_CHANNEL_FLAG_NOSTORE` correctly sets the noStore value

### API support
- [x] Added Python API bindings: https://github.com/meshcore-dev/meshcore_py/pull/31

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,44 @@
+{
+	"name": "MeshCore",
+	"image": "mcr.microsoft.com/devcontainers/python:3-bookworm",
+	"features": {
+		"ghcr.io/rocker-org/devcontainer-features/apt-packages:1": {
+			"packages": [
+				"sudo"
+			]
+		}
+	},
+	"runArgs": [
+		"--privileged",
+		// arch tty* is owned by uucp (986)
+		// debian tty* is owned by uucp (20) - no change needed
+		"--group-add=986",
+		"--network=host",
+		"--volume=/dev/bus/usb:/dev/bus/usb:ro"
+	],
+	"postCreateCommand": {
+		"platformio": "pipx install platformio"
+	},
+	"customizations": {
+		"vscode": {
+			"settings": {
+				"platformio-ide.disablePIOHomeStartup": true,
+				"editor.formatOnSave": false,
+				"workbench.colorCustomizations": {
+					"titleBar.activeBackground": "#0d1a2b",
+					"titleBar.activeForeground": "#ffffff",
+					"titleBar.inactiveBackground": "#0d1a2b99",
+					"titleBar.inactiveForeground": "#ffffff99"
+				}
+			},
+			"extensions": [
+				"platformio.platformio-ide",
+				"github.vscode-github-actions",
+				"GitHub.vscode-pull-request-github"
+			],
+			"unwantedRecommendations": [
+				"ms-vscode.cpptools-extension-pack"
+			]
+		}
+	}
+}
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ cmake-*
 .cache
 .ccls
 compile_commands.json
+.venv/
+venv/
--- a/boards/thinknode_m3.json
+++ b/boards/thinknode_m3.json
@@ -1,0 +1,72 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "nrf52840_s140_v6.ld"
+    },
+    "core": "nRF5",
+    "cpu": "cortex-m4",
+    "extra_flags": "-DNRF52840_XXAA",
+    "f_cpu": "64000000L",
+    "hwids": [
+      [
+        "0x239A",
+        "0x4405"
+      ],
+      [
+        "0x239A",
+        "0x0029"
+      ],
+      [
+        "0x239A",
+        "0x002A"
+      ]
+    ],
+    "usb_product": "elecrow_eink",
+    "mcu": "nrf52840",
+    "variant": "ELECROW-ThinkNode-M3",
+    "bsp": {
+      "name": "adafruit"
+    },
+    "softdevice": {
+      "sd_flags": "-DS140",
+      "sd_name": "s140",
+      "sd_version": "6.1.1",
+      "sd_fwid": "0x00B6"
+    },
+    "bootloader": {
+      "settings_addr": "0xFF000"
+    }
+  },
+  "connectivity": [
+    "bluetooth"
+  ],
+  "debug": {
+    "jlink_device": "nRF52840_xxAA",
+    "onboard_tools": [
+      "jlink"
+    ],
+    "svd_path": "nrf52840.svd",
+    "openocd_target": "nrf52.cfg"
+  },
+  "frameworks": [
+    "arduino"
+  ],
+  "name": "elecrow nrf",
+  "upload": {
+    "maximum_ram_size": 248832,
+    "maximum_size": 815104,
+    "speed": 115200,
+    "use_1200bps_touch": true,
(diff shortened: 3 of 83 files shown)


examples/companion_radio/DataStore.h
```diff
@@ -5,6 +5,16 @@
 #include <helpers/ChannelDetails.h>
 #include "NodePrefs.h"
 
+struct ChannelFlags {
```
kallanreed: This ends up being intrinsic to the ChannelDetails, define it in that header?

examples/companion_radio/DataStore.cpp
```diff
@@ -324,9 +324,11 @@ void DataStore::loadChannels(DataStoreHost* host) {
       uint8_t channel_idx = 0;
       while (!full) {
         ChannelDetails ch;
-        uint8_t unused[4];
+        ChannelHeader header;
+
+        bool success = (file.read((uint8_t*)&header, sizeof(ChannelHeader)) == sizeof(ChannelHeader));
+        ch.channel.flags.noStore = header.flags.noStore;
```
kallanreed: Is it possible add the header directly to the ChannelDetails instead? It would add 4 bytes, but depending on the alignment (I have no idea off the top of my head), it might be the same actual overhead as adding a single byte. I don't see any `#pragma pack` on these structs so it's probably whatever the platform defines.

kallanreed: `ch.channel.flags = header.flags;` so this "just works" when adding new flags later.

examples/companion_radio/DataStore.cpp
```diff
@@ -347,11 +349,12 @@ void DataStore::saveChannels(DataStoreHost* host) {
...
     uint8_t channel_idx = 0;
     ChannelDetails ch;
-    uint8_t unused[4];
-    memset(unused, 0, 4);
-
+    
     while (host->getChannelForSave(channel_idx, ch)) {
-      bool success = (file.write(unused, 4) == 4);
+      ChannelHeader header{};
+      header.flags.noStore = ch.channel.flags.noStore;
```
kallanreed: Same here.

examples/companion_radio/DataStore.h
```diff
@@ -5,6 +5,16 @@
...
+    bool noStore : 1;
+    uint8_t reserved : 7;  // remaining 7 bits unused
+};
+
+struct ChannelHeader {
+    ChannelFlags flags;
+    uint8_t unused[3];
+};
```
kallanreed, line 11: Because this is for serialization, it needs to be 4 bytes (for consistently with current data).
Add a static assert here to ensure ChannelHeader's size is what you expect.
`static_assert(sizeof(ChannelHeader) == 4, "ChannelHeader must be 4 bytes");

examples/companion_radio/DataStore.cpp
kallanreed: If you want to experiment with it, you can use static asserts on ChannelDetails to verify the size of the object matches what you expect.

swannman: Good suggestion. I tried it and two things felt off:
1. I'm not sure ChannelDetails should know about the "header" concept from the data storage layer. It's kind of a side-effect of maintaining backwards compatibility with previous firmware.
2. `ChannelDetails.channel` has a reference to the flags internally, so we'd be introducing two ways to access the same data (`ChannelDetails.header.flags` and `ChannelDetails.channel.flags`). We only have a reference to `ChannelDetails.channel` in `MyMesh::onChannelMessageRecv`, so we wouldn't be able to remove `ChannelDetails.channel.flags` to clean this up.

swannman:
@liamcottle would you like me to add `CMD_SET` and `CMD_GET` commands for this flag so it's ready for integration with clients? Just need to know how to test them if so. Thanks!

Update: the commands are now implemented and were tested using meshcore_py.

kallanreed (approved):

src/Mesh.h
```diff
@@ -4,10 +4,16 @@
 
 namespace mesh {
 
+struct ChannelFlags {
+    bool noStore : 1;
+    uint8_t reserved : 7;  // remaining 7 bits unused
+};
+
```
ripplebiz: This shouldn't be a concept for the core layer.
The GroupChannel class has only the bare info needed.

src/helpers/ChannelDetails.h
```diff
@@ -6,4 +6,4 @@
 struct ChannelDetails {
   mesh::GroupChannel channel;
   char name[32];
-};
```
ripplebiz, line 9: the flags byte would be better here.

ripplebiz:
Also, you really need to get buy-in from Liam, to support modifying the CMD_GET_CHANNEL and CMD_SET_CHANNEL companion protocol

swannman:
@ripplebiz thank you for the code review! Updated with your feedback.

> Also, you really need to get buy-in from Liam, to support modifying the CMD_GET_CHANNEL and CMD_SET_CHANNEL companion protocol

I think it makes sense to add new commands for getting and setting the flags for a given channel at a time. I'll work on that next!

swannman:
@ripplebiz I've added new CMD_GET_CHANNEL_FLAG_NOSTORE and CMD_SET_CHANNEL_FLAG_NOSTORE commands here, so there is no impact to the existing companion protocol. I've also published a PR that adds these to the Python API: https://github.com/meshcore-dev/meshcore_py/pull/31

Would you mind taking another look?

fdlamotte:
Just a little thought, maybe no_store channels could be handled only by the application ? (as there is no need to store message when connected to the app)

I've been thinking that an alternative way to get the same result as proposed here would be to register no_store channels only in the app and read them from rx_log

Then to send on these channels, there would be a new command to send a msg providing the pubkey of the channel where to send the message to 

And channels configured on the device would be dedicated to "store" channels (and we can lower their number, which would be another benefit), there would then be an unlimited number of no_store channels

swannman:
@ripplebiz or @liamcottle do you have any concerns about this change?

agessaman:
As someone who swaps back and forth between a Wismesh Tag and an Ikoka Stick, I support this change.

Beyond the mentioned #bot channel, we have another channel called #wardrive where users post their lat/long to fill out a coverage map. It's a bit spammy and low value to store, and will fill up the radio very quickly.

The default behavior is unchanged, but for power users in this situation this would mean that more messages on an alternate radio would be relevant when syncing.

swannman:
@liamcottle would you mind reviewing this when you have a moment? Thanks!

PR #1054, opened 2025-11-03
allow `set prv.key` via LoRa for easier repeater id collision fixup

recrof:
Remote-setting private key is crucial when operator needs to change public key of the node because of ID collision.
In case of repeater breach: setting different private key by the attacker has low gain - owner will be notified that the public key was changed as soon as repeater adverts again.


ripplebiz:
I just realised that changing the private key (ie. self_id) invalidates the ACL. For performance reasons, I decided to persist the shared_secret for each entry in the ACL.
The MyMesh::saveIdentity() method should call a (not yet written) acl.erase() method, to blow away the file.

recrof:
@ripplebiz does that mean you'll first implement the required changes before this PR can be merged?

liamcottle:
`set prv.key` is now supported via LoRa as of https://github.com/meshcore-dev/MeshCore/pull/1457

Issue #388, opened 2025-06-11
Turning off Meshcore on Heltec V3 not working


alrik11es:
As the title implies, I can't turn off my Heltec V3 running Meshcore. I've tried multiple versions. In Meshtastic, for example, it turns off by holding the function button for 3–5 seconds, but that doesn't seem to work here.

tarastella:
Let me say that it's not urgent.
I have a Muzi Works R1 with an external antenna, based on WisBlock RAK4631.
In Meshtastic, the application allows for the shutdown of the device. Pressing the button for 5 seconds will also shut it down.
It's not healthy for the device, but I'm just waiting for the battery to drain and that's it. However, if you ever consider fixing this for the Heltec V3, you might also want to take a look at these types of external devices. Thanks a million.

ErueWoad:
I would also like the ability to turn off my Heltec V3's

RodMyers:
adding my voice for this necessary feature

Jamira40:
Also adding to this feature. I have several Heltec v3 portable nodes and it would be neat to have ability put them to deep sleep. Its good for battery :)

fvanson:
+1
In my case for the Heltec T114. I am using the Heltec T114 for my mobile nodes, and being able to put them to deep sleep or standby mode would be super handy for at least the devices running the companion firmware. It would enable safe swapping and tinkling with antennas or when traveling (via air) you can turn off all radios.
Could this be please be implemented in a future firmware update?

d6e:
https://github.com/meshcore-dev/MeshCore/pull/603

4np:
AFAIK you also need to power off a device to switch antennas as removing an antenna from a device actively using Lora may damage the hardware. My T114 (companion) and RAK4631 (companion & repeating room server) devices can only power down by pulling the battery plug which is a bit cumbersome.

For repeaters / room servers, perhaps a command line command can be introduced to power down the radios for a given number of seconds (`lora off 120`, `ble off 60`)? Maybe with a led blinking slowly and more rapidly nearing the end of the timer?

Feika1981:
To turn off my heltec, I need to hold down the top button and then press the bottom one.

liamcottle:
These devices don't actually support power off. They're more of a "sleep" or "hibernation" state, where they still use power. The new UI that is in the dev branch, and should be available in v1.8.0 firmware will have the ability to hibernate the device. I'll close this out since that's been implemented already.

[...]

PR #1056, opened 2025-11-03
No wake up on receiving LoRa packet

vityevato:
In fact, it puts the device into a complete sleep. :)

The presented changes remove wakeup on LoRa activity.


IoTThinks:
Then how to poweroff but can be waken up by LoRa packets?
Should we have public hibernate method to deepsleep but can be waken up?

I'm pushing a PR to get repeaters to deepsleep.
So this PR may affect me as well.

vityevato:
I completely forgot about everything except the companion :) I’ll redo the pr

PR #1057, opened 2025-11-03
Disable screen switching on when connected

liquidraver:
screen flashes on if a message is received, even if we are connected to the node.  (and showing 0, because it synced them to the phone)

less battery usage, should slow down climate change by at least a second


liquidraver:
@fdlamotte can you look at this? the PR stuck for a while and it is a good UX/QoL change IMHO

fdlamotte:
Seems ok, I'll merge it

PR #1061, opened 2025-11-04
No wake up on receiving LoRa packet on companion firmware

vityevato:
In fact, it puts the device into a complete sleep in companion firmware. :)

The presented changes remove wakeup on LoRa activity for **companion firmware**.

@IoTThinks Here is a new proposal :)


IoTThinks:
I feel it should be more straightforward.

vityevato:
Not to the `dev` branch :(

PR #1063, opened 2025-11-04
Adding the ability to configure whether or not to wake up from a LoRa packet for the companion firmware

vityevato:
## Overview
This PR implements configurable wakeup control for companion radio devices, allowing users to choose whether the device can be woken from deep sleep by LoRa packet reception or only by the RESET button.

### 1. **Wakeup Configuration**
- Added `enable_wakeup` field to `NodePrefs` structure for persistent storage
- Implemented Rescue Mode command `set wakeup on/off` to control wakeup behavior
- Added `setEnableWakeup()` method to `MainBoard` interface for runtime control

### 2. **Auto-Sleep with Packet Activity Detection**
- Implemented auto-sleep timer (5 seconds) that activates when device wakes from LoRa packet
- Timer resets on any received packet to keep device awake during active communication
- Automatic deep sleep entry when timeout expires with no packet activity
- Overridden `onRecvPacket()` in `MyMesh` to handle timer reset logic

### 3. **Enhanced Deep Sleep Control (HeltecV3Board)**
- Modified `enterDeepSleep()` to accept optional `wakeup` parameter
- Conditional DIO1 configuration: only set up LoRa wakeup when enabled
- NSS pin hold is always maintained regardless of wakeup setting
- Proper GPIO deinitialization on wakeup to prevent conflicts

### 4. **Display Optimization**
- Display initialization now skipped when device wakes from LoRa packet
- Display only initializes on normal boot or user-triggered wakeup

### 5. **Improved CLI User Experience in Rescue Mode**
- Added `help` command listing all available Rescue Mode commands
- Enhanced error messages to guide users to help command
- Immediate feedback when wakeup setting is changed

**Default Behavior:**
- Wakeup is **disabled by default** (only RESET button wakes device)
- Auto-sleep activates only when woken by LoRa packet
- 5-second timeout with automatic reset on packet activity

## Use Cases

1. **Battery Conservation Mode** (`wakeup off`): Device stays in deep sleep until manually reset, maximizing battery life
2. **Active Monitoring Mode** (`wakeup on`): Device wakes on LoRa packets, processes them, and returns to sleep if no further activity

*This PR as a possible alternative implementation of #1061.*
*These changes in the context of the problem:https://github.com/meshcore-dev/MeshCore/issues/388#issuecomment-3240836807*

### DIFF
--- a/examples/companion_radio/DataStore.cpp
+++ b/examples/companion_radio/DataStore.cpp
@@ -225,6 +225,7 @@ void DataStore::loadPrefsInt(const char *filename, NodePrefs& _prefs, double& no
     file.read((uint8_t *)&_prefs.multi_acks, sizeof(_prefs.multi_acks));                   // 77
     file.read(pad, 2);                                                                     // 78
     file.read((uint8_t *)&_prefs.ble_pin, sizeof(_prefs.ble_pin));                         // 80
+    file.read((uint8_t *)&_prefs.enable_wakeup, sizeof(_prefs.enable_wakeup));             // 84
 
     file.close();
   }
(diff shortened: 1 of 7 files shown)


vityevato:
This is all pointless since the companion will never fall asleep in an active network :( Bad idea

PR #1069, opened 2025-11-06
T114: Allow GPS to sleep

rxkris:
Allow GPS to go to sleep on T114, VEXT needs to be active to control sleep modes of GPS module. If VEXT is turned off with GPS connected, GPS will be always active. Therefore use GPS_WAKEUP to enable/disable GPS instead of VEXT. This commit sets GPS_EN to GPS_WAKEUP.


weebl2000:
See also #1633 - I think this PR can be closed in favor of #1633 - it's a more generic fix.

weebl2000:
@liamcottle I think this PR can be closed since #1633 was merged and that fixes the power consumption issue for GPS.

PR #1073, opened 2025-11-07
we shouldn't leave debug switches uncommented on main/dev branches

liquidraver:


liquidraver:
we should disable debug flags when building for prod

PR #1075, opened 2025-11-08
Add statistics commands to return device metrics on companions

agessaman:
This adds command codes that return the same data as stats-core, stats-radio, and stats-packets do on repeaters and room servers. The responses are formatted in json, the same as the other devices.

- Added new commands for retrieving statistics: CMD_GET_STATS_CORE, CMD_GET_STATS_RADIO, and CMD_GET_STATS_PACKETS.
- Added matching response handling methods to format and send statistics data.
- Updated MyMesh constructor to initialize new member variables for managing statistics.
- Included StatsFormatHelper for formatting statistics replies.

These could be useful statistics to gather periodically to display device utilization in a client app. They would also definitely be useful for providing network metrics using a companion device as an observer device.

I have a parallel PR for meshcore_py ready to go that accesses these codes:
```INFO:meshcore:Scanning for devices...
INFO:meshcore:Found device: 00DDF382-9A5E-F28A-7589-7493F94BF0C6: MeshCore-HOWL
INFO:meshcore:BLE Connection started
Connected successfully!

Fetching core statistics...
📊 Core Statistics:
{
  "battery_mv": 4012,
  "uptime_secs": 1203,
  "errors": 0,
  "queue_len": 0
}

Fetching radio statistics...
📡 Radio Statistics:
{
  "noise_floor": -83,
  "last_rssi": -44,
  "last_snr": 12.25,
  "tx_air_secs": 0,
  "rx_air_secs": 38
}

Fetching packet statistics...
📦 Packet Statistics:
{
  "recv": 127,
  "sent": 2,
  "flood_tx": 2,
  "direct_tx": 0,
  "flood_rx": 126,
  "direct_rx": 1
}

Disconnecting...
Disconnected.

### DIFF
--- a/docs/stats_binary_frames.md
+++ b/docs/stats_binary_frames.md
@@ -6,37 +6,53 @@ Binary frame structures for companion radio stats commands. All multi-byte integ
 
 | Command | Code | Description |
 |---------|------|-------------|
-| `CMD_GET_STATS_CORE` | 56 | Get core device statistics |
-| `CMD_GET_STATS_RADIO` | 57 | Get radio statistics |
-| `CMD_GET_STATS_PACKETS` | 58 | Get packet statistics |
+| `CMD_GET_STATS` | 56 | Get statistics (2-byte command: code + sub-type) |
+
+### Stats Sub-Types
+
+The `CMD_GET_STATS` command uses a 2-byte frame structure:
+- **Byte 0:** `CMD_GET_STATS` (56)
+- **Byte 1:** Stats sub-type:
+  - `STATS_TYPE_CORE` (0) - Get core device statistics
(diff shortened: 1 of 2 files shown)


agessaman:
Updated to use binary frames to be consistent with other companion commands at Liam's suggestion.

ripplebiz:
All looks good, except I was chatting to Liam, and we're thinking prob best if the CMD frames could be 2 bytes, so 1 CMD_GET_STATS  code, then second byte with the sub-type. Could you do that?
Just don't want to use up all the available CMD_ codes too quickly.

agessaman:
Gotcha. That makes total sense. In order to use a similar structure, I also cut it down to a single response code (26) where the second byte verifies the response structure by including the same sub-type byte as the request.

Issue #1049, opened 2025-11-01
Powered off companion will be back to on when receiving a LoRa packet?


IoTThinks:
Hi guys,
I'm checking the code and see something may not right.

When we press the button to off the companion, it will call powerOff()
<img width="1388" height="929" alt="Image" src="https://github.com/user-attachments/assets/20a0f87b-10af-4a9f-85be-46a7c88fe7b9" />

Then it will call enterDeepSleep(0);
<img width="1378" height="918" alt="Image" src="https://github.com/user-attachments/assets/afefcf50-209e-4a26-8ab8-7a176085e3bf" />

Then it will reach here. 
The companion will wake up if it receives a LoRa packet.
`esp_sleep_enable_ext1_wakeup( (1L << P_LORA_DIO_1), ESP_EXT1_WAKEUP_ANY_HIGH);  // wake up on: recv LoRa packet`

When we poweroff the companion, the companion should wake up again ONLY IF we press the button again.
Thanks a lot.

IoTThinks:
I will make a pull request to fix it.

PR #1076, opened 2025-11-08
No wake up on receiving LoRa packet on companion firmware

vityevato:
In fact, it puts the device into a complete sleep in companion firmware. :)

The presented changes remove wakeup on LoRa activity for **companion firmware**.

and #1049


hubba1987:
I made some tests with these changes.

Power consumption:
while loading: 40 mA
working: 120 mA
in hibernation: 0.17 mA

mmatoszko:
@hubba1987 which device? what's the power characteristic for it before the change?

hubba1987:
> @hubba1987 which device? what's the power characteristic for it before the change?

Heltec V3

liamcottle:
Hello, thanks for the PR. Unfortunately this one didn't gain any traction, and there's been a lot of work in the repo related to sleep and power saving. I'll close this one out for now.

PR #1079, opened 2025-11-08
Support powersaving for repeaters

IoTThinks:
Hi all,
I'm adding a feature to let repeaters to deepsleep their MCUs to save power.
LoRa is still in RX mode.

I set the powersaving mode to Heltec v3 repeaters by default.
For Heltec v3, via battery supply, I can reduce the power consumption from 47mA down to 7mA.
The Heltec v3 will wake up when hearing a LoRa packet and repeats as normal.
Tested OK for a week.
So this feature makes Heltec v3 more practical as a repeater.

The side effects are:
* When the repeater wakes up, it will send and advert.
* The uptime is always a few seconds most of the time as the repeater is waken up when hearing a LoRa packet.


IoTThinks:
Sigh, I should sync from dev branch. To redo.

PR #1080, opened 2025-11-08
Support powersaving for repeaters. Heltec v3 repeaters reduce from 47mA down to 7mA

IoTThinks:
Hi all,
I'm adding a feature to let repeaters to deepsleep their MCUs to save power.
LoRa is still in RX mode.
So this feature makes Heltec v3 more practical as a repeater.

I set the powersaving mode to Heltec v3 repeaters by default.
For Heltec v3, via battery supply, I can reduce the power consumption from 47mA down to 7mA.
The Heltec v3 will wake up when hearing a LoRa packet and repeats as normal.
Tested OK for a week with both Meshcore's main and dev branch.

The side effects are:
* Edited: When the repeater wakes up, it will NOT send adverts. It only sends advert in reset or reboot.
* The uptime is always a few seconds most of the time as the repeater is waken up when hearing a LoRa packet.

### DIFF
--- a/examples/simple_repeater/main.cpp
+++ b/examples/simple_repeater/main.cpp
@@ -31,12 +31,23 @@ void setup() {
   board.begin();
 
 #ifdef DISPLAY_CLASS
-  if (display.begin()) {
-    display.startFrame();
-    display.setCursor(0, 0);
-    display.print("Please wait...");
-    display.endFrame();
-  }
+  #ifdef POWERSAVING_MODE
+    if (board.getStartupReason() != BD_STARTUP_RX_PACKET) { // Only display if not waken up from deepsleep
+      if (display.begin()) {
+        display.startFrame();
+        display.setCursor(0, 0);
+        display.print("Please wait...");
+        display.endFrame();
+      }
+    }
+  #else
+    if (display.begin()) {
+      display.startFrame();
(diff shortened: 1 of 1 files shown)


IoTThinks:
Edited a few more changes to optimized OLED, adverts and shorter operation time.
In normal start / reset:
* To wake up 30s to allow sending an advert.
* To power on OLED for a while.

When waken up from deepsleep:
* To sleep again in 5s
* No OLED
* No advert

Tested OK on Heltec v3.

IoTThinks:
Known issues:
* start ota will not work due to deepsleep.

IoTThinks:
To fix OTA this week.
When entering "start ota", to wait 15 minutes for next deepsleep.

aqua:
https://github.com/aqua/meshcore/tree/repeater-sleep has some rough suggested improvements -- you don't need to disable adverts, but can store their timings in the RTC's SRAM, deep sleep until one was due or we're woken by RX, then adjust the timings.  Not super clean and requires some plumbing, but shows the idea.  Also resets the sleep timer on serial RX, which is convenient when one is configuring the radio.

IoTThinks:
I will close this and will move to lightsleep version at 10mA instead.

PR #1100, opened 2025-11-13
Allow SF smaller than 7 to be saved

stphnrdmr:
Saving SF smaller than 7 fails as these checks prevent it, resulting in an error.


Cisien (approved):

liamcottle:
Thanks! That should do the trick :)

PR #1107, opened 2025-11-16
To support lightsleep for Heltec V3/v4 repeater to achieve 9mA power consumption

IoTThinks:
Hi all,
This PR is to allow Heltec v3 and v4 repeaters to save power down from 47mA to 9mA most of the time.
One 2500mAh can help to last 2 weeks for Heltec v3 and v4.
- To use light sleep to achieve 9mA
- No latency due to quicker wakeup
- First boot: To wake up for 30s to send an advert at 18th second.
- Light sleep: To wake up when receiving a LoRa packet and light sleep again after 5s.
- Periodically wakeup: To wake up every 30 minutes to do some periodically tasks and and lightsleep again after 5s.
- Do not sleep when WiFi is not off likely due to OTA via WiFi.
- To reboot to out of OTA and back to powersaving mode.

First advert and periodically adverts (if set) are working.
OTA via WiFi works too.

The code should be applicable to all ESP32-S3 MCUs.
We can move the code upstream. However, at this stage, I put at Heltec v3 and v4 first.
Thanks a lot.

### DIFF
--- a/src/helpers/ESP32Board.h
+++ b/src/helpers/ESP32Board.h
@@ -104,8 +104,6 @@ public:
   #endif
   }
 
-
-
   const char* getManufacturerName() const override {
     return "Generic ESP32";
   }


SaschaKt:
Works like a charm. Should be standard on all esp repeaters or possible to turn on/off via commandline

aqua:
This is a substantial improvement, bringing the v4 down to about 4.9wH/day (extrapolated from a 15 minute bench test consuming 0.040Wh) at near-idle conditions (about a 30% cycle range on a 2600mAh 18650 cell); even light-sleep should bring ESP32-based boards close to compact/cheap solar viability in middling latitudes.  It shouldn't be too hard to adapt for companions as well.

Quibbles, though: it's on by default and breaks the serial CLI on sleep.  That makes configuring a new board fiddly; consider either a config latch to enable powersaving mode, or disabling it when serial input is detected.

IoTThinks:
@aqua I have just made some changes to make the board awake for 2 minutes (instead of 30s) in first boot / restart to support Repeater setup via USB.
This is to do first time repeater setup via USB cable.

I tested and think 2 minutes is enough.

Serial CLI in Meshcore app works for me as it will send LoRa packets which wake up the board and the serial connection.

IoTThinks:
Yes, we have a POWERSAVING_MODE flag in the ini file.
However, I think powersaving mode is almost required for solar ESP32 based repeaters.
Else they will need huge solar panels to survive.

Yes, we can detect if an UART connection is connected to delay the sleep.
Nice idea.

image

SaschaKt:
> This is a substantial improvement, bringing the v4 down to about 4.9wH/day (extrapolated from a 15 minute bench test consuming 0.040Wh) at near-idle conditions (about a 30% cycle range on a 2600mAh 18650 cell); even light-sleep should bring ESP32-based boards close to compact/cheap solar viability in middling latitudes. It shouldn't be too hard to adapt for companions as well.
> 
> Quibbles, though: it's on by default and breaks the serial CLI on sleep. That makes configuring a new board fiddly; consider either a config latch to enable powersaving mode, or disabling it when serial input is detected.

You can flash the official version, configure it and then update with the mod. Maybe there can be implemented a cli command to switch on/off powersaving for all esp repeaters!? Maybe it is not always needed. A powersaving for companion will not work in my opinion because of BT is needed.

IoTThinks:
@SaschaKt You now have 2 minutes instead of 30s to setup a Repeater via Serial USB.

When we refresh the Repeater Setup page, the esp32 will be resetted and we have another 2 minutes to do the settings again.

After that, it will sleep after 5s.

SaschaKt:
> @SaschaKt You now have 2 minutes instead of 30s to setup a Repeater via Serial USB.
> 
> When we refresh the Repeater Setup page, the esp32 will be resetted and we have another 2 minutes to do the settings again.
> 
> After that, it will sleep after 5s.

Hi Kevin. I noticed with the v3 and v4 that the clock go ahead, the longer the running time. A normal reboot via cli doesn't set the clock to the 2024 Standard time so it is not possible to sync the clock (because time can not be synched if it is correct or ahead). Can you modify the CLI "clock Sync" so that it is always syncing?

It is interesting why the clock goes ahead, do you have noticed?

IoTThinks:
Let me monitor.
Usually, I keep power it on / off so it will be back to 2024. And I can always sync as it is 2024.

Let me sync clock and let it run.

By right, the clock of esp32 is not accurate. It will drift...

IoTThinks:
@SaschaKt You check time using clock CLI, right?

image

SaschaKt:
> @SaschaKt You check time using clock CLI, right?
> 
> 

I look at the Request Status page and in cli. The time goes ahead and cannot be synced. Only resetbutton on v3 sets it to 2024, the v4 is not in the near so I can't hit the reset button

IoTThinks:
@SaschaKt My Heltec v3's clock works fine for 2 hours. Let me monitor a day and see.
BTW, the clock in CLI is in UTC not in timezone.

How far is the drift in one day should I expect?

IoTThinks:
Let me explore and test to use a more accurate clock in light sleep.
To reduce the drift to around 20s/day instead of 3 minutes /day.

rtc_clk_slow_src_set(RTC_SLOW_FREQ_8MD256);

mikecarper:
https://analyzer.letsme.sh/node/7E7662676F7F0850A8A355BAAFBFC1EB7B4174C340442D7D7161C9474A2C9400

This is cougar; probably the most busy node in all of MeshCore with roughly 110 neighbors. The granularity is 10s; there are savings to be found even here 
image

cparen:
I just want to add my +1 to this change.

I've got my own private branch BLE companion that does lightsleep anytime the Bluetooth isn't connected, periodically waking up to allow reconnect. This would make it easier to keep my private branch in sync with main.

variants/heltec_v3/HeltecV3Board.h
```diff
@@ -76,6 +77,23 @@ class HeltecV3Board : public ESP32Board {
...
+    if (secs > 0) {
+      esp_sleep_enable_timer_wakeup(secs * 1000000); // To wake up every hour to do periodically jobs
+    }
+
+    esp_light_sleep_start();   // CPU enters light sleep
+  }
+
+  void sleep() override {
+    if (WiFi.getMode() == WIFI_MODE_NULL) { // WiFi is off ~ No active OTA, can go to sleep
+      enterLightSleep(1800); // To wake up every 30 minutes or when receiving a LoRa packet
```
cparen: Can I suggest moving the sleep durations as a parameter, and in milliseconds?

This sort of policy seems like it would be application dependent -- e.g. 30 minutes makes sense for a repeater, but I'm experimenting with a power saving companion where 30 minutes is way too long. Moving it into a parameter to sleep() would let the 30 min duration remain in the repeater code.

bosco688:
Works great on Heltec v4

variants/heltec_v3/HeltecV3Board.h
IoTThinks: The 30-minute is to wake up every 30 minutes to do periodically jobs like adverts.

Other times, it will go to sleep every 5s.

IoTThinks:
@cparen Yes, please create your own PR for companion.

I would love the powersaving for BLE companion.

variants/heltec_v3/HeltecV3Board.h
IoTThinks: Ok let me make the sleep duration as parameters for sleep method.

IoTThinks: Ok let me make sleep period as a parameter of sleep().

IoTThinks:
@mikecarper The dashboard looks awesome.

Looks like an ISP operator even this is P2P network.

variants/heltec_v3/HeltecV3Board.h
IoTThinks: @cparen I have added "secs" for sleep(). 
I think sleep duration in seconds is good enough. Deepsleep uses seconds too.
Yes, it is better to keep the 30 minutes wake up in repeater code.

https://github.com/meshcore-dev/MeshCore/pull/1107/files#diff-10a0092562d43e5bb73a25560e519d6106a0d09b143204ed36ae90d8a3ad24aaR128

image

image

image

IoTThinks:
> > @SaschaKt You now have 2 minutes instead of 30s to setup a Repeater via Serial USB.
> > When we refresh the Repeater Setup page, the esp32 will be resetted and we have another 2 minutes to do the settings again.
> > After that, it will sleep after 5s.
> 
> Hi Kevin. I noticed with the v3 and v4 that the clock go ahead, the longer the running time. A normal reboot via cli doesn't set the clock to the 2024 Standard time so it is not possible to sync the clock (because time can not be synched if it is correct or ahead). Can you modify the CLI "clock Sync" so that it is always syncing?
> 
> It is interesting why the clock goes ahead, do you have noticed?

@SaschaKt 
1. During light sleep, ESP32 will use slower and less accurate clock to save power.
That is why the clock is not as accurate as it is when in full operation.
And Heltec v3/v4 does not have external clock which is more accurate during sleeps.

Actually, why we need the clock anyway? In embedded system like LoRaWAN, we generally dont have clock at dumb devices like sensors and repeaters.

As long as the clock does not move ...backward, then it should be fine.

2. We can change code to sync the clock if the time is in advance.
However, this is usually not recommended due to potential file corruption or packet timestamp mismatch...
So I skip this approach.

3. Yes, it is interesting as MeshCore does not reset the time when we send reboot in MeshCore app.
From the code, yes, it seems it will keep the time during software reboot.
IMO, yes, it should reset everything the same as when we hit RESET button.

image

So I may leave the timing issue as it is now.

IoTThinks:
BTW, you are welcome to test my prebuilt binary files here.
Check for powersaving 05.
https://github.com/IoTThinks/EasySkyMesh/tree/main/firmware

mikecarper:
What are the odds of getting this working with the seeed esp33 node? https://www.seeedstudio.com/Wio-SX1262-with-XIAO-ESP32S3-p-5982.html

aqua:
> What are the odds of getting this working with the seeed esp33 node? https://www.seeedstudio.com/Wio-SX1262-with-XIAO-ESP32S3-p-5982.html

It works almost as-is on the seeed esp32 -- only problem on my test build is that it doesn't wake properly on LoRa packets.  Probably just incorrectly setting the wake conditions, but I haven't tried to troubleshoot it yet.

IoTThinks:
@aqua Let me enable power saving mode and compile for XIAO.

Currently, I set the flag to enable for Heltec v3/v4 only.

Did you make changes and compiled for Xiao?

IoTThinks:
Let me move the code upstream to apply to all ESP32 based board to avoid duplicated code.
To enable, we can set a flag to enable the powersaving feature in platformio.ini of each board.

aqua:
> @aqua Let me enable power saving mode and compile for XIAO.
> 
> Currently, I set the flag to enable for Heltec v3/v4 only.
> 
> Did you make changes and compiled for Xiao?

Yes, I just replicated sleep() and enterLightSleep() in XiaoS3WIOBoard and flashed to a Xiao S3 unit on a power meter.  The arguments to esp_sleep_enable_ext1_wakeup() will probably need to be customized for some boards (e.g. with user buttons) but that might be the only variant override required and the default behavior could go into ESP32Board for most of them.

IoTThinks:
@aqua I have moved up the code for sleep() and enterLightSleep() from Heltec v3/v4 to ESP32Board.h.
And I added   -D POWERSAVING_MODE=1 for Xiao Wio too.

For now, I believe all ESP32 should share the same sleep() and enterLightSleep() to avoid code duplication.
I believe most of the ESP32 based repeaters do not use buttons. PROG button does not have any effect too. 
They only use...RST button. 
We can add it in later phase or in variant if anyone wants.

For your testing, you can compile yourself or use my pre-compiled bin files at 
https://github.com/IoTThinks/EasySkyMesh/tree/main/firmware

It has the CPU temperature at Telemetry > Channel 1 too.
image

cparen:
> @cparen Yes, please create your own PR for companion.
> 
> I would love the powersaving for BLE companion.

yeah, thats just a kludge at the moment, i dont shut down bluetooth correctly or anything. But i do need shorter lightsleeps. Sleeping 10ms at a time still allows for polling other functions like retries, user button presses, etc. 30 minutes would leave dropped retries, and rn i dont think its worth it to try and set up a whole rtos worth of scheduling logic — as long as the sleeps are short, everything else works well enough.

except serial and Bluetooth, need to work on those still 😅

aqua:
Xiao S3 doesn't wake on packet with that build -- not surprising since it's the same wake bitmask just in a different place.  Might be a few days before I have time to fiddle with it, so anyone with differently-distributed free time, feel free.

IoTThinks:
@aqua Why it can wakeup your XIAO?

The pin mask is not correct?

mavericm1:
Been noticing some strange behavior.  I may need to get my sdr to double check but when a message is sent from my companion it isn't immediately repeated and it seems to wait until another message is sent.   

This happens after upgrading the v4 to this pr but a meshcore app update happened at the same time so it could be either causing this behavior

IoTThinks:
@mavericm1 Yes, please help to test with SDR.
Let me check with my SDR during weekend too.

Btw if so then MeshCore app will show the retries, right?

mavericm1:
> @mavericm1 Yes, please help to test with SDR. Let me check with my SDR during weekend too.
> 
> Btw if so then MeshCore app will show the retries, right?

No its a group text message being sent.  the app shows it as Sent and then it just sits there in that state for minutes or more until i send another group message and then all of a sudden it shows that previous message as heard repeat like i would expect

mavericm1:
> @mavericm1 Yes, please help to test with SDR. Let me check with my SDR during weekend too.
> 
> Btw if so then MeshCore app will show the retries, right?

confirmed on the SDR that it isn't repeating how it should 

https://limewire.com/d/qkXQP#RUy3lm9qNb

src/helpers/ESP32Board.h
```diff
@@ -42,6 +43,23 @@ class ESP32Board : public mesh::MainBoard {
   #endif
   }
 
+  void enterLightSleep (uint32_t secs) {
```
zjs81: A better solution would be to use the GPIO wakeup source 

GPIO Wakeup (Light-sleep Only)[](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/sleep_modes.html#gpio-wakeup-light-sleep-only)
In addition to EXT0 and EXT1 wakeup sources described above, one more method of wakeup from external inputs is available in Light-sleep mode. With this wakeup source, each pin can be individually configured to trigger wakeup on high or low level using [gpio_wakeup_enable()](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/gpio.html#_CPPv418gpio_wakeup_enable10gpio_num_t15gpio_int_type_t) function. Unlike EXT0 and EXT1 wakeup sources, which can only be used with RTC IOs, this wakeup source can be used with any IO (RTC or digital).

Should be like this 


```
void enterLightSleep(uint32_t s){
  pinMode(P_LORA_DIO_1,INPUT);
  esp_sleep_disable_wakeup_source(ESP_SLEEP_WAKEUP_ALL);
  gpio_wakeup_enable((gpio_num_t)P_LORA_DIO_1,GPIO_INTR_HIGH_LEVEL);
  esp_sleep_enable_gpio_wakeup();
  if(s>0) esp_sleep_enable_timer_wakeup((uint64_t)s*1000000ULL);
  esp_light_sleep_start();
}

void sleep(uint32_t s) override {
  if(WiFi.getMode()==WIFI_MODE_NULL) enterLightSleep(s);
}
```


Then at the top you gotta add. 

```
#include "esp_sleep.h"
#include "driver/gpio.h"
```

zjs81: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/sleep_modes.html#gpio-wakeup-light-sleep-only

SaschaKt:
> > @mavericm1 Yes, please help to test with SDR. Let me check with my SDR during weekend too.
> > Btw if so then MeshCore app will show the retries, right?
> 
> confirmed on the SDR that it isn't repeating how it should
> 
> https://limewire.com/d/qkXQP#RUy3lm9qNb

This behaviour I have noticed since a longer time and bevor Iotthinks modificated Firmware. Sometimes, mostly after a longer period of inactivity or when I write to another group it happens. Try please with the official latest release, I am sure this behaviour is generally a problem

mavericm1:
> > > @mavericm1 Yes, please help to test with SDR. Let me check with my SDR during weekend too.
> > > Btw if so then MeshCore app will show the retries, right?
> > 
> > 
> > confirmed on the SDR that it isn't repeating how it should
> > https://limewire.com/d/qkXQP#RUy3lm9qNb
> 
> This behaviour I have noticed since a longer time and bevor Iotthinks modificated Firmware. Sometimes, mostly after a longer period of inactivity or when I write to another group it happens. Try please with the official latest release, I am sure this behaviour is generally a problem

I run another repeater with same settings and official build vs the  v4 lightsleep build and it doesn't have this problem.

SaschaKt:
> > > > @mavericm1 Yes, please help to test with SDR. Let me check with my SDR during weekend too.
> > > > Btw if so then MeshCore app will show the retries, right?
> > > 
> > > 
> > > confirmed on the SDR that it isn't repeating how it should
> > > https://limewire.com/d/qkXQP#RUy3lm9qNb
> > 
> > 
> > This behaviour I have noticed since a longer time and bevor Iotthinks modificated Firmware. Sometimes, mostly after a longer period of inactivity or when I write to another group it happens. Try please with the official latest release, I am sure this behaviour is generally a problem
> 
> I run another repeater with same settings and official build vs the v4 lightsleep build and it doesn't have this problem.

Do you have more repeaters in range. With 3 repeaters I have much problems, I've tried all, set txdelay and so on. With one v4 as repeater it goes now very good. Maybe try flashing..my V4 has the light sleep on. The only thing is that the clock goes every day more ahead with v3/v4 and lightsleep hack

mavericm1:
> > > > > @mavericm1 Yes, please help to test with SDR. Let me check with my SDR during weekend too.
> > > > > Btw if so then MeshCore app will show the retries, right?
> > > > 
> > > > 
> > > > confirmed on the SDR that it isn't repeating how it should
> > > > https://limewire.com/d/qkXQP#RUy3lm9qNb
> > > 
> > > 
> > > This behaviour I have noticed since a longer time and bevor Iotthinks modificated Firmware. Sometimes, mostly after a longer period of inactivity or when I write to another group it happens. Try please with the official latest release, I am sure this behaviour is generally a problem
> > 
> > 
> > I run another repeater with same settings and official build vs the v4 lightsleep build and it doesn't have this problem.
> 
> Do you have more repeaters in range. With 3 repeaters I have much problems, I've tried all, set txdelay and so on. With one v4 as repeater it goes now very good. Maybe try flashing..my V4 has the light sleep on. The only thing is that the clock goes every day more ahead with v3/v4 and lightsleep hack

yes i have more than 1 repeater within range.  This isn't an issue in the  main branch and further talking about it in this PR isn't helping anything if you think its a bug in the main branch then report it as an issue.  If you want to fix your repeaters that are all in range then do `set int.thresh 14` on them

kinkywaterhose:
> > > > > > @mavericm1 Yes, please help to test with SDR. Let me check with my SDR during weekend too.
> > > > > > Btw if so then MeshCore app will show the retries, right?
> > > > > 
> > > > > 
> > > > > confirmed on the SDR that it isn't repeating how it should
> > > > > https://limewire.com/d/qkXQP#RUy3lm9qNb
> > > > 
> > > > 
> > > > This behaviour I have noticed since a longer time and bevor Iotthinks modificated Firmware. Sometimes, mostly after a longer period of inactivity or when I write to another group it happens. Try please with the official latest release, I am sure this behaviour is generally a problem
> > > 
> > > 
> > > I run another repeater with same settings and official build vs the v4 lightsleep build and it doesn't have this problem.
> > 
> > 
> > Do you have more repeaters in range. With 3 repeaters I have much problems, I've tried all, set txdelay and so on. With one v4 as repeater it goes now very good. Maybe try flashing..my V4 has the light sleep on. The only thing is that the clock goes every day more ahead with v3/v4 and lightsleep hack
> 
> yes i have more than 1 repeater within range. This isn't an issue in the main branch and further talking about it in this PR isn't helping anything if you think its a bug in the main branch then report it as an issue. If you want to fix your repeaters that are all in range then do `set int.thresh 14` on them

I ran into this issue trying to create power saving features, and worked around it by checking if there is work to be done in Dispatcher. If there is work to be done, don't sleep until the work is cleared up. (it required a bit of work, but corrected this issue since the radio sleeps too quick for the packet to be sent out)

src/helpers/ESP32Board.h
IoTThinks: @zjs81 I see.
There are some boards with bad design to use non RTC pins as wakeup pins.
For now, which boards are they in MeshCore's supported boards?

These boards will have issues if required to deepsleep as the wakeup pins without PULL UP/DOWN will be floating in deepsleep.

IoTThinks:
> I ran into this issue trying to create power saving features, and worked around it by checking if there is work to be done in Dispatcher. If there is work to be done, don't sleep until the work is cleared up. (it required a bit of work, but corrected this issue since the radio sleeps too quick for the packet to be sent out)

@kinkywaterhose Yes, this is a potential issue. Need to check if the radio is done with the tramission before sleep.
@mavericm1 Is your issue related to this matter, too?

mavericm1:
> > I ran into this issue trying to create power saving features, and worked around it by checking if there is work to be done in Dispatcher. If there is work to be done, don't sleep until the work is cleared up. (it required a bit of work, but corrected this issue since the radio sleeps too quick for the packet to be sent out)
> 
> @kinkywaterhose Yes, this is a potential issue. Need to check if the radio is done with the tramission before sleep. @mavericm1 Is your issue related to this matter, too?

I don't believe it sleeps in the middle of tx'ing but it does receive a packet and then it seems to sit on it until some unknown time like it is asleep or something.  usually if you send another message it will then wake up and tx that message and usually the one following.

IoTThinks:
Interesting, let me check over SDR.

We have tested to send from a companion to another companion via 2 hops with no issue.

Usually, this should be very straight forward with light sleep for esp32.

mavericm1:
> Interesting, let me check over SDR.
> 
> We have tested to send from a companion to another companion via 2 hops with no issue.
> 
> Usually, this should be very straight forward with light sleep for esp32.

Try using set int.thresh 14 on the repeater running this maybe that is the trigger that is causing it to act that way in combination to light sleep

src/helpers/ESP32Board.h
IoTThinks: @zjs81 Let me use and test your GPIO wakeup to maximize the compability for any ESP32-based board.
Thanks a lot. 👯

IoTThinks: @zjs81 I tested.
The wakeup via GPIO will not work for DIO1 of sx1262.

Dio1 of sx1262 will have short rising HIGH.
GPIO wake up may trigger many times and crash esp32 when trying to wakeup.
Let me see if any work around.

In my experience, I always use ext0 and ext1 for dio1 of sx1262.

IoTThinks:
> Xiao S3 doesn't wake on packet with that build -- not surprising since it's the same wake bitmask just in a different place. Might be a few days before I have time to fiddle with it, so anyone with differently-distributed free time, feel free.

@aqua Xiao S3 Wio uses GPIO39 as DIO1 and this is not a RTC pin.
So the wakeup via ext1 (best practice) does not work.

Actually, this is not a good hardware design. If going to **deep**sleep, this board can not wake up when a LoRa paket comes.
For light sleep, let me see how to workaround for Xiao S3.

IoTThinks:
To ensure the maximum compability.
- I removed the   -D POWERSAVING_MODE=1 for XIAO WIO S3 since its DIO1 (GPIO39) is not a RTC GPIO. This is a safety feature. We need to manually add this in the platform.ini to enable this feature. So default is for Heltec v3 and Heltec v4. Later we may add to all ESP32-S3 with RTC GPIO for DIO1.
- I added if defined to support ESP32-S3 only. Some ESP32-C3, C6 may not have EXT0 and EXT1. Sleep will skip for these boards.
- I added validation for RTC GPIO for DIO1. If the DIO1 is not RTC, sleep will skip for these boards.

I have compiled for ALL targets and all target can compile.
- Heltec_ct62_repeater_bridge_espnow has unrelated compilation issue. Not in scope. Skip.
- Heltec_Wireless_Tracker_repeater_bridge_espnow and Heltec_Wireless_Tracker_room_server refer to outdated  #include <helpers/HeltecV3Board.h>. Not in scope. Skip.

IoTThinks:
@mavericm1 I checked both SDR and power meter and found no issue.
Our environment is quite silent with around 5 repeaters around.

When I send message via direct chat or public channel, the repeaters repeat instantly.
All repeaters have the same default settings but with lightsleep.

1. First signal from T1000-e, and two Heltec v3 repeaters (with same powersaving 06) repeat instantly.
image

2. This is for usb power meter to one Heltec v3 repeater with light sleep.
The peak is from a TX, eg: The Heltec v3 with light sleep repeats the signal instantly.
The currently in light sleep is around 9mA and in TX around 100+ mA.
image

3. I have checked the code related to "set int.thresh "
I think int.thresh is inteference threshold. 
It means if the repeater detects the channel is busy, it will wait for the NEXT slot to send.

So it may due to your environment has high LoRa traffic, your repeater needs to wait a while before repeating the packet.
It may not relate to light sleep at all.
image

A potential problem is your repeater waits too long (more than 5 seconds) for the next available timeslot, then the repeater needs to go to sleep.
Then a potential fix is to wait until the repeating task completes before going to sleep again.
Then when wakeup the next time, the repeater will send the pending packet. 
At least the packet will not be lost.
=> I will find a way to improve this potential problem.

mavericm1:
in this situation you will not repro.  

Set a low power flashed v3 or v4 as a repeater on a non in use frequency and set int.thresh 14  and set a companion to the same frequency and use an SDR.   Send exactly 1 message and wait typically it will not repeat for quite awhile or until hearing another lora packet.  When you have other repeaters in the area and it hears them repeat the packet it causes it to send out the repeat because hearing the other repeats cause it not to sleep.

To show you the example before i turned off the repeaters near me from repeating using set repeat off

mavericm1:
The message/packet is not lost that is correct but its pretty frustrating because to force the repeater to send what you've already sent you must send yet another message.

IoTThinks:
@mavericm1 I dont get what you mean.

You mean the repeater repeats twice or the repeater needs to wait the next slot to repeat?

mikecarper:
Message gets stuck in the queue and doesn't go out until another event wakes it up. If I'm reading it correctly

IoTThinks:
> Message gets stuck in the queue and doesn't go out until another event wakes it up. If I'm reading it correctly

@mikecarper I guess so too. So let me fix the potential issue: "Delay the sleep until no more dispatcher task".
I suspect due to high traffic and high inteference, the repeater delays the repeat and can not finish repeating in 5s.

mavericm1:
> > Message gets stuck in the queue and doesn't go out until another event wakes it up. If I'm reading it correctly
> 
> @mikecarper I guess so too. So let me fix the potential issue: "Delay the sleep until no more dispatcher task". I suspect due to high traffic and high inteference, the repeater delays the repeat and can not finish repeating in 5s.

Yes what mike said is what i was describing here.  https://github.com/meshcore-dev/MeshCore/pull/1107#issuecomment-3597246193

IoTThinks:
@mavericm1 @mikecarper Your testing and feedback are well appreciated.

How about this? 
I will check if there is anything to send. 
If no, sleep. 
Else, it will awake for 5 more seconds.

I will test and see how it reacts first. Then let you test in the busy area.

image

Some friends here suggest me to implement this feature dynamically instead of hardcoded parameter.
So in CLI of repeater, we can do "set powersaving on" to turn on this feature. 
The default is well, off.

What do you think?

mavericm1:
> @mavericm1 @mikecarper Your testing and feedback are well appreciated.
> 
> How about this? I will check if there is anything to send. If no, sleep. Else, it will awake for 5 more seconds.
> 
> I will test and see how it reacts first. Then let you test in the busy area.
> image
> 
> Some friends here suggest me to implement this feature dynamically instead of hardcoded parameter. So in CLI of repeater, we can do "set powersaving on" to turn on this feature. The default is well, off.
> 
> What do you think?

Yes i think that is what makes the most sense that the dispatcher doesn't allow sleep until queue is empty

IoTThinks:
@mavericm1 I have tested. Looks OK. The getOutboundCount() is retrieved from Dispatcher.
Note the delay(100) is for debug log during testing only. This is to print the message properly before sleep.

In the photo, we can see when getOutboundCount=1, the sleep will be postponed from second 44 to second 49.
image

I put the pre-compiled bin file here for your testing.
They have the printout in the console in cause you want to check.
https://github.com/IoTThinks/EasySkyMesh/tree/main/firmware/Testing

Please note the logic:
- First boot: To wake up for 2 minutes to send an advert at 18th second and allow Repeater Setup via UART cable.
- Light sleep: To wake up when receiving a LoRa packet and light sleep again after 5s. To postpone 5 more seconds if there is pending outbound send.
- Periodically wakeup: To wake up every 30 minutes to do some periodically tasks and and try to lightsleep again after 5s.

Please help to test and feedback.
Thanks a lot.

IoTThinks:
I added some more codes to postpone sleep to further 5s if there is pending outbound.

Please note, the power consumption via USB is always higher (around 17mA) as the board needs to power uart chip, led...
9mA will be consumed if you power and measure from battery cable.
Thanks a lot.

dt267:
https://github.com/fdlamotte/MeshCore/blob/lightsleep_repeater/examples/simple_repeater/main.cpp

IoTThinks:
So good.
I'm in a correct track.

mavericm1:
does the clock reboot firmware also contain the sleep code  and dispatcher ?

mavericm1:
I flashed the low power updated bin you provided and it appears to have fixed the problem from what I see

IoTThinks:
> does the clock reboot firmware also contain the sleep code and dispatcher ?

Yes, the clock reboot firmware contains the sleep code and dispatcher.
This is for a friend want to reset the clock when reboot by Meshcore app and sync the time again.

IoTThinks:
> I flashed the low power updated bin you provided and it appears to have fixed the problem from what I see

Wow, that's great. Thanks a lot for your feedback.

hpux735:
I wasn't able to reproduce any savings with a v4.  Here's a screen shot of power analysis before:
image

And after applying this MR to a local branch.
ppk-20251210T222711

The quiescent current is 59mA in both cases.  Do you have any suggestions?  Or, perhaps, a bin file for the v4 that I could test to ensure that it's being compiled as you expect?

SaschaKt:
> I wasn't able to reproduce any savings with a v4. Here's a screen shot of power analysis before: image
> 
> And after applying this MR to a local branch. ppk-20251210T222711
> 
> The quiescent current is 59mA in both cases. Do you have any suggestions? Or, perhaps, a bin file for the v4 that I could test to ensure that it's being compiled as you expect?

I can tell you that it works perfect. This works only for repeaters, do you have flashed a repeater or do you try it for an companion? The bin file you can download under https://github.com/IoTThinks/MeshCore/releases/download/PowerSaving07/repeater-heltecv4-powersaving07.bin

hpux735:
I tried with the bin you linked to and got exactly the same results.

SaschaKt:
> I tried with the bin you linked to and got exactly the same results.

Have you tried another v4? Is the v4 maybe charging the battery if you measure over the Usb-C port (red led must be off)? It takes 2 minutes for the first sleep after switching on or rebooting

hpux735:
Ah, ok.  that's the ticket.  I didn't know you have to wait 2 minutes.
ppk-20251211T000014
ppk-20251210T235953

IoTThinks:
@hpux735 Yes, for first boot or after hard reset, have to wait 2 minutes to attempt to light sleep.
This is to allow Repeater Setup via UART (if any).

IoTThinks:
This is interesting.

We can convert all ESP32 boards from another mesh to become MeshCore repeaters at 9mA. Much lower than any NRF52 boards with BLE at 14+ mAh.

There is an unfair advantage of Meshcore repeater: It does not use BLE at all. Therefore, sleep is much easier and effective to reduce power.

Btw, you need to unplug the GPS module. This module uses 40mA!

examples/simple_repeater/MyMesh.cpp
```diff
@@ -1101,3 +1101,8 @@ void MyMesh::loop() {
   uptime_millis += now - last_millis;
   last_millis = now;
 }
+
+// To get the current pending outbound packets at Dispatcher
```
ripplebiz, line 1119: Is prob better to have this method sig:
bool hasPendingWork()

examples/simple_repeater/MyMesh.cpp
```diff
@@ -1101,3 +1101,8 @@ void MyMesh::loop() {
...
+int MyMesh::getOutboundCount (uint32_t now) const {
+  return _mgr->getOutboundCount(0xFFFFFFFF);
```
ripplebiz, line 1121: wondering whether this should also add 1 if Dispatcher::outbound != NULL ?

examples/simple_repeater/main.cpp
```diff
@@ -19,6 +19,11 @@ void halt() {
 
 static char command[160];
 
+#ifdef POWERSAVING_MODE
+  unsigned long lastActive = millis();   // mark last active time
```
ripplebiz: not sure, but prob safer to initialise lastActive in setup()... this assume the millis() function is ready and already set up. (might be, not sure)

examples/simple_repeater/main.cpp
```diff
@@ -117,4 +122,16 @@ void loop() {
   ui_task.loop();
 #endif
   rtc_clock.tick();
+
+#ifdef POWERSAVING_MODE
+  if (millis() - lastActive > nextSleepinSecs * 1000) {
```
ripplebiz: just wondering about the 49-day wrap around problem here. If you use the millisHasNowPassed() method, it takes this into account. (uses a 2's complement arithmetic trick)

examples/simple_repeater/MyMesh.cpp
IoTThinks: Ok, let me rename it to hasPendingWork().

examples/simple_repeater/MyMesh.cpp
IoTThinks: Why we need to add 1?
To include the current new packet in the count?

examples/simple_repeater/main.cpp
IoTThinks: Ok, let me take a look at millisHasNowPassed()
Thanks a lot.

IoTThinks: I will revisit this PR this weekend.
Sorry for the delay.

ViezeVingertjes:
> We can convert all ESP32 boards from another mesh to become MeshCore repeaters at 9mA. Much lower than any NRF52 boards with BLE at 14+ mAh.

Woah, 17 µA is not 14+ mAh haha... that's a whole different ballpark. That 17 µA is from the specs, as my meters wont even measure that low i dont know the exact number, but way below 1mA when it's on (in sleep with wake upon rx) until it gets an packet and repeats it.

mikecarper:
@IoTThinks would this also work for a g2?

IoTThinks:
@mikecarper This PR won't work for Station G2 as the board uses non-RTC GPIO 48 for DIO1.
Non-RTC GPIOs can not handle Rising High for DIO1 reliably. 
Let me revisit to support these boards.

image

RTC pins is up to GPIO21 only.
https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/peripherals/gpio.html#gpio-summary
image

MikesAllotment:
Can the same change be included for the Heltec V2 as well please?  Any changes for V3 / V4 should also be equally applicable for the V2 since that is also esp32-based.

examples/simple_repeater/MyMesh.cpp
IoTThinks: I have renamed it to hasPendingWork()

examples/simple_repeater/MyMesh.cpp
IoTThinks: If we need to add 1 to _mgr->getOutboundCount by any reason, _mgr->getOutboundCount should be fixed instead.

Since it is working, I would skip this for now.

examples/simple_repeater/main.cpp
IoTThinks: I put this before setup()
```
#ifdef POWERSAVING_MODE
  unsigned long lastActive = 0;          // mark last active time
  unsigned long nextSleepinSecs = 120;   // next sleep in seconds. First time is 2m to send advert and do repeater setup.
#endif
```

And put this in setup(), it should be OK.
```
#ifdef POWERSAVING_MODE
  lastActive = millis();   // mark last active time at boot
#endif
```

examples/simple_repeater/main.cpp
IoTThinks: I have changed this
`  if (millis() - lastActive > nextSleepinSecs * 1000) {`

To this
`  if (the_mesh.millisHasNowPassed(lastActive + nextSleepinSecs * 1000)) {`

It looks OK.
Let me test for a while today before committing new changes.
Thanks a lot.

IoTThinks:
Added changed as required by reviewers.
- Renamed getOutboundCount in MyMesh to hasPendingWork()
- Put lastActive to initialise in setup()
- Checked next sleep with millisHasNowPassed

https://github.com/meshcore-dev/MeshCore/pull/1107/commits/dd18ae7a3ed5d362055e91c46cabf42f6f1439d8

IoTThinks:
@MikesAllotment Actually, this PR is applicable for many ESP32-based boards as long as the DIO1 of these boards wire to RTC pin.
To limit the impact of the PR, I apply first to Heltec v3 and v4. Then I will have another PR to roll out to all ESP32-based board.

BTW, I have checked Heltec v2. 
The MCU is ESP32-D0WDQ6 (similar to a WROOM32). It uses GPIO26 for DIO1 which is a valid RTC pin.
image

So yes, this PR will work for Heltec v2.
image

Please help to unzip the file to bin, then flash this bin to see if it works for your Heltec v2.
To wait after 2 minutes to see if the power consumption is reduced to around 10mA from usb power and 9mA from battery power.
[repeater-heltecv2-powersaving.zip](https://github.com/user-attachments/files/24284613/repeater-heltecv2-powersaving.zip)

examples/simple_repeater/MyMesh.cpp
```diff
@@ -1115,3 +1115,8 @@ void MyMesh::loop() {
   uptime_millis += now - last_millis;
   last_millis = now;
 }
+
+// To get the current pending outbound packets at Dispatcher
+int MyMesh::hasPendingWork(uint32_t now) const {
```
fschrempf, line 1120: Remove the `now` parameter. It's not used.

src/helpers/ESP32Board.h
```diff
@@ -83,6 +104,8 @@ class ESP32Board : public mesh::MainBoard {
   #endif
   }
 
+
+
```
fschrempf:
```suggestion
```

Remove empty lines

variants/heltec_v3/platformio.ini
```diff
@@ -48,6 +48,7 @@ build_flags =
   -D MAX_NEIGHBOURS=50
 ;  -D MESH_PACKET_LOGGING=1
 ;  -D MESH_DEBUG=1
+  -D POWERSAVING_MODE=1
```
fschrempf, line 51: I don't really like the fact that we are enabling this for specific boards only. I know this can be rolled out to other boards later, but it will likely be confusing for some users if some boards show a different behavior than others (for example CLI accessibility after the initial 2 minute period).

examples/simple_repeater/MyMesh.cpp
IoTThinks: 1. This PR 1107 is much older than PR 1238.
In my PowerSaving fork, NRF52 is enjoying 8.5mA power consumption for a while.
All these stuff will be pushed after this PR is approved.

2. Last time, I'm thinking to use CLI to enable/disable powersaving mode.
However, I believe powersaving mode should be the default and built-in feature.
For IoT world, power saving is actually a must.

IoTThinks: @fschrempf We may just change the comment to become " To get the current pending works"
Or yes, just make it `int MyMesh::hasPendingWork() const {`

```
// To get the current pending outbound packets at Dispatcher
int MyMesh::hasPendingWork(uint32_t now) const {
```

fschrempf: You either have to remove the `now` parameter, or pass it on to `getOutboundCount()`. Keeping it without using it doesn't make sense.

variants/heltec_v3/platformio.ini
IoTThinks: Well, we will have power saving to all boards soon by default. All will have the same powersaving behaviors.
However due to large user base of MeshCore, we would like to see the impact of the change to Heltec v3/v4 first.

IoTThinks:
To cleanup and create another PR for clean state.

fschrempf:
> To cleanup and create another PR for clean state

Please, as I mentioned before: Just use `git rebase --interactive` and `git push --force`. You don't need a new PR for cleaning up.

IoTThinks:
> > To cleanup and create another PR for clean state
> 
> Please, as I mentioned before: Just use `git rebase --interactive` and `git push --force`. You don't need a new PR for cleaning up.

It is my preference.
I intend to do another approach with CLI.
So to close and to do another version to apply to all ESP32 boards.

fschrempf:
@IoTThinks Ok, feel free to cherry-pick the commits you need from my branch: https://github.com/fschrempf/MeshCore/commits/repeater-power-opt/

fschrempf:
> This PR 1107 is much older than PR 1238.
In my PowerSaving fork, NRF52 is enjoying 8.5mA power consumption for a while.
All these stuff will be pushed after this PR is approved.

Ok. The solution with `waitForEvent()` makes sense in general. But I don't see how to wake up after a certain timeout (maybe setting up an RTC interrupt) and how to avoid arbitrary wakeups by interrupts from other peripherals like sensors, etc.

Unfortunately when testing you `Powersave07` version, the lowest I can get is 9.8 mA on RAK4631 compared to 7.0 mA with https://github.com/meshcore-dev/MeshCore/pull/1238. Could it be that some interrupt is repeatedly waking up the device?

> Last time, I'm thinking to use CLI to enable/disable powersaving mode.
However, I believe powersaving mode should be the default and built-in feature.
For IoT world, power saving is actually a must.

Yes, I agree it should probably the default at some point. Still, I think it makes sense to have it opt-in at first and then make it the default later. And having it runtime-configurable makes sense e. g. for people who don't care about power savings but need serial CLI to be active all the time. But I guess it's best to let the maintainers decide what they prefer.

SaschaKt:
> > This PR 1107 is much older than PR 1238.
> > In my PowerSaving fork, NRF52 is enjoying 8.5mA power consumption for a while.
> > All these stuff will be pushed after this PR is approved.
> 
> Ok. The solution with `waitForEvent()` makes sense in general. But I don't see how to wake up after a certain timeout (maybe setting up an RTC interrupt) and how to avoid arbitrary wakeups by interrupts from other peripherals like sensors, etc.
> 
> Unfortunately when testing you `Powersave07` version, the lowest I can get is 9.8 mA on RAK4631 compared to 7.0 mA with #1238. Could it be that some interrupt is repeatedly waking up the device?
> 
> > Last time, I'm thinking to use CLI to enable/disable powersaving mode.
> > However, I believe powersaving mode should be the default and built-in feature.
> > For IoT world, power saving is actually a must.
> 
> Yes, I agree it should probably the default at some point. Still, I think it makes sense to have it opt-in at first and then make it the default later. And having it runtime-configurable makes sense e. g. for people who don't care about power savings but need serial CLI to be active all the time. But I guess it's best to let the maintainers decide what they prefer.

What do you mean with CLI? Serial CLI? Because the CLI over Lora works, and no power saving after reboot for 2 minutes for configuring via serial cable. Also we can turn ota mode via Lora admin on so the lightsleep is off and then should cli via serial be possible.CLI via serial is for the mostly users only needed when configuring after new flashing. When serial is on (what I think is not possible while lightsleep) then the powerconsumption will be higher

IoTThinks:
@SaschaKt @fschrempf A new PR is coming.
The PR is already 1-month old.

So let's refresh it with comments from reviewers and added new CLI to on/off powersaving mode instead of hardcoded built parameter.
And at clean state.

Merry Chirstmas everyone. 👯 
image

IoTThinks:
We can continue here.
Thanks a lot.
https://github.com/meshcore-dev/MeshCore/pull/1266

PR #1108, opened 2025-11-16
Add variant for RAK_killer PCB

Uroboros67:
Hello
Please add the "RAK_killer" PCB variant based on the nRF52840 and HT-RA62 (Ra-01SH-P in the PRO version).
https://github.com/Uroboros67/rak_killer
best regards
Jacek SQ9ETV


recrof:
Thanks for you entry, but we are unable to support niche boards, like this one and include them in our build process.

PR #1110, opened 2025-11-17
Adding save of the buzzer config to persist on reboot

zaquaz:


zaquaz:
This was a typing mistake, my bad.

PR #1116, opened 2025-11-18
Include battery voltage as feature 1 in adverts

chrisdavis2110:
Using the board's battery voltage, include it as feature 1 in AdvertDataBuilder. Instead of trying to log in and then trying to request repeater stats, which may fail, would be nice to have the battery voltage included in adverts.


liamcottle:
Hello, thanks for the contribution. Unfortunately advert packets are pretty large, and this isn't something we're looking to add to them at this time.

PR #1118, opened 2025-11-19
add heltec_v4 tft  expansion box

Quency-D:
The purchase link is below.
https://heltec.org/project/wifi-lora-32-v4-expansion-housing/


ripplebiz:
I had to reverse this commit, as it was causing Heltec V4 repeater to hang on startup

Quency-D:
@ripplebiz Hi, I tried the Heltec V4 repeater, and it started slowly the first time I downloaded the program, but it started quickly after subsequent power cycles or resets.

I suspect it might be because I added an I2C pin. Do you have any external devices connected?
```c
  -D ENV_PIN_SDA=4
  -D ENV_PIN_SCL=3
```
Or could you help us provide some other details?

PR #1123, opened 2025-11-20
heltec_t114: enable AUTO_SHUTDOWN_MILLIVOLTS for BLE companions

### LINKED ISSUE #1112: AUTO_SHUTDOWN_MILLIVOLTS for Heltec T114?
Hi,

it seems Heltec modules are compiled with AUTO_SHUTDOWN_MILLIVOLTS for companion, but Heltec T114 isn't. As you can connect LiPo to it like to other Heltecs, I think it should be treated the same way. IMHO it could save me (the T114 BLE companion user) some bad surprises :).

yarda:
It should fix #1112, but untested yet (I didn't let my battery to discharge so deep yet :) ).


ripplebiz:
Please change the base branch to 'dev'

Xinayder:
3.4V for auto shutdown seems too high. Flat LiPo batteries are safe up until 2.7V, which is the voltage most battery management system chips will cut off.

yarda:
All board variants in the meshcore repo use 3.3 or 3.4 V for auto shutdown. It maybe good to unify this. I selected the 3.4 V, but I haven't problem with a bit lower voltage. But the value should be safe for most of the batteries. E.g. my battery don't have any BM chip so there is nothing that would cut off the battery.  And nearly any battery can be connected to the T114 battery connector. IMHO this is the purpose of the auto shutdown feature, to save the battery from the quick deep discharge. I think this isn't the contest how deep the battery can be discharged and IMHO most of the companion users would prefer durability of the battery (number of cycles) not few more seconds uptime the deep discharge (bellow 3 V) could give them.

liamcottle:
> 3.4V for auto shutdown seems too high. Flat LiPo batteries are safe up until 2.7V, which is the voltage most battery management system chips will cut off.

I believe this voltage was chosen, as writing to flash under 3.3v can be unstable and potentially lead to file system corruption, which then requires the user to do a factory reset to resolve.

Safer to turn off early, than too late.

yarda:
I am fine with both 3.3 V or 3.4 V. I don't think that deeper discharge would give companion users any benefit. Just tell me which voltage to use or what to change in this PR.

yarda:
I finally tested it today and my T114 auto hibernated as expected when the battery voltage reached 3.4 V.

yarda:
Polite ping, is there anything still blocking this?

oltaco:
I'm not totally sure but I think the original decision was to only enable this on devices with epaper screens so the user would know not try to turn on the device while flat leading to more potential for filesystem damage?

There's a PR brewing to enable boot lockout below certain voltage thresholds which could be the change needed to make auto shutdown more acceptable on all devices? I'll admit that's just pure speculation on my part though.

weebl2000:
I think dev has a battery voltage cutoff now. Probably can close this PR? @ripplebiz

liamcottle:
Closing out due to no follow up from author.

PR #1129, opened 2025-11-21
promicro_washtastic variant proposal

dmduran12:
Although the Washtastic is a DIY board, it's a popular one with increasing deployments. A feature of the Washtastic, to increase its dynamic range, is a non-1:1 voltage bridge. This creates a challenging and frustrating scenario for repeater managers looking for accurate voltage readings, especially in weather conditions pushing the limits of the onboard mppt.

This PR is a very simple proposed addition, creating a promicro variant "promicro_washtastic" — the only change is the ADC_MULTIPLIER to more accurately report voltage for Wasthastic users looking for a firmware that "just works".

We recognize and empathize with efforts to decrease "bloat", but believe that Washtastic proliferation has reached a point where a bespoke firmware may be warranted.

Validated on a Wastastic v4. At charge cutoff, the MeshCore iOS app reported 4.06v while a direct multimeter probe of the 18650 taken at the same time reported 4.059v

Thank you for considering this minor but potentially very helpful addition to the supported devices.

### DIFF
--- a/variants/promicro_washtastic/PromicroBoard.cpp
+++ b/variants/promicro_washtastic/PromicroBoard.cpp
@@ -1,0 +1,83 @@
+#include <Arduino.h>
+#include "PromicroBoard.h"
+
+#include <bluefruit.h>
+#include <Wire.h>
+
+static BLEDfu bledfu;
+
+void PromicroBoard::begin() {    
+    // for future use, sub-classes SHOULD call this from their begin()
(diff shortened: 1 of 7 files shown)


oltaco:
Alternatively [PR #1133](https://github.com/meshcore-dev/MeshCore/pull/1133) would allow setting adc.multiplier via CLI instead.

ripplebiz:
PR #1133 will take care of this

PR #1130, opened 2025-11-21
Added buzzer config persistance across restart

zaquaz:
This set of commits adds persistence of buzzer selected state over restart.

- DataStore.cpp - added reading and writing of the buzzer_quiet state
- NodePrefs.h - added the buzzer_quiet field to the preferences
- MyMesh.h - made SavePrefs() into public to allow for saving the config on change, cleaned up an extra private:
- UITask.cpp - on switch of buzzer, saves the preferences file

Tested in T1000E

### DIFF
--- a/examples/companion_radio/DataStore.cpp
+++ b/examples/companion_radio/DataStore.cpp
@@ -200,11 +200,6 @@ void DataStore::loadPrefsInt(const char *filename, NodePrefs& _prefs, double& no
   File file = openRead(_fs, filename);
   if (file) {
     uint8_t pad[8];
-    
-    // Initialize defaults for any missing fields (backward compatibility)
-    memset(&_prefs, 0, sizeof(_prefs));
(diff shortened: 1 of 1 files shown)


examples/companion_radio/DataStore.cpp
```diff
@@ -200,6 +200,11 @@ void DataStore::loadPrefsInt(const char *filename, NodePrefs& _prefs, double& no
   File file = openRead(_fs, filename);
   if (file) {
     uint8_t pad[8];
+    
+    // Initialize defaults for any missing fields (backward compatibility)
+    memset(&_prefs, 0, sizeof(_prefs));
```
ripplebiz: Setting of defaults is already handled in MyMesh::MyMesh().  This blows away those defaults

zaquaz: removed this portion, this should be resolved now.

PR #1132, opened 2025-11-21
Removed halt and replaced it with a delay and reboot

wel97459:
Replaced the halt function with a delay and reboot, instead of an infinite loop. This removes an edge case were there's a bad brown out and the radio fails to initialize properly.

As @liamcottle suggested I have added an error message that is customizable to other failure cases. Available on node with a display enabled.

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -101,8 +101,18 @@ MyMesh the_mesh(radio_driver, fast_rng, rtc_clock, tables, store
 
 /* END GLOBAL OBJECTS */
 
-void halt() {
-  delay(10000);
+void delayedReboot(const char *msg, const uint32_t delayms) {
+  char tmp[32];
+  sprintf(tmp, "Rebooting in %3.1fs", delayms/1000.0);
+  #ifdef DISPLAY_CLASS
+    display.startFrame();
+    display.drawTextCentered(display.width() / 2, 24, msg);
+    display.drawTextCentered(display.width() / 2, 32, tmp);
+    display.endFrame();
+  #endif
+  MESH_DEBUG_PRINTLN(msg);
+  MESH_DEBUG_PRINTLN(tmp);
+  delay(delayms);
   board.reboot();
 }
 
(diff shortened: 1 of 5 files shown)


LitBomb:
halt means stop.  if you delay 10 seconds and then reboot, then it is a reboot and not a halt any more, is it?

fschrempf:
> This could keep some boards from booting when there's dirty power, and the radio chip fails to init correctly.

And I guess that's exactly what is intended. If there is a critical failure, there is no reason to continue. Rebooting could cause nasty boot loops that won't help either.

liamcottle:
For radio init failure, if the device has a screen, I'd probably prefer to show a message that says radio init failed.

wel97459:
The halt function is only called in one place in the whole code base, just for that radio init in the being of boot. It's definitely a contributing factor for solar nodes not booting up properly.

And I have already seen improvements in getting my repeaters to start from a BMS cut off.

I would also hazard to say that a rover on Mars would not have a function like halt or a call to a function that doesn't try to get the rover back to an operational status. Most repeaters are in places where reset buttons are a pain to get too.

And I thinking adding a message for nodes with a screens is a good idea with a count down to reset or something.

wel97459:
I'll change the function name. A halt function like this one is bad way to handle a radio init problem.
 > halt means stop.  if you delay 10 seconds and then reboot, then it is a reboot and not a halt any more, is it?

wel97459:
I personally would rather have a boot loop then a non-responsive repeater, with a chance of a fully functional boot.

> > This could keep some boards from booting when there's dirty power, and the radio chip fails to init correctly.
> 
> And I guess that's exactly what is intended. If there is a critical failure, there is no reason to continue. Rebooting could cause nasty boot loops that won't help either.

wel97459:
I removed the halt function and added delayedReboot. The function takes in a message string and delay amount, and displays the message like in the image.

Without DISPLAY_CLASS defined it just delays and reboots.

wel97459:
I cleaned up the code and smashed the commits.

PR #1133, opened 2025-11-21
Feature: configurable adc.multiplier for repeaters

oltaco:
This PR adds a new persistent preference for adc.multiplier to allow setting of the ADC multiplier value.

It came up that washtastic (a promicro based board) uses a different ADC multiplier to what is set on the standard promicro and to me it makes more sense to allow setting the adc.multiplier rather than adding another variant just to set a different ADC multiplier.

I added virtual methods for getAdcMultiplier() and setAdcMultiplier() to the MainBoard class, and examples are implemented in PromicroBoard.h. When adc.multiplier is set to 0 it uses the board definition, otherwise it uses whatever you have set in adc.multiplier. The value is constrained to between 0.0f and 10.0f.

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -335,7 +335,12 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
         sprintf(reply, "> %s", _prefs->bridge_secret);
 #endif
       } else if (memcmp(config, "adc.multiplier", 14) == 0) {
-        sprintf(reply, "> %s", StrHelper::ftoa(_prefs->adc_multiplier));
+        float adc_mult = _board->getAdcMultiplier();
+        if (adc_mult == 0.0f) {
+          strcpy(reply, "Error: unsupported by this board");
+        } else {
+          sprintf(reply, "> %.3f", adc_mult);
+        }
       } else {
         sprintf(reply, "??: %s", config);
       }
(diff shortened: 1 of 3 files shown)


variants/promicro/PromicroBoard.h
```diff
@@ -39,7 +40,18 @@ class PromicroBoard : public mesh::MainBoard {
...
+  }
+
+  void setAdcMultiplier(float multiplier) override {
+    if (multiplier == 0.0f) {
+      adc_mult = ADC_MULTIPLIER;}
+    else {
+      adc_mult = multiplier;
+    }
+  }
+  float getAdcMultiplier() const override {
```
ripplebiz, line 54: I don't think this is referenced anywhere?

examples/simple_repeater/MyMesh.cpp
```diff
@@ -733,6 +735,8 @@ void MyMesh::begin(FILESYSTEM *fs) {
   updateAdvertTimer();
   updateFloodAdvertTimer();
 
+  board.setAdcMultiplier(_prefs.adc_multiplier);
+
```
ripplebiz, line 739: Can you copy this same init to Room and Sensor firmwares too... as they share the _prefs and CommonCLI

variants/promicro/PromicroBoard.h
oltaco: You're right, I was thinking about calling it from CommonCLI and return the default board multiplier if the saved multiplier is 0.0f, what do you think?

src/helpers/CommonCLI.cpp
```diff
@@ -331,6 +334,8 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
       } else if (memcmp(config, "bridge.secret", 13) == 0) {
         sprintf(reply, "> %s", _prefs->bridge_secret);
 #endif
+      } else if (memcmp(config, "adc.multiplier", 14) == 0) {
+        sprintf(reply, "> %s", StrHelper::ftoa(_prefs->adc_multiplier));
```
ripplebiz: Yeah, could maybe call board.getAdcMultiplier() here

oltaco: Ok I tweaked the logic in CommonCLI and changed setAdcMultiplier to return a bool. Now if the board doesn't support these methods the CLI will let you know, so there is no ambiguity.

Setting adc.multiplier to 0 lets you know that it will use the default board multiplier as well.

Let me know if you think it's too over the top.

edited to add: If the multiplier is set to 0 and you do get adc.multiplier it will return the default setting for the board.

Issue #1014, opened 2025-10-25
Current to high generally in hibernating(turned off state)


SaschaKt:
The current is too high in Sleeping mode on a Companion, with an T114 it is 7,5mA, with an TEcho it is 8,4mA. When you calculate with 850mAh Battery on TEcho fully charged it is empty after 3 days without using. You can not store it for an blackout or when you don't use it. The T1000 is also empty after a few days like the Heltec v3.

To be honest, the first step I took a few months ago when I was looking into Meshcore was to flash my T1000 back to Meshtastic because it wouldn't turn off at all. Now, with the newer firmware, you can turn it off, but the power consumption is still half of what it used to be when in use.

I have not measured previously on meshtastic because of there was not such a behaviour draining the battery. It can be turned off for a long while and when you turn on it is still fully. It seems that the MT firmware is handling a hibernating mode more efficient.

* Unconnecting the battery is not convenient or practicable and also not good for the little connectors.
* To let it 24h on the charger (for sleeping?) as a second option is also not good, especially for the battery being held at 100%.
* Always running low as the third option and charging it is not convenient to have to look and remember every second day for charging it, and not good for the battery (300-500 cycles lifetime of a Lion)

1) It seems that hibernating is not implemented fully. It would be nice if it would be a top prio, it is an essential function to turn off a device. Each of us has more than one node, but we mainly use one and the others only when needed
2) It would also be nice if a (spare) repeater could be switched off for storage, but also efficiently and without draining the battery.

Vasili-Sk:
8 mA sounds like radio not sleeping. I remember pulling radio reset pin while testing power consumption on meshtastic, and got 400uA after that.

IoTThinks:
I checked poweroff function just now.

It will go to deepsleep BUT it will be back to normal mode when it receives any LoRa packets.

I think the code should separate the poweroff by button and other activities.

If off by buttons, then can only get back to normal mode by button press.

SaschaKt:
> I checked poweroff function just now.
> 
> It will go to deepsleep BUT it will be back to normal mode when it receives any LoRa packets.
> 
> I think the code should separate the poweroff by button and other activities.
> 
> If off by buttons, then can only get back to normal mode by button press.

Can you implement please a modification so that the companions are not waking up on packet receiving and last longer in power off? I think a complete deep sleep, also Lora switching in deep sleep? Thank you

IoTThinks:
[...]

PR #1138, opened 2025-11-22
companion: suspend radio when hibernating

yarda:
This should significantly reduce power consumption in the hibernation.

Also explicitly shutdown peripherals (GPS, LED) on Heltec T114.

Fixes: #1014

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -650,6 +650,7 @@ void UITask::shutdown(bool restart){
     _board->reboot();
   } else {
     _display->turnOff();
+    radio_driver.powerOff();
     _board->powerOff();
   }
 }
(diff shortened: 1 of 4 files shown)


yarda:
I am now testing it on Heltec T114 if it really saves power.

IoTThinks:
Have you shutdown the LoRa too?

fschrempf:
@yarda This is great! I recently provided [this PR](https://github.com/meshcore-dev/MeshCore/pull/1122) and it already has been tested on some devices with good results. I totally missed that there is a global instance of the `radio_driver` so I could actually call `sleep()` from the companion app code. That's what I've been looking for.

fschrempf:
@yarda I've integrated your changes in https://github.com/meshcore-dev/MeshCore/pull/1122 keeping your authorship. I hope this is ok for you.

I also changed it so, it always calls into `sleep()` as RadioLib should have it implemented for all modules. If it is not supported it will be a nop already in the lib.

fschrempf:
If @yarda agrees, we can close this in favor of #1122.

yarda:
@fschrempf thanks LGTM, closing.

Issue #989, opened 2025-10-21
GPS Mode setting does not persist after device reboot


salmancuso:
## Description
First, thank you to all the developers working on this project! The mesh networking capabilities are fantastic, and I really appreciate the work you're putting into MeshCore.

I'm experiencing an issue where the GPS Mode setting resets to "Disabled" every time my device reboots or powers back on. While I understand this might be intentional behavior based on the description in the UI, it creates a frustrating user experience that requires manual intervention after each reboot to restore GPS functionality.

![GPS Config Screen](https://stanford.edu/~smancuso/GPS-Issue-Meshcore.jpg "GPS Config Screen")

## Device Information
- **Hardware**: RAK WisBlock 4631
- **Firmware**: 1.9
- **App Version**: iPhone version 1.31.0. Apple MacBook Version 1.31.0

## Steps to Reproduce
1. Open Position Settings in the MeshCore app
2. Enable "GPS Mode" 
3. Confirm GPS is enabled and position updates are working
4. Reboot or power cycle the device
5. Check Position Settings after device comes back online

## Current Behavior
The GPS Mode setting reverts to "Disabled" after every device reboot. The UI even notes this behavior: *"When your MeshCore device reboots, GPS will be disabled."*

## Expected Behavior
The GPS Mode setting should persist across reboots. If I enable GPS, it should remain enabled until I explicitly disable it. This is the expected behavior for most device configuration settings.

## Impact
- Users must manually re-enable GPS after every reboot/power cycle
- Reduces the utility of GPS position broadcasting for mobile or remote deployments
- Creates potential for missed position updates if users forget to re-enable GPS
- Counterintuitive UX that differs from standard settings persistence patterns

## Suggested Solution
Persist the GPS Mode preference to non-volatile storage (EEPROM/flash) so it survives reboots, similar to how other device configuration settings are handled.

## Additional Context
This behavior seems inconsistent with other settings that do persist across reboots. Is there a technical reason for this current implementation, such as power consumption concerns or GPS initialization issues? If so, perhaps a compromise could be a user-configurable "GPS startup behavior" option?

Thanks again for your work on this project! Happy to provide any additional information or testing if needed.

liamcottle:
Hey there! I believe GPS state restoration was recently added for repeaters, sensors and room servers in https://github.com/meshcore-dev/MeshCore/pull/890, however this is still in the todo list for companions.

tgaguy:
[...]

PR #1156, opened 2025-11-29
Persist GPS enabled state to preferences

csrutil:
Add GPS configuration to NodePrefs structure and persist the GPS enabled state when toggled via UI. This ensures GPS settings are retained across device restarts.

Tested in WioTrackerL1_companion_radio_ble.

Scenario:

- Turn on the GPS in the GPS screen. Click `reset`, the device shows `gps on`.
- Turn off the GPS in the GPS screen. Click `reset`, the device shows `gps off`.

fixes https://github.com/meshcore-dev/MeshCore/issues/989

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -1525,6 +1525,17 @@ void MyMesh::handleCmdFrame(size_t len) {
       *np++ = 0; // modify 'cmd_frame', replace ':' with null
       bool success = sensors.setSettingValue(sp, np);
       if (success) {
+        #if ENV_INCLUDE_GPS == 1
+        // Update node preferences for GPS settings
+        if (strcmp(sp, "gps") == 0) {
+          _prefs.gps_enabled = (np[0] == '1') ? 1 : 0;
+          savePrefs();
+        } else if (strcmp(sp, "gps_interval") == 0) {
+          uint32_t interval_seconds = atoi(np);
+          _prefs.gps_interval = constrain(interval_seconds, 0, 86400);
+          savePrefs();
+        }
+        #endif
         writeOKFrame();
       } else {
         writeErrFrame(ERR_CODE_ILLEGAL_ARG);


liamcottle:
Looks pretty good thanks!

Would love if we could also get the new node prefs `gps_enabled` flag updated when the `CMD_SET_CUSTOM_VAR` is received from the companion app.

That command calls `sensors.setSettingValue("gps", ...)`, which allows the mobile app or other clients to turn gps on and off.

Being able to persist GPS across reboots when updating through the app, would be ideal imho.

https://github.com/meshcore-dev/MeshCore/blob/fe874032d56a4af67f176c9bdd8a942f90209f42/examples/companion_radio/MyMesh.cpp#L1522
https://github.com/meshcore-dev/MeshCore/blob/9405e8bee35195866ad1557be4af5f0c140b6ad1/src/helpers/sensors/EnvironmentSensorManager.cpp#L516

fdlamotte:
> Looks pretty good thanks!

Same for me, looks interesting

> Would love if we could also get the new node prefs `gps_enabled` flag updated when the `CMD_SET_CUSTOM_VAR` is received from the companion app.
> 
> That command calls `sensors.setSettingValue("gps", ...)`, which allows the mobile app or other clients to turn gps on and off.

This would definitely be better (and I think the only really valid way ...). But there are tricky things behind as sensors.setSettingValue is common to all firmware and companion has a distinct way to manage prefs ... 


Another thing is with the interval, ideally we should continue integrating @awolden work on #778, because there is no real interest in changing the update rate of gps if gps is not turned off between updates ;)

fdlamotte:
With latest commit, we now have basic gps prefs saved ok ...

I think the gps.interval pref can stay as it is, and when gps sleep is ok we can replace it with the sleeping behaviour

So for me it's ok

fdlamotte (approved):
looks good

liamcottle:
Great, thanks!

I've just tested this PR on a `Wio Tracker L1 Pro` and it enabled the GPS on reboot when saving through the app.

This feature will make a lot of users happy :)

Will wait for final review and merge from @ripplebiz.

liamcottle (approved):

PR #1160, opened 2025-11-29
Support for RAK11310 WisBlock

flol:
This is basically a copy of the `rpi_picow` and `waveshare_rp2040_lora` variants, spiced with some pinouts and voltage divider constants found in https://github.com/RAKWireless/RAK11300-AT-Command-Firmware and https://github.com/beegee-tokyo/SX126x-Arduino (both MIT license).

### DIFF
--- a/variants/rak11310/RAK11310Board.cpp
+++ b/variants/rak11310/RAK11310Board.cpp
@@ -1,0 +1,30 @@
+#include "RAK11310Board.h"
+
+#include <Arduino.h>
+#include <Wire.h>
+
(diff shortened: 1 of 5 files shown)


failMode:
Thanks for this! I've tested the Repeater build on a base unit (RF sheild on the board labeled "RAK11300") - it builds, flashes. boot and works perfectly. I'm happy to test out the other builds if it would assist getting this Pull over the line. In the meantime, I'll keep the Repeater build running to test stability etc.

PR #1164, opened 2025-11-30
CLI: Set current limit and RX boosted gain mode (SX126X and SX127X)

jbrazio:
This pull request adds support for configuring and persisting ~radio current limit and~ RX boosted gain settings for SX12xx series radios (SX1262, SX1268, SX1276) across the codebase. These settings are now stored in preferences, can be set and queried via CLI commands, and are applied during radio initialization.

-edit-
Removed the support for remotely setting the radio current limit as it proved to be rather useless in real the world application.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -802,20 +802,11 @@ MyMesh::MyMesh(mesh::MainBoard &board, mesh::Radio &radio, mesh::MillisecondCloc
 
   _prefs.adc_multiplier = 0.0f; // 0.0f means use default board multiplier
 
-#if defined(USE_SX1262) || defined(USE_SX1268) || defined(USE_SX1276)
-#if defined(SX126X_CURRENT_LIMIT)
-_prefs.sx12xx_current_limit = SX126X_CURRENT_LIMIT;
-#elif defined(SX127X_CURRENT_LIMIT)
-_prefs.sx12xx_current_limit = SX127X_CURRENT_LIMIT;
-#else
-_prefs.sx12xx_current_limit = 140; // mA
-#endif
 #if defined(USE_SX1262) || defined(USE_SX1268)
 #ifdef SX126X_RX_BOOSTED_GAIN
   _prefs.sx126x_rx_boosted_gain = SX126X_RX_BOOSTED_GAIN;
 #else
   _prefs.sx126x_rx_boosted_gain = 1; // enabled by default;
-#endif
 #endif
 #endif
 }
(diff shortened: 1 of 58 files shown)


fschrempf:
Instead of duplicating the setters and getters for each target, can we directly call the functions from the app code using the global `radio_driver` variable, similar to what was done here: https://github.com/meshcore-dev/MeshCore/commit/07e7e2d44bfd68abbe87f73a853b04d76b37ddf5?

fschrempf:
Can you explain why this is needed and what advantages these changes have? In what cases do these parameters need to be changed?

jbrazio:
> duplicating the setters and getters for each target

Didn't want to change the existing architecture: `radio_set_params`, `radio_set_tx_power` and `radio_new_identity` are all duplicated. If @ripplebiz agrees with it, I will refactor.

> why this is needed and what advantages these changes have

Are you serious on this one ?

fschrempf:
> If @ripplebiz agrees with it, I will refactor.

Got it, thanks!

> Are you serious on this one ?

Absolutely. Why does this need to be a user-facing parameter? Why is a hardcoded value per variant not enough? Sorry for my lack of knowledge here. I'm not an expert, just interested in what this is about.

jbrazio:
I would call it more "sysadmin facing" as the main target is repeaters: You might hang a repeater [with a proper antenna] in a difficult to access place to find out that there is a lot of QRM and RX is deaf; most of the situations where I faced this, disabling the Boosted Gain helped a lot with the problem.

When we use small out-of-tune antennas (the ones shipping with the boards) having a boost on the RX helps, this is usually the client situation. For repeater where we usually have proper antennas it's a question of testing with it ON and OFF and check what gives you better performance.

For the OCP there is no big use case but I had a couple of situations where tweaking the OCP helped with random resets of the radio.

fschrempf:
Thanks for the detailed explanations. Makes sense to me. I didn't know that disabling the RX boost can be benefitial for repeaters with proper antennas.

fdlamotte:
These enhancements are good (should be done for all radios ...)

but this is really up to @ripplebiz ;)

fdlamotte:
These enhancements are good (should be done for all radios ...)

but this is really up to @ripplebiz ;)

jbrazio:
@ripplebiz @fdlamotte @liamcottle This PR needs rebasing as it's pending arbitration since Nov 30 2025..

Could you please take a decision if it will be merged ?
Otherwise I will just close it and don't bother rebasing.

weebl2000:
Good feature. Especially boosted gain, not sure if the other params will need tweaking a lot.

JDat:
Why there are no patch for rpi_picow variant?

jbrazio:
Superseded by #1653.

PR #1171, opened 2025-12-03
variants: lilygo_techo: variant: Turn off leds on poweroff

luigi311:
Turns the leds off when putting the t-echo in hibernate mode


luigi311:
accidently built this on top of main, has been retargeted to dev and rebased

PR #1172, opened 2025-12-03
NRF52 Power related changes

oltaco:
This PR makes a couple of changes NRF52 power-related functions and clarifies DC/DC usage restrictions.

**Important:** [The NRF52840’s DC/DC converters *require external inductors and capacitors*](https://docs.nordicsemi.com/bundle/ps_nrf52840/page/power.html#ariaid-title3), so they are **not available on all boards**. The Heltec T114 and Xiao nRF52840 include the required components for REG1 DC/DC. I couldn't locate a schematic for the RAK WisMesh Tag and WM1110 to verify they have the required LC network, but I've left the fixed calls in anyway.
## Changes

1. **Remove unnecessary power-mode calls:**  
sd_power_mode_set(NRF_POWER_MODE_LOWPWR) was redundant because LOWPWR is already the default unless explicitly changed to CONSTLAT.

2. **Correct enabling of DC/DC converters:**  
DC/DC converters can only be configured directly through NRF_POWER registers when the SoftDevice is **not** running. When the SoftDevice is active, its API **must** be used.

### DIFF
--- a/variants/rak_wismesh_tag/RAKWismeshTagBoard.cpp
+++ b/variants/rak_wismesh_tag/RAKWismeshTagBoard.cpp
@@ -21,7 +21,15 @@ static void disconnect_callback(uint16_t conn_handle, uint8_t reason) {
 void RAKWismeshTagBoard::begin() {
   // for future use, sub-classes SHOULD call this from their begin()
   startup_reason = BD_STARTUP_NORMAL;
-  NRF_POWER->DCDCEN = 1;
+  
+  // Enable DC/DC converter for improved power efficiency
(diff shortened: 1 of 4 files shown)


fschrempf:
Strangely I could see a decrease of power consumption on the Xiao NRF52 even with SD enabled and `NRF_POWER->DCDCEN = 1`. But anyway, this looks like the correct approach.

oltaco:
It's entirely possible it was working correctly, it really depends on whether the softdevice is already enabled when that register change was being called.

Hopefully we can work towards a more unified power management system.

Issue #1067, opened 2025-11-06
[T1000-E] [iOS][Fw-crash] FW crashes on BT disconnects


Michael-Frank:
(Opening on behalf of other users not versed in github.)

While the recent fixes (i tracked some of them in #936 )  to BT and NRF file system corruption made the nightly firmware stable for T1000-E @ Android, the iOS users still have daily trouble.

In short: the firmware frequently crashes when connected to an ios device.

Anyone knows what happend to pr #766 ? it seams very related to the iOS specific stuff . @Cisien ? @ventz ?

ventz:
I have seen the same (FS) on the Nano G2 Ultra - it crashes daily (every few hours)

Michael-Frank:
@ventz  in #766 you mentioned you could provide the fixes in smaller increments.
Is the offer still standing?
I could build a custom firmware from your branch/pr and distribute  it to the local mesh for testing and collect feedback -  if that would help?

dosch:
Not sure if related to this issue, but yesterday I experienced multiple crashes with my seeed nanocap t-1000, using the Android app. The only way to revive the device was to hold the power-button while disconnecting and re-connecting the power. 

I found these lines in the log just before the crash happened;

```
[23:13:19] [BleConnection] connectionStateSubscription: BluetoothConnectionState.disconnected
[23:13:19] [BleConnection] disconnectReason=DisconnectReason{platform: ErrorPlatform.android, code: 8, LINK_SUPERVISION_TIMEOUT}
[23:13:19] [BleConnection] connectionStateSubscription: BluetoothConnectionState.disconnected
[23:13:19] [BleConnection] disconnectReason=DisconnectReason{platform: ErrorPlatform.android, code: 8, LINK_SUPERVISION_TIMEOUT}
[23:13:19] [ConnectScreen] ble device disconnected!
```

Michael-Frank:
@dosch 
do you use the current release version[ 1.9.0 ](https://github.com/meshcore-dev/MeshCore/releases/download/companion-v1.9.0/t1000e_companion_radio_ble-v1.9.0-b92d9bd.uf2) or the [nightly dev builds](https://files.brazio.org/meshcore/nightly/companion/)?

dosch:
the 1.9.0 version

Michael-Frank:
@dosch Then flash [nightly dev builds](https://files.brazio.org/meshcore/nightly/companion/) till the next official release - the android issues are allready fixed in dev.

[...]

Issue #915, opened 2025-10-11
Companion Node BLE Lockup, RAK4631, Heltec T114 (v2),Ikoka, Seeed t1000


MrAlders0n:
Several users, myself included, are still seeing companion node lockups on RAK4631, Heltec T114v2, Ikoka, and Seeed T1000E boards. When it happens, the device becomes completely unresponsive until it is power cycled.

This has been happening since at least 1.8.x and continues in 1.9.0. I am not sure if it started earlier than that. The same behavior does not show up on Heltec V3 companions running in the same environment.

Symptoms
	•	Only happens on iOS
	•	Happens specifically when leaving BLE range and then coming back into range
	•	The mobile app cannot reconnect after the freeze
	•	Blue LED on the RAK4631 goes solid instead of blinking
	•	Pressing reset reboots the device making it operational again.
	•	When frozen, button inputs and menu navigation do nothing
	•	Happens more often as the device contact list grows larger

Issue is being tracked in discord here:

https://discord.com/channels/1343693475589263471/1437565846435139668

MrAlders0n:
I just stumbled upon Issue 740 which actually sounds exactly like what our mesh is running into. This may be a duplicate at this point.

[Issue 740](https://github.com/meshcore-dev/MeshCore/issues/746)

mkwillis:
Can confirm T114 locking up as soon as I connect via BT to android app. 1.9.0 firmware and reflash not resolving the matter.

mkwillis:
Going back to this firmware did resolve the lockups:

https://github.com/meshcore-dev/MeshCore/issues/746#issuecomment-3296901192

MrAlders0n:
Hey @mkwillis what firmware version did you roll back to? The link you posted is an empty repo

mkwillis:
Sorry, link now fixed. I just uploaded the UF2 as discussed in issue 740. Apparently it is a power saving setting causing the BT connect lockups.

AJSNetworks:
I am also experience this a Heltec T114 on V 1.8.1-1130cf1. 

Randomly about once a day. 

Companion will become unresponsive over Bluetooth, blank screen and menu button. 

Pushing reboot resolves.

hotwolf:
I've made the following observation:
I have 4 Heltec T114 BT companions in use. Two of them have a GNSS module connected, the other two don't. 
The companions with out GNSS module run fine on the official 1.9.0 firmware.
[...]

PR #1177, opened 2025-12-04
Fix BLE semaphore leak and improve SerialBLEInterface

### LINKED ISSUE #746: Companion firmware v1.8.1 keeps crashing on a Heltec T114
I've installed Companion Firmware v1.8.1 on my HeltecT114, but it is not usable, because it freezes up and terminates the Bluetooth connection every 1 or 2 minutes. The same T114 hardware runs fine with companion firmware v1.7.4.
I have no problems running Companion Firmware v1.8.1 on my Heltec V3.

Please let me know is there is any information I can provide to help find the cause  of this issue.

liquidraver:
TL;DR
This PR fixes a critical semaphore leak in Bluefruit that cause device lockups with iOS, and completely refactor SerialBLEInterface to handle the messy reality (understatement) of BLE connections with our dear friends SoftDevice and Bluefruit.

(hopefully) Resolves #746 #915 and #1067

I try to write down the key points that changed, sorry for the diff mess, I needed to rearrange serialbleinterface for better visibility/flow trace for myself

If anyone is interested in the whole debug process that is going on for weeks, there are a lot to read on discord in the thread: https://discord.com/channels/1343693475589263471/1437565846435139668
A huge thanks to everyone who has been testing it and sticking with us through all the reflashing and tweaks!

The Semaphore Leak (The Big One)
Root Cause: When the BT disconnects unexpectedly (supervision timeout, out of range, etc.), the BLE_GATTS_EVT_HVN_TX_COMPLETE event may never fire. This leaves Bluefruit's _hvn_sem counting semaphore in a decremented state. Since BLEConnection objects are reused (destructor never called), the semaphore count is never restored. After a few disconnects, all semaphore counts are exhausted and notify() blocks forever resulting in device lockup.
Added "patch_bluefruit.py" build script that patches the Bluefruit library at compile time (massive thx to Liam for figuring out in the first place and oltaco for refining it)

Connection State Management
- Added "_conn_handle" to track the active connection handle (always the same, but more reliable to track the state)
- "isValidConnection()" method validates handle matches and connection is actually connected
- All callbacks check "isValidConnection()" before processing
- "requireWaitingForSecurity" parameter prevents race conditions during pairing / rapid disconnects eg. someone is at the edge of BT range (...or repeatedly slams the microwave door on the phone)
  * Rejects duplicate onSecured() callbacks
  * Prevents stale callbacks from previous connections
  * Ensures connection parameters are only updated once per connection
- "isConnected()" now checks both "_isDeviceConnected" AND "Bluefruit.connected() > 0" to prevent false positives

Queue Management
- Send queue handles partial writes and connection loss gracefully
- Overflow protection: drains BLE buffer if frame exceeds "MAX_FRAME_SIZE" (thanks to spacepc-de for the idea, this Supersedes PR: #872)
- Separate send_queue and recv_queue (both size 12, up from 4)
- Receive queue populated in onBleUartRX() callback, consumed in checkRecvFrame() (callback-driven, not polling)
- Proper queue shifting functions (shiftSendQueueLeft, shiftRecvQueueLeft)

Connection Parameter Hell
iOS ignores Peripheral Preferred Connection Parameters (PPCP) (why? because *beep* you, I know it better, that's why!)
Explicitly set connection parameters in two places:
- PPCP: Sets preferred params
- Explicit request an update (in onSecured)

Parameters: 15-30ms interval, 0 latency, 2s supervision timeout. - conforms https://developer.apple.com/accessories/Accessory-Design-Guidelines.pdf
(We can increase latency if we want to squeeze a little more out of our battery, but it's miniscule compared to LoRa's consumption)

Advertising Watchdog
Advertising can silently stop (BLE stack bug, SoftDevice magic, Bluefruit moodswings, etc.) Device becomes undiscoverable until BT disable/enable.
Health check in checkRecvFrame():
- Runs every 10 seconds and restores advertising
This is needed because we rely on Bluefruit.Advertising.restartOnDisconnect(true)
The SoftDevice manages a BLE state machine. Calling sd_ble_gap_adv_start() at the wrong time (e.g., while advertising is still stopping, during connection establishment, or before proper cleanup) can return BLE_ERROR_INVALID_STATE or cause unpredictable behavior.
Using restartOnDisconnect(true) is the proper way but there were edge cases when the advertisement did not start. This is the best solution choosing from the two methods.
- isAdvertising() method uses sd_ble_gap_adv_addr_get() to detect advertising state

Connection stuff
- onPairingComplete(): Disconnects if pairing fails
- disconnect(): New method to explicitly disconnect via sd_ble_gap_disconnect()

TX
- Partial write detection: if bleuart.write() returns less than frame length, frame is dropped (corrupted)
- Connection loss detection: if write fails and connection is lost, frame is dropped

RX
- Overflow protection: drains buffer if available > MAX_FRAME_SIZE
- Queue full protection: drains remaining data if queue full

Optional:
// Bluefruit.autoConnLed(false); - if uncommented, we can do whatever else we want with the BT led.

### DIFF
--- a/src/helpers/nrf52/SerialBLEInterface.cpp
+++ b/src/helpers/nrf52/SerialBLEInterface.cpp
@@ -1,193 +1,353 @@
 #include "SerialBLEInterface.h"
+#include <stdio.h>
+#include <string.h>
+#include "ble_gap.h"
+#include "ble_hci.h"
 
-static SerialBLEInterface* instance;
+#define BLE_HEALTH_CHECK_INTERVAL  10000  // Advertising watchdog check every 10 seconds
+
+static SerialBLEInterface* instance = nullptr;
 
 void SerialBLEInterface::onConnect(uint16_t connection_handle) {
-  BLE_DEBUG_PRINTLN("SerialBLEInterface: connected");
-  // we now set _isDeviceConnected=true in onSecured callback instead
+  BLE_DEBUG_PRINTLN("SerialBLEInterface: connected handle=0x%04X", connection_handle);
+  if (instance) {
+    instance->_conn_handle = connection_handle;
+    instance->_isDeviceConnected = false;
+    instance->clearBuffers();
+  }
 }
 
 void SerialBLEInterface::onDisconnect(uint16_t connection_handle, uint8_t reason) {
-  BLE_DEBUG_PRINTLN("SerialBLEInterface: disconnected reason=%d", reason);
-  if(instance){
-    instance->_isDeviceConnected = false;
-    instance->startAdv();
+  BLE_DEBUG_PRINTLN("SerialBLEInterface: disconnected handle=0x%04X reason=%u", connection_handle, reason);
+  if (instance) {
+    if (instance->_conn_handle == connection_handle) {
+      instance->_conn_handle = BLE_CONN_HANDLE_INVALID;
+      instance->_isDeviceConnected = false;
+      instance->clearBuffers();
+    }
   }
 }
 
 void SerialBLEInterface::onSecured(uint16_t connection_handle) {
-  BLE_DEBUG_PRINTLN("SerialBLEInterface: onSecured");
-  if(instance){
-    instance->_isDeviceConnected = true;
-    // no need to stop advertising on connect, as the ble stack does this automatically
+  BLE_DEBUG_PRINTLN("SerialBLEInterface: onSecured handle=0x%04X", connection_handle);
+  if (instance) {
+    if (instance->isValidConnection(connection_handle, true)) {
+      instance->_isDeviceConnected = true;
+      
+      // Connection interval units: 1.25ms, supervision timeout units: 10ms
+      // Apple: "The product will not read or use the parameters in the Peripheral Preferred Connection Parameters characteristic."
+      // So we explicitly set it here to make Android & Apple match
+      ble_gap_conn_params_t conn_params;
+      conn_params.min_conn_interval = 12;  // 15ms
+      conn_params.max_conn_interval = 24;  // 30ms
+      conn_params.slave_latency = 0;
+      conn_params.conn_sup_timeout = 200;  // 2000ms
+      
+      uint32_t err_code = sd_ble_gap_conn_param_update(connection_handle, &conn_params);
+      if (err_code == NRF_SUCCESS) {
+        BLE_DEBUG_PRINTLN("Connection parameter update requested: 15-30ms interval, 2s timeout");
+      } else {
+        BLE_DEBUG_PRINTLN("Failed to request connection parameter update: %lu", err_code);
+      }
+    } else {
+      BLE_DEBUG_PRINTLN("onSecured: ignoring stale/duplicate callback");
+    }
+  }
+}
+
+bool SerialBLEInterface::onPairingPasskey(uint16_t connection_handle, uint8_t const passkey[6], bool match_request) {
+  (void)connection_handle;
+  (void)passkey;
+  BLE_DEBUG_PRINTLN("SerialBLEInterface: pairing passkey request match=%d", match_request);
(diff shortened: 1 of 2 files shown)


liquidraver:
@fdlamotte Could you please review this?

recrof:
just my 2 cents: this change was tested pretty extensively with various radios by 10s of people, so it gets seal of approval from me.

liquidraver:
I can try to make the commits cleaner, but I'm not a git magician, hope I won't mess this up :)

fdlamotte:
Cool changes, that's really clean for integration !

oltaco:
Really nice, we've been testing this code for a while now and know it works well.
Awesome clean rebase and squash, if only more PRs looked like this 🥇 

I know this is a pedantic nitpick on my part, but there are a lot of unnecessary casts which maybe hurt readability and are a bit out of step with the rest of the codebase imo. Just my 2c.

src/helpers/nrf52/SerialBLEInterface.cpp
```diff
@@ -1,193 +1,353 @@
...
+    instance->clearBuffers();
+  }
 }
 
 void SerialBLEInterface::onDisconnect(uint16_t connection_handle, uint8_t reason) {
-  BLE_DEBUG_PRINTLN("SerialBLEInterface: disconnected reason=%d", reason);
-  if(instance){
-    instance->_isDeviceConnected = false;
-    instance->startAdv();
+  BLE_DEBUG_PRINTLN("SerialBLEInterface: disconnected handle=0x%04X reason=%u", connection_handle, (unsigned)reason);
```
oltaco: reason is already uint8_t, so the cast is unnecessary.

src/helpers/nrf52/SerialBLEInterface.cpp
```diff
@@ -1,193 +1,353 @@
...
+
+void SerialBLEInterface::onBLEEvent(ble_evt_t* evt) {
+  if (!instance) return;
+  
+  if (evt->header.evt_id == BLE_GAP_EVT_CONN_PARAM_UPDATE_REQUEST) {
+    uint16_t conn_handle = evt->evt.gap_evt.conn_handle;
+    if (instance->isValidConnection(conn_handle)) {
+      BLE_DEBUG_PRINTLN("CONN_PARAM_UPDATE_REQUEST: handle=0x%04X, min_interval=%u, max_interval=%u, latency=%u, timeout=%u",
+                       conn_handle,
+                       (unsigned)evt->evt.gap_evt.params.conn_param_update_request.conn_params.min_conn_interval,
```
oltaco: same with these ones

oltaco (changes requested):
@fdlamotte knows best, but I think maybe removing the unnecessary casts would help readability and keep it more in step with the rest of the codebase.

src/helpers/nrf52/SerialBLEInterface.cpp
liquidraver: right, removed

src/helpers/nrf52/SerialBLEInterface.cpp
liquidraver: removed this too

oltaco:
@liquidraver Very nice,  looks much more neat and tidy to me. 👍

I'm on phone at the moment, is the advertising check stuff new? I'm not familiar sd_ble_gap_adv_addr_get

liquidraver:
> @liquidraver Very nice, looks much more neat and tidy to me. 👍
> 
> I'm on phone at the moment, is the advertising check stuff new? I'm not familiar sd_ble_gap_adv_addr_get

sd_ble_gap_adv_addr_get is just used to check if we are advertising when we should (not connected, no handle, but enabled, etc). that command gives back error if we are not advertising so good for checking :)

PR #1180, opened 2025-12-05
Companion: Faster startup by calculating shared_secret on demand

oltaco:
The PR improves boot time on companions with a large amount of contacts.

Instead of calculating the ECDH shared secret for each contact at startup, I have added a bool shared_secret_valid to the ContactInfo struct which gets checked with a helper function BaseChatMesh::ensureSharedSecretIsValid() any time a shared secret is required.

Hopefully it passes review from the brains trust.

### DIFF
--- a/src/helpers/ContactInfo.h
+++ b/src/helpers/ContactInfo.h
@@ -9,9 +9,10 @@ struct ContactInfo {
   uint8_t type;   // on of ADV_TYPE_*
   uint8_t flags;
   int8_t out_path_len;
+  mutable bool shared_secret_valid; // flag to indicate if shared_secret has been calculated
   uint8_t out_path[MAX_PATH_SIZE];
   uint32_t last_advert_timestamp;   // by THEIR clock
-  uint8_t shared_secret[PUB_KEY_SIZE];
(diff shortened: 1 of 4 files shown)


src/helpers/ContactInfo.h
```diff
@@ -11,8 +11,9 @@ struct ContactInfo {
   int8_t out_path_len;
   uint8_t out_path[MAX_PATH_SIZE];
   uint32_t last_advert_timestamp;   // by THEIR clock
-  uint8_t shared_secret[PUB_KEY_SIZE];
+  mutable uint8_t shared_secret[PUB_KEY_SIZE];
   uint32_t lastmod;  // by OUR clock
   int32_t gps_lat, gps_lon;    // 6 dec places
   uint32_t sync_since;
+  mutable bool shared_secret_valid; // flag to indicate if shared_secret has been calculated
```
ripplebiz: If you move this field up to the other 1-byte fields (eg. after out_path_len) you'll save 4 bytes!

oltaco: Thanks for the tip, it's something I never think of. I've force-pushed the change :)

PR #1194, opened 2025-12-08
Added support for the nibble zero connect

ViezeVingertjes:
image

### DIFF
--- a/variants/nibble_zero_connect/platformio.ini
+++ b/variants/nibble_zero_connect/platformio.ini
@@ -1,0 +1,163 @@
+[nibble_zero_connect_base]
+extends = esp32_base
+board = esp32-s3-zero
+build_flags =
+  ${esp32_base.build_flags}
+  -I variants/nibble_zero_connect
+  -D NIBBLE_ZERO_CONNECT
+  -D P_LORA_DIO_1=4
+  -D P_LORA_NSS=10
+  -D P_LORA_RESET=6
+  -D P_LORA_BUSY=5
+  -D P_LORA_SCLK=12
+  -D P_LORA_MISO=13
+  -D P_LORA_MOSI=11
+  -D PIN_USER_BTN=1
+  -D PIN_BOARD_SDA=8
+  -D PIN_BOARD_SCL=7
+  -D PIN_STATUS_LED=39
+  -D P_LORA_TX_LED=39
+  -D DISPLAY_ROTATION=0
+  -D HAS_NEOPIXEL
+  -D NEOPIXEL_COUNT=1
+  -D NEOPIXEL_DATA=17
+  -D NEOPIXEL_TYPE=(NEO_GRB+NEO_KHZ800)
+  -D SX126X_DIO2_AS_RF_SWITCH=true
+  -D SX126X_DIO3_TCXO_VOLTAGE=1.8
+  -D SX126X_CURRENT_LIMIT=140
+  -D RADIO_CLASS=CustomSX1262
+  -D WRAPPER_CLASS=CustomSX1262Wrapper
+  -D LORA_TX_POWER=22
+  -D SX126X_RX_BOOSTED_GAIN=1
+build_src_filter = ${esp32_base.build_src_filter}
+  +<../variants/nibble_zero_connect>
+lib_deps =
+  ${esp32_base.lib_deps}
+  adafruit/Adafruit SSD1306 @ ^2.5.13
+  adafruit/Adafruit NeoPixel @ ^1.12.3
+
+[env:nibble_zero_connect_repeater]
+extends = nibble_zero_connect_base
+build_flags =
+  ${nibble_zero_connect_base.build_flags}
+  -D DISPLAY_CLASS=SSD1306Display
+  -D ADVERT_NAME='"Nibble Repeater"'
+  -D ADVERT_LAT=0.0
+  -D ADVERT_LON=0.0
+  -D ADMIN_PASSWORD='"password"'
+  -D MAX_NEIGHBOURS=50
+build_src_filter = ${nibble_zero_connect_base.build_src_filter}
+  +<helpers/ui/SSD1306Display.cpp>
+  +<../examples/simple_repeater>
+lib_deps =
+  ${nibble_zero_connect_base.lib_deps}
+  ${esp32_ota.lib_deps}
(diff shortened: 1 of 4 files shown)


ndom91:
Confirmed working! 🙏

ViezeVingertjes:
> Confirmed working! 🙏

Great! If you are actually using it with the flipper zero, then it might be cool to look into the SecureChat firmware. It needs a small modification to run over the TX/RX pins instead of USB Serial, but then you could also send messages etc from the Flipper. ^^

variants/nibble_zero_connect/platformio.ini
```diff
@@ -0,0 +1,163 @@
...
+  -D PIN_USER_BTN=1
+  -D PIN_BOARD_SDA=8
+  -D PIN_BOARD_SCL=7
+  -D PIN_STATUS_LED=39
+  -D P_LORA_TX_LED=39
+  -D DISPLAY_ROTATION=0
+  -D HAS_NEOPIXEL
+  -D NEOPIXEL_COUNT=1
+  -D NEOPIXEL_DATA=17
+  -D NEOPIXEL_TYPE=(NEO_GRB+NEO_KHZ800)
```
ndom91: I had to comment this out - it didnt like the `(` / `)` syntax

ViezeVingertjes: What `/` are you refering to exactly? Maybe im overlooking it, but i dont see it haha...
Builds and flashes fine as-is for me (on Windows though)

ndom91: Sorry haha the backtick formatting may have been confusing. The problematic chracters for me are the open parens and close parens

ndom91:
I'm not using it with a flipper zero though. Just wanted to separately compile the latest MeshCore fw for the nibble zero in the wifi variant, which nugget.dev seems to not offer.

ndom91:
Although everything seems to work, and I can see client and router adverts in the rx logs, I seemingly cannot discover any other contacts. Don't know if its an issue with these changes though as these are just the hardware definition stuffs, but figured I'd mention it in case anyone else has seen this on the nibble zero and knows a fix 😊

ViezeVingertjes:
> Although everything seems to work, and I can see client and router adverts in the rx logs, I seemingly cannot discover any other contacts. Don't know if its an issue with these changes though as these are just the hardware definition stuffs, but figured I'd mention it in case anyone else has seen this on the nibble zero and knows a fix 😊

I will test it once more when i get home, pretty sure i have seen that working, but wont hurt to confirm.

[Edit]
Works on my end (Companion BLE)

variants/nibble_zero_connect/platformio.ini
ViezeVingertjes: Check; i did see the same on Ubuntu. On Windows it did not care about it. Will fix.

variants/nibble_zero_connect/target.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+void radio_set_params(float freq, float bw, uint8_t sf, uint8_t cr) {
+  radio.setFrequency(freq);
+  radio.setSpreadingFactor(sf);
+  radio.setBandwidth(bw);
+  radio.setCodingRate(cr);
+}
+
+void radio_set_tx_power(uint8_t dbm) {
```
liamcottle: I believe this now needs to be `int8_t` instead of `uint8_t`.

liamcottle:
Hey, thanks for the PR :)

I know it's been a good 4 months since this was opened, but is this board purchasable anywhere?
It's very cute, but I'm wondering if it's also very limited in how many boards exist?

I'm thinking it should be left as a custom board that isn't automatically built, as our build/release flow already takes over an hour now, and if the board is only used by several people it would be best for us to avoid the increased build time.

If you still want this added in, I'd probably suggest putting an underscore `_` at the end of the target env names so it's excluded from the automatic build process. Anyone that has the board can build and flash, but it wouldn't be available on the flasher.

variants/nibble_zero_connect/target.h
```diff
@@ -0,0 +1,31 @@
...
+
+#ifdef DISPLAY_CLASS
+  extern DISPLAY_CLASS display;
+  extern MomentaryButton user_btn;
+#endif
+
+bool radio_init();
+uint32_t radio_get_rng_seed();
+void radio_set_params(float freq, float bw, uint8_t sf, uint8_t cr);
+void radio_set_tx_power(uint8_t dbm);
```
liamcottle: I believe this now needs to be int8_t instead of uint8_t.

liamcottle (changes requested):

skickar:
> Hey, thanks for the PR :)
> 
> I know it's been a good 4 months since this was opened, but is this board purchasable anywhere? It's very cute, but I'm wondering if it's also very limited in how many boards exist?
> 
> I'm thinking it should be left as a custom board that isn't automatically built, as our build/release flow already takes over an hour now, and if the board is only used by several people it would be best for us to avoid the increased build time.
> 
> If you still want this added in, I'd probably suggest putting an underscore `_` at the end of the target env names so it's excluded from the automatic build process. Anyone that has the board can build and flash, but it wouldn't be available on the flasher.

I'm the board dev! Currently can be bought from retia.io, so far there are 250+ of them, plus another 100 in production. It would be nice to be able to flash from the main site, but I understand. Making it easy to compile via our own github action would probably be my workaround.

TheNewsYorkTime:
I'm trying to flash my nibble to meshcore. Is anyone still working on this?

ndom91:
Just wanted to chime in again and say the latest version of this does still work on my nibble screen connect 🙏

ViezeVingertjes:
> Just wanted to chime in again and say the latest version of this does still work on my nibble screen connect 🙏

It shouldnt break anytime soon... thought i already addressed Liam's comments before, but apparantly not, so still need to hehe... 💯 

[Edit]
Comments resolved; code the comments were on didnt exist anymore, updated the env names as by @liamcottle 's suggestion, rebased.

liamcottle:
> [Edit] Comments resolved; code the comments were on didnt exist anymore, updated the env names as by @liamcottle 's suggestion, rebased.

Looks good, thanks for the adjustments. Will merge :)

Issue #1091, opened 2025-11-12
Autoadvert on boot for repeaters causes DDOS on mesh in case of boot loop


mattzzw:
We have a faulty repeater, probably stuck in a boot loop of some sorts.
This is resulting in an advert interval of up to  20 seconds, flooding the whole mesh.

I strongly propose to disable the "advert on boot" functionality for repeaters.

(Also, I propose to up the auto advert interval to 24h as default)

RobinWhitfield:
I can't come up with a good reason for advert-on-boot.

Personally I've not had any issues with frequently rebooting repeaters, but I regularly see new repeaters that clearly someone has not finished configuring, that is spamming out adverts with slightly nonsensical names or lacking information which then appears in another advert a few minutes to an hour later.

When building/configuring a new repeater (especially as a user new to MeshCore) it's pretty likely that you'll be rebooting the device numerous times, and it makes more sense to wait until that is done before manually sending an advert. Worst case no manual advert happens and the device then pings one out 12 hours later (or whatever the advert schedule is).

smeegle5000:
perhaps avert like 10 minutes after booting? when setting up a repeater its nice to not have to remember to hit a manual advert before closing it all up

jpds:
Far better to advert 30m-1h after boot, I've had this thought whilst replacing the antenna and then doing a firmware upgrade on my repeaters. 

Edit: And these things don't even have the correct time when they boot up.

mattzzw:
Also adding AG/Andrew's idea here: 

> Change it to advertise zero hop on boot
The situation is powering off and on a repeater in the field without computer to set it up or troubleshoot it.. say you don't have it on your node.

mikecarper:
Use the uptime stats and send out the zero hop advert after 10 minutes and the flood after 1 hour of uptime

k7cfo:
I like the advert on boot when setting up new repeaters

Beanow:
During a boot loop situation, a 0-hop advert may still add localized spamming.
And while a 10 minute delay reduces it, it doesn't solve the part those messages are wasteful.

## First setup use-case

Whilst setting up a new repeater, only really one advert needs to be received by your companion.
So why not cover that use case with on-demand discovery?

## On boot behavior

As for starting up the normal advert cadence, I'd select a **random** delay between 0 and `advert.interval` minutes.

If the repeater fails to stay up for the average of 6h and never gets to it's first advert, it clearly needs maintenance anyway.

[...]

PR #1199, opened 2025-12-09
Default to zero hop advert when booting + STEALTH_MODE flag

jbrazio:
This pull request updates the advertisement sending logic in the mesh examples to allow for both "flood" and "zero hop" advertisement modes. The changes introduce a new `flood` parameter to the `sendSelfAdvertisement` method, making it possible to specify the advertisement type. The initial advertisement sent during setup in all examples is now a zero hop advertisement.

The CLI command `"advert"` continues to use flood advertisements for backward compatibility, as the `flood` parameter is omitted (defaulting to `true`).

Adds a `STEALTH_MODE` switch to remove automatic node advertisement when booting and `DISCOVER_REQ` processing.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -703,8 +703,8 @@ MyMesh::MyMesh(mesh::MainBoard &board, mesh::Radio &radio, mesh::MillisecondCloc
   _prefs.bw = LORA_BW;
   _prefs.cr = LORA_CR;
   _prefs.tx_power_dbm = LORA_TX_POWER;
-  _prefs.advert_interval = 1;        // default to 2 minutes for NEW installs
-  _prefs.flood_advert_interval = 12; // 12 hours
+  _prefs.advert_interval = DEF_LOCAL_ADVERT_INTERVAL;
+  _prefs.flood_advert_interval = DEF_FLOOD_ADVERT_INTERVAL;
   _prefs.flood_max = 64;
   _prefs.interference_threshold = 0; // disabled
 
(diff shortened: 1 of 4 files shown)


fdlamotte:
This behavior has already been discussed (and the fact that bootlooping repeaters are an issue)

Another change that should be included upon @ripplebiz decision ...

jbrazio:
tbh I would prefer for repeaters not to issue any adverts during `setup()`.. nevertheless I kept it backward compatible but having less impact on the network by using zh.

mattzzw:
This would fix the DDOS/advert situation, see issue #1091

jbrazio:
Implements https://github.com/meshcore-dev/MeshCore/pull/1217#issuecomment-3654930555

```
  -D MIN_LOCAL_ADVERT_INTERVAL=60
  -D MAX_LOCAL_ADVERT_INTERVAL=240
  -D MIN_FLOOD_ADVERT_INTERVAL=3
  -D MAX_FLOOD_ADVERT_INTERVAL=48
```

mattzzw:
Why is local  in minutes and flood in hours?

jbrazio:
> Why is local in minutes and flood in hours?

@ripplebiz

ViezeVingertjes:
> tbh I would prefer for repeaters not to issue any adverts during `setup()`.. nevertheless I kept it backward compatible but having less impact on the network by using zh.

It needs to do it at least once for setup, otherwise the newly flashed repeater cannot be setup other than through USB but zero hop is much better than a flood advert for sure.

liamcottle:
Thanks for your work here. We have implemented similar changes in #1488

PR #1203, opened 2025-12-11
Fix GPS/Buzzer toggle UI popup

liquidraver:
Saving GPS state to flash introduced a small bug. (PR #1156)
The act of saving GPS preferences consumed most of the alert popup time so the screen only showed GPS: On / GPS: Off for a very brief period, not the full 800msec that was intended.

Moving alerts after saveprefs solves this.


liquidraver:
(this PR moves the buzzer alert after saveprefs too because it exhibits the same bug)

liquidraver:
sorry to always poke you @fdlamotte but can this be merged before the next release? the popups looking awful now on dev :)

liamcottle:
Thank you :) I noticed this on my unit too.

fdlamotte:
yes this was definitely necessary ;)

PR #1206, opened 2025-12-12
Added default temperature from ESP32 MCU and NRF52 MCU

IoTThinks:
I have cleaned up and creating a new PR to to display default temperature to all ESP32 and NRF52 boards.
- Added default temperature from ESP32 MCU and NRF52 MCU
- Added NRF52Board.h and NRF52Board.cpp
- Modified NRF52 variants to extend from NRF52Board to share common feature

Old PR is here: https://github.com/meshcore-dev/MeshCore/pull/1163.
Thanks a lot.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -174,9 +174,9 @@ int MyMesh::handleRequest(ClientInfo *sender, uint32_t sender_timestamp, uint8_t
     telemetry.reset();
     telemetry.addVoltage(TELEM_CHANNEL_SELF, (float)board.getBattMilliVolts() / 1000.0f);
 
-    float temperature = (float)board.getMCUTemperature();
+    float temperature = board.getMCUTemperature();
     if(!isnan(temperature)) { // Supported boards with built-in temperature sensor. ESP32-C3 may return NAN
-      telemetry.addTemperature(TELEM_CHANNEL_SELF, (float)board.getMCUTemperature()); // Built-in MCU Temperature
+      telemetry.addTemperature(TELEM_CHANNEL_SELF, temperature); // Built-in MCU Temperature
     }
 
     // query other sensors -- target specific


examples/simple_repeater/MyMesh.cpp
```diff
@@ -173,6 +173,12 @@ int MyMesh::handleRequest(ClientInfo *sender, uint32_t sender_timestamp, uint8_t
 
     telemetry.reset();
     telemetry.addVoltage(TELEM_CHANNEL_SELF, (float)board.getBattMilliVolts() / 1000.0f);
+
+    float temperature = (float)board.getMCUTemperature();
+    if(!isnan(temperature)) { // Supported boards with built-in temperature sensor. ESP32-C3 may return NAN
+      telemetry.addTemperature(TELEM_CHANNEL_SELF, (float)board.getMCUTemperature()); // Built-in MCU Temperature
```
liamcottle: Looks like you're calling `board.getMCUTemperature()` twice, instead of using the `temperature` variable that already did a read? Since those methods also return `float` type, probably don't need the type casts?

liamcottle (changes requested):

examples/simple_repeater/MyMesh.cpp
IoTThinks: Ooops, my bad.

IoTThinks: @liamcottle I have fixed it. 
Retested ok with RAK4631 and Heltec v3.
https://github.com/meshcore-dev/MeshCore/pull/1206/commits/2deb9cf1440a658c0b51cb824c6582eed7fb87f1

Thanks a lot.

liamcottle:
Looks good thanks!

My last comment is on the behaviour of boards that have an external temperature sensor installed as well.

We would now be using `TELEM_CHANNEL_SELF` for both the internal mcu temperature sensor, and any external temperature sensors.

https://github.com/meshcore-dev/MeshCore/blob/2deb9cf1440a658c0b51cb824c6582eed7fb87f1/examples/simple_repeater/MyMesh.cpp#L179C32-L179C50

https://github.com/meshcore-dev/MeshCore/blob/922e378be5df103d8e93c21e1ed24f99dd73ff86/src/helpers/sensors/EnvironmentSensorManager.cpp#L362

Wondering if this will cause any issues, and if so, which one takes priority unless we move external sensors to a separate channel.

IoTThinks:
I suggest to move the telemetry of external sensors to Channel 2.

That would be clean and clear.

What do you think?

IoTThinks:
From some friends, the later Temperature will override and take priority.

IoTThinks:
@liamcottle I managed to add a BME680 to a Heltec repeater.

This is from built-in sensor.
image

This is when we add BME680.
Yes, because we add the temperature of MCU before BME680.
The temperature of BME680 will override the temperature of MCU.
image

So the PR is quite safe.

liamcottle:
Great thanks, I'm happy with the default MCU temp sensor being overriden by the external sensors for now.

We can look to move external sensors to a separate channel in a future PR if that is preferred.

I like the idea of external sensors on a separate channel than "self", but to keep this PR minimal without a breaking change for existing users, let keep it as is.

Looks good to me :)

fizzyfuzzle:
@IoTThinks I just flashed the nightly (18-Dec-2025 02:08) to a WisMesh Tag and a t114, but I can not see the temperature in the telemetrics on the mobile app? Is there something additional I need to do to enable this feature? Thanks

Edit:
Ah I see this is only available in the Repeater Telemetrics, is there a reason for this? Would also be nice to have it in the companions I guess?

IoTThinks:
@fizzyfuzzle Yah, this PR is for repeaters only.

I push one by one to limit the impacts (if any).

Since it is merged, I may create another PR to push to companions and room servers.

Thanks for your suggestion.

IoTThinks:
Hi @liamcottle: This PR needs a fix before going to main.
- To fix the default temperature to be overridden by external sensors (if any)
- To get and average the temperature so it is more accurate, especially in low temperature
https://github.com/meshcore-dev/MeshCore/pull/1274

IoTThinks:
@fizzyfuzzle I tested on ESP32 companions and rooms.
For rooms, the temperature is OK. 
=> I may enable this PR for rooms.

For ESP32 companions, the temperature is shown quite high ~50+*C with room temperature is ~30*C. I guess it is due to no powersaving and high power consumption from BLE.
=> I need to do further test.

fizzyfuzzle:
@IoTThinks Thanks for testing, I tested it myself on the t114 (nrf52) companion, which seems to be giving the right temperature.
Guess the ESP32 might indeed be running hot because of the higher power consumption, which make sense :+1:

MikesAllotment:
I only just found this (thanks Recof) so I'm a bit late to the party ... I can't seem to get this to work with the Sensor firmware, which is kinda ironic.  Looks like it was implemented for Repeater and RoomServer - was it also implemented for Sensor firmware?

PR #1208, opened 2025-12-12
Add sensor support to Heltec t114

ngavars:
This PR adds a second I2C interface (Wire1) and uses EnvironmentSensorManager in t114 variant. This automatically adds support for all the sensors that EnvironmentSensorManager supports.
The Wire1 (and thus sensors) is currently enabled only for Heltec_t114_repeater and Heltec_t114_without_display_repeater targets.
Tested with t114 repeater (no screen, no gps). All the t114 build targets are building ok.

### DIFF
--- a/variants/t1000-e/T1000eBoard.h
+++ b/variants/t1000-e/T1000eBoard.h
@@ -34,7 +34,7 @@ public:
   }
 
   const char* getManufacturerName() const override {
-    return "Seeed Tracker T1000-e";
+    return "Seeed Tracker T1000-E";
   }
 
   int buttonStateChanged() {


ngavars:
Tested on t114 with GPS too - it's working.

IoTThinks:
Are you able to add sensor support to Heltec T114 with display?
For Heltec V3, I have issue when trying to support sensors on Wire1 while still keeping Wire for OLED.

Whenever, Wire1 is active, Wire instance seems clear / resets and can not be used.

ngavars:
> Are you able to add sensor support to Heltec T114 with display? For Heltec V3, I have issue when trying to support sensors on Wire1 while still keeping Wire for OLED.
> 
> Whenever, Wire1 is active, Wire instance seems clear / resets and can not be used.

My change adds EnvironmentSensorManager as a sensor manager class. It will only activate Wire1 and scan for sensors on boards without display because those boards are often used as repeaters. Repeaters often use sensors (current, temperature, humidity etc) and I was surprised that t114 implementation did not support any.
Heltec t114' s with display on the other hand are mostly used as companions and do not require any sensors (except for the GPS, of course), hence they will not jump though extra hoops, trying to init sensors. But this does not mean that they wouldn't work if pins are defined.

Regarding V3: if I remember correctly, it already has sensor support. Pins 17 and 18 are tiny and located in the middle of the board. Has something changed in this regard?

IoTThinks:
Ok, thanks a lot for sharing.
There is a bug when both Wire and Wire1 are active. Let me figure it out.

I use GPIO41 and 42 for Wire1 instead of the tiny GPIO17 and 18.

ngavars:
Rebased, all good.

ngavars:
Rebased

ngavars:
Rebased

PR #1219, opened 2025-12-14
add washtastic variant fix its ADC

mavericm1:
fixing washtastic's ADC and creating a variant.

Not sure if there is a better way to do this but tested on one of my boards and voltage is accurate now.


mavericm1:
didn't realize there is a cli command for this

PR #1220, opened 2025-12-14
Add admin packet priority for remote repeater management during congestion

ViezeVingertjes:
Prioritizes packets from ACL admins during network congestion to enable reliable remote repeater management.

### Changes
- Added virtual `getForwardPriority()` method to Mesh class for priority customization
- Repeater checks sender against ACL and assigns priority 0 to admin packets (REQ, RESPONSE, TXT_MSG, PATH types)

### Technical Details
Admin packets now get the same priority as direct-routed traffic (highest forwarding priority). The check happens during routing by comparing packet src_hash against ACL entries. Non-admin packets use default priority (unchanged).


ripplebiz:
This one doesn't really make sense, I'm afraid.

PR #1221, opened 2025-12-15
Refactor the saveIdentity method 

zjs81:
This change will allow uses to use these two new commands: 

regen key
Auto mode generates a new identity avoiding collisions with neighbors and reserved hashes (00, FF).
`regen key`
Response: OK - new ID: A3 (reboot to apply)

regen key XX
Vanity mode generates a new identity with a specific hash prefix (2 hex characters).
```
regen key A3
regen key 7F
regen key b2
```

Response: OK - new ID: A3 (reboot to apply)
If the target hash collides with a neighbor:
Response: WARNING: collision! new ID: A3 (reboot to apply)

The purpose of this is to reduce those annoying collisions that happen in the app and also to resolve times where repeaters with the same hash end up repeating a message that was intended for another repeater thus increasing airtime.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -882,8 +882,8 @@ void MyMesh::formatPacketStatsReply(char *reply) {
                                        getNumRecvFlood(), getNumRecvDirect());
 }
 
-void MyMesh::saveIdentity(const mesh::LocalIdentity &new_id) {
-  self_id = new_id;
+void MyMesh::saveIdentity(const mesh::LocalIdentity &new_id, bool apply_now) {
+  if (apply_now) self_id = new_id;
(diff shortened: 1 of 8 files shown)


zjs81:
https://github.com/user-attachments/assets/86a12ff4-4632-4240-961f-4fea8bd0c0ca

Example usage

oltaco:
Looks pretty cool!

You will need to regenerate the shared_secret for any ACL list entries though.

fschrempf:
Can you please use a proper formatting for your commit messages?

```
<subject Line>
<empty Line>
<multiline description, wrapped to approx. 70 chars>
```

That would help a lot to keep the git history readable. Thanks!

zjs81:
Thanks guys. @oltaco  I will take a look at the ACL list.

PR #1230, opened 2025-12-15
Removed halt and replaced it with a delay and reboot

wel97459:
Replaced the halt function with a delay and reboot, instead of an infinite loop. This removes an edge case were there's a bad brown out and the radio fails to initialize properly.

As @liamcottle suggested I have added an error message that is customizable to other failure cases. I also added debug output for nodes without a screen attached.

### DIFF
--- a/examples/simple_secure_chat/main.cpp
+++ b/examples/simple_secure_chat/main.cpp
@@ -550,8 +550,18 @@ StdRNG fast_rng;
 SimpleMeshTables tables;
 MyMesh the_mesh(radio_driver, fast_rng, rtc_clock, tables);
 
-void halt() {
-  delay(10000);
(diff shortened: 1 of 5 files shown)


examples/simple_secure_chat/main.cpp
```diff
@@ -550,16 +550,28 @@ StdRNG fast_rng;
 SimpleMeshTables tables;
 MyMesh the_mesh(radio_driver, fast_rng, rtc_clock, tables);
 
-void halt() {
-  while (1) ;
+void delayedReboot(const char *msg, const uint32_t delayms) {
```
446564: since these are the same in all roles, perhaps they can be moved to the MainBoard class?

liamcottle:
Closing due to old age and merge conflicts.

PR #1236, opened 2025-12-16
Text battery display option

Onefabis:
If user will define in build flags -D AUTO_SHUTDOWN_MILLIVOLTS=3100 and -D TEXT_BATTERY it will show battery level on the companion node as a text. Without AUTO_SHUTDOWN_MILLIVOLTS it will use default 3.0V as 0%. Without TEXT_BATTERY defined in platformio.ini it will draw default icon.


OverseerHSof (approved):
Thank you very much for this change

PR #1238, opened 2025-12-17
NRF52 Repeater Powersaving

fschrempf:
In order to reduce the power consumption of NRF52 repeaters, this implements CPU idling by suspending the main task that runs the Arduino `loop()`.

During the idle interval the CLI and the UI (if available) will be unresponsive. The RF module will be kept in RX mode and upon receiving a packet, the interrupt will cause the processing loop to continue immediately before going back to idle after all outgoing packets have been transferred.

On a RAK4631 repeater this can reduce the power consumption during RX mode from around 12 mA to around 7.5 mA.

Latest build artifacts from CI (updated 01/13/26): https://github.com/fschrempf/MeshCore/actions/runs/20971258478/artifacts/5118537314

Feedback, tests, reviews and questions welcome!

### DIFF
--- a/src/helpers/NRF52Board.cpp
+++ b/src/helpers/NRF52Board.cpp
@@ -19,6 +19,18 @@ static void disconnect_callback(uint16_t conn_handle, uint8_t reason) {
 
 void NRF52Board::begin() {
   startup_reason = BD_STARTUP_NORMAL;
+  rx_wait_sem = xSemaphoreCreateBinary();
+  xSemaphoreGive(rx_wait_sem);
+  xSemaphoreTake(rx_wait_sem, portMAX_DELAY);
+}
+
+void NRF52Board::sleep(uint32_t secs) {
+  MESH_DEBUG_PRINTLN("Suspending loop for powersaving, set wakeup timer in %u seconds", secs);
+  xSemaphoreTake(rx_wait_sem, pdMS_TO_TICKS(secs * 1000));
+}
+
+void NRF52Board::onRXInterrupt() {
+  xSemaphoreGive(rx_wait_sem);
 }
 
 void NRF52BoardDCDC::begin() {
--- a/src/helpers/NRF52Board.h
+++ b/src/helpers/NRF52Board.h
@@ -6,6 +6,9 @@
 #if defined(NRF52_PLATFORM)
 
 class NRF52Board : public mesh::MainBoard {
+private:
+  SemaphoreHandle_t rx_wait_sem;
+
 protected:
   uint8_t startup_reason;
 
@@ -14,6 +17,8 @@ public:
   virtual uint8_t getStartupReason() const override { return startup_reason; }
   virtual float getMCUTemperature() override;
   virtual void reboot() override { NVIC_SystemReset(); }
+  virtual void sleep(uint32_t secs) override;
+  virtual void onRXInterrupt() override;
 };
 
 /*


IoTThinks:
I intended to do the same before.
But it is quite hard to power down NRF52 and be waken up by RX events. The power consumption will be around 6mA.
NRF52 does not handle Rising High well.
Your PR looks complex. May be due to this issue.

So I ended up just do some power saving trick to keep the power down to 8.5mA with few code changes only.
The MCU is still running.

Hope we can push the power down more for NRF52

fschrempf:
> But it is quite hard to power down NRF52 and be waken up by RX events.

Yes, there is no real sleep mode for NRF52 so either put the CPU in idle to save power or shut it down completely. The latter requires full reinit at wakeup.

> The power consumption will be around 6mA.

In which case? With the MCU shut down and only the RF module running in RX mode? That would be in that ballpark, yes.

>  NRF52 does not handle Rising High well.

Sorry, I don't get what you mean here.

>  Your PR looks complex. May be due to this issue.

Actually it doesn't look complex to me at all. It's pretty straight forward. Instead of letting the CPU run continuously, it stops it and resumes it after the idle interval is over or a packet is received. What exactly looks complex to you?

> So I ended up just do some power saving trick to keep the power down to 8.5mA with few code changes only.

What "trick" would that be? Do you mean `waitForEvent()` [here](https://github.com/IoTThinks/MeshCore/commit/2c878edc45ac0dad6806811ac6e721c6b4c8cd33#diff-04e0d6ad57930580cf33631bfb7532c478ac1ab20b0d41ec0d319489cf207aedR100-R104)? It doesn't work for me. I'm still seeing around 14 mA @ 3.3V with your PowerSaving07 branch.

And if it would work, it would do the same thing: put the CPU in idle, right? My approach does it in a platform-agnostic way, which is better IMHO.

ngavars:
This feature has been long overdue. I tested it a bit with Heltec t114 (no screen) repeater. With stock FW I am seeing around 12 mA idle current on my cheap usb power meter. It is actually quite good - it used to be around 17 mA with stock FW not so long ago.

Then I flashed FW from this branch and immediately went for  "set idle.interval 10000". Now my cheapo usb meter shows current at 0 mA, which occasionally jumps briefly to 10 mA while idling. If I send a message from my companion then the repeater wakes, repeats and then goes back to idling. Now, the 0mA is probably just an artifact of my usb meter. I will try some additional tests tomorrow, but what I see so far - the power consumption at idle is noticeably lower and repeater still seems to be repeating messages and responding to admin commands etc

 What is the intended reasonable idle interval? Something like 1 to 3 seconds?

IoTThinks:
May be I will add the CLI for esp32 based boards.

Should we share the same CLI to enable/disable power saving to both esp32 and nrf52?

Like set powersave 1?

In your PR, you set the idle period? How about wake up period?

IoTThinks:
@ngavars You have to measure the current at the battery cable by power meter.

A usbc, it needs to turn on unneccessary components like uart chip, led...

fschrempf:
> May be I will add the CLI for esp32 based boards.
> Should we share the same CLI to enable/disable power saving to both esp32 and nrf52?

In my opinion we should use the same approach for ESP32 and NRF52 altogether. You should be able to take my generic implementation and simply add the `board.sleep()` implementation for ESP32 to be called before the loop is halted.

> Like set powersave 1?

What would that be good for? A single parameter that sets the length of the sleep/idle interval is enough. If set to zero (default) there is no change compared to the current implementation in the main branch. Repeater admins can then decide if they want to save power by increasing the interval.

> In your PR, you set the idle period? How about wake up period?

The idle interval is the time the CPU is idling/sleeping. The wake interval is hardcoded to five seconds or three minutes after reboot or after CLI activity. I don't think this needs to be a parameter for the user to change.

fschrempf:
> What is the intended reasonable idle interval? Something like 1 to 3 seconds?

The `idle.interval` setting is in seconds, so 10000 will give you around 2.8 hours. I'm currently running my test repeater with 1800 (30 minutes), but I'm not yet sure what would be a good value. If you have the value set very high, the automatic adverts of the repeater might be delayed accordingly (until the next wakeup happens). Apart from that I currently don't see any other downsides.

fschrempf:
@ngavars Thanks for testing by the way!

examples/simple_repeater/main.cpp
```diff
@@ -117,4 +126,11 @@ void loop() {
   ui_task.loop();
 #endif
   rtc_clock.tick();
+
+  if (the_mesh.getIdleInterval() && (millis() - active_timestamp > active_time_ms) &&
+      !the_mesh.getTotalOutboundCount()) {
```
mtlynch: Suggest reformatting this so that each clause is on its own line. Also, I think it's helpful practice to make operator precedence explicit rather than rely on the reader to remember C++ precedence rules, so suggest:

```cpp
if (the_mesh.getIdleInterval() &&
      ((millis() - active_timestamp) > active_time_ms) &&
      !the_mesh.getTotalOutboundCount()) {
    radio_driver.blockTaskUntilRXEvent(the_mesh.getIdleInterval() * 1000);
    ...
  }
```

examples/simple_repeater/main.cpp
```diff
@@ -19,6 +19,12 @@ void halt() {
 
 static char command[160];
 
+#define ACTIVE_TIME_MS_INUSE 3 * 60 * 1000; // 3 minutes
+#define ACTIVE_TIME_MS_IDLE  5 * 1000;      // 5 seconds
```
mtlynch: Is there a reason to use `#define` here instead of `constexpr`? We get more compile-time checks with `constexpr`.

```cpp
constexpr unsigned long ACTIVE_TIME_MS_INUSE = 3 * 60 * 1000; // 3 minutes
constexpr unsigned long ACTIVE_TIME_MS_IDLE  = 5 * 1000;      // 5 seconds
```

If we do use `#define`, we need to remove the trailing semicolon or it becomes part of the macro, which is not what we want (i.e., `unsigned long active_time_ms = ACTIVE_TIME_MS_INUSE;` expands to `unsigned long active_time_ms = 3 * 60 * 1000;;` with two trailing semicolons.

src/helpers/CommonCLI.h
```diff
@@ -48,6 +48,7 @@ struct NodePrefs { // persisted to file
   uint8_t advert_loc_policy;
   uint32_t discovery_mod_timestamp;
   float adc_multiplier;
+  uint32_t idle_interval; // in seconds
```
mtlynch: Could we make this `idle_interval_seconds` and not need the comment?

src/Dispatcher.h
```diff
@@ -167,6 +167,9 @@ class Dispatcher {
   Packet* obtainNewPacket();
   void releasePacket(Packet* packet);
   void sendPacket(Packet* packet, uint8_t priority, uint32_t delay_millis=0);
+  uint32_t getTotalOutboundCount() {
```
mtlynch: This is exposing more data than the API client currently needs. Could it simplify to `bool hasOutboundPackets() { ... }`

That simplifies the callsite to:

```c++
if (the_mesh.getIdleInterval() &&
      ((millis() - active_timestamp) > active_time_ms) &&
      the_mesh.hasOutboundPacket()) {
    radio_driver.blockTaskUntilRXEvent(the_mesh.getIdleInterval() * 1000);
    ...
  }
```

It's easier to reason about negating a `bool` than negating a `uint32_t`.

examples/simple_repeater/main.cpp
fschrempf: Thanks, I will change that. Actually I thought about it and then forgot. And semicolons are bogus of course.

src/Dispatcher.h
fschrempf: Good idea, yes.

examples/simple_repeater/main.cpp
fschrempf: I will try to improve the readibilty. Though this is what clang-format suggested using the project's configuration.

src/helpers/CommonCLI.h
fschrempf: We could and I thought about it. In the end I decided to follow the style of the existing variables.

fschrempf:
> I'm just a user sharing my thoughts and have no authority in this project.

Thanks for your review anyway! Very much appreciated!

src/Dispatcher.h
fschrempf: Now changed as proposed.

src/helpers/CommonCLI.h
fschrempf: I will leave that as is for now but happy to change it if others are in favor, too.

examples/simple_repeater/main.cpp
fschrempf: Now changed/fixed

examples/simple_repeater/main.cpp
fschrempf: Added the additional parentheses, but left the line breaks as clang-format suggested.

fschrempf:
I've had this running on my test repeater (RAK4630) for 3 days now and I can't see any negative or unexpected results. I currently do not have any ESP32 hardware to test, so if someone would be willing to help out that would be very much appreciated! :smiley:

SaschaKt:
> I've had this running on my test repeater (RAK4630) for 3 days now and I can't see any negative or unexpected results. I currently do not have any ESP32 hardware to test, so if someone would be willing to help out that would be very much appreciated! 😃

Can you attach/upload somewhere an bin file for v3 for testing?

fschrempf:
@SaschaKt You can find an archive with all repeater firmwares built by the GitHub CI here: https://github.com/fschrempf/MeshCore/actions/runs/20347701758/artifacts/4916433935

SaschaKt:
> @SaschaKt You can find an archive with all repeater firmwares built by the GitHub CI here: https://github.com/fschrempf/MeshCore/actions/runs/20347701758/artifacts/4916433935

I flashed it. Sorry for feedback after a few seconds. for v3 the powersaving is not so efficient like Iotthinks solution. I'm measuring over Usb-C and the consumption goes down from 48mA to 39mA with your solution, with Iotthinks solution it goes down to 13mA measured over Usb-C. Measuring on battery input pins will be a few mA less. I don't think that the combination of both solutions will give more power saving because I think that Lightsleep at esp's  with iotthink solution also reduce CPU activity

fschrempf:
@SaschaKt Thanks for testing! This is the expected behavior. This PR does not (yet) include the code for ESP32 sleep. And yes, the combination of both solutions won't give more power savings for ESP32 than https://github.com/meshcore-dev/MeshCore/pull/1107.

The reason for my PR is that it provides a generic approach that is also applicable for other platforms than ESP32 while still providing the possibility to be extended by platform-specific sleep.

If your test shows a slightly reduced power consumption and the repeater still works fine that's a success.

SaschaKt:
> @SaschaKt Thanks for testing! This is the expected behavior. This PR does not (yet) include the code for ESP32 sleep. And yes, the combination of both solutions won't give more power savings for ESP32 than #1107.
> 
> The reason for my PR is that it provides a generic approach that is also applicable for other platforms than ESP32 while still providing the possibility to be extended by platform-specific sleep.
> 
> If your test shows a slightly reduced power consumption and the repeater still works fine that's a success.

It works, I could go normal to cli and start OTA and flash with Iotthink variant again. Now again 13mA. 39mA is still too much. nrF have per default enabled power saving internally and are at the small consumption level, with tweaking maybe a few mA savings. But esp's needs an enabled lightsleep to be useful as a repeater. The advantage of an nrF is that still with enabled BT the power consumption is not noticeable higher then without. Where is the problem to have two power routines, one for esp's and one for NRFs? Also Iotthink made powersavings for nrF too which will reduce from 12.5 to 8.5mA, that's about 32%

fschrempf:
> Where is the problem to have two power routines, one for esp's and one for NRFs

@SaschaKt I think we are talking past each other. There is no problem here. What I want to achieve is a two step solution:

1. A platform-independent solution (with some power savings) with idling the CPU (without any sleep) that works in all cases and also for any future boards and platforms.
2. Additionally platform-specific sleep functions (with more power savings) if available.

Apart from the differences mentioned in https://github.com/meshcore-dev/MeshCore/pull/1107#pullrequestreview-3603540289 this is purely a strategic difference from what https://github.com/meshcore-dev/MeshCore/pull/1107 does.

SaschaKt:
> > Where is the problem to have two power routines, one for esp's and one for NRFs
> 
> @SaschaKt I think we are talking past each other. There is no problem here. What I want to achieve is a two step solution:
> 
> 1. A platform-independent solution (with some power savings) with idling the CPU (without any sleep) that works in all cases and also for any future boards and platforms.
> 2. Additionally platform-specific sleep functions (with more power savings) if available.
> 
> Apart from the differences mentioned in [#1107 (review)](https://github.com/meshcore-dev/MeshCore/pull/1107#pullrequestreview-3603540289) this is purely a strategic difference from what #1107 does.

You're right. I can confirm that with v3 the powerconsumption goes from 48mA to 39mA. That's an improvement..I hope that it would not interfere with hardware specific power saving ontop like iotthink has made..this has to be tested

ngavars:
I did some additional testing with power profiler and Heltec V3 board on stock firmware and on your firmware. Both measurements were captured after the initial boot/advert sequence, when the board starts idling. Heltec was powered through its battery connector with voltage set at 3307 mV. 

With stock firmware the avg current is 45 mA.
With power option firmware the avg current is also around 45 mA until approx. 3 minute mark, when it drops down to around  35 mA.

With stock firmware (latest from web flasher):
image

With power option firmware, idle.interval=1800:
image

Hope this helps. I can also test it with Promicro board if you want.

examples/simple_repeater/MyMesh.cpp
```diff
@@ -722,6 +722,7 @@ MyMesh::MyMesh(mesh::MainBoard &board, mesh::Radio &radio, mesh::MillisecondCloc
   _prefs.advert_loc_policy = ADVERT_LOC_PREFS;
 
   _prefs.adc_multiplier = 0.0f; // 0.0f means use default board multiplier
+  _prefs.idle_interval = 0;
```
4np: Perhaps add a comment to better explain what this means:

```suggestion
  // CPU Idle Interval (in seconds).
  //
  // If enabled (> 0), the CPU will start idling if there has not been any radio activity for
  // the specified number of seconds.
  //
  // Note: When the CPU is idling, remote management and / or UI will cease to function!
  _prefs.idle_interval = 0;
```

Or shorter:

```suggestion
  _prefs.idle_interval = 0; // CPU will idle when no RF activity during this interval (in sec, 0 = disabled)
```

4np (commented):
Thanks for working on this @fschrempf  👍  It would be good to see some more power saving for my Solar RAK19003 node :)

Having said that, I wonder if if would be better if the idle interval would be computed automatically rather than using a fixed duration. For example, in The Netherlands the mesh has grown exponentially. Where 2 months ago the mesh saw some activity, now I see hundreds of daily messages (let alone adverts) and (some parts of) Belgium and Germany have been connected. I wouldn't be surprised to see more of Germany and France starting to appear. An idle interval that worked a couple of weeks ago (maybe even days ago) may not work as well today.

Additionally, if CPU idling is triggered, how would one be able to enter remote management mode? IMHO when trying to use remote management or fetching telemetry, idling should be cancelled and the idle timer should be reset.

ngavars:
> Additionally, if CPU idling is triggered, how would one be able to enter remote management mode? IMHO when trying to use remote management or fetching telemetry, idling should be cancelled and the idle timer should be reset.

It will wake up on any RX event. Remote management works nicely, at least in all my tests so far.

ngavars:
I did some more testing with Xiao NRF52840. Here's a comparison of stock firmware (dev branch) and power option firmware (idle.interval=1800). In both cases the board is idling for 1 minute and there are no radio events.

Stock FW
image

idle.interval=1800
image

fschrempf:
> An idle interval that worked a couple of weeks ago (maybe even days ago) may not work as well today.

@4np How is the idle interval related to the mesh activity? During idle the repeater will still respond to incoming packets. Only outgoing packets will be left waiting until the timeout expires.

fschrempf:
> I did some more testing with Xiao NRF52840.

Thanks for the additional tests. This corresponds pretty nicely to the results of my own tests on RAK4631.

4np:
> @4np How is the idle interval related to the mesh activity? During idle the repeater will still respond to incoming packets.  Only outgoing packets will be left waiting until the timeout expires.

@fschrempf , what I meant is an idle interval that works today, may not work tomorrow. If I read the PR, and correct me if I'm wrong, in your case you set the idle interval to 30 mins. So if there's no rx during a 30 min window, the CPU will start idling. 

To illustrate the issue, this summer I didn't hear anything at all on MeshCore and there were almost no repeaters or companions. In September / October, we had a small regional mesh with several messages per hour. Today, we have [most of The Netherlands](https://map.meshcore.dev/?zoom=9&lat=52.1141&lon=5.3284) covered with the one of the most dense meshes in [Europe](https://map.meshcore.dev/?zoom=4&lat=50.2050&lon=12.5684) and I receive [several packets _per second_](https://analyzer.letsmesh.net/packets?region=netherlands).

So an idle timeout of, in your case, 30 mins that in my case would have been valid in September or summer would never trigger today.

As such IMHO it would be better to have an algorithm compute the best idle interval and have it adjust automatically on changing conditions. New repeaters pop up every single day.

ps. I think ill configured repeaters doing flood adverts at short intervals are the primary cause of the many packets per second, and putting a strain on the mesh. Reducing the number of configuration options could be beneficial as well.

SaschaKt:
> > @4np How is the idle interval related to the mesh activity? During idle the repeater will still respond to incoming packets.  Only outgoing packets will be left waiting until the timeout expires.
> 
> @fschrempf , what I meant is an idle interval that works today, may not work tomorrow. If I read the PR, and correct me if I'm wrong, in your case you set the idle interval to 30 mins. So if there's no rx during a 30 min window, the CPU will start idling.
> 
> To illustrate the issue, this summer I didn't hear anything at all on MeshCore and there were almost no repeaters or companions. In September / October, we had a small regional mesh with several messages per hour. Today, we have [most of The Netherlands](https://map.meshcore.dev/?zoom=9&lat=52.1141&lon=5.3284) covered with the one of the most dense meshes in [Europe](https://map.meshcore.dev/?zoom=4&lat=50.2050&lon=12.5684) and I receive [several packets _per second_](https://analyzer.letsmesh.net/packets?region=netherlands).
> 
> So an idle timeout of, in your case, 30 mins that in my case would have been valid in September or summer would never trigger today.
> 
> As such IMHO it would be better to have an algorithm compute the best idle interval and have it adjust automatically on changing conditions. New repeaters pop up every single day.
> 
> ps. I think ill configured repeaters doing flood adverts at short intervals are the primary cause of the many packets per second, and putting a strain on the mesh. Reducing the number of configuration options could be beneficial as well.

Idle time is how long the CPU should be idle and then wake up for doing stuff, or it is waken before throw an rx. After a Rx/tx I have measured on my own a fast falloff of the current and on Rx a short time normal state, repeat, idle again.

fschrempf:
@4np I think you misread the PR. The idle interval just determines how long the CPU will be in idle and therefore no processing of outgoing packets, serial user input or UI takes place. This interval can be interrupted at any time by incoming packets.

The time after which the idling starts is hardcoded to 3 minutes after startup and 5 seconds during normal operation.

So the worst that will happen in a busy mesh is that it will never go to idle because there are always incoming packets within the 5 second interval. But it will not affect the performance of the repeater or anything.

4np:
Thank you for the clarification @fschrempf  and @SaschaKt  👍

fschrempf:
With https://github.com/meshcore-dev/MeshCore/pull/1266 being merged in dev, this has now been reworked to work on top of this. I will let the new version run on my RAK4630 test repeater. Here is a [link to the repeater firmware archive](https://github.com/fschrempf/MeshCore/actions/runs/20656799231/artifacts/5007854971) from the CI for any of you who want to do another test themselves.

As a reminder, the new version requires powersaving to be turned on via CLI command `powersaving on`.

IoTThinks:
The PR looks great. 
I will do the test for this PR for my RAK4631 and T-Echo Lite next week.

Next week, I will push a PR to enable powersaving for ALL ESP32-based board too.
To do this, I also add 1 similar line for ESP32 in setFlag like you.
I guess the moderators will review that line in setFlag very carefully.

Hope we can have PowerSaving for both NRF52 and ESP32 repeaters soon.

4np:
> Here is a [link to the repeater firmware archive](https://github.com/fschrempf/MeshCore/actions/runs/20656799231/artifacts/5007854971) from the CI for any of you who want to do another test themselves.

It's running on my spare RAK based repeater and so far all seems to be working well.

fschrempf:
@4np Thanks for testing. On my repeater it seems like the node becomes unresponsive after some time. It might be unrelated, but I need to do some debugging.

4np:
@fschrempf this morning the repeater has become unresponsive for me as well, so there does appear to be some genuine issue.

It looks like on nRF52 platforms a timer callback runs in the interrupt/scheduler context, not in the main loop.

Another concern could be that [resumeLoop()](https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/cores/nRF5/main.cpp#L106) from setFlag() may be executing from a different context as well (`setFlag()` in RadioLib wrappers is commonly called from a DIO interrupt / radio IRQ path), although it looks like Adafruit's `resumeLoop()` handles that so that should not be a concern:
image

Also, the wake by RX logic does not cancel the timer so it keeps on running.

Lastly, it may be preferable to safeguard against invalid interval values. Maybe clamp them between a min and a max value (`max(min(maximum_number_of_secs, sec), minimum_number_of_seconds))`).

src/helpers/NRF52Board.cpp
```diff
@@ -17,10 +18,20 @@ static void disconnect_callback(uint16_t conn_handle, uint8_t reason) {
...
 
+static void sleep_timer_callback(TimerHandle_t xTimerID) {
+  resumeLoop();
+}
+
 void NRF52Board::begin() {
   startup_reason = BD_STARTUP_NORMAL;
 }
 
+void NRF52Board::sleep(uint32_t secs) {
```
4np, line 27: Psuedo code (I am not sure about the PIN numbers):

```C
// NRF52Board.cpp (file scope)

static SoftwareTimer sleep_timer;
static bool sleep_timer_inited = false;
static volatile bool sleeping = false;

static void sleep_timer_callback(TimerHandle_t) {
  sleeping = false;
  resumeLoop();
}

void NRF52Board::sleep(uint32_t secs) {
  if (secs == 0) return;

  // Convert seconds->ms with overflow clamp
  uint64_t ms64 = (uint64_t)secs * 1000ULL;
  uint32_t ms = (ms64 > 0xFFFFFFFFULL) ? 0xFFFFFFFFUL : (uint32_t)ms64;

  // Helper: detect SX1262 DIO1 asserted (pending/latched IRQ).
  auto dio1_high = []() -> bool {
#if defined(PIN_LORA_DIO1)
    return digitalRead(PIN_LORA_DIO1) == HIGH;
#elif defined(LORA_DIO1)
    return digitalRead(LORA_DIO1) == HIGH;
#else
    return false; // if unknown, we can't gate (better than breaking build)
#endif
  };

  // SX1262/RAK4630: if DIO1 is HIGH (IRQ latched), do NOT suspendLoop() or you can deadlock.
  if (dio1_high()) return;

  sleeping = true;

  // Cancel any previous schedule to avoid stale wake callbacks.
  sleep_timer.stop();

  // Initialize + (re)configure timer
  if (!sleep_timer_inited) {
    sleep_timer_inited = true;
  }
  sleep_timer.begin(ms, sleep_timer_callback, nullptr, /*repeating=*/false);
  sleep_timer.start();

  // Optional extra safety: re-check after arming timer to avoid a tiny race window
  // where an IRQ asserts DIO1 between the first check and suspendLoop().
  if (dio1_high()) {
    sleeping = false;
    sleep_timer.stop();   // don't leave a stray timer armed
    return;
  }

  suspendLoop();
}

void NRF52Board::wakeFromInterrupt() {
  // Optional micro-optimization: if we're not sleeping, no need to stop the timer.
  // But still safe to call resumeLoop() unconditionally.
  if (sleeping) {
    sleeping = false;
    sleep_timer.stop();   // prevent later spurious callback
  }
  resumeLoop();
}
```

So this should likely fix the unresponsive repeater issue:

```c
  // SX1262/RAK4630: if DIO1 is HIGH (IRQ latched), do NOT suspendLoop() or you can deadlock.
  if (dio1_high()) return;
```

4np (commented):
As the radio appears to become unresponsive after a period of low RX (overnight, at least in my case) it may be that the timer wasn’t actually armed (create/start/reset failed or got wedged), but `suspendLoop()` is still getting called. With no RX there’s no other wake source and the repeaters stays suspended forever. 

The question is why it does not wake by RX. It could be that the RX interrupt isn’t firing at all?

fschrempf:
> As the radio appears to become unresponsive after a period of low RX (overnight, at least in my case) it may be that the timer wasn’t actually armed (create/start/reset failed or got wedged), but suspendLoop() is still getting called. With no RX there’s no other wake source and the repeaters stays suspended forever.

In my case, I think I saw it come back to life when trying a bit later. That makes me think the timer wakeup works correctly but the RX wakeup fails. But I'm not sure at all.

> However, it also does not wake on RX so it may seem to point to the RX interrupt not firing at all? It looks like this may happen when the loop is suspended while the IRQ isn’t serviced/cleared, causing it to stay locked and resulting in the RX interrupt not being executed.

Hm, that could happen if we go to sleep with the interrupt being disabled. But I don't see why the interrupt would be disabled at that point. As far as I know everything runs synchronously and we are suspending the loop task from within the loop. The same flow has been working flawlessly with my previous implementation using a FreeRTOS semaphore to block the loop task (see https://github.com/fschrempf/MeshCore/commit/7e7f091a7daa113cef2c2c3ec93a0b362be9c589) which makes me think there might some other issue here.

4np:
It became unresponsive for me again, during daytime / peak RF. I'm going to test some code changes on my end as well.

4np:
I had another look and IMHO there are a few things that stand out:

- **Activity tracking:** `lastActive` is currently updated in the main loop, but it should reflect when the radio last received an RX packet. Consider exposing a method in `MyMesh` for this (for example `getLastRX() -> millis`).
- **Loop resumption:** In `RadioLibWrappers`, resuming the loop on RX makes sense, but it may be better to unify onRX in a single location (here or in `MyMesh`?)
- **Other considerations:**  
  - Disable sleep during OTA (otherwise BLE flashing will fail).  
  - Postpone sleep on Serial activity (you already have this).
- **Variable names:** Consider renaming for clarity (if found the current variable names a bit counter intuitive):  
  ```cpp
  // Interval of inactivity before sleeping
  constexpr unsigned long POWER_SAVING_TRIGGER_INTERVAL_SEC = 1 * 60; // 1 min
  // Duration of power-saving sleep
  constexpr unsigned long POWER_SAVING_SLEEP_DURATION_SEC = 30 * 60;  // 30 min
  ```
- Probably the 5s work interval can just be removed, pseudo code:
  ```cpp
  if (the_mesh.getNodePrefs()->powersaving_enabled &&
      the_mesh.millisHasNowPassed(the_mesh.getLastRX() + nextPowerSavingIntervalInSec * 1000),
      !the_mesh.hasPendingWork()) { ... }
  ```
- **Suggested approach:** Use an `onRX` event handler to reset a non-activity timer that triggers sleep rather than doing the heavy lifting inside the main loop.  

Thanks again for working on reducing power consumption! I think this will be a great addition, particularly for solar repeaters.

IoTThinks:
> * Probably the 5s work interval can just be removed

True.
However, usually when there is a LoRa packet, another packet is coming, too.
So wakeup for 5s will surely reduce latency if any, especially during tracing.

IoTThinks:
I have merged a minimum piece of this code.
Yes, it can achieve 6mA.

This check "digitalRead(PIN_LORA_DIO1) == HIGH" seems to fix the "stuck" problem.
Else if we keep "ping" the RAK4631 for a while, the board will be stuck at sleep mode.

BTW, we may need to move the definition of PIN_LORA_DIO1... into platform.ini instead.
Let me monitor it for a longer period.

The suspendLoop and resumeLoop are very neat approach.

4np:
> > * Probably the 5s work interval can just be removed
> 
> True. However, usually when there is a LoRa packet, another packet is coming, too. So wakeup for 5s will surely reduce latency if any, especially during tracing.

Yeah that makes sense.

> BTW, we may need to move the definition of PIN_LORA_DIO1... into platform.ini instead.
Let me monitor it for a longer period.

I have some changes where I added a `getLoRaDio1Pin` method, and implemented it in all the NRF52 variants. I ran the changes for a while with these and I don't think it became unresponsive, but maybe I didn't try long enough. Something still wasn't right because it looked like the `MESH_DEBUG_PRINTLN("Resuming work for %lu s", (unsigned long)POWER_SAVING_TRIGGER_INTERVAL_SEC);` in `main.cpp` kept repeating; first 2 lines, then 4 lines, then more lines. Maybe `millis()` in `main.cpp` vs `MyMesh` values are unreliable? Maybe it should use the RTC instead? I didn't have time to investigate further, maybe if I have some spare time today.

> Note: when the board starts sleeping, the `MESH_DEBUG_PRINTLN` that executed right before sleeping aren't flushed to the terminal so you only see them when the board wakes up again.

The core changes I made were these:

**NRF52Board.h**
```h
  // Return the LoRa DIO1 GPIO pin used for radio IRQ/wake on this board.
  // Boards without a known LoRa DIO1 pin should return -1.
  // This is used to gate sleep to avoid deadlocks when SX126x keeps DIO1 asserted.
  virtual int16_t getLoRaDio1Pin() const { return -1; }
```

and I implemented `getLoraDio1Pin()` for all the NRF52 variants (except for `minewsemi_me25ls01` is appears to always be `P_LORA_DIO_1`):

```zsh
$ grep -r "getLoRaDio1Pin" --after-context=2
./variants/wio-tracker-l1/WioTrackerL1Board.h:  int16_t getLoRaDio1Pin() const override {
./variants/wio-tracker-l1/WioTrackerL1Board.h-    return P_LORA_DIO_1;
./variants/wio-tracker-l1/WioTrackerL1Board.h-  }
--
./variants/xiao_nrf52/XiaoNrf52Board.h:  int16_t getLoRaDio1Pin() const override {
./variants/xiao_nrf52/XiaoNrf52Board.h-    return P_LORA_DIO_1;
./variants/xiao_nrf52/XiaoNrf52Board.h-  }
--
./variants/heltec_mesh_solar/MeshSolarBoard.h:  int16_t getLoRaDio1Pin() const override {
./variants/heltec_mesh_solar/MeshSolarBoard.h-    return P_LORA_DIO_1;
./variants/heltec_mesh_solar/MeshSolarBoard.h-  }
--
./variants/wio_wm1110/WioWM1110Board.h:  int16_t getLoRaDio1Pin() const override {
./variants/wio_wm1110/WioWM1110Board.h-    return P_LORA_DIO_1;
./variants/wio_wm1110/WioWM1110Board.h-  }
--
./variants/minewsemi_me25ls01/MinewsemiME25LS01Board.h:  int16_t getLoRaDio1Pin() const override {
./variants/minewsemi_me25ls01/MinewsemiME25LS01Board.h-    return LORA_DIO_1;
./variants/minewsemi_me25ls01/MinewsemiME25LS01Board.h-  }
--
./variants/thinknode_m1/ThinkNodeM1Board.h:  int16_t getLoRaDio1Pin() const override {
./variants/thinknode_m1/ThinkNodeM1Board.h-    return P_LORA_DIO_1;
./variants/thinknode_m1/ThinkNodeM1Board.h-  }
--
./variants/ikoka_nano_nrf/IkokaNanoNRFBoard.h:  int16_t getLoRaDio1Pin() const override {
./variants/ikoka_nano_nrf/IkokaNanoNRFBoard.h-    return P_LORA_DIO_1;
./variants/ikoka_nano_nrf/IkokaNanoNRFBoard.h-  }
--
./variants/promicro/PromicroBoard.h:  int16_t getLoRaDio1Pin() const override {
./variants/promicro/PromicroBoard.h-    return P_LORA_DIO_1;
./variants/promicro/PromicroBoard.h-  }
--
./variants/mesh_pocket/MeshPocket.h:  int16_t getLoRaDio1Pin() const override {
./variants/mesh_pocket/MeshPocket.h-    return P_LORA_DIO_1;
./variants/mesh_pocket/MeshPocket.h-  }
--
./variants/rak_wismesh_tag/RAKWismeshTagBoard.h:  int16_t getLoRaDio1Pin() const override {
./variants/rak_wismesh_tag/RAKWismeshTagBoard.h-    return P_LORA_DIO_1;
./variants/rak_wismesh_tag/RAKWismeshTagBoard.h-  }
--
./variants/ikoka_stick_nrf/IkokaStickNRFBoard.h:  int16_t getLoRaDio1Pin() const override {
./variants/ikoka_stick_nrf/IkokaStickNRFBoard.h-    return P_LORA_DIO_1;
./variants/ikoka_stick_nrf/IkokaStickNRFBoard.h-  }
--
./variants/ikoka_handheld_nrf/IkokaNrf52Board.h:  int16_t getLoRaDio1Pin() const override {
./variants/ikoka_handheld_nrf/IkokaNrf52Board.h-    return P_LORA_DIO_1;
./variants/ikoka_handheld_nrf/IkokaNrf52Board.h-  }
--
./variants/keepteen_lt1/KeepteenLT1Board.h:  int16_t getLoRaDio1Pin() const override {
./variants/keepteen_lt1/KeepteenLT1Board.h-    return P_LORA_DIO_1;
./variants/keepteen_lt1/KeepteenLT1Board.h-  }
--
./variants/lilygo_techo/TechoBoard.h:  int16_t getLoRaDio1Pin() const override {
./variants/lilygo_techo/TechoBoard.h-    return P_LORA_DIO_1;
./variants/lilygo_techo/TechoBoard.h-  }
--
./variants/heltec_t114/T114Board.h:  int16_t getLoRaDio1Pin() const override {
./variants/heltec_t114/T114Board.h-    return P_LORA_DIO_1;
./variants/heltec_t114/T114Board.h-  }
--
./variants/nano_g2_ultra/nano-g2.h:  int16_t getLoRaDio1Pin() const override {
./variants/nano_g2_ultra/nano-g2.h-    return P_LORA_DIO_1;
./variants/nano_g2_ultra/nano-g2.h-  }
--
./variants/lilygo_techo_lite/TechoBoard.h:  int16_t getLoRaDio1Pin() const override {
./variants/lilygo_techo_lite/TechoBoard.h-    return P_LORA_DIO_1;
./variants/lilygo_techo_lite/TechoBoard.h-  }
--
./variants/sensecap_solar/SenseCapSolarBoard.h:  int16_t getLoRaDio1Pin() const override {
./variants/sensecap_solar/SenseCapSolarBoard.h-    return P_LORA_DIO_1;
./variants/sensecap_solar/SenseCapSolarBoard.h-  }
--
./variants/rak4631/RAK4631Board.h:  int16_t getLoRaDio1Pin() const override {
./variants/rak4631/RAK4631Board.h-    return P_LORA_DIO_1; // GPIO 47
./variants/rak4631/RAK4631Board.h-  }
--
./variants/t1000-e/T1000eBoard.h:  int16_t getLoRaDio1Pin() const override {
./variants/t1000-e/T1000eBoard.h-    return P_LORA_DIO_1;
./variants/t1000-e/T1000eBoard.h-  }
```

The rest of the core changes I made:

**NRF52Board.cpp**

```ccp
static bool sleep_timer_inited = false;
static volatile bool sleeping = false;
static bool ota_enabled = false;

...

static void sleep_timer_callback(TimerHandle_t) {
  // Timer callbacks run in the FreeRTOS timer task, not in loop().
  // Only resume the loop if we are actually sleeping.
  if (!sleeping) return;

  MESH_DEBUG_PRINTLN("Waking from sleep by timer callback");
  sleeping = false;
  resumeLoop();
}

void NRF52Board::sleep(uint32_t secs) {
  if (secs == 0) return;

  // Don't sleep when OTA is enabled.
  if (ota_enabled) {
      MESH_DEBUG_PRINTLN("Skip sleeping %lu s (OTA enabled)", (unsigned long)secs);
      return;
  }

  int16_t dio1 = getLoRaDio1Pin();
  if (dio1 < 0) {
    MESH_DEBUG_PRINTLN("Could not sleep %lu s (unknown LoRa DIO1 pin)", (unsigned long)secs);
    return;
  }

  uint64_t ms64 = (uint64_t)secs * 1000ULL;
  uint32_t ms = (ms64 > 0xFFFFFFFFULL) ? 0xFFFFFFFFUL : (uint32_t)ms64;

  auto dio1_high = [&]() -> bool {
    return digitalRead((uint8_t)dio1) == HIGH;
  };

  if (dio1_high()) {
    MESH_DEBUG_PRINTLN("Skipped sleeping %lu s (DIO1 is HIGH - pending/latched IRQ)", (unsigned long)secs);
    return;
  }

  // From this point on, we logically consider ourselves asleep
  sleeping = true;

  // Lazily initialize the sleep timer once, and reuse it.
  if (!sleep_timer_inited) {
    sleep_timer.begin(ms, sleep_timer_callback, nullptr, /*repeating=*/false);
    sleep_timer_inited = true;
  } else {
    sleep_timer.stop();
    sleep_timer.setPeriod(ms);
  }

  sleep_timer.start();

  if (dio1_high()) {
    MESH_DEBUG_PRINTLN("Skipped sleeping %lu s (DIO1 went HIGH after arming timer)", (unsigned long)secs);
    sleeping = false;
    sleep_timer.stop();
    return;
  }

  MESH_DEBUG_PRINTLN("Sleeping for %lu s", (unsigned long)secs);

  suspendLoop();

  // Defensive: should never reach here
  sleeping = false;
}

...

bool NRF52BoardOTA::startOTAUpdate(const char *id, char reply[]) {
  ota_enabled = true;

  // Disable sleeping then OTA is enabled.
  if (sleep_timer_inited) {
    sleep_timer.stop();
    sleeping = false;
  }
  ...
}
```

**main.cpp**
```cpp
// Power saving (if enabled)
//
// The interval of no activity after which the board will sleep to save power.
constexpr unsigned long POWER_SAVING_TRIGGER_INTERVAL_SEC = 1 * 60; // 1 minute
// The sleep duration when the board has entered power saving sleep (or sooner on RX).
constexpr unsigned long POWER_SAVING_SLEEP_DURATION_SEC = 30 * 60;  // 30 minutes
// // When waking up from sleep, or when there is work pending, work a bit.
// constexpr unsigned long POWER_SAVING_WAKE_WORK_DURATION_SEC = 5;    // 5 seconds
// The interval (in sec) after which we will sleep (if powersaving is enabled).
unsigned long nextPowerSavingIntervalInSec = POWER_SAVING_TRIGGER_INTERVAL_SEC;

...
void loop() {
  ...
  if (the_mesh.getNodePrefs()->powersaving_enabled &&
      the_mesh.millisHasNowPassed(the_mesh.getLastRX() + nextPowerSavingIntervalInSec * 1000)) {
      // If we still have work to do, wait a little longer before sleeping.
      if (the_mesh.hasPendingWork()) {
        MESH_DEBUG_PRINTLN("Skipped powersaving (work pending, work another %lu s)", (unsigned long)POWER_SAVING_WAKE_WORK_DURATION_SEC);
        nextPowerSavingIntervalInSec += POWER_SAVING_WAKE_WORK_DURATION_SEC;
      } else {
        MESH_DEBUG_PRINTLN("Start powersaving");

        // Sleep and wake up after the sleep duration, or when receiving a LoRa packet.
        board.sleep(POWER_SAVING_SLEEP_DURATION_SEC);

        MESH_DEBUG_PRINTLN("Resuming work for %lu s", (unsigned long)POWER_SAVING_TRIGGER_INTERVAL_SEC);
        nextPowerSavingIntervalInSec = POWER_SAVING_TRIGGER_INTERVAL_SEC;
      }
  }
}
  ```
  
  **MyMesh.h**
  ```h
    unsigned long lastRX; // The last time a RX event happened.
    ...
      
    public:
    ...
    unsigned long getLastRX() {
      return lastRX;
    }
  ```
  
And in **MyMesh.cpp** I set `lastRX = millis();` where RX happened. However, probably it's better to use RadioLibWrapper's setState for this when there is RX.

Hopefully this will be helpful to move this PR along :)

**Comment by fschrempf** [comment]
I've updated this PR. Here is a summary of the changes compared to the previous version:

* fix `SoftwareTimer` handling (create only one instance and update it for each sleep)
  * this was likely the reason for the lockups before as I wasn't aware that `SoftwareTimer::begin()` calls `xTimerCreate()` each time
* Add generic way to prevent sleep for critical operations and use this to prevent sleep after OTA update has been activated
* Extend `MyMesh::hasPendingWork()` to take ongoing packet handling into account
* Cosmetics: Use a board-specific callback `onRXInterrupt()` in the ISR and rename the ISR

@IoTThinks @4np Thanks for your comments. With these fixes, I don't think it is necessary to explicitly check the DIO1 pin or introduce any other workarounds.

This has been running just fine on my repeater for the past two days without any noticeable issues.

Here is a link with the firmware binaries from the CI run: https://github.com/fschrempf/MeshCore/actions/runs/20864519291/artifacts/5081189520.

**Comment by fschrempf** [comment]
@4np One more thing: Do you use any AI tools for you reviews and code snippets? If yes, please make sure that you will mention this in the future. I want to know when I'm dealing with feedback generated by a LLM. If this is not the case, just ignore my comment. Thanks!

**Comment by 4np** [comment]
@fschrempf, actually, I wrote these. The [one with the bullet points](https://github.com/meshcore-dev/MeshCore/pull/1238#issuecomment-3715585518) I wrote, but I did ask ChatGPT to rewrite what I wrote and summarize it more consistently. 

> With these fixes, I don't think it is necessary to explicitly check the DIO1 pin or introduce any other workarounds.

It could indeed very well be that the many timers was the cause of the unresponsiveness. Good to hear you have good results :)

I'll give the new firmware build a go.

**Comment by fschrempf** [comment]
@4np Thanks for clarifying!

**Comment by 4np** [comment]
@fschrempf I am sorry to report that my RAK19007 / RAK4631 is again unresponsive so it looks like your latest changes did not solve the issue.

Battery was at 99% when I flashed the firmware yesterday evening, this morning it's again unresponsive.

**Comment by fschrempf** [comment]
@4np Thanks for testing! That's strange. On my end with the old version the problem occurred after a few hours or a day at the latest. With the new version the repeater has worked fine for four days in a row now.

I will try to do some further investigations.

**Comment by 4np** [comment]
@fschrempf I applied a different implementation checking for `DIO1 == LOW`, and for now I only implemented it for the RAK4631 variant. I am testing it at the moment on my repeater to see if it fixes the unresponsiveness for me. This is [the patch](https://github.com/user-attachments/files/24552005/MeshCore-PR1238-Patch.diff.zip) with the changes I made.

> Note: this is the RAK4631 repeater build: [RAK_4631_repeater-adhoc-202601111311-firmware.zip](https://github.com/user-attachments/files/24552015/RAK_4631_repeater-adhoc-202601111311-firmware.zip)

**Comment by 4np** [comment]
@fschrempf: So far so good, 19h uptime without stalls... It looks like sleeping when DIO1 is HIGH is the cause of the unresponsiveness.

As this is related to SX126x, I wonder if the same deadlock may happen in the ESP32 implementation or if it's unique to NRF52?

> With the new version the repeater has worked fine for four days in a row now.

The mesh here is pretty [crowded](https://analyzer.letsmesh.net/map?lat=52.06857&long=5.64739&zoom=9) and [busy](https://analyzer.letsmesh.net/packets?region=netherlands), maybe it's more quiet on your end?

**Comment by fschrempf** [comment]
So I've been digging into the deadlock issue a bit more and got a pretty clear image of what is happening and how it can be prevented.

## 1. `suspendLoop()` and `resumeLoop()` in NRF52 Arduino Core

The code uses the FreeRTOS task API to suspend and resume the task that runs the main loop.

```c
void suspendLoop(void)
{
  vTaskSuspend(_loopHandle);
}

void resumeLoop(void)
{
  if ( isInISR() ) 
  {
    xTaskResumeFromISR(_loopHandle);
  } 
  else
  {
    vTaskResume(_loopHandle);
  }
}
```

This works fine in general, but brings in some issues, especially the one [mentioned in the docs](https://www.freertos.org/Documentation/02-Kernel/04-API-references/02-Task-control/08-xTaskResumeFromISR):

> `xTaskResumeFromISR()` is generally considered a dangerous function because its actions are not latched. For this reason it should definitely not be used to synchronise a task with an interrupt if there is a chance that the interrupt could arrive prior to the task being suspended, and therefore the interrupt being lost.

## 2. Triggering the deadlock

When going to sleep we call `suspendLoop()` from within the loop task that gets suspended. On the other hand the `resumeLoop()` is fully asynchronous called by the ISRs. Calling `resumeLoop()` when the task is not yet or currently in the process of being suspended apparently causes the lockup. Unfortunately we can't make it 100% safe unless we disable the interrupts for a certain amount of time which risks missing incoming packets. Trying to sync the suspend in the loop and the resume in the ISR always bears the risk of leaving a small time window in which the triggering of the ISR would cause a lockup.

## 3. Solution using FreeRTOS API

We maybe could guard the `resumeLoop()` by a check of the loop task being actually suspended. As that would require to use the FreeRTOS API anyway, it seems much better to go for a more elegant solution that is also recommended by the FreeRTOS docs. Using a [direct task notification](https://www.freertos.org/Documentation/02-Kernel/02-Kernel-features/03-Direct-to-task-notifications/03-As-counting-semaphore) is efficient and fast according to the docs.

Edit: Using a direct task notification is not possible as we need the task handle of the loop task which is declared static in the core and the API function (`xTaskGetHandle()`) is disabled in `FreeRTOSConfig.h`. A good alternative is a binary semaphore. 

It works like this:

1. Initializing a binary semaphore and "taking" it once in the setup code
2. Blocking the loop task by taking the semaphore a second time in `sleep()`
3. Resuming the loop by "giving" the semaphore from the radio ISR
4. We don't need a wakeup timer as we can use the timeout parameter in `xSemaphoreTake()`

Yes, we use the FreeRTOS API directly instead of the higher level Arduino core functions. But I think that's acceptable as it should provide a reliable and working solution.

fschrempf:
I rebased the PR and implemented the following changes compared to the previous version:

1. Fix a bug in RadioLibWrappers.cpp where the bitmask of `STATE_TX_WAIT` was wrong which led to `onRXInterrupt()` being also triggered on TX ready interrupts (see https://github.com/meshcore-dev/MeshCore/pull/1238/commits/72a7335434d8e5c8e675ddcd7d1f9fd3a8a931eb).
2. Use direct task notifications for suspending and resuming the loop task as explained [above](https://github.com/meshcore-dev/MeshCore/pull/1238#issuecomment-3740222763) (see https://github.com/meshcore-dev/MeshCore/pull/1238/commits/56213ca3f479d5882e0db9fcc7720a5ba325bdf6).

I noticed that for some reason the serial debug messages on the USB CDC interface of the RAK4631 (using TinyUSB library) stop after the the first suspend/resume cycle. I couldn't find the reason for this yet. If anyone got an idea or could check if they see the same problem that would be appreciated.

Here is a link to the CI build: https://github.com/fschrempf/MeshCore/actions/runs/20933125598/artifacts/5103765633

@4np Sorry for proposing yet another solution. I would have been fine with adding the DIO1 == LOW check as a fix but after trying to understand the issue I think this still has the potential of locking up, although the probability is likely lower than without the check.

4np:
@fschrempf sure no problem, I think your investigation makes sense. I'll give the new build a go and see how it behaves on my end. 

In any case, my repeater has been up for 1 day and 18 hours without locking up with the DIO1 changes.

4np:
@fschrempf your CI build does not contain the RAK4631 repeater, it [failed to build](https://github.com/fschrempf/MeshCore/actions/runs/20933125598/job/60148828798#step:4:12118) (as did the other NRF52 variants):

```log
.pio/build/RAK_4631_repeater/src/helpers/NRF52Board.cpp.o: In function `NRF52Board::begin()':
NRF52Board.cpp:(.text._ZN10NRF52Board5beginEv+0xa): undefined reference to `xTaskGetHandle'
.pio/build/RAK_4631_repeater/src/helpers/NRF52Board.cpp.o: In function `NRF52BoardDCDC::begin()':
NRF52Board.cpp:(.text._ZN14NRF52BoardDCDC5beginEv+0x14): undefined reference to `xTaskGetHandle'
.pio/build/RAK_4631_repeater/src/helpers/NRF52Board.cpp.o: In function `virtual thunk to NRF52BoardDCDC::begin()':
NRF52Board.cpp:(.text._ZTv0_n80_N14NRF52BoardDCDC5beginEv+0x1c): undefined reference to `xTaskGetHandle'
collect2: error: ld returned 1 exit status
*** [.pio/build/RAK_4631_repeater/firmware.elf] Error 1
========================= [FAILED] Took 54.80 seconds =========================

Environment        Status    Duration
-----------------  --------  ------------
RAK_4631_repeater  FAILED    00:00:54.799
==================== 1 failed, 0 succeeded in 00:00:54.799 ====================
```

fschrempf:
> @fschrempf your CI build does not contain the RAK4631 repeater, it [failed to build](https://github.com/fschrempf/MeshCore/actions/runs/20933125598/job/60148828798#step:4:12118) (as did the other NRF52 variants):

Oh dear! I didn't notice because the job status is green. And locally my build worked because I had modified my Arduino core manually some weeks ago when I was doing some experiments. Therefore my local `FreeRTOSConfig.h` contained `#define INCLUDE_xTaskGetHandle 1` whereas the official version doesn't have `xTaskGetHandle` and so this approach won't work.

Therefore I now reverted to the original solution (used in the very first version of this PR, sometimes the first ideas are the best ;)) using a FreeRTOS binary semaphore. This is equally good and simple and also a recommended solution in the FreeRTOS docs.

Sorry for the additional noise. I updated the [explanatory post](https://github.com/meshcore-dev/MeshCore/pull/1238#issuecomment-3740222763) above accordingly.

Here are the latest binaries from the CI build: https://github.com/fschrempf/MeshCore/actions/runs/20971258478/artifacts/5118537314

4np:
I just flashed `RAK_4631_repeater-opt-1ae1d7b`, fingers crossed! 🤞🏻 ;)

fschrempf:
@4np Ok, but don't spend too much efforts in testing. Currently it looks like we will go for @IoTThinks alternative solution in https://github.com/meshcore-dev/MeshCore/pull/1353.

4np:
Maybe I should test #1353 instead then 😄

SaschaKt:
> Maybe I should test #1353 instead then 😄

From 8mA to 4mA, very good, I'm using it from the first version primary with v3/v4 (48mA down to 9-10mA) as solar repeater

fschrempf:
We are working on getting https://github.com/meshcore-dev/MeshCore/pull/1353 merged which provides an alternative solution. I will revisit this PR to get the remaining fixes/changes merged once we are done there.

fschrempf:
Superseded by https://github.com/meshcore-dev/MeshCore/pull/1565 and https://github.com/meshcore-dev/MeshCore/pull/1562

PR #1242, opened 2025-12-18
Disable the blue LED on RAK19003 and 19007

ivmden:
Adding a flag to the build to disable the Blue LED on the RAK19007 and 19003 boards, and a call in SerialBLEInterface.cpp to turn the LED off.

Tested on a 19007 and 19003.


liamcottle:
Howdy! Thanks for the contribution :)

I would probably opt to keep the LED enabled in default builds for now, as it's a very good indicator of when we have connection issues or the board locks up etc. We've had many support cases with connection issues and users have been able to tell us the LED is stuck on even when disconnected. It would be great to keep this for now.

Would you be willing to remove the `-D DISABLE_BLUE_LED=1` from `platformio.ini`, or at least comment it out, and when you build your own firmware you can set this flag in the build script?

liquidraver:
also remove this to keep it clean :)
   // If we want to control BLE LED ourselves, uncomment this:
  // Bluefruit.autoConnLed(false);

ivmden:
> I would probably opt to keep the LED enabled in default builds for now, as it's a very good indicator of when we have connection issues or the board locks up etc. We've had many support cases with connection issues and users have been able to tell us the LED is stuck on even when disconnected. It would be great to keep this for now.

(Please read this as conversational, not argumentative). You've also had multiple people asking about how to turn the LED off via discord.  How can we strike a balance of keeping good debug tools and keeping energy usage reasonable/users happy?  I tried getting it to turn off after the connection was well established but struggled with it.

Could a configuration: Off when connected, Flashing when trying to pair be a reasonable compromise?

> also remove this to keep it clean :) // If we want to control BLE LED ourselves, uncomment this: // Bluefruit.autoConnLed(false);

Done.

4np:
Can't it be a console command (and app setting) to disable the led? My solar repeater doesn't need the LED and it probably saves additional power usage if it is disabled, but by default it's probably best to have it enabled. To make it platform agnostic maybe introduce a generic property disable all unnecessary bells & whistles, then platform specific implementations can disable -say- the led or -say- the buzzer.

ivmden:
@4np These changes do not apply to anything but Bluetooth companion nodes.  Your solar repeater should not have this blue LED enabled.

@liamcottle I'd love some feedback from you from my above questions.  Many thanks!

liamcottle:
Her there, closing out due to old age, merge conflicts. Unfortunately the changes I requested weren't made so can't be accepted at this time.

PR #1244, opened 2025-12-19
Techo hibernation

ViezeVingertjes:


examples/companion_radio/ui-new/UITask.cpp
```diff
@@ -688,7 +688,13 @@ void UITask::shutdown(bool restart){
   if (restart) {
     _board->reboot();
   } else {
+    delay(2000);
```
fschrempf, line 691: Do we need such a long delay here?

ViezeVingertjes: Probably not, its just so the "Hibernating..." message has some time. But still need to measure it, it seems that i had different results with latest version from the version vs dev, the issue with the LED etc that kept drawing power might already been solved in another way in the time between.

PR #1247, opened 2025-12-19
Fixed T1000-E temperature, lux and BME280 sensor reading accuracies

entr0p1:
- Fix T1000-E thermistor math by clamping the divider supply to the regulated 3V3 rail for consistent temp readings across battery voltages
- Power and sample the light sensor correctly (enable 3V3 rail, set ADC pins to input at boot)

### DIFF
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -192,6 +192,13 @@ bool EnvironmentSensorManager::begin() {
   if (BME280.begin(TELEM_BME280_ADDRESS, TELEM_WIRE)) {
     MESH_DEBUG_PRINTLN("Found BME280 at address: %02X", TELEM_BME280_ADDRESS);
     MESH_DEBUG_PRINTLN("BME sensor ID: %02X", BME280.sensorID());
+    // Reduce self-heating: single-shot conversions, light oversampling, long standby.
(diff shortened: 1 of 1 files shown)


IoTThinks:
Could you compile a zip or uf2?
We can test for you.

entr0p1:
Sure, here's the BLE companion:

[firmware.uf2.zip](https://github.com/user-attachments/files/24258551/firmware.uf2.zip)

Let me know if you want a different flavour.

IoTThinks:
This is my T1000e with black battery.
Before your patch.
image

This is your patch.
The temperature is acceptable correct around 28-30*C.
The LUX is better from 0 to 10. Still a bit low indoor. May be it is due to T1000-e'case?

image

entr0p1:
That's actually a quirk of the sensor - it returns a "percentage" of light (whatever that's supposed to mean), not a raw lux value. I had the same confusion. 

I tested returning the value as a percentage which works, but the label for the sensor in the app just shows as "Percentage" not "Luminosity". So I deemed it too confusing and rolled that part back.

entr0p1:
Have added in a second fix - this time for the BME280. Have tested on the T-echo and temp readings are looking a lot more accurate now. If anyone else can test a node with BME280 that would be awesome (can provide firmware file, just specify which).

IoTThinks:
I have tested the BME280 and BME680 yesterday with similar initiation yesterday.
Not much different with and without the initiation when I tested.

Issue #604, opened 2025-08-12
LED is still on - Seed Xiao nRF52840


michalsiman:
I built a client using the Seed Xiao nRF52840 board together with the Wio-SX1262. It includes a battery and an external card, all in the size of a credit card. The LED on the Xiao board lights up red (probably when receiving an advert or a message) and then stays on until I manually send an advert or a message. Even though I’ve already viewed the advert or message in the app (synchronization has occurred), the LED remains lit. I believe this is a bug — in my opinion, the red LED should turn off once synchronization with the app is complete (after which the app takes over the indication of new messages, adverts, or contacts). Currently, the red LED stays on (until I transmit something), which I think unnecessarily drains the battery. Thank you.

446564:
Is this a custom build, or a kit built from the two dev boards they sell?

michalsiman:
It s a kit as Seed sell together, Xiao nRF52840 board and Wio-SX1262 board. I talking about internal LED on Xiao board. There isnt custom changes and stock firmware flash from web flasher, without any changes. All is original and stock.

446564:
ok, just confused as you said with an external card the size of a credit card so thought you had some custom hardware going on.

michalsiman:
Ah, so, no, creditcard size is only 3D printed box, but inside is standard hw xiao without any changes. just connected battery and antenna. Its my fault in description. Thx

entr0p1:
FWIW I've had this happen on repeater firmware as well for the Xiao nRF52840+Wio SX1262, but its intermittent.

entr0p1:
Found it. If the radio fails to TX, it will re-idle the radio (so it goes back to RX mode) but it doesn't call the function that turns the LED back off. If this is actually the problem, then its not just a Xiao thing but any board with an LED.

Problem is here:
https://github.com/meshcore-dev/MeshCore/blob/2228214ded57c2761312730cdeae14b2b31bc5a3/src/helpers/radiolib/RadioLibWrappers.cpp#L131C5-L141C2

We don't call `_board->onAfterTransmit();` (the function the turns off the LED in the board variant files) after calling `idle();`.

 I'll submit a PR with the fix.

liamcottle:
Fixed in #1254

PR #1253, opened 2025-12-20
FIX: TX led stuck on when TX fails

entr0p1:
Per issue https://github.com/meshcore-dev/MeshCore/issues/604 - TX LED gets stuck on if StartTransmit() fails since we don't call `_board->onAfterTransmit();` in this scenario.


entr0p1:
Whoops.. goofed when creating the branch and included unrelated commits. I'll try that again.

PR #1254, opened 2025-12-20
Fix TX LED stuck on when StartTransmit() fails

entr0p1:
Per issue https://github.com/meshcore-dev/MeshCore/issues/604 - TX LED gets stuck on if StartTransmit() fails since we don't call _board->onAfterTransmit(); in this scenario.


IoTThinks:
@entr0p1 I'm just curious. 
What is the possible reason to have StartTransmit() fail?

Is this due to hardware failure or something?
Thanks a lot.

PR #1266, opened 2025-12-23
Added powersaving to all ESP32 boards with RTC-supported DIO1

IoTThinks:
Added powersaving to all ESP32 boards (Heltec v2, v3, v3, wsl3...) with RTC-supported DIO1
- First sleep (if powersaving is enabled) is 2 minutes from boot. Then after that, they will try to sleep every 5s.
- In powersaving, the ESP32 boards will consume only 9mA. Much less than 50mA without powersaving.
- Added CLI powersaving, powersaving on and powersaving off to manage powersaving mode.
- Support for boards with DIO1 wiring to non-RTC GPIO (Xiao S3...) will be added after this PR. For now, they will continue to work but without any powersaving.

This is the old PR with all comments and discussion. 
Thanks a lot for your supports, comments and reviews.
Merry Chirstmas to everyone.
https://github.com/meshcore-dev/MeshCore/pull/1107

This is CLI for powersaving.
image

### DIFF
--- a/src/helpers/ESP32Board.h
+++ b/src/helpers/ESP32Board.h
@@ -8,7 +8,7 @@
 #include <rom/rtc.h>
 #include <sys/time.h>
 #include <Wire.h>
-#include <WiFi.h>
+#include "esp_wifi.h"
 #include "driver/rtc_io.h"
 
 class ESP32Board : public mesh::MainBoard {
@@ -65,8 +65,12 @@ public:
   }
 
   void sleep(uint32_t secs) override {
-    if (WiFi.getMode() == WIFI_MODE_NULL) { // WiFi is off ~ No active OTA, safe to go to sleep
-      enterLightSleep(secs);                // To wake up after "secs" seconds or when receiving a LoRa packet
+    // To check for WiFi status to see if there is active OTA
+    wifi_mode_t mode;
+    esp_err_t err = esp_wifi_get_mode(&mode);
+    
+    if (err != ESP_OK) {          // WiFi is off ~ No active OTA, safe to go to sleep
+      enterLightSleep(secs);      // To wake up after "secs" seconds or when receiving a LoRa packet
     }
   }
 


SaschaKt:
I'm using your modification since the first day and helped to improve with testing and suggestions. Your mod is running on 2x v4 and 3x v3 Repeaters without problems. One of the repeater is solar based with only 5W Panel and 6000mAh battery and installed on the roof. I had in 24h no noticeable battery draining, it was very cloudy and over the day I think it was enough to support the v3 and now over night the accu is still at 78% like it was before 12h earlier. Thank you for your work, heltec is since your modification a really good and cheap repeater. It should be integrated it official firmware for all esps, and the cli powersaving on / off is perfect also for those connected to AC power.

Little question. Is when CLI powersaving on still active after reboot or plugging off from power? The last set should be saved and be active also after a reboot or power on

IoTThinks:
> Little question. Is when CLI powersaving on still active after reboot or plugging off from power? The last set should be saved and be active also after a reboot or power on

Hi @SaschaKt, 
Yes, the last powersaving setting is saved and will continue after you reboot the repeater.
E.g: If you set `powersaving on,`  then you reboot the repeater.
After boot, the powersaving is still on and effective.

examples/simple_repeater/MyMesh.cpp
```diff
@@ -1115,3 +1115,8 @@ void MyMesh::loop() {
   uptime_millis += now - last_millis;
   last_millis = now;
 }
+
+// To get the current pending work
+int MyMesh::hasPendingWork() const {
```
csrutil: I guess we could check `digitalRead(P_LORA_DIO_1) == HIGH` and `digitalRead(P_LORA_BUSY) == HIGH` to make sure there are no pending work.

examples/simple_repeater/main.cpp
```diff
@@ -19,12 +19,19 @@ void halt() {
 
 static char command[160];
 
+// For power saving
+unsigned long lastActive = 0; // mark last active time
```
csrutil, line 23: lets say we could just define only one variable `last_activity_at`. that variable will be updated if

- user send data through serial
- user press user button

in the loop, we put the device into sleep mode if now - last_activity_at > 60s.

src/helpers/ESP32Board.h
```diff
@@ -47,6 +49,27 @@ class ESP32Board : public mesh::MainBoard {
...
+
+      if (secs > 0) {
+        esp_sleep_enable_timer_wakeup(secs * 1000000); // To wake up every hour to do periodically jobs
+      }
+
+      esp_light_sleep_start(); // CPU enters light sleep
+    }
+#endif
+  }
+
```
csrutil, line 66: We could just have one `enterLightSleep` function here, the sleep is a little confusing. could move the WiFi check to the main loop before entering sleep.

csrutil:
Thank you so much, this will help a lot of people. Very great work!

src/helpers/ESP32Board.h
```diff
@@ -47,6 +49,27 @@ class ESP32Board : public mesh::MainBoard {
     return temperatureRead();
   }
 
+  void enterLightSleep(uint32_t secs) {
+#if defined(CONFIG_IDF_TARGET_ESP32S3) // Supported ESP32 variants
```
csrutil: We should be able to wake up if user press the user button.

IoTThinks: This is a repeater not a companion.

Why you want to press button to wake a repeater up?

I believe in the current UI of a repeater, the two buttons have no functionalities except the Reset button.

-
For companion, it uses deepsleep for poweroff now.
It does not use sleep/light sleep.
So adding wakeup via buttons has no use now.

We may add later if there is a real need.

examples/simple_repeater/MyMesh.cpp
IoTThinks: I think no need.
The check for BUSY or similar activities should be handled by the below Dispatcher's getOutboundCount instead.

If a packet is sending, the outbound count should be 1 or more, then there is no sleep.

src/helpers/ESP32Board.h
IoTThinks: Besides lightsleep, we have deepsleep also.

For me, sleep() is for users to understand.

The correct term for these actions is light sleep.

Just like we are pointing poweroff() to deepsleep now.


Then later, we may implement for NRF52.

Then sleep will point to System idle on.

examples/simple_repeater/main.cpp
IoTThinks: Yes, it is a way of thought.

However, for this phase. The current naming is approriate.

We can always extend the feature or modify it in future if there is real need.

IoTThinks:
> Thank you so much, this will help a lot of people. Very great work!

Yah, it is helping many friends to survive winter now.

The work is intially submitted by me.
However, a lot of reviewers have put in valuable improvement to the codes too.
So now it becomes our work instead.

I hope we can have an initial "framework" to implement to powersaving to all MCUs and companions.

At least, we have sleep() in the Mainboard for all boards to inherit now.

Then we can move on and improve it further.

Thanks a lot for reviewing.

Merry Chirstmas.

Cisien:
For consistency, the cli command should probably be `get powersaving` and `set powersaving on/off`

examples/simple_repeater/main.cpp
SaschaKt: > lets say we could just define only one variable `last_activity_at`. that variable will be updated if
> 
> * user send data through serial
> * user press user button
> 
> in the loop, we put the device into sleep mode if now - last_activity_at > 60s.

60 sec are to much, my repeater will never go to lightsleep then because in a minute comes a dozen of packets. 5 sec. is a proven and good setting

weebl2000:
See also https://github.com/IoTThinks/MeshCore/pull/1

IoTThinks:
> See also https://github.com/IoTThinks/MeshCore/pull/1

I already replied.

IoTThinks:
> For consistency, the cli command should probably be `get powersaving` and `set powersaving on/off`

May be, a bit long. And the code is a bit messy for get and set area.

But how about this?
image

weebl2000:
> > For consistency, the cli command should probably be `get powersaving` and `set powersaving on/off`
> 
> May be, a bit long. And the code is a bit messy for get and set area. :D
> 
> But how about this? image

The cli isn't very consistent to begin with. There are some commands that work without set/get and some that work with. 😂

src/helpers/ESP32Board.h
IoTThinks: I have thought of moving Wifi to main.cpp
However, if we move WiFi to main.cpp, then we need to mask it for ESP32 only as NRF52 does not have WiFi.

So I think it makes sense to leave WiFi at ESP32Board.h.

IoTThinks:
@liamcottle Should we use "set powersaving on/off" and "get powersaving"? Or "powersaving" and "powersaving on/off"?
Thanks a lot.

IoTThinks:
@weebl2000 I merged and pushed your suggested changes to this PR.
1. Added powersaving_enabled sanitization - CommonCLI.cpp:98 - prevents garbage values from uninitialized prefs
2. Reordered struct field - CommonCLI.h - moved powersaving_enabled to match serialization order (after bridge_secret)

For this one, I modified hasPendingWork() to return bool instead.
4. Renamed method - MyMesh.cpp/h - hasPendingWork() → getPendingWorkCount() for clarity

src/helpers/CommonCLI.cpp
```diff
@@ -65,7 +65,7 @@ void CommonCLI::loadPrefsInt(FILESYSTEM* fs, const char* filename) {
     file.read((uint8_t *)&_prefs->bridge_baud, sizeof(_prefs->bridge_baud));                       // 131
     file.read((uint8_t *)&_prefs->bridge_channel, sizeof(_prefs->bridge_channel));                 // 135
     file.read((uint8_t *)&_prefs->bridge_secret, sizeof(_prefs->bridge_secret));                   // 136
-    file.read(pad, 4);                                                                             // 152
+    file.read((uint8_t *)&_prefs->powersaving_enabled, sizeof(_prefs->powersaving_enabled));       // 152
```
ripplebiz, line 68: there should be a file.read(pad, 3);  here.

src/helpers/CommonCLI.cpp
```diff
@@ -145,7 +147,7 @@ void CommonCLI::savePrefs(FILESYSTEM* fs) {
     file.write((uint8_t *)&_prefs->bridge_baud, sizeof(_prefs->bridge_baud));                       // 131
     file.write((uint8_t *)&_prefs->bridge_channel, sizeof(_prefs->bridge_channel));                 // 135
     file.write((uint8_t *)&_prefs->bridge_secret, sizeof(_prefs->bridge_secret));                   // 136
-    file.write(pad, 4);                                                                             // 152
+    file.write((uint8_t *)&_prefs->powersaving_enabled, sizeof(_prefs->powersaving_enabled));       // 152
```
ripplebiz, line 151: there should be a file.write(pad, 3);  here.

src/helpers/ESP32Board.h
```diff
@@ -47,6 +49,27 @@ class ESP32Board : public mesh::MainBoard {
...
+        esp_sleep_enable_timer_wakeup(secs * 1000000); // To wake up every hour to do periodically jobs
+      }
+
+      esp_light_sleep_start(); // CPU enters light sleep
+    }
+#endif
+  }
+
+  void sleep(uint32_t secs) override {
+    if (WiFi.getMode() == WIFI_MODE_NULL) { // WiFi is off ~ No active OTA, safe to go to sleep
```
ripplebiz: Would be better for code size to not bring in the WiFi dependency, and hit the IDF function directly (esp_wifi_get_mode()), which is this apparently:
wifi_mode_t WiFiGenericClass::getMode()
{
    if(!lowLevelInitDone || !_esp_wifi_started){
        return WIFI_MODE_NULL;
    }
    wifi_mode_t mode;
    if(esp_wifi_get_mode(&mode) != ESP_OK){
        log_w("WiFi not started");
        return WIFI_MODE_NULL;
    }
    return mode;
}

src/helpers/CommonCLI.cpp
IoTThinks: Let me adjust the code and test.
Thanks a lot for your review.

src/helpers/CommonCLI.cpp
IoTThinks: Let me adjust the code and test.
Thanks a lot for your review.

src/helpers/ESP32Board.h
IoTThinks: Let me adjust the code and test.
Thanks a lot for your review.

src/helpers/CommonCLI.cpp
IoTThinks: @ripplebiz: Just to confirm. 
You want like this?
I tested on Heltec v3 / RAK4631 and it works fine.

```
    file.read((uint8_t *)&_prefs->bridge_secret, sizeof(_prefs->bridge_secret));                   // 136
    file.read((uint8_t *)&_prefs->powersaving_enabled, sizeof(_prefs->powersaving_enabled));       // 152
    file.read(pad, 3);                                                                             // 153
    file.read((uint8_t *)&_prefs->gps_enabled, sizeof(_prefs->gps_enabled));                       // 156
```

src/helpers/CommonCLI.cpp
IoTThinks: @ripplebiz: Just to confirm. 
You want like this?
I tested on Heltec v3 / RAK4631 and it works fine.

```
    file.write((uint8_t *)&_prefs->bridge_secret, sizeof(_prefs->bridge_secret));                   // 136
    file.write((uint8_t *)&_prefs->powersaving_enabled, sizeof(_prefs->powersaving_enabled));       // 152
    file.write(pad, 3);                                                                             // 153
    file.write((uint8_t *)&_prefs->gps_enabled, sizeof(_prefs->gps_enabled));                       // 156
```

src/helpers/CommonCLI.cpp
IoTThinks: Added file.read(pad, 3); as required.

src/helpers/CommonCLI.cpp
IoTThinks: Added file.write(pad, 3);  as required.

src/helpers/ESP32Board.h
IoTThinks: Done.
https://github.com/meshcore-dev/MeshCore/pull/1266/commits/d911a34eeb959c566d7d1c4751eb361db5b7b24f

IoTThinks:
Thanks a lot for all reviews and approvals.

nakoeppen:
@IoTThinks I know you have been getting a lot of praise for this already, but just wanted to also reaffirm that I got to try this out over the past couple of days and I really think it is going to be revolutionary to the Mesh project. If I understand the power savings right, it seems like Solar Repeaters that get consistent traffic may not benefit much from the power savings, but with MeshCore being far more reliant on infrastructure, this is going to make it a lot easier for people to build a starter network in their area (myself included). 

Personally, I am ecstatic about having a solution where I can benefit from the increased TX output of the Heltec V4 (or even the far lower cost of the Heltec V3) while consuming only roughly twice the power if the NRF-based chips. It's also really neat to see the Meshtastic community stirring up about this news as well. Truthfully, I thought the 80% power savings would be primarily on-paper, but even I have been noticing the same results (albeit on a node that seldom needs to transmit). Thanks a lot for your open-source contribution!

IoTThinks:
@nakoeppen Hi, this is a community effort to make this come true.

A few friends here in Munich - a busy area for Meshcore, have reported good improvement for Heltec v4. 
The average power consumption is 18mA in busy area.
So a big reduction from original 50mA.

For less busy area, around 10mA on average.

6v 5w and 6000mA battery works perfect for Heltec v3, v4 now in Europe winter.

In MeshCore Facebook, I have released the firmare to work for all Esp32-based repeaters.

Let the group to test then I will push a PR here.

Btw, there is another PR from a friend here is to reduce NRF52 power consumption down to 7.5mA.

I'm looking at the powersaving for companions now.
Hope we can make the battery longer to be more more useful.

mikecarper:
I'm testing out the g2 right now. Cut's power usage at least in half.

fschrempf:
@ripplebiz This PR added a CLI command `powersaving` that only handles the setting/getting of a boolean preference parameter. Is this the correct approach? Why not using a config value that can be handled using the `get/set` command. From a user-perspective this seems much more reasonable to me.

Is there any concept or scheme for the CLI what should be a separate command and what can simply be a config parameter? Should we change that before the next release?

mikecarper:
I believe gps acts similar. But we could change this. Also we could make this available via Lora only as turning it on via usb serial will cut the connection making it hard to continue configuring the node

fschrempf:
@mikecarper Ok. [This change](https://github.com/meshcore-dev/MeshCore/pull/1238/commits/fbcb16a2b4251160fdd2ec9074e5a0c750096abb) I proposed as part of https://github.com/meshcore-dev/MeshCore/pull/1238 should prevent the serial CLI from dying when the option is turned on.

fizzyfuzzle:
@IoTThinks I want to test this on a Heltec V4 with added RTC on pins 17+18, but I don't really understand what this 'with RTC-supported DIO1' means?
Seems the Heltec V4 has this in platform.ini  '-D P_LORA_DIO_1=14', does that mean I also need to wire 'something' to that IO poort to?

Thanks in advance for all the PR's you are creating :ok_hand:

IoTThinks:
@fizzyfuzzle You may try this instead.

You dont need to wire anything for your boards (including Heltec v4).

If you want a pre-compiled bin file, let me know.

https://github.com/meshcore-dev/MeshCore/pull/1353

pki791:
@IoTThinks 
Hi, can You explain what we need to have the Xiao ESP32S3+Wio working with powersaving?
The DIO_1 seems connected to pin GPIO39. In case it really needs to be connected to a different pin, that is possible because DIO_1 is also on pin 1 on one of the pinheaders. I can proceed tests if needed, have all equipment on hand, compiling enviroment also ready.

----- edit
I understand little bit more, DIO_1 needs to be connected to a GPIO which is a RTC_GPIO, my proposal is to go for GPIO1=D0 which is mechanically a straight connection as the pins are directly above each other. After changing
``` -D P_LORA_DIO_1=D0  ; was 39``` and connecting DIO_1-D0 powersaving works.

Also my proposal to add directly the definition for V_BAT, when user adds voltage divider - which fits nicely direct to pin D3.
```
-D PIN_VBAT_READ=D3 ; test battery voltage
```

LanWolf:
Testing powersaving on the nightly espnow bridge build on a Heltec v4, but that seem to stop the espnow bridge from working.
Please advise ?

pon-wessel:
> Testing powersaving on the nightly espnow bridge build on a Heltec v4, but that seem to stop the espnow bridge from working. Please advise ?

I don't know what the ESPNOW bridge does exactly, but if it requires any serial output (like BLE, WiFi, USB) it's probably not compatible with powersaving.

IoTThinks:
> @IoTThinks Hi, can You explain what we need to have the Xiao ESP32S3+Wio working with powersaving? The DIO_1 seems connected to pin GPIO39. In case it really needs to be connected to a different pin, that is possible because DIO_1 is also on pin 1 on one of the pinheaders. I can proceed tests if needed, have all equipment on hand, compiling enviroment also ready.
> 
> ----- edit I understand little bit more, DIO_1 needs to be connected to a GPIO which is a RTC_GPIO, my proposal is to go for GPIO1=D0 which is mechanically a straight connection as the pins are directly above each other. After changing ` -D P_LORA_DIO_1=D0 ; was 39` and connecting DIO_1-D0 powersaving works.
> 
> Also my proposal to add directly the definition for V_BAT, when user adds voltage divider - which fits nicely direct to pin D3.
> 
> ```
> -D PIN_VBAT_READ=D3 ; test battery voltage
> ```

@pki791 You may want to change the DIO1 to a RTC GPIO.
Or use this PR instead. You only need to merge the ESP32Board.h.
I will do a cleanup for this PR this week.

https://github.com/meshcore-dev/MeshCore/pull/1353/changes#diff-bf2a85f9de86622aa1706e48181c52038e533cae13e3f52f5cd2a2a063b61dc8

IoTThinks:
> Testing powersaving on the nightly espnow bridge build on a Heltec v4, but that seem to stop the espnow bridge from working.
> Please advise

@LanWolf Yes, correct.
The UART will stop during a sleep. 
We will put a fix for that: Either to disable sleep for bridge repeater or improve to allow sleep properly in repeater.

Do you need powersaving for this espnow bridge repeater?
I guess you will use main power for such case?

LanWolf:
> > Testing powersaving on the nightly espnow bridge build on a Heltec v4, but that seem to stop the espnow bridge from working.
> > Please advise
> 
> @LanWolf Yes, correct. The UART will stop during a sleep. We will put a fix for that: Either to disable sleep for bridge repeater or improve to allow sleep properly in repeater.
> 
> Do you need powersaving for this espnow bridge repeater? I guess you will use main power for such case?

At this moment there are 2 solar powered repeater-bridge setup, they stand about 10 meters apart, and I would like to bridge them. This way we can use 433 to have longer and stable range to a neighbour city, and the 868 will be used locally.
Using mains is not really an option on this location, and I would suppose in a shtf situation bridges and repeaters really need to be on solar, due to brownouts to keep coms going especially to tell the outside world what is going on etc

IoTThinks:
@LanWolf Ok, let me try to add and test uart wakeup for ESP32-based repeaters
For now, please disable it with "powersaving off".

PR #1272, opened 2025-12-26
Prevent flood packet routing loops

ViezeVingertjes:
Adds path-based loop detection to prevent nodes from forwarding packets they've already forwarded. Drops packets when node ID appears in path and any adjacent hop (before or after) is a known neighbor, eliminating duplicate IDs in flood paths while preserving alternative routes.

image

### DIFF
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -333,6 +333,7 @@ DispatcherAction Mesh::routeRecvPacket(Packet* packet) {
         if (i >= PATH_HASH_SIZE) {
           const uint8_t* prev_hop = &packet->path[i - PATH_HASH_SIZE];
           if (isKnownNeighbor(prev_hop)) {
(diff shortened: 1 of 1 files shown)


jbrazio:
@ViezeVingertjes why did you close this PR ?

ViezeVingertjes:
> @ViezeVingertjes why did you close this PR ?

It does work; but i also have seen that it might introduce new problem scenario's. For example; in our Dutch mesh, where every ID (single byte ID) as 5+ duplicates, we now are at the numbers and density, that even my neigbhours-neighbours are duplicates of my ID. In that case, this will detect it as 'already sent' and discard it, which might cause loss of packets.

I think this solution might be a more performant version of the Hashtable though once we get 2 byte ID's, so we dont have to store a whole bunch of hashes in the memory.

PR #1274, opened 2025-12-27
To fix MCU Temperature for repeaters

IoTThinks:
To fix MCU Temperature for repeaters:
- To fix the default temperature to be overridden by external sensors (if any). In MeshCore, the first value for the data type has higher priority and the later value will be ignored. E.g: If there is BME280/680, the temperature of BME280/680 will be shown instead of MCU Temperature.
- To get and average the temperature so it is more accurate, especially in low temperature.

### DIFF
--- a/src/helpers/ESP32Board.h
+++ b/src/helpers/ESP32Board.h
@@ -44,7 +44,14 @@ public:
 
   // Temperature from ESP32 MCU
   float getMCUTemperature() override {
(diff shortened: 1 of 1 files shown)


IoTThinks:
Tested with hardcoded value from getMCUTemperature and the temperature from BME280/680 is shown instead of MCU Temperature.

PR #1282, opened 2025-12-29
Added i2c_probe for sensors and fixed BME680 to support Wire1

IoTThinks:
Added i2c_probe for sensors and fixed BME680 to support Wire1
- Fixed BME680 to support Wire1
- Added i2c_probe to check I2C addresses before triggering begin() for sensors. This is to speed up startup process, to avoid unnecessary initiation of unused sensors and to avoid bugs due to sensor codes of unused sensors
- Added LPS22HB as local library as public Arduino_LPS22HB breaks Wire connection if Wire1 connection is initiated.

In MeshCore > platform.ini, there are two faulty libraries:
-  -D ENV_INCLUDE_LPS22HB=1 ; Crash Wire connection => I added a local LPS22HB to fix this issue.
-  -D ENV_INCLUDE_VL53L0X=1 ; Spam "[Wire.cpp:513] requestFrom(): i2cRead returned Error -1" => i2c_prode will mitigate this as most people do not use VL53L0X

### DIFF
--- a/lib/Arduino_LPS22HB/Arduino_LPS22HB.h
+++ b/lib/Arduino_LPS22HB/Arduino_LPS22HB.h
@@ -1,0 +1,25 @@
+/*
+  This file is part of the Arduino_LPS22HB library.
+  Copyright (c) 2019 Arduino SA. All rights reserved.
+
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation; either
+  version 2.1 of the License, or (at your option) any later version.
+
+  This library is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  Lesser General Public License for more details.
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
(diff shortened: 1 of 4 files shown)


IoTThinks:
For quick reference, this is the undesired code in Arduino_LPS22HB
https://github.com/arduino-libraries/Arduino_LPS22HB/blob/master/src/BARO.cpp#L59

image

And this is the change in the local Arduino_LPS22HB
image

IoTThinks:
@fschrempf Would you mind reviewing this PR?
This is to fix LPS22HB to break Wire when Wire1 is in use.
I also want to use i2c_probe to skip initializing not-in-use sensors to speed up boot times and to avoid bug from these sensors to affect the main loop.

Furthermore, I need this i2c_probe to automatically detect and use 0x76 or 0x77 for BME280/680 as these modules may have different addresses.

Thanks a lot.

fschrempf:
@IoTThinks I don't really know why you are requesting a review specifically from me. Anyway, I would love to help, but I haven't worked with sensors and I currently don't have time to dive into another part of the code.

IoTThinks:
> @IoTThinks I don't really know why you are requesting a review specifically from me. Anyway, I would love to help, but I haven't worked with sensors and I currently don't have time to dive into another part of the code.

Yah, no problem.
Since, you are one of the reviewers putting a lot of detailed and useful comments.
See you in other PRs . :)

IoTThinks:
I feel quite annoyed with this unhandled error.
It has been 2 months since it was reported.

image

IoTThinks:
To be replaced by a new PR.

PR #1287, opened 2025-12-30
Sx1262 Improvements - LoRa Radio reset

entr0p1:
Several fixes/improvements for SX1262 radios:

1. Implemented radio resets for some TX and RX fault conditions. Radio will reset, then re-init and go back to receive state. Fault codes that will trigger reset:
RADIOLIB_ERR_TX_TIMEOUT
RADIOLIB_ERR_SPI_CMD_FAILED
RADIOLIB_ERR_SPI_CMD_INVALID
RADIOLIB_ERR_SPI_WRITE_FAILED
RADIOLIB_ERR_SPI_READ_FAILED
RADIOLIB_ERR_CHIP_NOT_FOUND
RADIOLIB_ERR_DEVICE_BUSY

On SX1262-based variants without the P_LORA_RESET pin configured, this *should* NO-OP.

2. Added SX1262 DC-DC enable code (separate to the board's DC-DC) and platform.ini flag to enable it on supported nodes. Enabled on RAK4631 as this is confirmed supported.

### DIFF
--- a/src/Dispatcher.h
+++ b/src/Dispatcher.h
@@ -57,6 +57,11 @@ public:
   virtual void onSendFinished() = 0;
 
   /**
+   * \brief  optional hook for handling radio TX/RX fault recovery
+  */
+  virtual void onTXRXFault() { }
+
+  /**
    * \brief  do any processing needed on each loop cycle
    */
   virtual void loop() { }
--- a/variants/rak4631/RAK4631Board.h
+++ b/variants/rak4631/RAK4631Board.h
@@ -7,7 +7,7 @@
 // LoRa radio module pins for RAK4631
 #define  P_LORA_DIO_1   47
 #define  P_LORA_NSS     42
-#define  P_LORA_RESET  RADIOLIB_NC   // 38
+#define  P_LORA_RESET  38
 #define  P_LORA_BUSY    46
 #define  P_LORA_SCLK    43
 #define  P_LORA_MISO    45
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -80,6 +80,7 @@ void Dispatcher::loop() {
       MESH_DEBUG_PRINTLN("%s Dispatcher::loop(): WARNING: outbound packed send timed out!", getLogDateTime());
 
       _radio->onSendFinished();
+      _radio->onTXRXFault();
       logTxFail(outbound, 2 + outbound->path_len + outbound->payload_len);
 
       releasePacket(outbound);  // return to pool
--- a/src/helpers/radiolib/CustomSX1262Wrapper.h
+++ b/src/helpers/radiolib/CustomSX1262Wrapper.h
@@ -19,6 +19,21 @@ public:
     int sf = ((CustomSX1262 *)_radio)->spreadingFactor;
     return packetScoreInt(snr, sf, packet_len);
   }
+  void onTXRXFault() override {
+    auto radio = static_cast<CustomSX1262 *>(_radio);
+    int16_t result = radio->reset();
+    if (result != RADIOLIB_ERR_NONE) {
+      MESH_DEBUG_PRINTLN("CustomSX1262Wrapper: reset failed (%d)", result);
+      return;
+    }
+    if (!radio->std_init()) {
+      MESH_DEBUG_PRINTLN("CustomSX1262Wrapper: re-init failed");
+      return;
+    }
+    // Rebind ISR and restart receive after reset.
+    RadioLibWrapper::begin();
+    startRecv();
+  }
   virtual void powerOff() override {
     ((CustomSX1262 *)_radio)->sleep(false);
   }
--- a/src/helpers/radiolib/RadioLibWrappers.cpp
+++ b/src/helpers/radiolib/RadioLibWrappers.cpp
@@ -24,6 +24,32 @@ void setFlag(void) {
   state |= STATE_INT_READY;
 }
 
+// Decide which radio TX/RX errors warrant a reset (hardware/timeout), versus logic/parameter errors.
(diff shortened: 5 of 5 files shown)


src/helpers/BaseChatMesh.h
```diff
@@ -101,7 +101,7 @@ class BaseChatMesh : public mesh::Mesh {
   virtual void onSignedMessageRecv(const ContactInfo& contact, mesh::Packet* pkt, uint32_t sender_timestamp, const uint8_t *sender_prefix, const char *text) = 0;
   virtual uint32_t calcFloodTimeoutMillisFor(uint32_t pkt_airtime_millis) const = 0;
   virtual uint32_t calcDirectTimeoutMillisFor(uint32_t pkt_airtime_millis, uint8_t path_len) const = 0;
-  virtual void onSendTimeout() = 0;
+  virtual void onRadioFault() = 0;
```
fschrempf: Wouldn't it be better to move `onRadioFault()` from `BaseChatMesh` to the `Mesh` base class as it makes sense to use it in all roles (companion, repeater, etc.)? Also we probably can use a default implementation that prints the error log and don't need any overrides in the subclasses anymore.

src/Dispatcher.h
```diff
@@ -56,6 +56,11 @@ class Radio {
   */
   virtual void onSendFinished() = 0;
 
+  /**
+   * \brief  optional hook for handling radio timeouts/fault recovery (eg. TX timeout or RX start failure).
+  */
+  virtual void onRadioFault() { }
```
fschrempf: Maybe use a different name for the `onRadioFault` method in the Radio class to avoid confusion with the one in the Mesh class? The "radio" in the name is probably redundant so it could be just `onFault()`.

variants/rak4631/RAK4631Board.h
```diff
@@ -7,7 +7,7 @@
 // LoRa radio module pins for RAK4631
 #define  P_LORA_DIO_1   47
 #define  P_LORA_NSS     42
-#define  P_LORA_RESET  RADIOLIB_NC   // 38
+#define  P_LORA_RESET  38
```
fschrempf, line 10: Good catch! I also have that here in https://github.com/meshcore-dev/MeshCore/pull/1238: https://github.com/meshcore-dev/MeshCore/pull/1238/commits/13e3d9aa514013729dfec97dd043aa26767b1bcb

src/helpers/radiolib/CustomSX1262.h
```diff
@@ -55,6 +55,10 @@ class CustomSX1262 : public SX1262 {
         return false;  // fail
       }
     
+#ifdef SX126X_USE_DCDC
+      setRegulatorMode(RADIOLIB_SX126X_REGULATOR_DC_DC);
```
fschrempf: I was always assuming the DC/DC would already get enabled implicitly, but I never really checked. With this change I don't see a decrease of the supply current on RAK4631 in RX mode (it looks like it's even slightly higher), but I'm not sure if this is even supposed to lower the power consumption.

It seems like there is a shorter wrapper `setRegulatorDCDC()` that you could use.

fschrempf (commented):
Great work! Here are some things I've noticed.

And maybe you could clean up the commits a little bit so there is one commit per logical change. You can just force-push if you cleaned up the history.

entr0p1:
> Great work! Here are some things I've noticed.

Thanks heaps for your time reviewing this! Great points all around and I'll respond/fix accordingly. 

> And maybe you could clean up the commits a little bit so there is one commit per logical change. You can just force-push if you cleaned up the history.

Good call - I seem to have a concrete skull when it comes to learning how Git works haha :) will fix that up too, cheers

src/helpers/radiolib/CustomSX1262.h
entr0p1: You're correct - I'll fix this. I got a bum steer; appears Semtech doesn't turn on the DC-DC regulator by default, but Radiolib does. I'll remove this.

src/Dispatcher.h
entr0p1: Noted, I've made it `onTXRXFault()` just so we still have some idea on what it is for.

src/helpers/BaseChatMesh.h
entr0p1: This was a heavy-handed find and replace (rookie mistake) and not intended to be modified. I've undone this now.

entr0p1:
@fschrempf would you mind taking another look for me now? All your comments should be resolved.

src/helpers/radiolib/CustomSX1262.h
fschrempf: > but Radiolib does

Are you sure? I don't think it does. There is `useRegulatorLDO` in `SX126x::begin()` but it seems like we are not using this overloaded method. Anyway we can investigate further and improve that later if necessary.

fschrempf (approved):
Looks good to me!

entr0p1:
Been running this build on a RAK4631 repeater for nearly a month now, suffice to say no new issues introduced.

weebl2000:
@entr0p1 Is this PR still needed? What problems does it fix.

446564:
And if you still want to merge can you resolve conflicts?

entr0p1:
> @entr0p1 Is this PR still needed? What problems does it fix.

Needed is subjective, but I would say yes. It will reset the LoRa radio if it ends up in a non-functional state. It covers more than just stuck TX which last I checked was all we did (but I haven't checked in a while). Even then, we just forced it back to RX mode rather than resetting the radio. 




> And if you still want to merge can you resolve conflicts?

Can do.

PR #1294, opened 2025-12-31
Add NVS partition reset to ESP factory reset

liquidraver:
NVS partition is not touched by software initiated factory reset, so clearing out bluetooth bonding files can be only done with a flash erase now.

We should format NVS partition too to clear out everything, to match nRF factory reset behaviour.

Also disable serial interface when issuing factory reset command, because now the code re-start advertising when the phone app disconnects (so we are advertising during format)

Full debug from a xiao c3 from issuing the factory reset command to reboot:

[2025-12-31 13:29:07.2412 BLE: readBytes: sz=6, hdr=51
[2025-12-31 13:29:07.2415 DEBUG: Factory reset: disabling serial interface to prevent reconnects (BLE/WiFi)
[2025-12-31 13:29:07.2416 BLE: SerialBLEInterface: disable
[2025-12-31 13:29:07.2430 [310655][E][esp32-hal-misc.c:128] disableCore0WDT(): Failed to remove Core 0 IDLE task from WDT
[2025-12-31 13:29:07.2915 BLE: SerialBLEInterface: disconnected conn_handle=1 reason=0x216 (initiated by remote)
[2025-12-31 13:29:22.9804 ESP-ROM:esp32c3-api1-20210207

### DIFF
--- a/examples/companion_radio/DataStore.cpp
+++ b/examples/companion_radio/DataStore.cpp
@@ -65,6 +65,7 @@ void DataStore::begin() {
 
 #if defined(ESP32)
   #include <SPIFFS.h>
+  #include <nvs_flash.h>
 #elif defined(RP2040_PLATFORM)
   #include <LittleFS.h>
 #elif defined(NRF52_PLATFORM) || defined(STM32_PLATFORM)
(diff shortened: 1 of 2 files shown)


recrof:
would also add that flashing merged bin and/or using flash erase in web flasher will also clear NVS partition.

liquidraver:
> would also add that flashing merged bin and/or using flash erase in web flasher will also clear NVS partition.

yes, the ultimate nuke is flash erase from webflasher, this is mainly to make ESP reset equal to nRF (so we won't retain security information like BLE bondings on factory reset)

examples/companion_radio/DataStore.cpp
```diff
@@ -178,6 +179,15 @@ bool DataStore::formatFileSystem() {
 #endif
 }
 
+bool DataStore::clearNVS() {
```
ripplebiz: prob simpler to make this part of the formatFileSystem() method.

liquidraver: right, I just thought at first, we will be using only clearing nvs in the future and that idea stuck.

PR #1295, opened 2025-12-31
Merge littlefs 1.7.0 with adafruit's 1.6.0

liquidraver:
Since Adafruit forked littlefs there were a couple of patches, this script will update adafruit framework's littlefs to 1.7 retaining modifications that adafruit did in the fork.

it should bring corruption prevention on renames.

### DIFF
--- a/arch/nrf52/extra_scripts/littlefs_patch/lfs.c
+++ b/arch/nrf52/extra_scripts/littlefs_patch/lfs.c
@@ -1,0 +1,2586 @@
+/*
+ * The little filesystem
+ *
+ * Copyright (c) 2017, Arm Limited. All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+#include "lfs.h"
+#include "lfs_util.h"
+
+#include <inttypes.h>
+
+
+/// Caching block device operations ///
+static int lfs_cache_read(lfs_t *lfs, lfs_cache_t *rcache,
+        const lfs_cache_t *pcache, lfs_block_t block,
+        lfs_off_t off, void *buffer, lfs_size_t size) {
+    uint8_t *data = buffer;
+    LFS_ASSERT(block < lfs->cfg->block_count);
+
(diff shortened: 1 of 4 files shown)


recrof:
this change was part of series of btfixvXX releases.. we don't have conclusive evidence if it helped to prevent corruptions, but it's good to be close to upstream version, and we at least know it did not introduce any regressions during those few weeks we tested it.

liquidraver:
this will need a refactor to only patch the relevant lines in the unlikely event adafruit patches upstream (haven't in ages), but that's for 2026, this overwrite is faster

liamcottle:
Is it possible for us to fork the original repos, apply any needed patches as git commits, and use the forked repos instead of using a python script to swap out code files at build time? (Same comment for that Bluefruit patch) :)

liquidraver:
> Is it possible for us to fork the original repos, apply any needed patches as git commits, and use the forked repos instead of using a python script to swap out code files at build time? (Same comment for that Bluefruit patch) :)

of course it would be more in the github spirit, but I had reasons to do it like this:
original PR's upstream are not merged for months (years) now and maintaining a whole fork for the framework just for a few commits is overkill imho.
(https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/831)

I can maintain 2 python files and tune them according to upstream changes (delete them if something get's merged etc)

as for bluefruit:
well, someone should PR it upstream too so we can delete the patch, but I'm not confident in myself comprehending the whole thing enough to do it ;)

liquidraver:
anyway, if one of you guys decide to make an "official" fork of the adafruit framework, I'll gladly commit the build time patches there

weebl2000:
@liamcottle 
> Is it possible for us to fork the original repos, apply any needed patches as git commits, and use the forked repos instead of using a python script to swap out code files at build time? (Same comment for that Bluefruit patch) :)

Done, see #1922

geeksville:
i'm glad this was useful for ya'll.

PR #1297, opened 2025-12-31
Implement token bucket duty cycle enforcement

### LINKED ISSUE #817: Feature: Proper Duty Cycle Calculation
Basic idea: let’s assume 10% duty cycle in Europe calculated rolling over last hour.

1 hour = 3600sec = 3.600.000 ms -> 10% duty cycle = 360.000ms
We now need 2 variables: tx_budget and last_update

So we start with a full tx_budget of 360.000ms at last_update 0

Now we send first message which takes 200ms 
-> reduce tx_budget by 200 ms -> 359.800ms
-> set last_update to now()

Next message 100secs after first one:
-> check difference between last_update and now()  = 100sec
-> during that time we had no TX so we can add 10% of that ( 10 sec / 10000ms ) back to our tx_budget
-> tx_budget = 360.000ms

Now we send 100 messages in a row each 200ms
-> tx_budget reduced by (100x200ms) 20.000ms
-> tx_budget 340.000ms
-> last_update = now()

5sec later next package which is bigger and takes 300ms:
-> time difference last_uodate and now = 5sec -> add 0.5sec/500ms to budget
-> tx_budget 340.500ms
-> reduce budget by the now needed 300ms
-> tx_budget 340.200ms

That way you always have a rolling window defined by your budget and refilled by duty cycle percentage of passed time without TX.

Image

ViezeVingertjes:
Replaces per-transmission delay with a rolling window token bucket for better regulatory compliance.

Resolves #817.

### DIFF
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -20,12 +20,34 @@ void Dispatcher::begin() {
   _err_flags = 0;
   radio_nonrx_start = _ms->getMillis();
 
+  duty_cycle_window_ms = getDutyCycleWindowMs();
+  float duty_cycle = 1.0f / (1.0f + getAirtimeBudgetFactor());
+  tx_budget_ms = (unsigned long)(duty_cycle_window_ms * duty_cycle);
+  last_budget_update = _ms->getMillis();
+
   _radio->begin();
   prev_isrecv_mode = _radio->isInRecvMode();
 }
 
 float Dispatcher::getAirtimeBudgetFactor() const {
-  return 2.0;   // default, 33.3%  (1/3rd)
+  return 1.0;
+}
+
+void Dispatcher::updateTxBudget() {
+  unsigned long now = _ms->getMillis();
+  unsigned long elapsed = now - last_budget_update;
+
+  float duty_cycle = 1.0f / (1.0f + getAirtimeBudgetFactor());
+  unsigned long max_budget = (unsigned long)(getDutyCycleWindowMs() * duty_cycle);
+
+  unsigned long refill = (unsigned long)(elapsed * duty_cycle);
+  tx_budget_ms += refill;
+
+  if (tx_budget_ms > max_budget) {
+    tx_budget_ms = max_budget;
+  }
+
+  last_budget_update = now;
 }
 
 int Dispatcher::calcRxDelay(float score, uint32_t air_time) const {
@@ -61,11 +83,24 @@ void Dispatcher::loop() {
   if (outbound) {  // waiting for outbound send to be completed
     if (_radio->isSendComplete()) {
       long t = _ms->getMillis() - outbound_start;
-      total_air_time += t;  // keep track of how much air time we are using
+      total_air_time += t;
       //Serial.print("  airtime="); Serial.println(t);
 
-      // will need radio silence up to next_tx_time
-      next_tx_time = futureMillis(t * getAirtimeBudgetFactor());
+      updateTxBudget();
+
+      if (t > tx_budget_ms) {
+        tx_budget_ms = 0;
+      } else {
+        tx_budget_ms -= t;
+      }
+
+      if (tx_budget_ms < 100) {
+        float duty_cycle = 1.0f / (1.0f + getAirtimeBudgetFactor());
+        unsigned long needed = 100 - tx_budget_ms;
+        next_tx_time = futureMillis((unsigned long)(needed / duty_cycle));
+      } else {
+        next_tx_time = _ms->getMillis();
+      }
 
       _radio->onSendFinished();
       logTx(outbound, 2 + outbound->path_len + outbound->payload_len);
@@ -224,9 +259,20 @@ void Dispatcher::processRecvPacket(Packet* pkt) {
 }
 
 void Dispatcher::checkSend() {
-  if (_mgr->getOutboundCount(_ms->getMillis()) == 0) return;  // nothing waiting to send
-  if (!millisHasNowPassed(next_tx_time)) return;   // still in 'radio silence' phase (from airtime budget setting)
-  if (_radio->isReceiving()) {   // LBT - check if radio is currently mid-receive, or if channel activity
+  if (_mgr->getOutboundCount(_ms->getMillis()) == 0) return;
+  
+  updateTxBudget();
+  
+  uint32_t est_airtime = _radio->getEstAirtimeFor(MAX_TRANS_UNIT);
+  if (tx_budget_ms < est_airtime / 2) {
+    float duty_cycle = 1.0f / (1.0f + getAirtimeBudgetFactor());
+    unsigned long needed = est_airtime / 2 - tx_budget_ms;
+    next_tx_time = futureMillis((unsigned long)(needed / duty_cycle));
+    return;
+  }
+  
+  if (!millisHasNowPassed(next_tx_time)) return;
+  if (_radio->isReceiving()) {
     if (cad_busy_start == 0) {
       cad_busy_start = _ms->getMillis();   // record when CAD busy state started
     }
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -39,15 +39,15 @@ void Dispatcher::updateTxBudget() {
 
   float duty_cycle = 1.0f / (1.0f + getAirtimeBudgetFactor());
   unsigned long max_budget = (unsigned long)(getDutyCycleWindowMs() * duty_cycle);
-
   unsigned long refill = (unsigned long)(elapsed * duty_cycle);
-  tx_budget_ms += refill;
-
-  if (tx_budget_ms > max_budget) {
-    tx_budget_ms = max_budget;
+  
+  if (refill > 0) {
+    tx_budget_ms += refill;
+    if (tx_budget_ms > max_budget) {
+      tx_budget_ms = max_budget;
+    }
+    last_budget_update = now;
   }
-
-  last_budget_update = now;
 }
 
 int Dispatcher::calcRxDelay(float score, uint32_t air_time) const {
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -8,7 +8,9 @@
 
 namespace mesh {
 
-#define MAX_RX_DELAY_MILLIS   32000  // 32 seconds
+#define MAX_RX_DELAY_MILLIS        32000  // 32 seconds
+#define MIN_TX_BUDGET_RESERVE_MS   100    // min budget (ms) required before allowing next TX
+#define MIN_TX_BUDGET_AIRTIME_DIV  2      // require at least 1/N of estimated airtime as budget before TX
 
 #ifndef NOISE_FLOOR_CALIB_INTERVAL
   #define NOISE_FLOOR_CALIB_INTERVAL   2000     // 2 seconds
@@ -94,9 +96,9 @@ void Dispatcher::loop() {
         tx_budget_ms -= t;
       }
 
-      if (tx_budget_ms < 100) {
+      if (tx_budget_ms < MIN_TX_BUDGET_RESERVE_MS) {
         float duty_cycle = 1.0f / (1.0f + getAirtimeBudgetFactor());
-        unsigned long needed = 100 - tx_budget_ms;
+        unsigned long needed = MIN_TX_BUDGET_RESERVE_MS - tx_budget_ms;
         next_tx_time = futureMillis((unsigned long)(needed / duty_cycle));
       } else {
         next_tx_time = _ms->getMillis();
@@ -264,9 +266,9 @@ void Dispatcher::checkSend() {
   updateTxBudget();
   
   uint32_t est_airtime = _radio->getEstAirtimeFor(MAX_TRANS_UNIT);
-  if (tx_budget_ms < est_airtime / 2) {
+  if (tx_budget_ms < est_airtime / MIN_TX_BUDGET_AIRTIME_DIV) {
     float duty_cycle = 1.0f / (1.0f + getAirtimeBudgetFactor());
-    unsigned long needed = est_airtime / 2 - tx_budget_ms;
+    unsigned long needed = est_airtime / MIN_TX_BUDGET_AIRTIME_DIV - tx_budget_ms;
     next_tx_time = futureMillis((unsigned long)(needed / duty_cycle));
     return;
   }
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -732,7 +732,7 @@ MyMesh::MyMesh(mesh::Radio &radio, mesh::RNG &rng, mesh::RTCClock &rtc, SimpleMe
 
   // defaults
   memset(&_prefs, 0, sizeof(_prefs));
-  _prefs.airtime_factor = 1.0; // one half
+  _prefs.airtime_factor = 1.0;
   strcpy(_prefs.node_name, "NONAME");
   _prefs.freq = LORA_FREQ;
   _prefs.sf = LORA_SF;
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -689,7 +689,7 @@ MyMesh::MyMesh(mesh::MainBoard &board, mesh::Radio &radio, mesh::MillisecondCloc
 
   // defaults
   memset(&_prefs, 0, sizeof(_prefs));
-  _prefs.airtime_factor = 1.0;   // one half
+  _prefs.airtime_factor = 1.0;
   _prefs.rx_delay_base = 0.0f;   // turn off by default, was 10.0;
   _prefs.tx_delay_factor = 0.5f; // was 0.25f
   _prefs.direct_tx_delay_factor = 0.2f; // was zero
--- a/examples/simple_room_server/MyMesh.cpp
+++ b/examples/simple_room_server/MyMesh.cpp
@@ -597,7 +597,7 @@ MyMesh::MyMesh(mesh::MainBoard &board, mesh::Radio &radio, mesh::MillisecondCloc
 
   // defaults
   memset(&_prefs, 0, sizeof(_prefs));
-  _prefs.airtime_factor = 1.0;   // one half
+  _prefs.airtime_factor = 1.0;
   _prefs.rx_delay_base = 0.0f;   // off by default, was 10.0
   _prefs.tx_delay_factor = 0.5f; // was 0.25f;
   _prefs.direct_tx_delay_factor = 0.2f; // was zero
--- a/examples/simple_secure_chat/main.cpp
+++ b/examples/simple_secure_chat/main.cpp
@@ -280,7 +280,7 @@ public:
   {
     // defaults
     memset(&_prefs, 0, sizeof(_prefs));
-    _prefs.airtime_factor = 2.0;    // one third
+    _prefs.airtime_factor = 1.0;
     strcpy(_prefs.node_name, "NONAME");
     _prefs.freq = LORA_FREQ;
     _prefs.tx_power_dbm = LORA_TX_POWER;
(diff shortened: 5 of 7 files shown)


ViezeVingertjes:
We had a discussion in the Dutch community today. Some people didn't care much, others wanted a proper implementation. I'm neutral either way (not even sure if I enabled it myself). Perhaps this could start a broader discussion to decide it once and for all.

LitBomb:
how does this work for USA where there is no duty cycle limit?

ViezeVingertjes:
> how does this work for USA where there is no duty cycle limit?

The same as it is currently, airtime factor already limits it to the same amount, just less reliable. So both now and with this method, they would need to disable it. (set af 0)

LitBomb:
whatever the change is, it should not limit duty cycle at any percentage for regions that don't have duty cycle limit.

ViezeVingertjes:
> whatever the change is, it should not limit duty cycle at any percentage for regions that don't have duty cycle limit.

Its probably higher than the current cap you guys have. But a valid point for another issue on if and what the default should be. Perhaps it should be set with the region or something.

src/Dispatcher.cpp
```diff
@@ -20,12 +20,34 @@ void Dispatcher::begin() {
...
   _radio->begin();
   prev_isrecv_mode = _radio->isInRecvMode();
 }
 
 float Dispatcher::getAirtimeBudgetFactor() const {
-  return 2.0;   // default, 33.3%  (1/3rd)
+  return 2.0;
+}
+
+void Dispatcher::updateTxBudget() {
```
4np, line 38: There is race potential here. Are we sure this is always called from an Arduino single loop context? Or can this also be called by timers, callbacks or interrupts?

ViezeVingertjes: Good catch on thinking about thread safety! However, this is not a concern in this case. The updateTxBudget() method is only called from within the Dispatcher::loop() context, which is always executed from the Arduino single-threaded loop() function.

recrof:
@ViezeVingertjes I welcome proper implemetation of hourly duty cycle for europe, but this needs to be set to 0 by default. we don't want US, AU, NZ and other countries be limited to 10% by default.

ViezeVingertjes:
> @ViezeVingertjes I welcome proper implementation of hourly duty cycle for Europe, but this needs to be set to 0 by default. we don't want US, AU, NZ and other countries be limited to 10% by default.

Hmm, but default AF was 1.0 right? So i feel 10% would reflect the original 'duty cycle' more than 0%?
But disabled would be fine with me too though: eec5ef085364770676496081fa2c478b0d36ccd2

In practice it should already transmit more than with the original AF, so if anything, it shouldn't be less after this change.

recrof:
> Hmm, but default AF was 1.0 right? So i feel 10% would reflect the original 'duty cycle' more than 0%? But disabled would be fine with me too though: [eec5ef0](https://github.com/meshcore-dev/MeshCore/commit/eec5ef085364770676496081fa2c478b0d36ccd2)

yes, you are right. AF=0 can be dangerous.. AF=1 is optimal, sorry for that.

ViezeVingertjes:
> > Hmm, but default AF was 1.0 right? So i feel 10% would reflect the original 'duty cycle' more than 0%? But disabled would be fine with me too though: [eec5ef0](https://github.com/meshcore-dev/MeshCore/commit/eec5ef085364770676496081fa2c478b0d36ccd2)
> 
> yes, you are right. AF=0 can be dangerous.. AF=1 is optimal, sorry for that.

Haha check! thought maybe my thought train was wrong, reverted it -> 1b959a913924136aabe528bb58ecd60594a1607e

recrof:
> Haha check! thought maybe my thought train was wrong, reverted it -> [1b959a9](https://github.com/meshcore-dev/MeshCore/commit/1b959a913924136aabe528bb58ecd60594a1607e)

you reverted it to 9 tho? we need it at 1

ViezeVingertjes:
> > Haha check! thought maybe my thought train was wrong, reverted it -> [1b959a9](https://github.com/meshcore-dev/MeshCore/commit/1b959a913924136aabe528bb58ecd60594a1607e)
> 
> you reverted it to 9 tho? we need it at 1

Yes; i went off from numbers i saw on my repeaters which maxed out on 9%, but there are more things at play on there. Updated it to 1.0 and cleaned up the commits. a43b0e5aab823f0b8c15e86cbaac40d5d583cb38

fschrempf:
@ViezeVingertjes You can simply drop the last two commits, right? First you change it from 1 to 9 and then back from 9 to 1. Except for the `simple_secure_chat` which was at 2 for some reason at might want to be aligned with the others.

ViezeVingertjes:
> @ViezeVingertjes You can simply drop the last two commits, right? First you change it from 1 to 9 and then back from 9 to 1. Except for the `simple_secure_chat` which was at 2 for some reason at might want to be aligned with the others.

Done.

LitBomb:
how does a user know when they exceed the air time factor limit?  just a failed send on the radio?  any return error code to the clients so the clients can put up error messages to let the users know?

ViezeVingertjes:
> how does a user know when they exceed the air time factor limit? just a failed send on the radio? any return error code to the clients so the clients can put up error messages to let the users know?

Nothing changed on that behavior; the new calculation allows for bursts if there is budget, otherwise it will act like the 'old' AF waits, as its a sliding window. So its much more responsive while still utilizing the queue etc like it did before. I am not aware of any error code that is returned or feedback.

Obviously you wont notice a difference when there is no duty cycle applicable in your country (or you have it disabled anyway like most people), but on the default it feels like it responds pretty much instantly most of the time, until the budget is depleted, then it feels like it did in the current release.

marcelverdult:
thanks so much for implementing my request!

marcelverdult:
@liamcottle @andymux this would be quiete important for countries with strict dutycycle limits. the current implementation with the waits throws away valuable tx time.

marcelverdult:
@liamcottle this one would be very important for europe as well

recrof:
@ripplebiz are there any roadblocks preventing merging this request?

jbrazio:
@ViezeVingertjes just to make sure, the % calculation is still the same as the original one ?
I.e. AF1 = 50%, AF9 = 10% ?

ViezeVingertjes:
> @ViezeVingertjes just to make sure, the % calculation is still the same as the original one ? I.e. AF1 = 50%, AF9 = 10% ?

Yes, AF is preserved as the way to configure it as that is what is currently supported/known already. It should match the actually Duty Cycle more closely now, while also allowing for bursts instead of doing static waits after each transmission.

src/Dispatcher.cpp
```diff
@@ -20,12 +20,34 @@ void Dispatcher::begin() {
...
+  unsigned long max_budget = (unsigned long)(getDutyCycleWindowMs() * duty_cycle);
+
+  unsigned long refill = (unsigned long)(elapsed * duty_cycle);
+  tx_budget_ms += refill;
+
+  if (tx_budget_ms > max_budget) {
+    tx_budget_ms = max_budget;
+  }
+
+  last_budget_update = now;
```
weebl2000: When `updateTxBudget()` is called on fast loop iterations (1-5ms elapsed) and `duty_cycle` is small (e.g. 0.1 for factor=9), `(unsigned long)(elapsed * duty_cycle)` truncates to 0 every iteration. Since `last_budget_update` is always advanced to `now`, the lost fractional milliseconds never accumulate — the budget effectively stops refilling.

Fix: only advance `last_budget_update` when `refill > 0`, so elapsed time accumulates across iterations until it produces a non-zero refill. Max error at any point is <1ms of budget.

```suggestion
  unsigned long refill = (unsigned long)(elapsed * duty_cycle);
  if (refill > 0) {
    tx_budget_ms += refill;
    if (tx_budget_ms > max_budget) {
      tx_budget_ms = max_budget;
    }
    last_budget_update = now;
  }
```

weebl2000 (commented):
Budget refill truncation fix

weebl2000 (commented):
Budget refill truncation fix

ViezeVingertjes:
> Budget refill truncation fix

Applied the fix; havent seen it happen so far, but a good catch! :)
Makes sense, as AF 9 would pretty much disable your repeater at this time, so we have AF on 0 or at most at 1.0 in NL.

src/Dispatcher.cpp
```diff
@@ -61,11 +83,24 @@ void Dispatcher::loop() {
...
-      next_tx_time = futureMillis(t * getAirtimeBudgetFactor());
+      updateTxBudget();
+
+      if (t > tx_budget_ms) {
+        tx_budget_ms = 0;
+      } else {
+        tx_budget_ms -= t;
+      }
+
+      if (tx_budget_ms < 100) {
```
weebl2000: Feel like 100 is bit of a magic number. Perhaps we can clarify why 100ms?

src/Dispatcher.cpp
```diff
@@ -224,9 +259,20 @@ void Dispatcher::processRecvPacket(Packet* pkt) {
...
 void Dispatcher::checkSend() {
-  if (_mgr->getOutboundCount(_ms->getMillis()) == 0) return;  // nothing waiting to send
-  if (!millisHasNowPassed(next_tx_time)) return;   // still in 'radio silence' phase (from airtime budget setting)
-  if (_radio->isReceiving()) {   // LBT - check if radio is currently mid-receive, or if channel activity
+  if (_mgr->getOutboundCount(_ms->getMillis()) == 0) return;
+  
+  updateTxBudget();
+  
+  uint32_t est_airtime = _radio->getEstAirtimeFor(MAX_TRANS_UNIT);
+  if (tx_budget_ms < est_airtime / 2) {
```
weebl2000: Can we clarify the 2, it feels a bit like a magic number.

weebl2000 (approved):
✔️ Looks good. Maybe clarify two magic numbers with a small comment.

I think this is a big improvement for bursty workloads like room servers.

src/Dispatcher.cpp
ViezeVingertjes: https://github.com/meshcore-dev/MeshCore/pull/1297/changes/519b97a90ab9786e65423ac0d9dd7a9fd92dd180

src/Dispatcher.cpp
ViezeVingertjes: Half isnt too bad; but perhaps this helps -> https://github.com/meshcore-dev/MeshCore/pull/1297/changes/519b97a90ab9786e65423ac0d9dd7a9fd92dd180

weebl2000:
Love it, immediately clarifies what the numbers do, thanks.

JvM-nl:
It feels like that the boost make the repeater also react better.
With this or I got almost all the time a heard in flood.
Before I need a few re-send for a heard

weebl2000:
Running this for a while now, I feel like it's a no-brainer. Better token bucket in practice and adhering to duty cycle more accurately.

weebl2000:
@liamcottle Let's merge this? Some arguments below:

1. current duty cycle implementation is fundamentally broken for EU compliance
The per-packet delay approach doesn't actually enforce a duty cycle — it enforces a spacing. If a node is idle for 50 minutes then sends a burst, the delay model has no memory of that idle time. Conversely, after a single short packet, it imposes an unnecessarily long silence even though the node is nowhere near the hourly limit. The token bucket correctly models the regulatory requirement: X% of airtime within a rolling window.

2. huge throughput improvement for bursty workloads, i.e. room servers/repeaters
The current model forces AF × airtime silence after every packet, even if the node has been idle for an hour. The token bucket lets nodes use their accumulated budget for bursts, then naturally throttle once depleted. User @JvM-nl confirmed: "With this I got almost all the time a heard in flood. Before I needed a few re-sends."

3. code is well-reviewed and field-tested, I've been running this for 2 months as have many others and it's good to go.
4. for anyone who doesn´t care about duty cycle nothing really changes. set af 0 still disables enforcement altogether. Existing CLI is unchanged
5. token bucket is the textbook algorithm for rate limiting. The current per-packet delay is a bad approximiation that over-restricts (blocking bursts) and under-enforces (no actual rolling window...)
6. the changes are contained, literally all the logic changes are handled in Dispatcher, there's no new dependencies, no wire format changes, no client protocol changes, fully backwards compatible. Only thing new is getReminaingTxBudget() and it's additive to the rest
7. this PR has been open for 2+ months without any blocking objections, all review commentary was addressed.

tl;dr this replaces a fundamentally incorrect duty cycle model with the standard tocken bucket algorithm that actually enforces the regulatory requirements and actually **improves  throughput** for bursty workloads while being **more compliant**, not less. The diff is tiny at +71, -14 LOC. 

Please let's merge this, it's time.

liamcottle:
> @liamcottle Let's merge this? Some arguments below:

This will also need review from @ripplebiz, as he's the core maintainer of the Dispatcher system.

We are looking to do a v1.14.0 release later tonight, and I don't want to merge any new PRs that have a large footprint right now. Dev branch seems pretty stable right now.

One thing I'd like to see, is a simpler command, such as `set dutycycle 10` for 10%, and `set dutycycle 50` for 50%.

The airtime factor calculations have always confused me, and when I initially added support for it in the app, I removed it because it didn't make sense in the user interface. Having a cleaner command make the user experience nicer.

We can probably look at this properly after v1.14.0 is released.

weebl2000:
My 2 cents would be to merge it into v1.14.0😋. I'm really confident it's only going to improve the mesh as a whole.

4np:
Let me chip in another 2 cents that I agree with @liamcottle that a dedicated `set dutycycle x` would make sense as `set af` is confusing, but that can be implemented later. It would be good to have this merged _before_ `v1.14.0` is tagged as it would, as @weebl2000 explained, solve a number of existing issues that are probably the source of our (±  [3212](https://analyzer.letsmesh.net/nodes/repeaters) node) Dutch mesh (and other large meshes?) going on its knees.

Right now, often you need to resend messages. As people will think they are not heard they will use **#bot** and/or **#test** to see if they are heard, causing massive unnecessary chatter from several bots causing a lot of mesh overhead. If users see their messages get repeated, they are less likely to 'test', which in turn alleviates the mesh.

fschrempf:
Now that 1.14 is out, I would suggest to merge this rather sooner than later. This way we can get additional test coverage from people using dev until the next release is getting tagged.

recrof:
> Now that 1.14 is out, I would suggest to merge this rather sooner than later. This way we can get additional test coverage from people using dev until the next release is getting tagged.

agree. @liamcottle @ripplebiz, can we make it happen please?

liamcottle:
If this one can't be resolved before merging;

> One thing I'd like to see, is a simpler command, such as `set dutycycle 10` for 10%, and `set dutycycle 50` for 50%.

Can someone please provide a map/list of airtime factor values and how they map to duty cycle values.

Unless the user experience around this can be improved, it's unlikely I would add UI to the app. Users should be working with duty cycle numbers, since that's what's described in radio regulations etc.

Regarding merging, I'd like for @ripplebiz to review the code first. I'll link this to him in a DM as well.

weebl2000:
> If this one can't be resolved before merging;
> 
> > One thing I'd like to see, is a simpler command, such as `set dutycycle 10` for 10%, and `set dutycycle 50` for 50%.
> 
> Can someone please provide a map/list of airtime factor values and how they map to duty cycle values.
> 
> Unless the user experience around this can be improved, it's unlikely I would add UI to the app. Users should be working with duty cycle numbers, since that's what's described in radio regulations etc.
> 
> Regarding merging, I'd like for @ripplebiz to review the code first. I'll link this to him in a DM as well.

I think it's good to replace af with dutycycle, but let's do it in a separate PR. This one has been open for 2 months and is good to go as it is, if we start changing things we need to test it all over.

image

recrof:
> If this one can't be resolved before merging;
> 
> > One thing I'd like to see, is a simpler command, such as `set dutycycle 10` for 10%, and `set dutycycle 50` for 50%.
> 
> Can someone please provide a map/list of airtime factor values and how they map to duty cycle values.

af vs duty cycle examples:
af 1: dc 50%
af 2: dc 33%
af 3: dc 25%
...
af 9: dc 10%
dc = 100/(af+1)

weebl2000:
I can make a PR to add dutycycle command and deprecate af.

jbrazio:
I would merge this as it is.
In another PR we can then handle the different behavior for the new command because it's not only the logic that will be reversed i.e. `af 9` vs `dc 0.1` or `dc 10` but it would be great for the app support it under the radio settings and radio profiles.

recrof:
I'm also for merging this one, as it silently fixes all repeaters that have af=9 and get performance hit because how it's implemented

weebl2000:
See #1961 which builds on this PR and adds get/set dutycycle command.

dreirund:
> *One thing I'd like to see, is a simpler command, such as `set dutycycle 10` for 10%, and `set dutycycle 50` for 50%.*
> 
> *The airtime factor calculations have always confused me*

Thinking of the "Airtime Factor" as the amount of time-units to be "silent" I can grasp it intuitively:

E.g. when I am only allowed to send 20% of the time, i.e. only 1/5th, 1 time unit I am allowed to send and the 4 remaining I am not allowed to send, to think of it. So, `af=4`.

Regards!

ripplebiz:
I am _for_ this change, in principle, but I'm a little skeptical of the implementation, particularly all the arithmetic with the millis(). So, I did an experiment and modified the ArduinoHelpers.h, starting the millis clock at 5 minutes before the 49-day/32-bit wrap-around event:

class ArduinoMillis : public mesh::MillisecondClock {
public:
  unsigned long getMillis() override { return 4294667296 + millis(); }  // TEMP!!   5 minutes before 49-day wrap around
};

And, for five minutes I couldn't login to my test repeater, but as soon as 5 minutes passed it all worked again. 
Please look into this, and how to work _with_ the 49-day wrap-around.

beachmiles:
> > Hmm, but default AF was 1.0 right? So i feel 10% would reflect the original 'duty cycle' more than 0%? But disabled would be fine with me too though: [eec5ef0](https://github.com/meshcore-dev/MeshCore/commit/eec5ef085364770676496081fa2c478b0d36ccd2)
> 
> yes, you are right. AF=0 can be dangerous.. AF=1 is optimal, sorry for that.

There are people setting dutycycle to 100 now in the US thinking this is the new default and I am suspecting this could be bad and the dutycycle default still should be 50 for the US.
Can you or anyone else clarify why setting af=0 (dutycycle 100) is bad? Thanks!

recrof:
> There are people setting dutycycle to 100 now in the US thinking this is the new default and I am suspecting this could be bad and the dutycycle default still should be 50 for the US. Can you or anyone else clarify why setting af=0 (dutycycle 100) is bad? Thanks!

with dutycycle at 100%(af=0), repeater malfunction may cause non-stop tx. In very busy networks, well placed super busy repeaters may suffocate smaller ones with non-stop traffic. 50% is giving others chance to tx. those are just hypothetical scenarios as if you're getting close to 50% air utilization, mesh is long gone.

ViezeVingertjes:
Another side-effect is; we have a spammer that bombs the mesh daily with tens of thousands of flood packets in a timespan of an hour or 2-3, now because most repeaters are on 10% in this country, it somewhat dampens the effects... The mesh is still crawling, but the routes around the spammer hit the duty cycle and then it slows down.

Imagine if we did not have it set...

beachmiles:
> > There are people setting dutycycle to 100 now in the US thinking this is the new default and I am suspecting this could be bad and the dutycycle default still should be 50 for the US. Can you or anyone else clarify why setting af=0 (dutycycle 100) is bad? Thanks!
> 
> with dutycycle at 100%(af=0), repeater malfunction may cause non-stop tx. In very busy networks, well placed super busy repeaters may suffocate smaller ones with non-stop traffic. 50% is giving others chance to tx. those are just hypothetical scenarios as if you're getting close to 50% air utilization, mesh is long gone.

I was trying to look at the code to see if the dutycyle is over the setpoint if it queues the packet for a later time or if it discards it. If it queues it for a later time that then it seems like we should set the default dutycycle to maybe be something like 30 instead of 50?

Is there any reason why it would be beneficial to go above 30-50 dutycycle? 
If not maybe the code should probably only allow a max dutycycle of 30-50 as people here in the US seem to really want to go higher than 50 thinking that its allowed and its going to improve service.

Issue #1140, opened 2025-11-22
Remote Management via app unresponsive


mjvolante:
Whilst logging into a node via Remote Management is usually successful, once in, the commands time out consistently. It can take many, many attempts to sometimes get a successful responsive (if you're lucky). The CLI is similar but slightly better with maybe a success on the 6th or 7th attempt. Once a CLI command works, the preset commands then seem to work OK.

4np:
AFAIK remote management works over LoRa, not BLE. Maybe mesh packets are being dropped?

mjvolante:
The issue even happens when there is a very strong LoRa link to the Repeater with no other LoRa packets potentially causing problems. Takes many many attempts and a CLI command seems to free things up.

recrof:
> The issue even happens when there is a very strong LoRa link to the Repeater with no other LoRa packets potentially causing problems. Takes many many attempts and a CLI command seems to free things up.

and on top of that, it will work on different companion just next to the first one. it's very hard to simulate tho.

4np:
A LoRa radio cannot receive when it is busy sending, so repeaters will occasionally be deaf to Mesh packets. Could that potentially be at play here? Does changing txdelay / rxdelay and / or direct.txdelay / direct.rxdelay cause a change in the behavior? And what if you disable repeat mode (set repeat off) to try to rule out tx induced deafness?

What hardware you using @mjvolante ?

mjvolante:
This behaviour happens with no TX whatsoever.

the-antz:
I have the same issue on a Station G2, currently on 1.10.0 but had that on 1.9.0, too.

Pings reliably come back with a SNR of >10dB (direct link), login and status page refresh work basically every time, but cli commands or preset commands in the settings tab fail most of the time, with the exception of requesting the neighbour list, which usually works fine. Once one request succeeds, further requests will work fine for about one minute, then it's back to failing, even for short packets like for get-version.

Something else I noticed is that when I refresh the status page, send a get-version request, then refresh the status page again, it shows the router received 2 packets and sent 1.
Does someone know when the packet counters are updated? Is that before or after the request is dealt with?

LitBomb:
Scott posted on discord that he pushed a fix to the dev branch for this.  

> I just pushed to dev a pretty important fix if you've ever been frustrated with remote repeater admin, where the login works, but then CLI commands just get no responses, and Get Stats just gets timeouts. Maybe get the next nightly, and let me know how it goes.

[...]

Issue #983, opened 2025-10-21
Cannot remote admin Repeater from companion


askpatrickw:
Hardware: Seeed Sensecap Solar
Version: 1.9.1

Some of in us PugetMesh have noticed this issue on a few repeaters.

The repeater works as a repeater fine.
You can login to it from a companion.
You can request stats.
You cannot:

- Send any CLI commands. No response.
- Use any of the settings pages. Time out.

Also tried:

- Rebooting
- Upgrading from 1.9.0 to 1.9.1

Tested Companion:

**Heltec v3**
  - USB connection to MacOS
  - Meschcore web app over USB

**LilyGo TDeck**

--- 

I have pretty easy access to my repeater if there is a debug build or logs I can gather.

askpatrickw:
This might be a dupe of https://github.com/meshcore-dev/MeshCore/issues/947 

That other issues assumes this is a Heltec specific bug, but I can reproduce from a TDECK as well.

lone-baggie:
same issue on two Seeed Xiao NRF52 repeaters  From Android companion, you can only read the status. If you try anything else, it will time out. However, if I use a web client companion and connect to a node via USB, I can then control the Seeed repeaters without issue

smiba:
Same issue here with the Uniteng Station G2 (ESP32) as repeater, and Uniteng Nano G2 Ultra (nRF) as client

recrof:
newest repeater firmware(1.11.0) has some bugfixes that might help with this problem

smiba:
Unfortunately the 1.11.0 firmware is the one I'm running, only started using MeshCore today.

recrof:
> Unfortunately the 1.11.0 firmware is the one I'm running, only started using MeshCore today.

then I'm sure this is something different.

Jesse-Dur:
Any updates? I am having the same issue on a NRF52 repeater. Remote administration was working and then started the exact same timeout on Settings & CLI pages (no settings changed for this to start occuring).

liamcottle:
Try resetting the path. Mobile app v1.35.0 has a new "Reset Path" button in the top right dropdown menu when viewing the repeater settings tab. I've had to do this a few times, even after a successful login.

Jesse-Dur:
Hi Liam,
I'm on version v1.35.0 on the android app & have also tried on webapp on a Mac (Edge browser) also tried from 2 different companion nodes.
[...]

PR #1299, opened 2026-01-01
BUGFIX: Companion packet timestamp mismatch trips replay protection

entr0p1:
Initial login packet and REQ packets use the timestamp from the companion node's RTC. Most subsequent text packets using the timestamp from the companion node's mobile app. Drift between these timestamps can trip the replay protection and cause timeouts after login on the client side. 

Fix changes this logic to send the node's RTC timestamp for TXT_TYPE_CLI_DATA messages instead of the timestamp from the app (matches the sendRequest() code logic).

Ideally we would use the same timestamp everywhere and this does work in my testing, however it breaks the "Heard <n> repeats" in the mobile app (presumably because the timestamp is part of the packet hash and the heard packet doesn't match the one the app thinks was sent).

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -903,6 +903,7 @@ void MyMesh::handleCmdFrame(size_t len) {
       int result;
       uint32_t expected_ack;
       if (txt_type == TXT_TYPE_CLI_DATA) {
+        msg_timestamp = getRTCClock()->getCurrentTimeUnique(); // Use node's RTC instead of app timestamp to avoid tripping replay protection
         result = sendCommandData(*recipient, msg_timestamp, attempt, text, est_timeout);
         expected_ack = 0; // no Ack expected
       } else {


liamcottle:
Regarding the heard repeats in the app, ideally we don't break that functionality.

This is probably the reason for these related issues linked below:
- https://github.com/meshcore-dev/MeshCore/issues/983
- https://github.com/meshcore-dev/MeshCore/issues/1140

The app makes heavy use of sender timestamp and message content as a unique id for messages.
Mostly used for DM dedupe and heard repeats in the mobile app.

However for CLI commands, this change is probably OK, since the app doesn't care about the outgoing timestamp for CLI commands.

entr0p1:
Hey mate!

> Regarding the heard repeats in the app, ideally we don't break that functionality.
Yep, agreed. A fix isn't a fix if it breaks something else. 

> This is probably the reason for these related issues linked below:
> 
>     * [Cannot remote admin Repeater from companion #983](https://github.com/meshcore-dev/MeshCore/issues/983)
> 
>     * [Remote Management via app unresponsive #1140](https://github.com/meshcore-dev/MeshCore/issues/1140)
Highly likely. So far my tests have had significant improvements remotely managing repeaters. I discovered the issue by accident because I was sending commands that were timing out to a repeater I just so happened to be debugging at the time. Saw the debug log for replay protection on every packet I sent that was timing out. 


> The app makes heavy use of sender timestamp and message content as a unique id for messages. Mostly used for DM dedupe and heard repeats in the mobile app.
> 
> However for CLI commands, this change is probably OK, since the app doesn't care about the outgoing timestamp for CLI commands.

Yep, I found that out the hard way when I tested it for any text packet. Ended up narrowing the change to just the CMD packets which seems to not break anything in the app. 

I do think for consistency sake it would be good to use the RTC from the node for all packets, if that can be coordinated in a way that the app knows its happening. Having said that I'm not sure if that's even technically possible or introduces an unreasonable level of complexity.

IoTThinks:
It seems my CLI is back responsive on my T1000-e companion.
I used to reboot my T1000-e companion every 2 days when the CLI keeps ignore my CLI commands.

Let me monitor more.

PR #1307, opened 2026-01-02
Add Washtastic with E22P to Firmware list on Flasher.

yellowcooln:
The ProMicro nRF52 is used on the [Washtastic](https://github.com/valzzu/meshtastic-pcbs/tree/main/WashTastic) devices. And with the release of the E22P chip, we have had to modify the firmware to make it work.

This PR is asking that a custom firmware under the flasher be created for the Washtastic board when paired with a E22P.

We have created a firmware and testing it here in [Greater Boston Mesh](https://bostonme.sh/). 

Firmware: https://github.com/yellowcooln/MeshCore-washtastic-e22p/releases/tag/v.1.11.0


aeblyve:
Firmware works well for me in testing on E22P.

gfdrich:
I have tested this out on my E22P as well and is working for me too

PR #1308, opened 2026-01-03
Fix: WiFi Interface Frame Parsing

liamcottle:
This PR implements frame header parsing for the TCP/WiFI interface used in the companion firmware.

Currently, it's possible for the WiFi firmware to crash if you send TCP packets faster than the firmware can process them.
This happens due to `client.available()` having more bytes available than the `MAX_FRAME_SIZE` allows for.
The current implementation tries to copy all of the available bytes into the buffer, resulting in overflow/stack smashing.

I've added parsing of the protocol provided frame header, which tells us how many bytes we should expect to follow.
If we haven't received enough bytes yet, which could happen on slow links, it will wait for the next loop and check again.

I've been working on fast channel syncing in the mobile app, which works on BLE connections, however noticed the WiFi firmware would crash due to the large amount of bytes available.

In addition to this, I've also added new checks to ensure frames that are of unexpected type, or are larger than the `MAX_FRAME_SIZE` are skipped.

Requesting review from @fdlamotte as you were the original author ;)

Screenshot 2026-01-03 at 1 12 27 AM

### DIFF
--- a/src/helpers/esp32/SerialWifiInterface.cpp
+++ b/src/helpers/esp32/SerialWifiInterface.cpp
@@ -108,17 +108,8 @@ size_t SerialWifiInterface::checkRecvFrame(uint8_t dest[]) {
         if(client.available() >= frame_header_length){
 
           // read frame header
-          uint8_t frame_header[3];
-          client.readBytes(frame_header, frame_header_length);
-
-          // parse frame header
-          uint8_t frame_type = frame_header[0];
-          uint16_t frame_length;
-          memcpy(&frame_length, &frame_header[1], 2);
-
-          // we have received a frame header
-          received_frame_header.type = frame_type;
-          received_frame_header.length = frame_length;
+          client.readBytes(&received_frame_header.type, 1);
+          client.readBytes((uint8_t*)&received_frame_header.length, 2);
 
         }
 
@@ -161,12 +152,8 @@ size_t SerialWifiInterface::checkRecvFrame(uint8_t dest[]) {
           return 0;
         }
 
-        // read frame data
-        uint8_t frame_data[MAX_FRAME_SIZE];
-        client.readBytes(frame_data, frame_length);
-
-        // copy frame to provided buffer
-        memcpy(dest, frame_data, frame_length);
+        // read frame data to provided buffer
+        client.readBytes(dest, frame_length);
 
         // ready for next frame
         resetReceivedFrameHeader();


src/helpers/esp32/SerialWifiInterface.cpp
```diff
@@ -86,13 +89,83 @@ size_t SerialWifiInterface::checkRecvFrame(uint8_t dest[]) {
...
+          uint8_t frame_header[3];
+          client.readBytes(frame_header, frame_header_length);
+
+          // parse frame header
+          uint8_t frame_type = frame_header[0];
+          uint16_t frame_length;
+          memcpy(&frame_length, &frame_header[1], 2);
+
+          // we have received a frame header
+          received_frame_header = new FrameHeader();
```
ripplebiz: Shouldn't use dynamic allocation (other than in setup). Is a basic no-no in this repo.

src/helpers/esp32/SerialWifiInterface.cpp
```diff
@@ -86,13 +98,82 @@ size_t SerialWifiInterface::checkRecvFrame(uint8_t dest[]) {
...
+            uint8_t skip[1];
+            int skipped = client.read(skip, 1);
+            frame_length -= skipped;
+          }
+          resetReceivedFrameHeader();
+          return 0;
+        }
+
+        // read frame data
+        uint8_t frame_data[MAX_FRAME_SIZE];
```
ripplebiz: could be a bit kinder on the stack and remove frame_data local array.
then just do:
   client.readBytes(dest, frame_length);

src/helpers/esp32/SerialWifiInterface.cpp
```diff
@@ -86,13 +98,82 @@ size_t SerialWifiInterface::checkRecvFrame(uint8_t dest[]) {
...
+          client.readBytes(frame_header, frame_header_length);
+
+          // parse frame header
+          uint8_t frame_type = frame_header[0];
+          uint16_t frame_length;
+          memcpy(&frame_length, &frame_header[1], 2);
+
+          // we have received a frame header
+          received_frame_header.type = frame_type;
+          received_frame_header.length = frame_length;
```
ripplebiz: could simplify this with:
   client.readBytes(&received_frame_header.type, 1);
   client.readBytes(&received_frame_header.length, 2);

fdlamotte (approved):
Hi,
Good for me, hopefully it'll solve some issues @awolden had in meshcore_py with persistent connexions ;)

PR #1309, opened 2026-01-03
Only add adverts that have location

theraspb3rry:
disclaimer: im not a c developer

Friend of mine (silentlightning of the VIC mesh) does lots of repeater installs and is running into issues with contact limits. To reduce this he suggested to only add repeaters with a location, and i've built and tested this on a t114 in companion mode and it does seem to work, I only received adverts from repeaters (and clients actually) that had locations.

Note that this code enables this functionality by default (because i think the app needs to be told of the function, and this was a quick thing, so i just set it to true to flash it and test it)

let me know if you need anything else :)

edit: i dont know how far you guys would take this, but you could also have an option for example to accept zero-hop adverts, but not repeated adverts to allow people to discover each other easier without getting full contact lists. At that point you could even make a "filter ruleset" that filters incoming adverts based on name (contains, prefix, suffix), if it has a location or not, its path and path length, etc.

### DIFF
--- a/src/helpers/BaseChatMesh.cpp
+++ b/src/helpers/BaseChatMesh.cpp
@@ -104,13 +104,20 @@ void BaseChatMesh::onAdvertRecv(mesh::Packet* packet, const mesh::Identity& id, 
       return;
     }
 
+    if (isAutoAddLocEnabled() && !parser.hasLatLon()) { return; } // Ignore advert if it has no Location
+
     is_new = true;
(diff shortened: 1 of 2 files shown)


liamcottle:
Hey, thanks for the PR.

I'm not sure this is something we are looking to add in at this stage.
There have already been a few changes to the automatic contact adding, which allows for automatically replacing oldest non-favourite contacts when the contacts list is full and a new contact is discovered.

I'll close this out for now :)

PR #1314, opened 2026-01-03
Add GPS support for SenseCap Solar variant

Confituurke:
This PR adds GPS support for the SenseCap Solar variant using the L76KB GPS module.

## Changes
- Enable GPS support via `ENV_INCLUDE_GPS` flag in platformio.ini
- Add MicroNMEALocationProvider and wire it to EnvironmentSensorManager in target.cpp
- Use existing GPS pin definitions in variant.h (TX, RX, EN, STANDBY)
- Define `GPS_BAUD_RATE` in variant.h for EnvironmentSensorManager (Serial1 baud rate)
- Use `PIN_GPS_EN` instead of `GPS_EN` in variant.h for consistency with other GPS pins and variants
- Remove obsolete GPS init from variant.cpp (MicroNMEALocationProvider manages the enable pin)
- GPS can be enabled/disabled via device settings; `gps off` powers down the module to save current

## Hardware
- **GPS Module:** L76KB
- **GPS Pins:**
  - TX: Pin 7 (Serial1 RX)
  - RX: Pin 6 (Serial1 TX)
  - EN: Pin 18 (GPS enable/power control)
  - STANDBY: Pin 0 (not used)

## Testing
- GPS functionality tested and working
- GPS can be toggled on/off via device settings
- Power is saved when GPS is disabled

### DIFF
--- a/variants/sensecap_solar/platformio.ini
+++ b/variants/sensecap_solar/platformio.ini
@@ -8,7 +8,6 @@ build_flags = ${nrf52_base.build_flags}
   -I lib/nrf52/s140_nrf52_7.3.0_API/include/nrf52
   -I variants/sensecap_solar
   -I src/helpers/nrf52
-  -D ENV_INCLUDE_GPS=1
   -D NRF52_PLATFORM=1
   -D RADIO_CLASS=CustomSX1262
   -D WRAPPER_CLASS=CustomSX1262Wrapper


Habbie:
My SenseCap P1 Pro is now running a build of this PR. GPS works, giving me location and time. I do not have a BME280 I think (at least, I did not add one myself - I get a temperature reading but not humidity).

Thank you!

Confituurke:
> My SenseCap P1 Pro is now running a build of this PR. GPS works, giving me location and time. I do not have a BME280 I think (at least, I did not add one myself - I get a temperature reading but not humidity).
> 
> 
> 
> Thank you!

Thank you for testing! Much appreciated.

To the repo maintainers, if any changes are needed, let me know.

radiobuisje-cmyk:
I've also installed it, and I'm getting a proper lock on satellites, time, and position. Turning it off with GPS on/off also works.

Now I just need to find someone who can fix the PWR on/off button :)

Habbie:
I wonder (and I can't compare to before - this GPS patched build is the first thing I installed after unpacking my sensecap) if getting location + time at some frequency (don't know how often this happens) is drawing too much power for the panel+battery to keep up in this winter weather.

Habbie:
Based on the discussion in https://github.com/meshcore-dev/MeshCore/pull/1353 I will likely remove the GPS from my SenseCap. Of course this PR is still useful - some people may have a different power setup, or want the GPS because their SenseCap is on a boat or something.

In a build with this PR, if GPS is not turned on, does the GPS still draw power?

Habbie:
> In a build with this PR, if GPS is not turned on, does the GPS still draw power?

I guess

> EN: Pin 18 (GPS enable/power control)

answers that!

radiobuisje-cmyk:
I also haven't really looked at how much more this requires from the battery. And with the changeable weather here, it is also difficult to measure. But with GPS switched on, I can easily get more than a week of uptime without sun. And charging when the sun shines on it is actually quite fast.

etrever:
GPS enabled is working well for me here at 50 degrees north, panel is still keeping up with charging the batteries despite some days of cloud and trees shading the unit parts of the day.  Happy to have the clock set automatically now.

Habbie:
Are you sure it's actually setting the clock automatically outside of an explicit `gps sync` command? Somebody mentioned to me that there appears to be no code to pick the clock up automatically/periodically when GPS is on.

etrever:
Yup.   It pulls the time from the gps.

```
> gps off
ok
> clock
00:18 - 27/1/2026 UTC
> reboot
> clock
10:52 - 15/5/2024 UTC
> gps
off
> clock
10:53 - 15/5/2024 UTC
> gps on
ok
> clock
10:53 - 15/5/2024 UTC
> gps
on, active, no fix, 0 sats
> gps
on, active, fix, 5 sats
> clock
00:20 - 27/1/2026 UTC
```

dowjames:
why hasn't this been merged yet? It works great!

alexsporn:
Rebased this locally on top of 1.12.0 and it works great.

Confituurke:
# PR #1314 – Comment (follow-up)

**Follow-up (branch updated):**

- **Version:** Bumped to **v1.12.0**.
- **Cleanup:** Removed the unnecessary `GPS_EN` alias in `variant.h`.
- **LED defines:** Corrected LED naming in `variant.h` / `variant.cpp`:
  - **White LED** (pin 11) → `LED_WHITE` renamed from `LED_GREEN`
  - **Blue LED** (pin 12, LoRa TX) → `LED_BLUE`

No change to GPS behavior; only version bump, alias removal, and fix for LED naming/defines.

---

Users who want working buttons to activate/deactivate GPS, send advert, or power down the solar node can use the branch: [feature/sensecap-headless-btn-led](https://github.com/aXistem-dev/MeshCore/tree/feature/sensecap-headless-btn-led). We are not opening a PR for that branch yet, as it introduces more extensive changes to source libraries and shared files.

recrof:
the only problem I have with this PR is that GPS does not power off itself after getting the correct time. repeater should never move, so it doesn't make much sense to keep it on all the time.

dowjames:
Who cares? You can turn it off manually or leave it on. Not like it uses a lot of energy.

There are mobile repeaters too.. which would need it on.

recrof:
> Who cares? You can turn it off manually or leave it on. Not like it uses a lot of energy.

I care.

dowjames:
I get that, but that behaviour you want isn't implemented in any of the devices currently.
So it's really a moot point.

Confituurke:
> the only problem I have with this PR is that GPS does not power off itself after getting the correct time. repeater should never move, so it doesn't make much sense to keep it on all the time.

I appreciate and understand your feedback.

We can implement a new `GPS_OFF_AFTER_FIX` flag for repeaters and room server with GPS for example.
However this will introduce something new to `src/helpers/sensors/EnvironmentSensorManager.cpp`, and I wasn't eager to develop something having an impact on this level of the codebase. But I'll look into it.

Another question we should also ask ourself is; do we want it to autostart GPS on reboot or not.
Since it will turn itself down after clock sync, it will be off through a reboot too. So we might want to implement a second flag for that as well.

Is this the right direction?

recrof:
> Is this the right direction?

My idea was to keep the gps flag unchanged after fix/sync/shutdown, promising to always sync clock after reboot. only not doing anything with gps if gps is set to off.

Confituurke:
@recrof 
Thanks for the feedback on the GPS power off idea

So the plan is more to have a **power-save option**: when GPS is on (per the setting), we get a fix, sync the clock, then power down the GPS to save current. We **keep the `gps` setting as-is**, so on next boot we bring GPS up again for clock sync and then power it down again. So it’s not “GPS on all the time”, just on when we need it.

We can do it without big changes (e.g. an extra flag for the build), but then we have “user wants GPS on” vs “hardware is actually on or off” — they can diverge. And as you've seen with @dowjames ' reply, it doesn't fit for everyone. Needs a bit of thinking in settings, and we might add something like **`gps saver on`** / **`gps saver off`**.

Give me a couple of days to let this sync in and find a good approach.

In the meantime I’ve pushed what I have (a bit more heavy-handed than I’d like). In [b7d71cb](https://github.com/aXistem-dev/MeshCore/commit/b7d71cba246fc754fb82cf5c28a65c40fe98d410) and [99454dc](https://github.com/aXistem-dev/MeshCore/commit/99454dc1d1f6d90b9b20df2a09adbeb85f8286f9): with **`GPS_POWER_SAVE`** and **`gps on`** + **`gps saver on`** you get the behavior you asked for. One catch: **`gps`** status shows **`off`** after we power down, so it’s not obvious if that’s “hardware off for power-save” or “user turned GPS off”. I want to make that clearer when I refine it.

recrof:
thanks @Confituurke, makes sense.

spfmoby:
Something interesting could also be to work on a way to sync the clock by turning the GPS on once a day or something like that to correct time drift due to the powersaving modes.
I have no idea how to implement that, maybe based on what "24 hours seems to be according to the internal clock" even if it has drifted, it is not very important, the main goal being having :
- only very few gps on time to save power (could be useful for every repeater with gps, not only the P1)
- from time to time a clock sync with the GPS (and a position update). It can be 12 hours or 3 days I don't think it is very important.
- it may also help to set an accurate altitude for nodes with a BME280, because the altitude is off because it has no way to be calibrated.

recrof:
> way to sync the clock by turning the GPS on once a day or something like that to correct time drift due to the powersaving modes. 

the main problem with this is: you can't set time backwards as it breaks timers and statistics. we would need complete overhaul of how time is stored/used in MC for this to be viable. right now, if you set the time that is in past, board needs to be rebooted.

variants/sensecap_solar/platformio.ini
```diff
@@ -8,7 +8,7 @@ build_flags = ${nrf52_base.build_flags}
   -I lib/nrf52/s140_nrf52_7.3.0_API/include/nrf52
   -I variants/sensecap_solar
   -I src/helpers/nrf52
-  -UENV_INCLUDE_GPS
+  -D ENV_INCLUDE_GPS=1
```
alexsporn: this env and the MicroNMEA lib_dep should already be part of `sensor_base`, so there should be no need to add these lines here

Confituurke: Thank you for the feedback. I will modify this since it is indeed provided in `sensor_base`.
After reviewing it seems `stevemarple/MicroNMEA` is also redundant.

Your feedback is appreciated.

Confituurke:
Closing this PR since the same approach was merged in #1589.

We will open a new PR for the GPS power-saving feature (boot-only mode for repeaters).

PR #1316, opened 2026-01-03
Allow negative tx

weebl2000:
Change `tx_power_dbm` from `uint8_t` to `int8_t` to support negative TX power values. SX1262 for example allows down to -9 dBm.

- `NodePrefs.tx_power_dbm`, `setTxPower()`, and `radio_set_tx_power()` across all variants: `uint8_t` → `int8_t`
- TX power constraint lower bound: `1` → `-9`
- `get tx` CLI prints as signed value

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -1228,10 +1228,11 @@ void MyMesh::handleCmdFrame(size_t len) {
       writeErrFrame(ERR_CODE_ILLEGAL_ARG);
     }
   } else if (cmd_frame[0] == CMD_SET_RADIO_TX_POWER) {
-    if (cmd_frame[1] > MAX_LORA_TX_POWER) {
+    int8_t power = (int8_t)cmd_frame[1];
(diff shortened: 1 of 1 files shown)


liamcottle:
Hey, thanks for the PR. Have you tested setting `-9` tx power through the app and confirmed that it saves as still transmits?

weebl2000:
> Hey, thanks for the PR. Have you tested setting `-9` tx power through the app and confirmed that it saves as still transmits?

Had to parse as signed int, then it works. But the app needs to be changed to read the response as signed int too. Setting -9 now shows as 247 dBm - because the app is reading it as an unsigned int I guess.

Just a visual bug though, my companion still sends at -9 dBm and that works.


https://github.com/user-attachments/assets/feab8ca3-de13-4998-a6e9-6e5b6ba2140d

liamcottle:
Great, thanks! Merged in. I've also updated the app to read as a signed int. This should be working through the web app if you force reload the page. It will also be available from the next mobile app update v1.39.0, older clients will just show as a high positive number.

PR #1318, opened 2026-01-04
Add null checks for PMU allocation in TBeamBoard

cj-vana:
## Summary
Add null safety checks after `new` operations for PMU initialization in TBeamBoard.cpp.

## Changes
- Check if `new XPowersAXP2101()` succeeded before calling `init()`
- Check if `new XPowersAXP192()` succeeded before calling `init()`

## Why
In embedded systems with limited RAM, `new` can fail and return nullptr. Dereferencing nullptr causes a crash. This adds defensive null checks to prevent potential crashes.

```cpp
// Before
if (!PMU->init()) {

// After  
if (!PMU || !PMU->init()) {
```

### DIFF
--- a/src/helpers/esp32/TBeamBoard.cpp
+++ b/src/helpers/esp32/TBeamBoard.cpp
@@ -131,7 +131,7 @@ bool TBeamBoard::power_init()
     #else
       PMU = new XPowersAXP2101(PMU_WIRE_PORT, PIN_BOARD_SDA, PIN_BOARD_SCL, I2C_PMU_ADD);
     #endif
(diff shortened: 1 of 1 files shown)


weebl2000:
@cj-vana
The null check after new is unnecessary here.
In this build configuration new does not return nullptr, instead allocation failure triggers abort (or throws std::bad_alloc in standard C++).

Since new is not used with std::nothrow, the !PMU branch cannot occur, so if (!PMU || !PMU->init()) is effectively equivalent to if (!PMU->init()).

@liamcottle I think this one can be closed.

PR #1326, opened 2026-01-06
Fix: allow clock sync to correct time drift after sleep

csrutil:
ESP32's internal RTC drifts during light sleep due to the 150kHz RC oscillator. When the device clock runs ahead, the previous logic rejected correct time from mobile because "clock cannot go backwards". This prevented fixing drifted clocks.

Removed the forward-only restriction from both `clock sync` and `time` commands to allow syncing to the correct time regardless of direction.

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -135,15 +135,10 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
       _callbacks->sendSelfAdvertisement(1500);  // longer delay, give CLI response time to be sent first
       strcpy(reply, "OK - Advert sent");
     } else if (memcmp(command, "clock sync", 10) == 0) {
-      uint32_t curr = getRTCClock()->getCurrentTime();
-      if (sender_timestamp > curr) {
-        getRTCClock()->setCurrentTime(sender_timestamp + 1);
-        uint32_t now = getRTCClock()->getCurrentTime();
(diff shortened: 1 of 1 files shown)


IoTThinks:
I have tried similar approach before with "clock fsync".
The clock is synced and shown correctly internally in ESP32.

However, when I login into the repeater, the Status still shows old time for Last Login.
I need to reboot the repeater to see the synced Last Login.

Have you managed to figure out / fix it?

leon332157:
Bump! Very annoyed by the fact that I can't sync my clock backwards

IoTThinks:
Go ahead with the PR. 
Eventually, any boards can drift forward or backward. It is just a matter of time.
So this PR will be helpful.

When we sync backward, then we need to wait for the current time to reach the drift time to be shown correctly on Last login.
No big issue.

On the other hand, I'm working on a fix to minimize the time drift.

beachmiles:
Have to type "**clkreboot**" in the repeater console and only then can you resync the clock and the app and/or firmware won't complain about the clock already being synced.
The esp32-s3 at least keeps its time when doing a soft reboot so the wrong future clock will persist.
Looks like the powersaving mode is what is causing this clock drift as I didn't experience this behavior.
This is an unfortunate bug

IoTThinks:
> Looks like the powersaving mode is what is causing this clock drift as I didn't experience this behavior.
> This is an unfortunate bug

Cosmestic bug for now.
I will push a fix to avoid / reduce much of the time drift.

PR #1331, opened 2026-01-06
OFFLINE_QUEUE_SIZE for Heltec Wifi companions

an0key:
OFFLINE_QUEUE_SIZE missing from h3/h4 wifi companions, causing them only store 16 messages. Adding this to the build flags (as they are for the BLE) resolves this.


olanwe:
Isn’t this change also relevant for the other WiFi companions as well? It seems to be missing in all other environments too.

an0key:
> Isn’t this change also relevant for the other WiFi companions as well? It seems to be missing in all other environments too.

I only have a Heltec V3 and V4 I am able to test the change on, so thats all I have modified.

liamcottle:
Thanks! Looks like we missed the WiFi variants when the upgrade was done. Was wondering why my remote WiFi node never had much history... Merged :)

olanwe:
Hi @liamcottle, shouldn’t we apply this change to the other WiFi companions as well? Unfortunately, I only have a v3 and v4 available for testing, but the ESPs themselves are the same. (Though it’s possible that some specific variants don’t support this queue size?)

Issue #1223, opened 2025-12-15
Proposal: Probabilistic Flood Forwarding for Advertisement Packets


ViezeVingertjes:
Large mesh networks experience severe congestion from advertisement floods. Every node broadcasts advertisements periodically (every 6-12 hours), and with traditional flooding, these propagate through the entire network with 100% forwarding probability at every hop.

## Proposed Solution

Implement **hop-based probabilistic forwarding** for flood advertisement packets. Forwarding probability decreases exponentially with hop count, dramatically reducing redundant transmissions while maintaining good network coverage.

**Formula**: `P(h) = 0.308^(h-1)` where h = hop count

## Results

Based on genetic algorithm optimization across realistic network sizes (25-800 nodes, averaged across all scenarios):

| Metric | Current (Baseline) | With Probabilistic Forwarding | Improvement |
|--------|-------------------|-------------------------------|-------------|
| **Transmissions per advert** | 6,545 | 452 | **93.1% reduction** |
| **Network coverage** | 100% | 68.2% | Sufficient for adverts |
| **Average max hops** | 7.4 | 4.2 | **43% faster** |
| **Efficiency** | 9.4% | 55.8% | **6× improvement** |

### Real-World Impact Example

**Scenario**: Medium-sized mesh network  
**Network size**: 200 nodes  
**Flood advertisement interval**: 12 hours

**Calculation of daily advertisement load**:
- Each node floods 1 advert every 12 hours = 2 adverts/node/day
- Network total: 200 nodes × 2 adverts/day = **400 advertisements/day**

**Impact per single advertisement**:
- Baseline: 6,545 transmissions per advert
- Optimized: 452 transmissions per advert
- Reduction: 6,093 fewer transmissions (**93.1% reduction**)

**Daily network-wide impact**:

| Metric | Before (Baseline) | After (Probabilistic) | Calculation |
|--------|-------------------|----------------------|-------------|
| Transmissions per advert | 6,545 | 452 | From simulation |
| **Total transmissions/day** | **2,618,000** | **180,800** | 400 adverts × transmissions |
| **Transmission reduction** | **93.1%** | **-** | (2,618,000 - 180,800) / 2,618,000 |

This massive reduction frees up airtime for actual messages, dramatically reducing congestion and delays.

## How It Works

The forwarding probability decreases rapidly after the first hop:

| Hops | Probability | Meaning |
|------|-------------|---------|
| 0    | 100%        | Source always transmits |
[...]

PR #1338, opened 2026-01-07
Limit flood advert packet forwarding

mattzzw:
This implements #1223.
See also discussion in #1217 about the flood advert issue.
Thanks to @ViezeVingertjes for taking the time to come up with the proposal.

Added cli command:
`set flood.advert.base x: float in [0,1], defaults to 0.308`, 0 will disable flood advert forwarding, 1 forwards all.


michielappelman:
I had spent some tokens on getting this implemented, but now I'm not sure the probabilistic forwarding is a good way forward, it inherently makes it less predictable which might confuse/frustrate users. It's easier to say for example: "we limit adverts to 5 hops".

Also, consider from Discord: "Advert traffic should be mostly local. This is exactly what this algorithm does: 100% for the 1st hop, 31% for the 2nd, 10% for the 3rd, 3% for the 4th." - if those are the numbers that means that a repeater 2 hops over will receive approx. 1 in 10 flood adverts from another repeater. If those repeaters really want to exchange flood adverts, that would mean they would just try a bunch of times because at some point 1 will get through. In the mean time, they flooded the closer repeaters multiple times.

This is theoretical, but not unimaginable. I think predictability should prevail (we have enough uncertainty in the mesh already with current airtime challenges).

ViezeVingertjes:
I ran numerous simulations for those numbers, its probabilistic, so those percentages only are for 1 route.

{9727F71C-8CB6-467D-A762-D9C5FFF44703}

Because there are many routes for each message, it adds up to these numbers:

{01423655-6A5F-498F-ADA0-EE84B7030544}

But i agree, this isn't going to be accepted quickly hehe... for now, in NL we just drop all adverts, which is much easier and people will just have to deal with sending their adverts through another way.

mattzzw:
> If those repeaters really want to exchange flood adverts, that would mean they would just try a bunch of times because at some point 1 will get through. In the mean time, they flooded the closer repeaters multiple times.

This is not how it works. Flood packets are not waiting for an ACK like direct packets. They are fire-and-forget. If a flood advert is being dropped, the sending repeater will not know and will not retry sending.

michielappelman:
> > If those repeaters really want to exchange flood adverts, that would mean they would just try a bunch of times because at some point 1 will get through. In the mean time, they flooded the closer repeaters multiple times.
> 
> This is not how it works. Flood packets are not waiting for an ACK like direct packets. They are fire-and-forget. If a flood advert is being dropped, the sending repeater will not know and will not retry sending.

I know that's not how it works from a technical perspective, I was thinking more about the human-level: think 2 operators that are trying to learn each other's repeater. Instead of giving them a chance to exchange flood adverts, let's just force them to exchange information out-of-band.

mattzzw:
Thats why we have trace packets.

src/helpers/CommonCLI.cpp
```diff
@@ -642,6 +650,15 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
           _prefs->adc_multiplier = 0.0f;
           strcpy(reply, "Error: unsupported by this board");
         };
+      } else if (memcmp(config, "flood.advert.base ", 18) == 0) {
+        float f = atof(&config[18]);
+        if((f > 0) || (f<1)) {
```
weebl2000:
```suggestion
        if(f >= 0 && f <= 1) {
```
This needs to be &&

`f>0 || f<1` is effectively just `true` - all real numbers satisfy the condition.

examples/simple_repeater/MyMesh.cpp
```diff
@@ -403,6 +403,14 @@ bool MyMesh::allowPacketForward(const mesh::Packet *packet) {
...
     return false;
   }
+  // Limit flood advert paket forwarding using a probabilistic reduction defined by P(h) = 0.308^(hops-1)
+  // https://github.com/meshcore-dev/MeshCore/issues/1223
+  double_t roll_dice = (double)rand() / RAND_MAX;
+  double_t forw_prob = pow(_prefs.flood_advert_base, packet->path_len - 1);
+  if (packet->getPayloadType() == PAYLOAD_TYPE_ADVERT && packet->isRouteFlood() && roll_dice > forw_prob)
+    return false;
+
+  // all other packets
```
weebl2000:
```suggestion
  // Limit flood advert packet forwarding using a probabilistic reduction defined by P(h) = base^(hops-1)
  // https://github.com/meshcore-dev/MeshCore/issues/1223
  if (packet->getPayloadType() == PAYLOAD_TYPE_ADVERT && packet->isRouteFlood()) {
    double roll_dice = (double)rand() / RAND_MAX;
    double forw_prob = pow(_prefs.flood_advert_base, packet->path_len - 1);
    if (roll_dice > forw_prob)
      return false;
  }

```

pow() and rand() are calculated for every forwarded packet, but we only need it for flood advert. Also fixes 'paket' typo.

examples/simple_room_server/MyMesh.cpp
```diff
@@ -277,6 +277,15 @@ uint32_t MyMesh::getDirectRetransmitDelay(const mesh::Packet *packet) {
...
   if (packet->isRouteFlood() && packet->getPathHashCount() >= _prefs.flood_max) return false;
+
+  // Limit flood advert paket forwarding using a probabilistic reduction defined by P(h) = 0.308^(hops-1)
+  // https://github.com/meshcore-dev/MeshCore/issues/1223
+  double_t roll_dice = (double)rand() / RAND_MAX;
+  double_t forw_prob = pow(_prefs.flood_advert_base, packet->path_len - 1);
+  if (packet->getPayloadType() == PAYLOAD_TYPE_ADVERT && packet->isRouteFlood() && roll_dice > forw_prob)
+    return false;
+
+  // all other packets
```
weebl2000:
```suggestion

  // Limit flood advert packet forwarding using a probabilistic reduction defined by P(h) = base^(hops-1)
  // https://github.com/meshcore-dev/MeshCore/issues/1223
  if (packet->getPayloadType() == PAYLOAD_TYPE_ADVERT && packet->isRouteFlood()) {
    double roll_dice = (double)rand() / RAND_MAX;
    double forw_prob = pow(_prefs.flood_advert_base, packet->path_len - 1);
    if (roll_dice > forw_prob)
      return false;
  }

```

same as repeater, we only need it for flood adverts.

weebl2000 (changes requested):
Please see my comments. the `f>0 || f<1` statement is a bug - it's effectively just if (true)

mattzzw:
thanks @weebl2000 for reviewing

dreirund:
Note: [&nearr;&hairsp;Alternative suggestion](https://github.com/meshcore-dev/MeshCore/issues/2322):

Instead of probabilistic advert suppression that starts to increase with the second hop, unconditionally repeat the first advert heard, remember, and ignore subsequent adverts of the same ID (long enough ID prefix) until some time has passed.

PR #1343, opened 2026-01-08
Configurable packet forwarding for repeaters (types and hops)

michielappelman:
Adds configurable packet filtering for MeshCore repeaters to reduce airtime congestion on large mesh networks. This allows network operators to selectively control which packet types their repeaters forward, with granular sub-type filtering for advert packets and distance-based hop count limits.

## Motivation

Large mesh networks suffer from airtime congestion (see discussion in meshcore-dev#1217) when repeaters forward all packet types indiscriminately. This feature allows operators to:
- Block group messages on public repeaters
- Filter advert types by device category (chat, sensor, room, repeater)
- Limit advert propagation distance with hop count filtering (always excluding companion adverts)
- Emergency stop: disable all forwarding instantly with preserved filter configuration

## Core Changes

**New Header: `src/helpers/PacketTypeNames.h`**
- Centralized packet and advert type name definitions
- Provides helper functions for type name lookup

**New Preferences (offsets 170-173):**
- `uint16_t repeat_packet_types` - Bitmask for 16 packet types (bit=1 means allow)
- `uint8_t repeat_advert_types` - Bitmask for 5 advert sub-types (bit=1 means allow)
- `uint8_t advert_max_hops` - Maximum hop count for advert packets excluding from companions (0 = unlimited)

**Repurposed Preference:**
- `bool disable_fwd` - Now acts as a global on/off switch independent of bitmask settings

**Extended `allowPacketForward()` in MyMesh.cpp:**
1. Check flood-mode constraints
2. **Check global disable_fwd switch (preserves bitmask)**
3. Check packet type bitmask
4. For adverts: check sub-type bitmask and hop count limit (if not from a companion)
5. Return true to forward, false to block

## Design Decisions

 - **Preference Retention**: The `disable_fwd` preference now acts as a global on/off switch that preserves all bitmask filter settings. Users can configure granular filtering, then toggle repeating on/off without losing their custom configuration.
 - **Backwards compatible**: Migrates `disable_fwd=true` from old firmware to `repeat_packet_types=0x0000` on first load
 - **Positive Logic**: bit=1 means "allow" (intuitive: `set repeat on` enables forwarding)
 - **Permissive Default**: 0xFFFF repeats everything (backward compatible behavior)
 - **Auto-Enable Parent Type**: When enabling an advert sub-type, the advert packet type is automatically enabled if currently off.
 - **Flood-Mode Guard**: Hop count filtering only applies to flood-mode adverts (excluding companion adverts), since path_len is not meaningful for direct-mode packets.

## CLI Reference

- `get repeat` - Show global state (ON/OFF) and configured packet type filters
- `get repeat advert` - Show which advert sub-types are configured
- `get repeat advert.max_hops` - Show hop count limit
- `set repeat on` - Enable forwarding globally (uses configured packet filters)
- `set repeat off` - Disable forwarding globally (packet filters preserved)
- `set repeat {type} {on|off}` - Configure specific packet type filter
- `set repeat advert.{subtype} {on|off}` - Configure specific advert sub-type filter
- `set repeat advert.max_hops {0-255}` - Set hop count limit (0 = unlimited) for adverts (excluding from companions)

## Example Workflow

```bash
# Configure granular filters
set repeat txt on
set repeat grp.txt off
set repeat advert.sensor on
set repeat advert.chat off

# Check configuration
get repeat
> ON (filters: txt,advert)

# Temporarily disable repeating (settings preserved)
set repeat off
> OK - repeating disabled (packet filters preserved)

# Re-enable repeating (custom filters still active)
set repeat on
> OK - repeating enabled (using configured packet filters)

# Verify filters were preserved
get repeat
> ON (filters: txt,advert)
```

## Testing

All repeater builds pass and functionality has been validated on a Heltec v3 and T114 (no display).

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -318,15 +318,13 @@ File MyMesh::openAppend(const char *fname) {
 }
 
 bool MyMesh::allowPacketForward(const mesh::Packet *packet) {
-  if (packet->isRouteFlood() && packet->path_len >= _prefs.flood_max) return false;
-  if (packet->isRouteFlood() && recv_pkt_region == NULL) {
-    MESH_DEBUG_PRINTLN("allowPacketForward: unknown transport code, or wildcard not allowed for FLOOD packet");
+  if (_prefs.disable_fwd) {
     return false;
   }
 
-  // Check global disable_fwd switch first
-  if (_prefs.disable_fwd) {
-    MESH_DEBUG_PRINTLN("allowPacketForward: forwarding globally disabled");
+  if (packet->isRouteFlood() && packet->path_len >= _prefs.flood_max) return false;
+  if (packet->isRouteFlood() && recv_pkt_region == NULL) {
+    MESH_DEBUG_PRINTLN("allowPacketForward: unknown transport code, or wildcard not allowed for FLOOD packet");
     return false;
   }
 
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -356,8 +356,9 @@ bool MyMesh::allowPacketForward(const mesh::Packet *packet) {
       }
     }
 
-    // Check max hops for advert packets (0 = no limit)
-    if (_prefs.advert_max_hops > 0 && packet->path_len > _prefs.advert_max_hops) {
+    // Check max hops for flood-mode advert packets (0 = no limit)
+    // path_len is only meaningful for flood packets, not direct-mode packets
+    if (packet->isRouteFlood() && _prefs.advert_max_hops > 0 && packet->path_len > _prefs.advert_max_hops) {
       MESH_DEBUG_PRINTLN("allowPacketForward: advert exceeded max hops (%d > %d)",
                         packet->path_len, _prefs.advert_max_hops);
       return false;
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -332,7 +332,7 @@ bool MyMesh::allowPacketForward(const mesh::Packet *packet) {
     return false;
   }
 
-  // For advert packets, also check advert sub-type filtering
+  // For advert packets, check sub-type filtering and max hops
   if (pkt_type == PAYLOAD_TYPE_ADVERT) {
     const int app_data_offset = PUB_KEY_SIZE + 4 + SIGNATURE_SIZE;
 
@@ -353,15 +353,17 @@ bool MyMesh::allowPacketForward(const mesh::Packet *packet) {
                             mesh::getAdvertTypeName(adv_type));
           return false;
         }
-      }
-    }
 
-    // Check max hops for flood-mode advert packets (0 = no limit)
-    // path_len is only meaningful for flood packets, not direct-mode packets
-    if (packet->isRouteFlood() && _prefs.advert_max_hops > 0 && packet->path_len > _prefs.advert_max_hops) {
-      MESH_DEBUG_PRINTLN("allowPacketForward: advert exceeded max hops (%d > %d)",
-                        packet->path_len, _prefs.advert_max_hops);
-      return false;
+        // Check max hops for flood-mode advert packets (0 = no limit)
+        // path_len is only meaningful for flood packets, not direct-mode packets
+        // CHAT adverts are excluded from hop count filtering
+        if (packet->isRouteFlood() && adv_type != ADV_TYPE_CHAT && _prefs.advert_max_hops > 0 &&
+            packet->path_len > _prefs.advert_max_hops ) {
+          MESH_DEBUG_PRINTLN("allowPacketForward: advert exceeded max hops (%d > %d)",
+                            packet->path_len, _prefs.advert_max_hops);
+          return false;
+        }
+      }
     }
   }
 
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -324,6 +324,12 @@ bool MyMesh::allowPacketForward(const mesh::Packet *packet) {
     return false;
   }
 
+  // Check global disable_fwd switch first
+  if (_prefs.disable_fwd) {
+    MESH_DEBUG_PRINTLN("allowPacketForward: forwarding globally disabled");
+    return false;
+  }
+
   // Check if this packet type is allowed to be repeated
   uint8_t pkt_type = packet->getPayloadType();
   if (pkt_type < mesh::MAX_PACKET_TYPES && (_prefs.repeat_packet_types & (1 << pkt_type)) == 0) {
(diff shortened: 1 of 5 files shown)


examples/simple_repeater/MyMesh.cpp
```diff
@@ -311,12 +311,53 @@ File MyMesh::openAppend(const char *fname) {
...
+          MESH_DEBUG_PRINTLN("allowPacketForward: %s advert type not allowed to repeat",
+                            mesh::getAdvertTypeName(adv_type));
+          return false;
+        }
+      }
+    }
+
+    // Check max hops for flood-mode advert packets (0 = no limit)
+    // path_len is only meaningful for flood packets, not direct-mode packets
+    if (packet->isRouteFlood() && _prefs.advert_max_hops > 0 && packet->path_len > _prefs.advert_max_hops) {
```
josch0: Hi Michiel,

I have been following the discussions in the German MeshCore Telegram group.

I have a comment on your code:
The restriction on the maximum number of hops affects all flood adverts, including those from companions. I think this is wrong. The adverts from companions are used to exchange the key for direct communication. This should not be restricted under any circumstances, as it is a fundamental functionality of MeshCore.
What use are 64 hops for message packets if the key is dropped after 10 hops, for example?

I would add the following to the if statement here: adv_type != ADV_TYPE_CHAT

michielappelman: That's a good point. What about other adverts? Eg. Room servers, should they be limited? Thinking about the best way to implement this.

josch0: My personal opinion is that repeaters, room servers, and sensors can all be restricted in the same way, because they usually have a regional relevance and there is no need to propagate the adverts across the entire network. (If you ask me, sensors shouldn’t send any adverts at all, or shouldn’t even exist in MeshCore.)

**Here is an additional thought:**
The number of hops does not say anything about the actual distance traveled. Based on the GPS coordinates transmitted in the advert, forwarding could also be blocked if the repeaters are more then X kilometers apart. This way, all repeaters in the immediate vicinity would be visible, regardless of how many hops away they are. And if a repeater does not include coordinates in the advert, the hop count can still be used as a fallback.

This would only require a few more lines in your code, but it would make it more transparent how far one can “see” a repeater (e.g., 50 km as a default).

image

image

usrflo: Measuring reachable repeater nodes I think 50km is a well selected distance; getting the public key by a repeater advert enables IMHO a repeater login and control commands and this is limited to reachable nodes anyway. The reachable repeaters are marked blue in the following image, the radius is at 50km.

grafik

In general I'd like to see limitations of floods by a different approach that considers positional information and still allows long distance communication. I will add an idea next.

michielappelman: I tend to agree (without having extensive experience with other adverts so far) so I've changed the code to exclude companion adverts. We can always adapt later.

With regard to your second idea, I am firmly opposed to having any type of GPS location exchange in repeater adverts: I think we should remove it from the promiscuous adverts completely, and allow it to be pulled from a repeater only when another nodes asks for it. It adds to many unnecessary bits on the air, and a repeater and its state should be kept as minimal as possible.

josch0: > With regard to your second idea, I am firmly opposed to having any type of GPS location exchange in repeater adverts: I think we should remove it from the promiscuous adverts completely, and allow it to be pulled from a repeater only when another nodes asks for it. It adds to many unnecessary bits on the air, and a repeater and its state should be kept as minimal as possible.

As long as the adverts are sent with GPS in their current form, you could use GPS in your PR as an interim solution for decision-making instead of just relying on hops.

You are already talking about a more advanced solution beyond this PR:
Removing GPS from the advert would make the advert even more useless than it already is. Why would we even need the flood advert at all then? I would rather suggest completely removing flood adverts from repeaters. Instead, one could rely entirely on a pull-based approach, where you click on a repeater in the path of a message and send a "whois." This way, you would even have the direct path to the repeater and wouldn’t need flooding at all.

michielappelman: I prefer not to introduce GPS decision-making in this PR to keep it simple. Let's start with this and we can improve in further PRs.

And I think we're actually thinking along the same lines 😊 Why DO we need flood adverts from repeaters, indeed! But let's also leave that idea out of this PR which can be seen as a short-term tool to help certain mesh operators. In the mid/long-term we will need to rethink this whole advert-flooding thing indeed.

etrever:
After setting up repeating preferences, are they retained  or restored after a “set repeat off” and “set repeat on”?   Description in pull request says they are probably lost.  (On phone right now, hard to look at code)

For user/admin convenience I’d want the global on/off toggle to not clear the custom settings.

michielappelman:
> After setting up repeating preferences, are they retained or restored after a “set repeat off” and “set repeat on”? Description in pull request says they are probably lost.

I didn't think of that! That means we'll continue to use the `disable_fwd` preference for what we've been using it. Let me implement this.

wesselah:
Just too make the comment, this change is potentially dangerous, i agree that flood adverts must be controlled and limited that is the only way to enjoy meshcore to the fullest.
But this changes makes it so complicated for users who just managed to get there  repeater up and running to controll flood behaviour .
The point is that they need to read tfm manual which at its best explains a little bit 
This change can destroy the mesh in crucial locations when that owner of the repeater fucks up which can easily be done .  Meshcore needs to be simple to use which it basically is at the moment so don't kill it by making it difficult. 
why not do automated throtteling  as suggested in other pull request 
And don't get me wrong the excessive flood averts that we see  need to be handled , but  not by depending  on the knowledge or willingness of people but in general so decide make it a mesh that stutters or make a meah that just  works by making important decisions that also supports big meshes

michielappelman:
Thinking about the comment above by @wesselah made me re-evaluate the proper way to solve this situation. Just blocking flood adverts from repeaters (or allowing operators to block them, as suggested in this PR) _will_ solve the flooding issue, but does not provide a good alternative to users to figure out their local/regional mesh. Instead, we should _default_ to blocking flood adverts from repeaters (without making it configurable) and allow users to request those adverts on-demand. 

This has now been proposed as part of #1374 and this PR is left as a draft for context, but can be closed when appropriate.

wesselah:
I appreciate the time you took to answer. I also think that you are creative comming up with solutions to damm or minimize flood adverts in whatever form because that is needed or meshes will collapse because of it own success. 
one sugestion i would like to make is that building solutions which need manual intervention by people who do not know as much as you about how the protocol works always seem to dammed to fail. I will try to explain why.
In general people are lazy esspecially when things just work as meshcore does at the moment. Looking away of the problems arising in big meshes. They do not read the fucking manual or try to understand why things are what they are. 
Why with all your knowledge can't you come up with a solution that does not need any manual intervention and just regulates the number of flood adverts passing thru a repeater  when its idle, busy or very busy taking into concern the configured airtime that is allowed because of local regulations.
If you differentiate based on what kind of flood advert it is and automatically throtle those flood adverts no manuel intervention is needed no pulling of information just the mesh will adapt to the traffic flows just like traffic lights which adjest traffic based on the number of cars arriving.
Again great to have involved software engineers who care but think about meshcore users as regular people not as tech safy engineers 
I hope i did give some ideas and take care

PR #1347, opened 2026-01-08
Powersaving on companion when serial is off (ble/wifi)

weebl2000:
**Build firmware:** [Build from this branch](https://mcimages.weebl.me/?commitId=powersaving-ble-companion)

**Update 27 February 2026**

I've been running this for nearly two months now on a range of ESP32 devices and have seen no issues. I think it's good to merge.

## Summary

Adds power saving to the companion radio when BLE/WiFi is disabled by the user. When the serial interface is off (and GPS is not enabled), the device enters ESP32-S3 light sleep for 30-minute intervals, significantly reducing power consumption for battery-powered companion nodes that are left running as pure relays.

## Changes

**Light sleep in companion radio** (`examples/companion_radio/main.cpp`)
- Enters 30-minute light sleep cycles when `serial_interface.isEnabled()` is false and GPS is off
- Wakes on: LoRa packet (ext1), timer (30min), or button press (GPIO wakeup)
- Stays awake briefly after wake to process any pending work before sleeping again
- Initial 2-minute delay before first sleep to allow setup/connection

**WiFi radio actually turns off** (`src/helpers/esp32/SerialWifiInterface.cpp/h`)
- `disable()` now calls `WiFi.disconnect(true)` and `WiFi.mode(WIFI_OFF)` to power down the radio
- `enable()` reconnects using stored SSID/password credentials
- Previously `disable()` only set a flag — the WiFi radio stayed powered on

**Button wakeup with correct polarity** (`src/helpers/ESP32Board.h`)
- `enterLightSleep()` accepts optional `pin_wake_btn` parameter (backward-compatible)
- Uses GPIO wakeup (`GPIO_INTR_LOW_LEVEL`) for button — correct for active-LOW buttons
- Keeps ext1 (`ANY_HIGH`) for LoRa DIO1 only — these can't share a mask since they need opposite trigger levels

**Pending work detection** (`examples/companion_radio/MyMesh.cpp/h`)
- `hasPendingWork()` checks outbound packet queue to avoid sleeping while packets are queued

## Notes

- Only affects ESP32-S3 targets (gated by `CONFIG_IDF_TARGET_ESP32S3`)
- Boards without `PIN_USER_BTN` gracefully skip button wakeup
- GPS-enabled nodes skip sleep entirely (GPS needs continuous operation)
- No changes to the repeater sleep path — `sleep()` still calls `enterLightSleep(secs)` without button

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -2257,7 +2257,7 @@ bool MyMesh::advert() {
   }
 }
 
-// To check if there is pending work
+// Check if there is pending work (packets to send or pending contacts write)
 bool MyMesh::hasPendingWork() const {
   return _mgr->getOutboundTotal() > 0 || dirty_contacts_expiry != 0;
 }
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -2418,7 +2418,7 @@ bool MyMesh::advert() {
   }
 }
 
-// To check if there is pending work
+// Check if there is pending work (packets to send or pending contacts write)
 bool MyMesh::hasPendingWork() const {
   return _mgr->getOutboundTotal() > 0 || dirty_contacts_expiry != 0;
 }
--- a/src/helpers/ESP32Board.h
+++ b/src/helpers/ESP32Board.h
@@ -15,6 +15,7 @@
 #include "soc/rtc.h"
 #include "esp_system.h"
 #include <driver/rtc_io.h>
+#include <driver/gpio.h>
 
 class ESP32Board : public mesh::MainBoard {
 protected:
@@ -68,6 +69,28 @@ public:
 
   uint32_t getIRQGpio() override {
     return P_LORA_DIO_1; // default for SX1262
+  }
+
+  void enterLightSleep(uint32_t secs, int pin_wake_btn = -1) {
+#if defined(CONFIG_IDF_TARGET_ESP32S3) && defined(P_LORA_DIO_1) // Supported ESP32 variants
+    if (rtc_gpio_is_valid_gpio((gpio_num_t)P_LORA_DIO_1)) { // Only enter sleep mode if P_LORA_DIO_1 is RTC pin
+      esp_sleep_pd_config(ESP_PD_DOMAIN_RTC_PERIPH, ESP_PD_OPTION_ON);
+
+      esp_sleep_enable_ext1_wakeup((1L << P_LORA_DIO_1), ESP_EXT1_WAKEUP_ANY_HIGH); // Wake on LoRa packet
+
+      // Wake on button press (active-LOW: pin is HIGH when idle, LOW when pressed)
+      if (pin_wake_btn >= 0) {
+        gpio_wakeup_enable((gpio_num_t)pin_wake_btn, GPIO_INTR_LOW_LEVEL);
+        esp_sleep_enable_gpio_wakeup();
+      }
+
+      if (secs > 0) {
+        esp_sleep_enable_timer_wakeup(secs * 1000000); // To wake up every hour to do periodically jobs
+      }
+
+      esp_light_sleep_start(); // CPU enters light sleep
+    }
+#endif
   }
 
   void sleep(uint32_t secs) override {
--- a/src/helpers/ESP32Board.h
+++ b/src/helpers/ESP32Board.h
@@ -16,6 +16,7 @@
 #include "esp_system.h"
 #include <driver/rtc_io.h>
 #include <helpers/KeyValueStore.h>
+#include <driver/gpio.h>
 
 class ESP32Board : public mesh::MainBoard {
 protected:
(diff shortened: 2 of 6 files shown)


examples/companion_radio/MyMesh.cpp
```diff
@@ -1886,3 +1886,8 @@ bool MyMesh::advert() {
     return false;
   }
 }
+
+// Check if there is pending work (packets to send)
+bool MyMesh::hasPendingWork() const {
+  return _mgr->getOutboundCount(0xFFFFFFFF) > 0;
+}
```
fschrempf, line 2424: Instead of duplicating the same function from the repeater code, can we please move it to the base class `mesh::Mesh` to share the code for all apps/roles.

src/helpers/ESP32Board.h
```diff
@@ -56,11 +56,17 @@ class ESP32Board : public mesh::MainBoard {
...
 #if defined(CONFIG_IDF_TARGET_ESP32S3) && defined(P_LORA_DIO_1) // Supported ESP32 variants
     if (rtc_gpio_is_valid_gpio((gpio_num_t)P_LORA_DIO_1)) { // Only enter sleep mode if P_LORA_DIO_1 is RTC pin
       esp_sleep_pd_config(ESP_PD_DOMAIN_RTC_PERIPH, ESP_PD_OPTION_ON);
-      esp_sleep_enable_ext1_wakeup((1L << P_LORA_DIO_1), ESP_EXT1_WAKEUP_ANY_HIGH); // To wake up when receiving a LoRa packet
+
+      // Configure wakeup sources: LoRa packet and optionally button
+      if (pin_wake_btn < 0) {
+        esp_sleep_enable_ext1_wakeup((1L << P_LORA_DIO_1), ESP_EXT1_WAKEUP_ANY_HIGH); // Wake on LoRa packet only
+      } else {
+        esp_sleep_enable_ext1_wakeup((1L << P_LORA_DIO_1) | (1L << pin_wake_btn), ESP_EXT1_WAKEUP_ANY_HIGH); // Wake on LoRa packet OR button
```
fschrempf: Do all ESP32 boards have the user button wired as active-high? Probably we should also consider the case where it is active-low.

examples/companion_radio/MyMesh.cpp
weebl2000: Good point, made it DRY.

src/helpers/ESP32Board.h
weebl2000: I've implemented it to default to HIGH for pressed if it's not defined. If we want to force maintainers to think about this we could opt for only waking on button presses if the pressed is defined. Let me know what you think.

weebl2000:
Moved button handling logic to ESP32Board - this has the added benefit that repeaters will also wake up when you press the button (they do not currently).

weebl2000:
@fschrempf Actually basically all devices are low when pressed. Updated the code accordingly.

On my Heltec v4 the screen turns off when I disable bluetooth.

weebl2000:
I have confirmed this saves a bunch of milliamps when bluetooth is disabled, but even more so look at pr #1686. When bluetooth is enabled but phone isn't connected, we sleep for 3 seconds out of every and this saves about 80% power while still having bluetooth enabled. This PR works great but I'd suggest merging #1686 for even more power saving.

IoTThinks:
> **Button wakeup with correct polarity** (`src/helpers/ESP32Board.h`)
> 
> * `enterLightSleep()` accepts optional `pin_wake_btn` parameter (backward-compatible)
> * Uses GPIO wakeup (`GPIO_INTR_LOW_LEVEL`) for button — correct for active-LOW buttons
> * Keeps ext1 (`ANY_HIGH`) for LoRa DIO1 only — these can't share a mask since they need opposite trigger levels

Hi @weebl2000,
I intended to do for companion but using dynamic power management.
Eager to test yours and hope it can be in dev soon.

How much is power consumption during sleep? Still 10mA as same as repeater.
So when it sleeps, it will wakeup on: LoRa packet (ext1), timer (30min), or button press (GPIO wakeup)
=> During the wakeup, will the companion push the latest messages to the phone app?
if yes, then we don't need to press the button on the companion frequently.

In my latest PR, I shifted from ext1 use GPIO wakeup instead of ext1 so the sleep is available to all esp32 boards.
Ext1 is available to RTC GPIOs only and some Esp32 boards only have ext0.

https://github.com/meshcore-dev/MeshCore/pull/1687/changes#diff-bf2a85f9de86622aa1706e48181c52038e533cae13e3f52f5cd2a2a063b61dc8R85

IoTThinks:
I'm testing PowerSaving for Heltec v3 companion with BLE **on** .
Able to reduced from 120mA down to 40mA with BLE on. So 3000mAh battery can last Heltec v3 companion 3+ days.
Trying to reduce it to 20mA.

Yes, BLE connection to MeshCore app still works as normal.
I would say it much harder but more fun and helpful to keep BLE on.

weebl2000:
> I'm testing PowerSaving for Heltec v3 companion with BLE **on** . Able to reduced from 120mA down to 40mA with BLE on. So 3000mAh battery can last Heltec v3 companion 3+ days. Trying to reduce it to 20mA.
> 
> Yes, BLE connection to MeshCore app still works as normal. I would say it much harder but more fun and helpful to keep BLE on.

See #1686

IoTThinks:
@weebl2000 My PowerSaving reduces power consumption to connected and fully functional BLE companions.
Your PRs only save battery when no BLE or BLE is disconnected.

A normal sleep as in repeaters can not save power without breaking BLE connection.
Nevermind, we can go with different approaches.

IoTThinks:
Ah, your PowerSaving still can provide extra power saving when BLE is not connected.
My PowerSaving provides power saving when BLE is connected.

The code is not overlapped / conflicted anyway. They complement each other. 👯‍♂️

PR #1350, opened 2026-01-09
Sync gps time every 30min

weebl2000:
## Summary

Re-syncs the RTC clock from GPS every 30 minutes to combat clock drift, which is especially significant when powersaving is enabled.

Currently, GPS time sync only happens once on the initial fix. After that, the ESP32's internal oscillator drifts and the RTC gradually becomes inaccurate. This adds periodic re-synchronization while the GPS has a valid fix.

## Changes

- Added `_last_time_sync` to track when the last GPS→RTC sync occurred
- Added `TIME_SYNC_INTERVAL` constant (30 minutes)
- Re-enables `_time_sync_needed` flag periodically so the existing sync logic fires again
- First sync still happens immediately on boot (unchanged behaviour)
- Requires 3 consecutive seconds of valid GPS fix before syncing (existing `time_valid > 2` guard)

## Notes

- `millis()` overflow after 49 days is handled correctly by unsigned subtraction
- GPS spoofing is not a new concern — if GPS is spoofed from the start, the initial sync would already be wrong
- Only active when `_clock` is non-NULL (i.e. when an RTC is configured)

---
**Build firmware:** [Build from this branch](https://mcimages.weebl.me/?commitId=sync-gps-time-30min)

### DIFF
--- a/src/helpers/sensors/MicroNMEALocationProvider.h
+++ b/src/helpers/sensors/MicroNMEALocationProvider.h
@@ -43,6 +43,8 @@ class MicroNMEALocationProvider : public LocationProvider {
     int _pin_en;
     long next_check = 0;
     long time_valid = 0;
+    unsigned long _last_time_sync = 0;
+    static const unsigned long TIME_SYNC_INTERVAL = 1800000; // Re-sync every 30 minutes
 
 public :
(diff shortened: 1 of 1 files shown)


JugglerLKR (approved):

JugglerLKR (approved):

JugglerLKR (approved):

JugglerLKR (approved):

IoTThinks:
@weebl2000 So at least 1.5 hours is required to have GPS to sync the time to the devices, right?
Thanks a lot.

PR #1351, opened 2026-01-09
Support GPS properly on Heltec Wireless Tracker v1.x

weebl2000:
**Update 27 February 2026**

After running for nearly two months this seems battled tested, also received feedback from multiple users GPS keeps working and GPS is properly disabled and not causing phantom drain.

## Summary

Power usage is high On HTWT even with GPS disabled, it's not being disabled properly.

Adds proper GPS support to the Heltec Wireless Tracker **v1.x** (HTIT-Tracker V0.5 schematic).

- Passes `&rtc_clock`, `GPS_RESET`, `GPS_EN`, and `&board.periph_power` to `MicroNMEALocationProvider`
- Delegates power management to `MicroNMEALocationProvider::begin()`/`stop()` which use `RefCountedDigitalPin` to safely share VEXT with the display
- Adds `getLocationProvider()` override so firmware can access the GPS provider
- Enables `ENV_INCLUDE_GPS=1` and `GPS_BAUD_RATE=115200` in the build config

### Pin mapping (verified against V0.5 schematic)

| ESP32 GPIO | Net Name | Function |
|---|---|---|
| GPIO33 | GPS_TX | UART Transmit (ESP32 → GPS) |
| GPIO34 | GPS_RX | UART Receive (GPS → ESP32) |
| GPIO35 | GPS_EN | Power enable — N-ch MOSFET (Q2) drives P-ch high-side switch, active HIGH |
| GPIO36 | GPS_RST | Hardware reset, active LOW |
| GPIO37 | PPS | Pulse Per Second (GPS → ESP32, not used in this PR) |

### GPS power control

Unlike earlier assumptions, the v1 hardware **does** have independent GPS power control via GPIO35. The `stop()` path de-asserts GPS_EN to cut GPS power independently of the display, then releases VEXT. This avoids wasting ~25-30mA on an idle GPS chip while the display is active.

### Note

Passes `&rtc_clock` to the location provider, enabling GPS-to-RTC time sync. Complementary with #1364 and #1350 in the GPS chain.

---
**Build firmware:** [Build from this branch](https://mcimages.weebl.me/?commitId=heltec_wireless_tracker_gps)


weebl2000:
Confirmed this works as expected. With this firmware deployed: with GPS disabled and then GPS enabled. GPS fix works and this is the difference in power consumption.

PXL_20260217_035457850 MP
PXL_20260217_035514461 MP

wikiyu:
Heltec Tracker V1.2 (as on board)
IMG_20260226_120316464_HDR

Works correctly, GPS fix goes onto app on android and then is correctly sent to other devices.

Toggling on/off of GPS works, and power usage changes accordingly

pon-wessel:
> Works correctly, GPS fix goes onto app on android and then is correctly sent to other devices.
> 
> Toggling on/off of GPS works, and power usage changes accordingly

Thanks for testing!

PR #1353, opened 2026-01-09
To support PowerSaving for all ESP32-based repeaters and NRF52-based repeaters.

IoTThinks:
Hi all,
This PR is to add PowerSaving for all ESP32-based repeaters and NRF52-based repeaters including 4 boards with old SX1276.
- To support PowerSaving for all ESP32-based repeaters and NRF52-based repeaters.
- To add flag RADIO_SX1276 to support PowerSaving for SX1276. SX1276 uses DIO0 to wakeup MCU instead of DIO1
- To expose LoRa pins at platformio.ini instead of header files
- When BLE is enabled on NRF52, Powersaving will be skipped as same as when enabling WiFi on ESP32. This is to allow OTA in powersaving mode.

**Credits:**
(12 Jan 2026) In latest approach, I use hardware and software events instead of SuspendLoop and ResumeLoop to avoid deadlock issue.
- I have adopted neat sleep/wakeup using `SuspendLoop `and `ResumeLoop` from Mr. Fschrempf and the check of HIGH-level DIO1 to prevent deadlock from Mr. 4np at [NRF52 Repeater Powersaving](https://github.com/meshcore-dev/MeshCore/pull/1238). Thanks a lot.
- However, I implemented our own style as same as PowerSaving for ESP32 to have a similar architecture for powersaving. The code should be simple, effective and consistent for all ESP32 and NRF52 boards.

**Known issue:**
- ESP32-based repeaters may have the clock drift forward while NRF52-based repeaters may have the clock slowdown. This is due to inaccurate / slow clock during sleep. We will improve on this issue, however it may be a cosmetic bug for now.

**Testing:**
I have tested the change with main branch and achieve 10mA for ESP32-based repeaters and 6mA for NRF52-based repeaters.
Some kind friends have tested boards with SX1276 working.

After the merge to dev, I have compiled 29 boards and confirm no compilation error. I have tested the merged code and it worked for Heltec v3, v4 and RAK4631.
- Heltec_v3_repeater Heltec_WSL3_repeater heltec_v4_repeater LilyGo_T3S3_sx1262_repeater
- T_Beam_S3_Supreme_SX1262_repeater Tbeam_SX1262_repeater Station_G2_repeater Station_G2_logging_repeater
- Xiao_C3_repeater Xiao_S3_WIO_repeater Heltec_Wireless_Tracker_repeater heltec_tracker_v2_repeater
- RAK_4631_repeater SenseCap_Solar_repeater Heltec_t114_repeater t1000e_repeater
- Xiao_nrf52_repeater LilyGo_T-Echo_repeater LilyGo_T-Echo-Lite_repeater ProMicro_repeater
- WioTrackerL1_repeater Heltec_mesh_solar_repeater RAK_WisMesh_Tag_repeater ThinkNode_M1_repeater
- Mesh_pocket_repeater
- Heltec_v2_repeater LilyGo_T3S3_sx1276_repeater Tbeam_SX1276_repeater LilyGo_TLora_V2_1_1_6_repeater

Heltec v3 and RAK 4631 in PowerSaving mode
image image

This is a RAK4631 repeater with PowerSaving in action: Sleep at 6mA, wakeup when a LoRa packet comes, process it, wake up for 5s and go to sleep again.

https://github.com/user-attachments/assets/fbb698ce-f551-4900-bf2b-1c9945d6f3a4

### DIFF
--- a/src/helpers/ESP32Board.h
+++ b/src/helpers/ESP32Board.h
@@ -56,6 +56,10 @@ public:
     return raw / 4;
   }
 
+  uint32_t getIRQGpio() {
+    return P_LORA_DIO_1; // default for SX1262
+  }
+
   bool safeToSleep() {
     // Check for RX status
     gpio_num_t wakeupPin = (gpio_num_t)getIRQGpio();
--- a/src/helpers/NRF52Board.h
+++ b/src/helpers/NRF52Board.h
@@ -61,6 +61,7 @@ public:
   const char* getShutdownReasonString(uint8_t reason) override;
 #endif
 
+  uint32_t getIRQGpio() { return P_LORA_DIO_1; } // default for SX1262
   virtual bool safeToSleep() override;
   virtual void sleep(uint32_t secs) override;
 };
--- a/src/MeshCore.h
+++ b/src/MeshCore.h
@@ -51,7 +51,7 @@ public:
   virtual void onAfterTransmit() { }
   virtual void reboot() = 0;
   virtual void powerOff() { /* no op */ }
-  virtual uint32_t getIRQGpio() { return P_LORA_DIO_1; } // default for SX1262
+  virtual uint32_t getIRQGpio() { return -1; } // not supported. DIO1 (SX1262) and DIO0 (SX127x)
   virtual bool safeToSleep() { return false; }
   virtual void sleep(uint32_t secs)  { /* no op */ }
   virtual uint32_t getGpio() { return 0; }


src/helpers/radiolib/RadioLibWrappers.cpp
```diff
@@ -22,6 +22,18 @@ static
 void setFlag(void) {
   // we sent a packet, set the flag
   state |= STATE_INT_READY;
+
+  // To avoid ISR flood during wakeup due to HIGH LEVEL interrupt
+#if defined(ESP32) && defined(RADIO_SX1276) && defined(P_LORA_DIO_0) // SX1276
+  gpio_set_intr_type((gpio_num_t)P_LORA_DIO_0, GPIO_INTR_POSEDGE);
+#elif defined(ESP32) && defined(P_LORA_DIO_1) // SX1262
+  gpio_set_intr_type((gpio_num_t)P_LORA_DIO_1, GPIO_INTR_POSEDGE);
+#endif
```
fschrempf: This looks like an awful hack. Anyway can we try to keep board-specific code out of this generic class.

src/helpers/NRF52Board.cpp
```diff
@@ -101,4 +104,37 @@ bool NRF52BoardOTA::startOTAUpdate(const char *id, char reply[]) {
...
+static void wakeUpCallback(TimerHandle_t xTimer) {
+  // To wake up based on timer
+  resumeLoop();
+  wakeupTimer.stop();
+}
+
+void NRF52Board::enterLightSleep(uint32_t secs) {
+#if defined(P_LORA_DIO_1)
+  // To prevent to enter suspendLoop when the loop has not processed the pending RX
+  if(digitalRead(P_LORA_DIO_1) == HIGH) {
```
fschrempf: I don't think we need this. This should also do the trick and is much cleaner: https://github.com/meshcore-dev/MeshCore/pull/1238/commits/5a673866c8b88599f3309d587644414d0df8ae2c.

src/helpers/NRF52Board.cpp
```diff
@@ -101,4 +104,37 @@ bool NRF52BoardOTA::startOTAUpdate(const char *id, char reply[]) {
...
+    return;
+  }
+
+  // To wake up periodically to do scheduled jobs
+  wakeupTimer.stop();
+  wakeupTimer.begin(secs * 1000, wakeUpCallback, nullptr, false);
```
fschrempf: `SleepTimer::begin()` creates a new timer instance on each call, so we mustn't call it here. See here for a better solution: https://github.com/meshcore-dev/MeshCore/pull/1238/commits/a32700726a5d68a0d6e590209469b32e3d64f6d9.

src/helpers/NRF52Board.cpp
```diff
@@ -101,4 +104,37 @@ bool NRF52BoardOTA::startOTAUpdate(const char *id, char reply[]) {
...
+  suspendLoop();
+#endif
+}
+
+void NRF52Board::sleep(uint32_t secs) {
+  // To check if the BLE is powered and looking for/connected to a phone
+  uint8_t sd_enabled;
+  sd_softdevice_is_enabled(&sd_enabled); // To set sd_enabled to 1 if the BLE stack is active.
+
+  if (!sd_enabled) { // BLE is off ~ No active OTA, safe to go to sleep
```
fschrempf: Instead of checking for the softdevice, I would rather implement a generic mechanism to prevent the sleep which can be used in other cases too. See https://github.com/meshcore-dev/MeshCore/pull/1238/commits/fe17894c52dd64e36d5c585e64693e58e29b470a.

fschrempf (commented):
@IoTThinks Thanks for your work! As I have already posted a solution for the sleep implementation for NRF52 in https://github.com/meshcore-dev/MeshCore/pull/1238 before, would you mind rebasing your work onto the latest version of this PR?

src/helpers/NRF52Board.cpp
IoTThinks: wakeupTimer is a static object.
Why do wakeupTimer.begin create a new timer instance on each call?

I don't want to put wakeupTimer at NRF52Board object as not everyone uses powersaving.

src/helpers/radiolib/RadioLibWrappers.cpp
IoTThinks: It is same as putting ResumeLoop in SetFlag.
The code must be in Radiolib's ISR to be effective.

Even if you rename the SetFlag to RadioISR, well, basically, it is still ISR of Radiolib.

src/helpers/NRF52Board.cpp
IoTThinks: Well, this is only 1 line and easy to understand.
Your proposed radio_driver may work too.

Let me explore if this works better and safer.
`return (_mgr->getOutboundCount(0xFFFFFFFF) > 0) || !radio_driver.isInRecvMode();`

src/helpers/NRF52Board.cpp
IoTThinks: I would love to have consistent approach for ESP32 and NRF52.
In ESP32, we don't have "enterSleep" and "preventSleep".

I think we need to keep thing simple and effective.
This code works not only for OTA, but also for repeaters with WiFi or BLE active due to any reasons.

src/helpers/radiolib/RadioLibWrappers.cpp
fschrempf: I know, that's not what I'm talking about. My complaint is about the following two things:

1. reconfiguring the interrupt type from within the ISR seems a bit weird, but if that's the solution I'm ok with it
2. adding board-specific code in a generic file (`RadioLibWrappers.cpp`) is not so nice. Better call into the board driver. That's why I removed `resumeLoop()` from `setFlag` and introduced a new method in the board class in my PR.

src/helpers/NRF52Board.cpp
fschrempf: > Why do wakeupTimer.begin create a new timer instance on each call?

Because that's how it's implemented. See https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/cores/nRF5/utility/SoftwareTimer.cpp#L39. Each time you call `begin()` a new instance is created and the old instances are not removed therefore potentially causing a memory leak. I think this is poorly implemented in the Arduino core, but anyway.

> I don't want to put wakeupTimer at NRF52Board object as not everyone uses powersaving.

I don't think it hurts that much. Stack space will be reserved for this anyway.

src/helpers/NRF52Board.cpp
fschrempf: You seem to miss the point. The mentioned commit from my PR implements an easy and straight-forward way in the BaseBoard base class, that can be used by all board types. Your implementation is specific to ESP32 and can only be used from inside the board class.

src/helpers/NRF52Board.cpp
IoTThinks: Ok, let me try to put wakeupTimer in NRF52Board as you suggested.

IoTThinks:
To fix the memory leaking for SoftwareTimer, to perform internal testing and to let the community to do beta test for NRF52-based repeaters.
I will push the fix to this PR this week.

fschrempf:
> Ok, let me try to put wakeupTimer in NRF52Board as you suggested.

I've already done that work in https://github.com/meshcore-dev/MeshCore/pull/1238. There is no reason to reimplement my suggestions. Please, just use my branch and add your commits on top.

src/helpers/NRF52Board.cpp
IoTThinks: Thanks for your explanation.
I'm going with the check on wakeupTimer.getHandle() instead.

src/helpers/NRF52Board.cpp
IoTThinks: I would have same "sleep" methods for ESP32 and NRF52.

IoTThinks:
Let me and some friends in busy-traffic areas test to use a simpler but more robust implementation without SuspendLoop, ResumeLoop and not touching setFlag for NRF52 repeaters.
I will push another implementation here in a few days.

IoTThinks:
I have pushed the change to implement power saving for NRF52 to use System-Idle On instead.

examples/simple_repeater/MyMesh.cpp
```diff
@@ -1119,5 +1119,5 @@ void MyMesh::loop() {
 
 // To check if there is pending work
 bool MyMesh::hasPendingWork() const {
-  return _mgr->getOutboundCount(0xFFFFFFFF) > 0;
+  return (_mgr->getOutboundCount(0xFFFFFFFF) > 0) || (!radio_driver.isInRecvMode());
```
fschrempf: This is copied from my PR. If you take over commits from other people, then please do `git cherry-pick` so their authorship persists. Thanks!

src/helpers/radiolib/RadioLibWrappers.cpp
fschrempf: Point 2 above is still valid IMHO

src/helpers/NRF52Board.cpp
```diff
@@ -101,4 +101,29 @@ bool NRF52BoardOTA::startOTAUpdate(const char *id, char reply[]) {
...
+  // To mark the start of the sleep
+  uint32_t startTime = millis();
+
+  // To wake up when a LoRa packet comes or sleep timeout. Safe to 49-day overflow
+  while (digitalRead(P_LORA_DIO_1) == LOW && (millis() - startTime < secs*1000)) {
+    // To enter System-on idle
+    __SEV(); // To clear any stale event flag
+    __WFE(); // To consume that event
+    __WFE(); // To actually puts CPU to sleep
+  }
```
fschrempf: This puts the CPU to sleep and wakes up every millisecond through the systick to check if the timeout has elapsed or the radio IRQ is active. The IRQ for the radio listens on the rising edge of DIO1.

1. What if the rising edge happens during the sleep interval? Is the ISR still triggered or is there a chance to miss the IRQ?
2. What is the power consumption compared to the solution that suspends the loop task?

IoTThinks: Same power consumption as SuspendLoop.
And no deadlock.

examples/simple_repeater/MyMesh.cpp
IoTThinks: I have reverted MyMesh.cpp to exclude radio_driver.isInRecvMode().
It still works fine.

You may want to commit this change to my PR to include this line.

src/helpers/NRF52Board.cpp
IoTThinks: The ISR will still fire, even if the rising edge happens while the CPU is in __WFE() sleep.

IoTThinks:
> can you at least mention in the description how your solution works and what benefits it has over the one in my PR? Thanks!

I did mention in the PR. Thanks a lot.
image

However, in my yesterday proposed solution, I don't use these methods any more.
I have changed the solution approach to use safer hardware / software events instead.

IoTThinks:
I have put in changes to implement light sleep for ESP32 without changing setFlag at all.

src/helpers/radiolib/RadioLibWrappers.cpp
IoTThinks: I have put in changes to implement light sleep for ESP32 without changing setFlag at all.

variants/heltec_v2/platformio.ini
```diff
@@ -183,7 +184,6 @@ build_flags =
   -D WIFI_DEBUG_LOGGING=1
   -D WIFI_SSID='"myssid"'
   -D WIFI_PWD='"mypwd"'
-  -D OFFLINE_QUEUE_SIZE=256
```
fschrempf, line 186: This shouldn't be removed, right?

variants/lilygo_tlora_v2_1/platformio.ini
```diff
@@ -136,7 +137,6 @@ build_flags =
   -D WIFI_SSID='"ssid"'
   -D WIFI_PWD='"password"'
   -D WIFI_DEBUG_LOGGING=1
-  -D OFFLINE_QUEUE_SIZE=256
```
fschrempf, line 139: This shouldn't be removed, right?

src/helpers/ESP32Board.h
```diff
@@ -57,18 +57,39 @@ class ESP32Board : public mesh::MainBoard {
...
-    if (rtc_gpio_is_valid_gpio((gpio_num_t)P_LORA_DIO_1)) { // Only enter sleep mode if P_LORA_DIO_1 is RTC pin
-      esp_sleep_pd_config(ESP_PD_DOMAIN_RTC_PERIPH, ESP_PD_OPTION_ON);
-      esp_sleep_enable_ext1_wakeup((1L << P_LORA_DIO_1), ESP_EXT1_WAKEUP_ANY_HIGH); // To wake up when receiving a LoRa packet
+#if defined(RADIO_SX1276) && defined(P_LORA_DIO_0) // SX1276
+    gpio_num_t wakeupPin = (gpio_num_t) P_LORA_DIO_0;
+#elif defined(P_LORA_DIO_1) // SX1262
+    gpio_num_t wakeupPin = (gpio_num_t) P_LORA_DIO_1;
+#else
+    return; // Not supported
+#endif
```
fschrempf: I really don't like the fact that we add another define `RADIO_SX1276` which doesn't add any new information. We already set `RADIO_CLASS` and `WRAPPER_CLASS` to specify the radio module type.

And the information of the correct IRQ pin is already stored within the `Module` RadioLib instance. We could use `mod.getIrq()` to retrieve it but unfortunately it's currently not accessible in a generic way. So this probably needs some refactoring to work and we don't want this to delay the implementation of the powersaving feature.

Can we at least use a `getIrqPin()` method in the board class that defaults to DIO1 and is overridden to DIO0 for the SX1276 boards?

src/helpers/ESP32Board.h
```diff
@@ -57,18 +57,39 @@ class ESP32Board : public mesh::MainBoard {
...
 
-      if (secs > 0) {
-        esp_sleep_enable_timer_wakeup(secs * 1000000); // To wake up every hour to do periodically jobs
-      }
+    // To configure GPIO wakeup
+    esp_sleep_enable_gpio_wakeup();
+    gpio_wakeup_enable((gpio_num_t) wakeupPin, GPIO_INTR_HIGH_LEVEL); // To wake up when receiving a LoRa packet
```
fschrempf: Remove the `gpio_num_t` cast. It's not needed.

src/helpers/ESP32Board.h
```diff
@@ -57,18 +57,39 @@ class ESP32Board : public mesh::MainBoard {
...
+    // MCU to enter light sleep
+    esp_light_sleep_start();
+    Serial.println(esp_sleep_get_wakeup_cause());
+
+    // To avoid ISR flood during wakeup due to HIGH LEVEL interrupt
+#if defined(ESP32) && defined(RADIO_SX1276) && defined(P_LORA_DIO_0) // SX1276
+    gpio_set_intr_type((gpio_num_t)P_LORA_DIO_0, GPIO_INTR_POSEDGE);
+#elif defined(ESP32) && defined(P_LORA_DIO_1) // SX1262
+    gpio_set_intr_type((gpio_num_t)P_LORA_DIO_1, GPIO_INTR_POSEDGE);
 #endif
```
fschrempf: This whole block can be replaced with a single line


```suggestion
    gpio_set_intr_type(wakeupPin, GPIO_INTR_POSEDGE);
```

src/helpers/ESP32Board.h
```diff
@@ -57,18 +57,39 @@ class ESP32Board : public mesh::MainBoard {
...
+
+     // To disable CPU interrupt servicing
```
fschrempf:
```suggestion
     // To enable CPU interrupt servicing
```

src/helpers/NRF52Board.cpp
```diff
@@ -1,6 +1,6 @@
 #if defined(NRF52_PLATFORM)
 #include "NRF52Board.h"
-
+#include "nrf_sdm.h"
```
fschrempf: I don't think we need this, do we?

src/helpers/NRF52Board.cpp
fschrempf: Ok, good.

src/helpers/NRF52Board.cpp
fschrempf: Ok

src/helpers/NRF52Board.cpp
```diff
@@ -101,4 +101,29 @@ bool NRF52BoardOTA::startOTAUpdate(const char *id, char reply[]) {
...
   return true;
 }
+
+void NRF52Board::enterLightSleep(uint32_t secs) {
+#if defined(P_LORA_DIO_1)
+  // To mark the start of the sleep
+  uint32_t startTime = millis();
+
+  // To wake up when a LoRa packet comes or sleep timeout. Safe to 49-day overflow
+  while (digitalRead(P_LORA_DIO_1) == LOW && (millis() - startTime < secs*1000)) {
```
fschrempf: What about the boards with DIO0 interrupt instead of DIO1? The `digitalRead(P_LORA_DIO_1)` would be incorrect in this case, right?

As mentioned above we could use a `getIrqPin()` function to retrieve the correct pin.

examples/simple_repeater/MyMesh.cpp
fschrempf: > It still works fine.

Yeah it works, but I did some debugging and I think the formally proper way is to only sleep in receive mode. Otherwise I see the sleep being entered at the wrong time when starting up, but we can change that later.

IoTThinks:
> I see you did a force-push, but the commits are still not separated properly. If you need help doing it, feel free to ask.

I did force-push to fix some simple missing codes.
Let the code being reviewed for a while, then we can start over clean.
I don't like rebase actually. 
Startover from heavily reviewed and commented code is faster and cleaner.

And clean PRs are usually approved faster. :D

> One more thing: You always begin your comments end commit subjects with "To ...". This doesn't really makes sense. Just leave out the "to" and state your changes in an imperative way.

Well, if you don't like then I will not include the "To..."
It is the style rather than sense.

examples/simple_repeater/MyMesh.cpp
fschrempf: > You may want to commit this change to my PR to include this line.

I can't commit to your PR unless you give me the permissions to your fork, but that's not useful. You can simply add my fork as a git remote (`git remote add fschrempf https://github.com/fschrempf/MeshCore.git`) and then fetch my branch (`git fetch fschrempf`) and then cherry-pick my commit to your branch (`git cherry-pick 5d37a372b7c74661955f9b82f428c78c94083e77`).

You can also leave it out for now and I will add it later in another PR.

fschrempf:
> I did mention in the PR. Thanks a lot.

I know, thanks. If FOSS developers work on the same topic and take over work from each other they usually credit by taking the commits with the original authorship.

> I have put in changes to implement light sleep for ESP32 without changing setFlag at all.

Ok, good idea. I think we can do it like this.

fschrempf:
>  It is the style rather than sense.

Ok, maybe. To me it sounds awkward and I've never seen this style elsewhere. But maybe that's just me...

fschrempf:
>  Let the code being reviewed for a while, then we can start over clean.

Ok, do as you like.

IoTThinks:
@fschrempf Thanks a lot for your comprehensive reviews.
I appreciate your great effort.

Since, the code for lightsleep for esp32 and NRF52 have changed.
I will let my boards to run a while this week to make sure they are stable.
And wait for reviews from other friends here.

Then I will make a final clean push.
Thanks a lot.

4np:
@IoTThinks , thanks for the work here :) I'll test this out on my RAK repeater.

I was thinking if it would be good to show how much the repeater has slept on the repeater stats to get some insight. Could very well be added in some other PR of course.

src/helpers/ESP32Board.h
fschrempf: > Can we at least use a getIrqPin() method in the board class that defaults to DIO1 and is overridden to DIO0 for the SX1276 boards?

One more idea: Use a new define like `LORA_IRQ_PIN` in `platformio.ini` and set it to either `P_LORA_DIO_0` or `P_LORA_DIO_1`. This can be used everywhere where the IRQ pin is needed.

IoTThinks:
> I was thinking if it would be good to show how much the repeater has slept on the repeater stats to get some insight. Could very well be added in some other PR of course.

Hi @4np,
It is a nice suggestion.

No idea if we can change the repeater stats on our own. Or we need help from MeshCore app developer. 
Let me check on that.

CLI is surely possible.
There are many "hidden" CLI commands which are not suggested by Command help of MeshCore app.

IoTThinks:
> I know, thanks. If FOSS developers work on the same topic and take over work from each other they usually credit by taking the commits with the original authorship.

@fschrempf: Oh, I see.
I would love to have your authorship and your valuable improvements in the PR too.
Let's work together in the final push.

src/helpers/NRF52Board.cpp
4np: In [#1238](https://github.com/meshcore-dev/MeshCore/pull/1238#issuecomment-3734469458) I floated the idea to use [safeToSleep](https://github.com/user-attachments/files/24552005/MeshCore-PR1238-Patch.diff.zip). This could easily be implemented in a generic manner by adding it to `MeshCore.h` and overriding it in ESP32 as well as NRF52 (or even specific variants if necessary ). Pseudo patch:

```patch
diff --git a/src/MeshCore.h b/src/MeshCore.h
index b52ac21..a0ea11b 100644
--- a/src/MeshCore.h
+++ b/src/MeshCore.h
@@ -59,6 +59,11 @@ public:
   virtual void enterSleep(uint32_t secs) {
     if (!prevent_sleep) sleep(secs);
   }
+  virtual bool safeToSleep() { return true; }
   virtual void preventSleep() { prevent_sleep = true; }
   virtual uint32_t getGpio() { return 0; }
   virtual void setGpio(uint32_t values) {}
```

With your changes moving the DIO1 pin configuration from the header files into platform.io the `safeToSleep` method should probably only require implementation in the `NRF52Board.h` and `ESP32Board.h`. Pseudo patch:

```patch
diff --git a/src/helpers/NRF52Board.h b/src/helpers/NRF52Board.h
index f187242..fac7fc3 100644
--- a/src/helpers/NRF52Board.h
+++ b/src/helpers/NRF52Board.h
@@ -16,6 +16,11 @@ public:
   virtual void sleep(uint32_t secs) override;
   virtual void onRXInterrupt() override;
   virtual bool safeToSleep() override { return digitalRead(P_LORA_DIO_1) == LOW; }
 };
```

src/helpers/ESP32Board.h
```diff
@@ -57,18 +57,39 @@ class ESP32Board : public mesh::MainBoard {
...
+    if (secs > 0) {
+      esp_sleep_enable_timer_wakeup(secs * 1000000ULL); // To wake up periodically to do scheduled jobs
     }
+
+    // To disable CPU interrupt servicing
+    noInterrupts();
+
+    // MCU to enter light sleep
+    esp_light_sleep_start();
+    Serial.println(esp_sleep_get_wakeup_cause());
```
4np: Probably it's better to use the `MESH_DEBUG_PRINTLN` function:

```suggestion
    MESH_DEBUG_PRINTLN(esp_sleep_get_wakeup_cause());
```

src/helpers/NRF52Board.h
```diff
@@ -14,6 +14,8 @@ class NRF52Board : public mesh::MainBoard {
   virtual uint8_t getStartupReason() const override { return startup_reason; }
   virtual float getMCUTemperature() override;
   virtual void reboot() override { NVIC_SystemReset(); }
+  virtual void enterLightSleep(uint32_t secs);
+  virtual void sleep(uint32_t secs) override;
```
4np, line 66: Why do we need 2 functions, it looks like all can be implemented inside `sleep()`? In case we do want 2 functions, I don't think `enterLightSleep()` should have public visibility.

4np (commented):
I had a look at the changes and provided some -minor- feedback.

src/helpers/ESP32Board.h
IoTThinks: Let me remove the Serial.print.

src/helpers/NRF52Board.h
IoTThinks: Yes, you may be right.

As there is hibernate / poweroff points to deepsleep, so I made sleep() to point to lightsleep.

May be we can have sleep() only and remove enterLightSleep().

IoTThinks:
@4np Thanks a lot for review.

I will fix for some feedbacks and will check how much changes if using safeToSleep.

h0lad:
I just profiled this with my rusty old Nordic Power Profiler II.

Looks good so far! Makes  repeaters finally competitive to our Mesthastic colleagues. 

It reduced the idle consumption from ~26mA to most times sleeping with 7.5-8mA.

Hardware: HT-CT62 (ESP32C3 + SX1262) <- 3.0V LDO <- TI BQ25185 <- Nordic Power Profiler II (3.7V)

Without Patch
image

With Patch
image

h0lad:
When  implementing please keep in mind hat we also have the RAK3172 in the 'variants'. It's a  ST STM32WLE55C based device.  This one is still missing a power saving implementation  - it's on my list of looking deeper into it but my schedule is a bit too tight right now.

IoTThinks:
@h0lad Is RAK3172 repeater compilable now?

Last time, it is not compilable when I try.

Thanks for testing.
So it works for esp32c3.

IoTThinks:
@h0lad Are you able to compile for RAK3172?
Last time, I can use RadioLib and stm32duino for it.

image

4np:
@IoTThinks , `rak3x72_repeater` compiles for me (on macOS), but it fails on `mergebin`:

<details>
 <summary>Do not know how to make File target `mergebin'.  Stop.</summary>

```zsh
$ FIRMWARE_VERSION="adhoc-$(date +"%Y%m%d%H%M")" sh build.sh build-firmware rak3x72_repeater
...
Building .pio/build/rak3x72_repeater/firmware.bin
========================================= [SUCCESS] Took 10.47 seconds =========================================

Environment       Status    Duration
----------------  --------  ------------
rak3x72_repeater  SUCCESS   00:00:10.472
========================================== 1 succeeded in 00:00:10.472 ==========================================
Processing rak3x72_repeater (board: rak3172; platform: platformio/ststm32@19.1.0; framework: arduino)
-----------------------------------------------------------------------------------------------------------------
Verbose mode can be enabled via `-v, --verbose` option
CONFIGURATION: https://docs.platformio.org/page/boards/ststm32/rak3172.html
PLATFORM: ST STM32 (19.1.0) > BB-STM32WL
HARDWARE: STM32WLE5CCU 48MHz, 64KB RAM, 224KB Flash
DEBUG: Current (stlink) External (jlink, stlink)
PACKAGES:
 - framework-arduinoststm32 @ 0.0.0
 - framework-cmsis @ 2.50900.0 (5.9.0)
 - toolchain-gccarmnoneeabi @ 1.120301.0 (12.3.1)
LDF: Library Dependency Finder -> https://bit.ly/configure-pio-ldf
LDF Modes: Finder ~ chain, Compatibility ~ soft
Found 25 compatible libraries
Scanning dependencies...
Dependency Graph
|-- SPI @ 1.1.0
|-- Wire @ 1.0.0
|-- RadioLib @ 7.5.0
|-- Crypto @ 0.4.0
|-- RTClib @ 2.1.4
|-- Melopero RV3028 @ 1.2.0
|-- CayenneLPP @ 1.6.1
|-- Adafruit Little File System Libraries @ 0.11.0
|-- Adafruit BusIO @ 1.17.2
|-- ed25519
|-- nrf52
Building in release mode
*** Do not know how to make File target `mergebin' (~/Developer/MeshCore/mergebin).  Stop.
========================================== [FAILED] Took 1.13 seconds ==========================================

Environment       Status    Duration
----------------  --------  ------------
rak3x72_repeater  FAILED    00:00:01.130
===================================== 1 failed, 0 succeeded in 00:00:01.130 =====================================
```

</details>

Something else that came up in talks with other MeshCore users is that some have a GPS inside their repeaters and use the GPS for location and clock sync. In the PRs we may have overlooked this use case.

- Does GPS support low power sleep?
- If so, how long will it take to regain a fix and / or update the clock?
- Should the wake-up work time interval change from 5s to whatever duration the GPS requires to do its work?
- If GPS isn't put to sleep, does it keep doing its job while ESP32 / NRF52 is sleeping?

Or should power saving be unsupported altogether in the case a GPS is present?

IoTThinks:
@4np Let me check the code to understand the behaviour of a GPS module during the current sleep.

My guess now is MCU will sleep and GPS may not sleep.

For repeaters, GPS may be mainly for time sync as the position is quite stationary.

Btw, I still think we should remove time in repeaters.

IoTThinks:
@4np This is RAK4631 repeater with GPS module from RAK. 
With GPS module => 44mA. Nearly same as Heltec v3 with no GPS module at 57mA. It defeats the purpose of saving 6mA and waste 32 mA for GPS module.
Without GPS module => 12mA by default.

Although we may optimize code for GPS module to ... sleep more, I still think GPS module for repeaters is not neccessary as we actually do not need "time" for repeaters.

However, let me check what is the behavior now.

image

h0lad:
I can compile the RAK3172 here (Opensuse Thumbleweed)  without problems. Even build a energy harvester PCB for it - runs fine!

Only the power consumption is a bit problematic: the STM32WLE55C sucks around 14mA while hanging in Idle without power saving.


```
Building in release mode
Checking size .pio/build/rak3x72_repeater/firmware.elf
Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
RAM:   [==        ]  22.2% (used 14528 bytes from 65536 bytes)
Flash: [========  ]  79.2% (used 181696 bytes from 229376 bytes)
```

IoTThinks:
> I can compile the RAK3172 here (Opensuse Thumbleweed) without problems. Even build a energy harvester PCB for it - runs fine!
> 
> Only the power consumption is a bit problematic: the STM32WLE55C sucks around 14mA while hanging in Idle.
> 
> ```
> Building in release mode
> Checking size .pio/build/rak3x72_repeater/firmware.elf
> Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
> RAM:   [==        ]  22.2% (used 14528 bytes from 65536 bytes)
> Flash: [========  ]  79.2% (used 181696 bytes from 229376 bytes)
> ```

Love to see RAK3172 working.
You use the default platformio.ini from MeshCore's main, right?
I'm on PlatformIO Windows, btw.

By right, RAK3172 can sleep very well as it is STM32.

h0lad:
> By right, RAK3172 can sleep very well as it is STM32.

Yes. The STM32W-  and the newer STM32U-series have the "sleep very deep and wake up extremely fast" in their  silicon blood. 

>  You use the default platformio.ini from MeshCore's main, right?

image

I can compile the main and your branch without problems.

Maybe give it a try with K/X/Ubuntu 24.04.x - it's pretty much the industrial standard for embedded development until the next LTS from Ubuntu is being released.

IoTThinks:
Ok, let me try the RAK3172.

IoTThinks:
I deleted framework-arduinoststm32 folder and try again.
PlatformIO can download the zip and unzip again.
It can find the tools folder and compile OK.
May be last time, the unzip process was interrupted since it took some time to complete.

Love to know STM32 boards are supported. 
Thanks millions. 👯 👯 👯 

Next time, we will include STM32 repeaters in the powersaving.
At least, this PR doesn't break anything for RAK3x72.

Let's focus on ESP32 and NRF52 this round and the GPS behaviors.
image

4np:
My RAK19007 / RAK4631 has been running fine for over 2 days. I don't have any feedback on power consumption as I don't measure that, but objectively it looks like the battery is draining more slowly.

IoTThinks:
@4np: Please help to monitor.
Thanks a lot.

h0lad:
Maybe it's a bit offtopic:

Does Meshcore have some kind of statistics interface that can be read out remotely? Then I could pull the data and put into a TSDB and visualize it using Grafana.

src/helpers/NRF52Board.cpp
IoTThinks: Let me make to support for SX127x to be more sure.
Even though, there is no NRF52 board with SX127x now.

```
void NRF52Board::enterLightSleep(uint32_t secs) {
#if defined(RADIO_SX1276) && defined(P_LORA_DIO_0) // SX1276
  gpio_num_t wakeupPin = (gpio_num_t)P_LORA_DIO_0;
#elif defined(P_LORA_DIO_1) // SX1262
  gpio_num_t wakeupPin = (gpio_num_t)P_LORA_DIO_1;
#else
  return; // Not supported
#endif

  // To mark the start of the sleep
  uint32_t startTime = millis();

  // To wake up when a LoRa packet comes or sleep timeout. Safe to 49-day overflow
  while (digitalRead(wakeupPin) == LOW && (millis() - startTime < secs * 1000)) {
    // To enter System-on idle
    __SEV(); // To clear any stale event flag
    __WFE(); // To consume that event
    __WFE(); // To actually puts CPU to sleep
  }
}
```

4np:
> @4np: Please help to monitor.
> Thanks a lot.

Maybe [Jorijn](https://github.com/jorijn)'s polling based [MeshCore Stats](https://github.com/jorijn/meshcore-stats) project could be useful for monitoring power consumption.

IoTThinks:
@4np
Hi, 
I have added getIRQGpio and safeToSleep. Merged enterLightSleep to sleep.
Tested OK with Heltec v3 and RAK4631.
Able to compile for TBeam SX1262 and SX1276.

Please help to review. :)

fschrempf:
@IoTThinks Thanks for your work. Here is what I would change: https://github.com/fschrempf/MeshCore/commit/2c2e9e5a5670e32d3ebd66bf2eb021f770cf5789.

* Proper use of OOP scheme for `getIRQGpio()` instead of preprocessor logic (`RADIO_SX1276`)
* some clang-format fixups

IoTThinks:
Let me take a look.
Thanks a lot.

4np:
Other than what @fschrempf said I have no other points 👍

variants/promicro/platformio.ini
```diff
@@ -4,6 +4,18 @@ board = promicro_nrf52840
...
   -D PROMICRO
+  -D P_LORA_NSS=13
+  -D P_LORA_DIO_1=11
+  -D P_LORA_RESET=10
+  -D P_LORA_BUSY=16
+  -D P_LORA_MISO=15
+  -D P_LORA_SCLK=12
+  -D P_LORA_MOSI=14
+  -D SX126X_POWER_EN=21
+  -D SX126X_RXEN=2
```
oltaco: Can the comments be added back please? They are helpful for knowing which pin is which.

oltaco:
Thanks for the work on this, what is the current status of this PR? I take it that this supercedes https://github.com/meshcore-dev/MeshCore/pull/1238 is that correct?

I must be blind, but I can't see where sleep actually gets called, maybe this PR is dependant on https://github.com/meshcore-dev/MeshCore/pull/1238? edit: If this is dependant on the https://github.com/meshcore-dev/MeshCore/pull/1238 can I request that the PRs be combined so that the overall change is easier to quantify?

What is the situation with clock drift, is that still an issue?

I noticed that this sets the lora interrupt pin as a wakeup source on ESP32 but seemingly not on NRF52, is there any reason for the difference?

IoTThinks:
@oltaco The parts for esp32 work solid.

For NRF52, still in testing and revising phase.

ESP32 and NRF52 are a bit different on how they sleep.

fschrempf:
@oltaco https://github.com/meshcore-dev/MeshCore/pull/1238 provides an alternative approach and additional fixes. I switched https://github.com/meshcore-dev/MeshCore/pull/1238  to draft now as we decided to first get this one ready. There is no dependency between the PRs.

The sleep approach for NRF52 is to put in in idle and wakeup on every interrupt event. This includes the systick interrupt every millisecond and the GPIO interrupt from the radio module. There is no need to explicitly enable the wakeup sources.

IoTThinks:
For NRF52, I have put in some improvements locally to avoid deadlock or livelock in high LoRa traffic.
Implementing sleep without touching setFlag method is a bit tricky.

Let me see how to simulate to generate around 20K packets / days.
To match the high traffic in Munich.
Thanks a lot.

4np:
I just ran into an unresponsive repeater (RAK19007 / RAK4631 - Jan 14 build). Honestly, I am unsure if it's a genuine issue as it could just be that the battery drained, but after putting it back onto power I was also unable to access it. Only after reset. I'm going to uptake the latest changes and re-flash my spare repeater.

fschrempf:
> For NRF52, I have put in some improvements locally to avoid deadlock or livelock in high LoRa traffic.

@IoTThinks Please share what exactly the problems are and how your changes look like so we can understand what is going on and how to make it work.

> Implementing sleep without touching setFlag method is a bit tricky.

What do you mean with this?

IoTThinks:
For NRF52, I have committed a fix to poll wakeupPin correctly. https://github.com/meshcore-dev/MeshCore/pull/1353/commits/26339a756238f4b42e2ce10724bf8b2e21fff237
This to fix the hang / unresponsiveness issue for repeater in high LoRa traffic location (5K sent packets and 20K received packets/day)

The old code may hang after a few days after logining into Remote Management and click around for a while.

4np:
I guess I ran into that. Running the latest changes now.

oltaco:
> @4np Hi, I have added getIRQGpio and safeToSleep. Merged enterLightSleep to sleep. Tested OK with Heltec v3 and RAK4631. Able to compile for TBeam SX1262 and SX1276.
> 
> Please help to review. :)

Yes, this is what I was getting at when I asked why you weren't setting the pin as a wake-up source like you were for esp32 😉

IoTThinks:
@4np Please help to monitor and update us.
Does your site have high LoRa traffic?
I'm creating a program sending LoRa storm of raw LoRa packets at the same frequency settings to test the stability of NRF52's sleep.

@oltaco They are similar now.
However, ESP32 and NRF52 have different MCU architecture. So sometimes, they dont have equivalent APIs.

IoTThinks:
It seems no visible time drift on NRF52 due to sleep.
Events from NRF52 are quite noisy (from DIO1, RTC, UART...), so __WFE sleeps and wakes up frequently. That causes RTC has some room to advance and keep the time correctness.
On NRF52, the main MCU pauses and wakeup frequently.

On ESP32, the main MCU sleeps all the way and the ULP (Ultra low processor) wakes and keep the time using slow RTC which causes time drift forward. 
We may want to improve the time drift on ESP32 sleep.

IoTThinks:
@4np I have put in the getIRQDio and safeToSleep as you suggested.
I will check the GPS issue during sleep within a few days. I remember you mentioned it.

Thanks a lot.

4np:
@IoTThinks , thank you for your work on this!

> Does your site have high LoRa traffic?

Yeah there's quite [a lot](https://analyzer.letsmesh.net/packets?region=netherlands) of traffic [here](https://analyzer.letsmesh.net/map?lat=52.2026&long=5.74049&zoom=8).

ps. I don't see South East Asia on the [map](https://analyzer.letsmesh.net/map?lat=31.50336&long=80.97388&zoom=2), perhaps it's worth considering adding an [observer](https://analyzer.letsmesh.net/observer/onboard).

IoTThinks:
@4np Is your NRF52 repeater with new powersaving code still alive?

> ps. I don't see South East Asia on the [map](https://analyzer.letsmesh.net/map?lat=31.50336&long=80.97388&zoom=2), perhaps it's worth considering adding an [observer](https://analyzer.letsmesh.net/observer/onboard).

Let me see how to add in.
Not much traffic in South East Asia yet.
The map looks nice.

Well, there are always interesting stuffs in MeshCore everyday to me.

IoTThinks:
GPS vs. Powersaving is interesting.
For the current PowerSaving, there is NO code to off the GPS module.
So even in light-sleep mode for ESP32 and System-on idle for NRF52, GPS module will still awake and work when MCU goes to sleep.
Yes, GPS module will continue to consume around 30mA.

When in sleep, the MCU will ignore the data from Serial1 (for GPS).
The data from Serial1 may be overflown.

So the repeaters will get stale GPS data and timing.
The repeaters will read GPS data and timing when there is a LoRa packet or every 30minutes.
The impact should be minimal as the repeaters never move and we dont need too accurate timing.

At least, Powersaving code will not break GPS sync.
Well, you or I can test the tracing if having time.

--
In next version, we can off GPS module if the GPS module has EN pin.

Heltec T114 has EN pin to on and off GPS module
image

RAK4631 does not have EN pin to on and off GPS module
image

IoTThinks:
To generate load, I use 2 Heltec v3 as LoRa spammers to replay the flood adverts every randomly 50ms to 6s.
Well, of course, repeaters will not forward is due to replay attack prevention.
However, the packets still reach repeaters and repeaters will have to process them. So, the test for high RX load is still valid.
The advert is very huge at 117 bytes. We may need to reduce the payload length to serve more devices.

For 14 hours, 16K packets. Yes, most of them are duplicated packets.
The RAK4631 with Powersaving 11.2.2 is still running fine.
image image

Our MeshCore friends in high traffic locations report "running fine" too.
Let me merge the remaining changes from reviewers about formatting and getIRQGpio.

IoTThinks:
Hi all,
I have rebased the PR based on the latest dev.
I hope I do not miss any code from our contributors. Please help to check.
- I haved added fixes from reviewers @fschrempf , @4np and @oltaco . Thanks a lot for your effort.
- I fixed the incorrect naming of DIO1 pointing to DIO0 for Heltec v2. So I fixed the deepsleep for Heltec v2 to use DIO0 instead of DIO1. SX127x uses DIO0.
- So far, things look good for Heltec v3, v4, RAK4631, T114 and SenseCap Solar Node P1. Some repeaters are in high traffic locations. Thanks all friends during the testing. 

After rebase, the code still can run well on Heltec v3 and RAK4631.

Known issues:
- When a repeater sleeps, UART console will be disconnected. Workaround: To run "powersaving off" or just reset the board.
- For board with CDC (such as Heltec v4), when it sleeps, the UART console will be disconnected and a firmware flash will fail. Workaround: To run "powersaving off" or just reset the board or run "start ota" to flash via OTA.
- Time drift in Power Saving mode: There is time drift forward to ESP32-based repeaters. We will improve it in the next PR. There is no time drift for NRF52-based repeaters.

mikecarper:
@IoTThinks I was wondering about the 5 second stay awake. Could this awake time be reduced to the minimum needed + a small 0.1s buffer?

IoTThinks:
@mikecarper I think it is possible to reduce the wakeup time.

Let me try 1s wakeup period.

mikecarper:
Is there a way to know in advance how much time is needed to stay awake or a way to check the stack every 100ms

IoTThinks:
> Is there a way to know in advance how much time is needed to stay awake or a way to check the stack every 100ms

We can to check time-on-air of a LoRa packet based on SF, BW and payload length. https://www.loratools.nl/#/airtime
Else we need to more strictly check if the LoRa has finished sending.

For now, 5 seconds is a safe number.
Are your repeaters unable to have any sleep for 5-second period?

aqua:
Been soak-testing this PR on half a dozen Xiao nRF repeaters on a ~1-4%util mesh for over a month; I had a single node deadlock once, on a build that preceded the fix in 7082d80.  I've yet to see a deadlock since then, though that's obviously had less soak time.

IoTThinks:
> Been soak-testing this PR on half a dozen Xiao nRF repeaters on a ~1-4%util mesh for over a month; I had a single node deadlock once, on a build that preceded the fix in [7082d80](https://github.com/meshcore-dev/MeshCore/commit/7082d80c2bd2af0000b518c179bdca1c260a460a). I've yet to see a deadlock since then, though that's obviously had less soak time.

Great news. Thanks for your testing.
My Xiao NRF52 is on the way to me. So it is great to have confirmation from your Xiao NRF52.

IoTThinks:
Hmm, the fix to MeshCore.h to use `virtual uint32_t getIRQGpio() { return P_LORA_DIO_1; }` makes D1, D2... for Xiao NRF52, C3... non-compilable.

IoTThinks:
I don't want to include Arduino.h for MeshCore.h to reduce any possible impact.

I have fixed getIRQGpio() returns default DIO1 at ESP32Board and NRF52Board](https://github.com/meshcore-dev/MeshCore/commit/b75d601c0e3a406829b549765192940527f7216d)

STM32Boards do not use getIRQGpio() at the moment. So it is ok.
We may implement getIRQGpio() in later phase if required.

I'm able to compile most if not all repeaters.
```
# 12 boards
sh build.sh build-firmware \
Heltec_v3_repeater Heltec_WSL3_repeater heltec_v4_repeater LilyGo_T3S3_sx1262_repeater \
T_Beam_S3_Supreme_SX1262_repeater Tbeam_SX1262_repeater Station_G2_repeater Station_G2_logging_repeater  \
Xiao_C3_repeater Xiao_S3_WIO_repeater Heltec_Wireless_Tracker_repeater heltec_tracker_v2_repeater

# 13 boards - NRF52
sh build.sh build-firmware \
RAK_4631_repeater SenseCap_Solar_repeater Heltec_t114_repeater t1000e_repeater \
Xiao_nrf52_repeater LilyGo_T-Echo_repeater LilyGo_T-Echo-Lite_repeater ProMicro_repeater \
WioTrackerL1_repeater Heltec_mesh_solar_repeater RAK_WisMesh_Tag_repeater ThinkNode_M1_repeater \
Mesh_pocket_repeater

# 6 boards
sh build.sh build-firmware \
Heltec_v2_repeater LilyGo_T3S3_sx1276_repeater Tbeam_SX1276_repeater LilyGo_TLora_V2_1_1_6_repeater \
Heltec_Wireless_Paper_repeater Heltec_Wireless_Paper_repeater_bridge_rs232

# 6 boards - NRF52
sh build.sh build-firmware \
ikoka_nano_nrf_22dbm_repeater ikoka_nano_nrf_30dbm_repeater ikoka_nano_nrf_33dbm_repeater \
ikoka_stick_nrf_22dbm_repeater ikoka_stick_nrf_30dbm_repeater ikoka_stick_nrf_33dbm_repeater
```

IoTThinks:
@mikecarper To go to an extreme, I set the board to wakeup for ...0 second.
POWERSAVING_WAKEUP_PERIOD_SECS=0

Basically, the repeater will loop to do some works once then back to sleep. 
By right, if there is ongoing RX, it will skip for next loop.
Not sure about TX.

My Heltec v3 and RAK4631 still work. Seems no visible faulty LoRa packets yet.

https://github.com/meshcore-dev/MeshCore/pull/1353/changes#diff-10a0092562d43e5bb73a25560e519d6106a0d09b143204ed36ae90d8a3ad24aaR25

You can compile on your own: POWERSAVING_WAKEUP_PERIOD_SECS=0
Or what is your board model? I can compile for you.

IoTThinks:
Tested for a night with LoRa spammer, still work fine.
hasPendingWork() seems to protect the TX well.

POWERSAVING_WAKEUP_PERIOD_SECS=0 is very effective when there are many duplicated packets:
- Repeaters only wake up instantly to check if they need to repeat (TX), esle they will be back to sleep instantly in around 10 loops.
- Power meter can see 6mA most of the time as the repeaters are back to sleep so fast.

mikecarper:
@IoTThinks If I'm reading this right 2 power saving modes would be possible

1. Normal where we power up long enough for the max packet length to come in and process it ASAP.
2. Even deeper sleep where you quickly wake and set another sleep that will wait till the max time a packet could take and then wake up again and then process it. 

This is using 256 bytes

=POWER(2,{Spreading Factor (5-12)})/{Bandwidth (7.81-500)} * (28.25 + MAX( CEILING( (2092 - 4*{Spreading Factor (5-12)}) / (4*{Spreading Factor (5-12)}), 1) * 8, 0))

So some examples
SF8 BW62.5 CR8: 2,245ms
SF7 BW62.5 CR8: 1,270ms
SF5 BW125 CR8: 220ms
SF7 BW500 CR8: 159ms

fschrempf:
I've been testing the last few versions of this PR here and I have to say that I *did* see lockups after some days, even after the changes introduced in https://github.com/meshcore-dev/MeshCore/commit/7082d80c2bd2af0000b518c179bdca1c260a460a. So I'm currently not really confident that this is all as it should be. Also I don't quite follow how @IoTThinks came up with the changes in https://github.com/meshcore-dev/MeshCore/commit/7082d80c2bd2af0000b518c179bdca1c260a460a. Why do we need to reconfigure the IRQ pin? Why do we need to poll the IRQ pin additionally with interrupts turned off? How does this intend to solve the problem?

I've updated my repeater to the latest force-push (rebased to v1.12) and will continue the testing. If this continues to cause problems, I'm thinking about resurrecting my alternative approach from https://github.com/meshcore-dev/MeshCore/pull/1238 as with all the additional workarounds needed here, this now seems like a much easier (and maybe also more reliable) approach.

oltaco:
@fschrempf Yes I feel that a more simple straightforward approach to these things would be best.

I'm personally also against these features being enabled by default, as sleeping and radio tend to have issues at the best of times from what I gather. Even when it should work fine on paper it often does not due to things outside our control, like cold/warm start of receive paths from sleep for example.

4np:
The potential of deadlocks also makes me a bit hesitant. I'm not so keen of having to unmount my zip-tied solar repeater from the roof when a deadlock happens. Personally I find reliability more important than saving a couple of mA.

IoTThinks:
> @IoTThinks If I'm reading this right 2 power saving modes would be possible
> 
> 1. Normal where we power up long enough for the max packet length to come in and process it ASAP.
> 2. Even deeper sleep where you quickly wake and set another sleep that will wait till the max time a packet could take and then wake up again and then process it.
> 
> This is using 256 bytes
> 
> =POWER(2,{Spreading Factor (5-12)})/{Bandwidth (7.81-500)} * (28.25 + MAX( CEILING( (2092 - 4*{Spreading Factor (5-12)}) / (4*{Spreading Factor (5-12)}), 1) * 8, 0))
> 
> So some examples SF8 BW62.5 CR8: 2,245ms SF7 BW62.5 CR8: 1,270ms SF5 BW125 CR8: 220ms SF7 BW500 CR8: 159ms

@mikecarper Yes, 5s is a safe number for now.
During my testing with wakup period =0s, it seems hasPendingWork() is effective to prevent going to sleep when there is pending TX.

The point is we only need to wakeup for some more time if the repeater wants to send a packet.
We should not wake long if this is a duplicated packet which don't want to repeat.

IoTThinks:
> I'm personally also against these features being enabled by default, as sleeping and radio tend to have issues at the best of times from what I gather. Even when it should work fine on paper it often does not due to things outside our control, like cold/warm start of receive paths from sleep for example.

@oltaco Well, it is not the default now.
And I think we should keep the Power Saving as option.

Btw, in LoRaWAN world, deepsleep is always the default in Class A for all devices.
Therefore, LoRaWAN devices can work and sleep for at least 3 years for 3000mAh non-chargeable battery.

IoTThinks:
@fschrempf 

> I've been testing the last few versions of this PR here and I have to say that I _did_ see lockups after some days, even after the changes introduced in [7082d80](https://github.com/meshcore-dev/MeshCore/commit/7082d80c2bd2af0000b518c179bdca1c260a460a). So I'm currently not really confident that this is all as it should be. Also I don't quite follow how @IoTThinks came up with the changes in [7082d80](https://github.com/meshcore-dev/MeshCore/commit/7082d80c2bd2af0000b518c179bdca1c260a460a). Why do we need to reconfigure the IRQ pin? Why do we need to poll the IRQ pin additionally with interrupts turned off? How does this intend to solve the problem?

This is not reconfiguring the pin the wakeup pin.
This is to setup sense event for the wakeup pin to makesure the HIGH event of wakeup pin will not be missed and wakeup the NRF52.

`nrf_gpio_cfg_sense_input(wakeupPin, NRF_GPIO_PIN_PULLDOWN, NRF_GPIO_PIN_SENSE_HIGH);`

I have fixed it to `nrf_gpio_cfg_sense_input(wakeupPin, NRF_GPIO_PIN_NOPULL, NRF_GPIO_PIN_SENSE_HIGH);` as some boards such as SenseCap Solar Node P1 may have this wakeup pin already pulled down.

noInterrupts() is to disable ISR servicing shortly to let NRF52 wakeup and check wakeup pin reliably.

> I've updated my repeater to the latest force-push (rebased to v1.12) and will continue the testing. If this continues to cause problems, I'm thinking about resurrecting my alternative approach from #1238 as with all the additional workarounds needed here, this now seems like a much easier (and maybe also more reliable) approach.

I would love to have the implementation to stay isolated in a few functions so it won't affect other critical functions as setFlag...

Yes, please go ahead with the testing. So far, no more reports to me yet. However, more testing is always better.
I don't mind which approach will go to main as long as MeshCore eventually has the best / lowest power consumption for repeaters, room servers and companions.

During "events", we may not have the luxury of main power or many sun lights.

oltaco:
> Btw, in LoRaWAN world, deepsleep is always the default in Class A for all devices.
> Therefore, LoRaWAN devices can work and sleep for at least 3 years for 3000mAh non-chargeable battery.

That's not a fair comparison. LoRaWAN Class A devices sleep until they wake up to transmit and then only listen for a response for a brief time after they transmit before going back to deep sleep. The gateway will be always awake listening for their transmissions.

IoTThinks:
> > Btw, in LoRaWAN world, deepsleep is always the default in Class A for all devices.
> > Therefore, LoRaWAN devices can work and sleep for at least 3 years for 3000mAh non-chargeable battery.
> 
> That's not a fair comparison. LoRaWAN Class A devices sleep until they wake up to transmit and then only listen for a response for a brief time after they transmit before going back to deep sleep. The gateway will be always awake listening for their transmissions.

@oltaco Yes, correct.
What I meant is sleep and deepsleep are reliable on its own. Just the integration into the current code requires testing.

h0lad:
> The potential of deadlocks also makes me a bit hesitant. I'm not so keen of having to unmount my zip-tied solar repeater from the roof when a deadlock happens. Personally I find reliability more important than saving a couple of mA.

I see this power saving functionality as very important feature for Meshcore. Because this enables us to build very compact repeaters.

IoTThinks:
6v 2W panel is reliable for Heltec V3 in my tropical country even in cloudy days.
image

Now, I'm trying with 6v 1W panel for Heltec v3 (10mA) and RAK4631 (6mA).
I may try Xiao C3 + WIO too, as they are cheaper.
image

I agree we need to test more to ensure reliability.
However, PowerSaving enables us to use very small repeaters to deploy any where.

So we can have performance mode (powersaving off) and low power (powersaving on).

IoTThinks:
So far so good.
Friends with T114 (NRF52) repeaters report running fine, no hang issue.

I will replace the OTA detection by setting a flag instead of using esp_wifi to save some DRAM for TBeam.
Then I will put fix to avoid time drift for ESP32-based repeaters too.

src/helpers/ESP32Board.h
```diff
@@ -56,29 +56,63 @@ class ESP32Board : public mesh::MainBoard {
...
-      if (secs > 0) {
-        esp_sleep_enable_timer_wakeup(secs * 1000000); // To wake up every hour to do periodically jobs
-      }
+    // Check for WiFi status to see if there is active OTA
+    wifi_mode_t mode;
+    esp_err_t err = esp_wifi_get_mode(&mode);
 
-      esp_light_sleep_start(); // CPU enters light sleep
+    if (err == ESP_OK) { // WiFi is on
+      return false;
```
weebl2000, line 75: If WiFi was initialized but then stopped, `esp_wifi_get_mode()` returns `ESP_OK` with `mode == WIFI_MODE_NULL`, which would incorrectly block sleep.

```suggestion
    if (err == ESP_OK && mode != WIFI_MODE_NULL) { // WiFi is active
      return false;
    }
```

variants/rak4631/platformio.ini
```diff
@@ -7,7 +7,14 @@ build_flags = ${nrf52_base.build_flags}
   -I variants/rak4631
   -D RAK_4631
   -D RAK_BOARD
-  -D NRF52_POWER_MANAGEMENT
+  -D P_LORA_DIO_1=47
```
weebl2000, line 10: `-D NRF52_POWER_MANAGEMENT` was removed here but other NRF52 variants (`heltec_t114`, `xiao_nrf52`) still have it. This flag gates LPCOMP voltage-based shutdown/wake and boot voltage checking in `RAK4631Board.cpp`. Looks like an accidental deletion — should be kept.

```suggestion
  -D RAK_BOARD
  -D NRF52_POWER_MANAGEMENT
  -D P_LORA_DIO_1=47
```

weebl2000:
Might be good to split up things in this PR to different PRs for easier review.

- power-saving logic
- pin migration from headers to platformio.ini
- SX1276 DIO0 support
- Heltec V2 deep sleep bug fix

weebl2000:
Perhaps this fixes the deadlock - I don't have nRF52 device so I cannot test it.

https://github.com/IoTThinks/MeshCore/pull/2

fschrempf:
> Perhaps this fixes the deadlock - I don't have nRF52 device so I cannot test it.

@weebl2000 With https://github.com/meshcore-dev/MeshCore/pull/1562 already being merged in dev we should focus on testing the solution in dev and see if any deadlocks or other problems occur.

If there is anything left that needs to fixed in dev or if any changes in this PR are still relevant, they should probably be split and proposed as new PRs as you suggested.

IoTThinks:
@weebl2000 There is no more deadlock for NRF52 since PowerSaving 11.2.2+.
So far, the community has done more testing for PowerSaving 11 in reality.

However, since another NRF52 powersaving has been merged to the dev, I will drop the code for NRF52 powersaving and keep ESP32 powersaving + some improvement.

I will close this PR and create a new PR for ESP32-based repeater this week.
Thanks a lot.

IoTThinks:
> Might be good to split up things in this PR to different PRs for easier review.
> 
> * power-saving logic
> * pin migration from headers to platformio.ini
> * SX1276 DIO0 support
> * Heltec V2 deep sleep bug fix

@weebl2000 I would drop "pin migration from headers to platformio.ini" as I will drop the code for NRF52.
The rest, I believe should come into 1 PR as the community expects to have powersaving for all ESP32-based boards.

After dropping code for NRF52, the changes wont be so much to review.
Don't worry. 👯‍♂️

IoTThinks:
The cleanup PR is here.
For esp32 only.
Thanks a lot.

https://github.com/meshcore-dev/MeshCore/pull/1687

PR #1362, opened 2026-01-10
Do not save contacts to FS when nothing has changed

weebl2000:
Specifically, when auto-add is off and a new contact is discovered this currently triggers a write. That write has no effect, other than reducing flash life and causing more potential corruption.

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -358,7 +358,11 @@ void MyMesh::onDiscoveredContact(ContactInfo &contact, bool is_new, uint8_t path
     memcpy(p->path, path, p->path_len);
   }
 
-  if (!is_new) dirty_contacts_expiry = futureMillis(LAZY_CONTACTS_WRITE_DELAY); // only schedule lazy write for contacts that are in contacts[]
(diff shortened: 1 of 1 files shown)


oltaco:
Nice work!

ripplebiz:
@oltaco Do you know if this change still makes senses _after_ your PR (now merged) ?

oltaco:
> @oltaco Do you know if this change still makes senses _after_ your PR (now merged) ?

The idea is still sound but the code will need to be tweaked. I have another PR which I will submit in the next hour that makes another small change to the onAdvertRecv() logic and I will add a change to that so that we only schedule the lazy write when the contacts file has actually changed 👌

weebl2000:
Rebased the PR on dev.

oltaco:
This can be closed, it may not look like it but this is already the behaviour now that #1408 is merged.

This PR would infact revert to saving contacts when nothing has changed because is_new now means "was not added to contacts[]" as described in the comments in current dev.

pon-wessel:
> This can be closed, it may not look like it but this is already the behaviour now that #1408 is merged.
> 
> This PR would infact revert to saving contacts when nothing has changed because is_new now means "was not added to contacts[]" as described in the comments in current dev.

Thanks, had a look at dev, looks good.

PR #1363, opened 2026-01-10
Robust contacts - backup first before saving and fallback to backup file

weebl2000:
I've noticed multiple times on Heltec v4 and Heltec Wireless Tracker v1.2 that my contacts sometimes (partially) disappear. I think this is caused by overwriting immediately.

This PR addresses that problem by first copying to backup, then saving and renaming.

### DIFF
--- a/examples/companion_radio/DataStore.cpp
+++ b/examples/companion_radio/DataStore.cpp
@@ -267,42 +267,57 @@ void DataStore::savePrefs(const NodePrefs& _prefs, double node_lat, double node_
 }
 
 void DataStore::loadContacts(DataStoreHost* host) {
-File file = openRead(_getContactsChannelsFS(), "/contacts3");
-    if (file) {
-      bool full = false;
-      while (!full) {
-        ContactInfo c;
-        uint8_t pub_key[32];
-        uint8_t unused;
+  FILESYSTEM* fs = _getContactsChannelsFS();
+  File file = openRead(fs, "/contacts3");
 
-        bool success = (file.read(pub_key, 32) == 32);
-        success = success && (file.read((uint8_t *)&c.name, 32) == 32);
-        success = success && (file.read(&c.type, 1) == 1);
-        success = success && (file.read(&c.flags, 1) == 1);
-        success = success && (file.read(&unused, 1) == 1);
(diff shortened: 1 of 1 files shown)


oltaco:
Thanks for your efforts, atomic writes would be great but unfortunately it can't work for NRF52 devices due to space constraints. Maybe this can be closed and moved to issues as a discussion point?

liamcottle:
I'll close as requested by @oltaco :) We don't have GitHub Discussions enabled at this time, and as far as I know, you can't convert a PR to a discussion. If it was an issue we could, but I think for now discussions will remain disabled to avoid an extra space we have to monitor.

pon-wessel:
@oltaco @liamcottle what about implementing only for ESP32?

oltaco:
> @oltaco @liamcottle what about implementing only for ESP32?

I'm not against using it where it works, it's a pity we don't have space on most nrf52 devices to allow it or it would probably already be done. Even better if we could do chunked writes with the chunks aligned to flash erase size so we don't overwrite the entire contacts file every time. Maybe down the track. 👍

I'm not totally up on esp32 but I thought spiffs didn't have quite such a problem with corruption due to power loss on write?

weebl2000:
> > @oltaco @liamcottle what about implementing only for ESP32?
> 
> I'm not against using it where it works, it's a pity we don't have space on most nrf52 devices to allow it or it would probably already be done. Even better if we could do chunked writes with the chunks aligned to flash erase size so we don't overwrite the entire contacts file every time. Maybe down the track. 👍
> 
> I'm not totally up on esp32 but I thought spiffs didn't have quite such a problem with corruption due to power loss on write?

Let me look into chunking. Not sure about resilience but noticed a few times contacts (partially) disappearing after power cycle.

oltaco:
> Let me look into chunking.

Unfortunately nrf52 is still going to require more space in order to do chunking, because of the metadata space requirements of the extra files involved.

It's not impossible but will require discussions to decide on a way forward.

> Not sure about resilience but noticed a few times contacts (partially) disappearing after power cycle.
Yes I think spiffs is better in this regard. On nrf52 power loss on write will lose the entire file, and if you're unlucky the entire filesystem 🥲

PR #1369, opened 2026-01-11
Fix: Send device discovery notifications when contact list is full

agessaman:
### Problem
When the contact list reached MAX_CONTACTS, new device discoveries were ignored and the app was not notified. This behavior is particularly frustrating for new users who miss out on mesh discovery because of a full contacts list.

### Solution
When the table is full, create a temporary `ContactInfo` and call `onDiscoveredContact()` to notify the app, matching manual-add mode behavior and send `PUSH_CODE_NEW_ADVERT` with the full contact info allowing the Discovery List to continue to populate.

### Changes
- `BaseChatMesh::onAdvertRecv()`: When `num_contacts >= MAX_CONTACTS`, create a temporary `ContactInfo` and call `onDiscoveredContact()` instead of returning early.
- `MyMesh::onDiscoveredContact()`: Remove check for `!isAutoAddEnabled()` and send full contact if `is_new` is true.

### Behavior
- Auto-add enabled + table not full: Add contact and notify
- Auto-add enabled + table full: Notify only with temporary `ContactInfo` using `PUSH_CODE_NEW_ADVERT`
- Auto-add disabled: Always notify with `PUSH_CODE_NEW_ADVERT` (existing behavior, unchanged)

Connection checks remain in `onDiscoveredContact()`.

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -263,7 +263,7 @@ bool MyMesh::isAutoAddEnabled() const {
 
 void MyMesh::onDiscoveredContact(ContactInfo &contact, bool is_new, uint8_t path_len, const uint8_t* path) {
   if (_serial->isConnected()) {
-    if (!isAutoAddEnabled() && is_new) {
+    if (is_new) {
       writeContactRespFrame(PUSH_CODE_NEW_ADVERT, contact);
     } else {
       out_frame[0] = PUSH_CODE_ADVERT;
--- a/src/helpers/BaseChatMesh.cpp
+++ b/src/helpers/BaseChatMesh.cpp
@@ -117,7 +117,6 @@ void BaseChatMesh::onAdvertRecv(mesh::Packet* packet, const mesh::Identity& id, 
     } else {
       // Table is full, but still notify app about discovered contact
       // Use a temporary ContactInfo since we can't add to permanent list
-      // (Connection check happens inside onDiscoveredContact, same as manual-add mode)
       ContactInfo ci;
       memset(&ci, 0, sizeof(ci));
       ci.id = id;


agessaman:
I just took a look at the potential impacts on meshcore_py for @fdlamotte.

1. Packet parsing: In `reader.py` (lines 84-112), both `CONTACT` and `PUSH_CODE_NEW_ADVERT` packets are parsed identically, and `out_path_len = -1` is handled (lines 94-95).
2. Event handling: `PUSH_CODE_NEW_ADVERT` dispatches `EventType.NEW_CONTACT` (line 109), which is stored in `_pending_contacts`. This matches Liam's client pattern.

I don't think there are any code changes necessary to implement.

- No examples directly use `NEW_CONTACT` events, so no example code breaks.
- The library already distinguishes between permanent contacts (`CONTACTS` event) which won't be sent after contacts are full and discovered contacts (`NEW_CONTACT` event) which will continue to fill the discovery list.
- The `out_path_len = -1` case is handled correctly.

People writing client code will need to deduplicate NEW_CONTACT events, but I'm guessing that's already implemented to prevent repeated notifications when auto-add is turned off.

oltaco:
This looks good, I like the refactor. 👌
I just submitted a PR that might supercede it, https://github.com/meshcore-dev/MeshCore/pull/1379. I'll look at pinching the refactor!

oltaco:
I just double checked the logic in my PR and it looks like I already implemented this, you must have telepathic powers or something 😂
https://github.com/meshcore-dev/MeshCore/blob/b23e62266fa15b8e47669080527e53005bb213cb/src/helpers/BaseChatMesh.cpp#L134-L139

agessaman:
Indeed you did. It needed doing, and I'm glad it is getting done. I'll leave this here and close it out when yours gets merged.

welcomereality:
> Indeed you did. It needed doing, and I'm glad it is getting done. I'll leave this here and close it out when yours gets merged.

https://github.com/meshcore-dev/MeshCore/pull/1379 got merged btw

PR #1371, opened 2026-01-11
Refactor LBT to use explicit CAD and RSSI channel checks

ViezeVingertjes:
Refactors LBT into three explicit checks before transmitting:
- Check if radio is mid-receive (IRQ flags)
- Run active CAD scan
- Check RSSI vs noise floor threshold

### DIFF
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -224,25 +224,33 @@ void Dispatcher::processRecvPacket(Packet* pkt) {
 }
 
 void Dispatcher::checkSend() {
-  if (_mgr->getOutboundCount(_ms->getMillis()) == 0) return;  // nothing waiting to send
-  if (!millisHasNowPassed(next_tx_time)) return;   // still in 'radio silence' phase (from airtime budget setting)
-  if (_radio->isReceiving()) {   // LBT - check if radio is currently mid-receive, or if channel activity
+  if (_mgr->getOutboundCount(_ms->getMillis()) == 0) return;
+  if (!millisHasNowPassed(next_tx_time)) return;
+
+  bool channel_busy = _radio->isReceivingPacket();
+  if (!channel_busy) {
+    int cad_result = _radio->scanChannelActivity();
+    if (cad_result > 0) {
+      channel_busy = true;
+    } else if (_radio->isChannelActive()) {
+      channel_busy = true;
+    }
+  }
+
+  if (channel_busy) {
     if (cad_busy_start == 0) {
-      cad_busy_start = _ms->getMillis();   // record when CAD busy state started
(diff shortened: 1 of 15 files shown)


oltaco:
This has been discussed before, and the conclusion was there is no need for active CAD because we are already in listening mode.

The semtech documentation is explicit that active CAD is useful if your device spends most of the time sleeping and wakes up periodically to transmit, but we are not sleeping so there is no benefit.

ViezeVingertjes:
> This has been discussed before, and the conclusion was there is no need for active CAD because we are already in listening mode.
> 
> The semtech documentation is explicit that active CAD is useful if your device spends most of the time sleeping and wakes up periodically to transmit, but we are not sleeping so there is no benefit.

Still, it looks like packets are colliding on the SDR and just talking over each other. Do you have a link to that documentation? As far as I know, we only have `int.thresh`, which tries to reduce collisions but is off by default and rarely used. I’m trying to figure out the right combo of settings so we don’t have to retry sending so often.

oltaco:
https://www.semtech.com/uploads/technology/LoRa/cad-ensuring-lora-packets.pdf

_CAD is an energy-efficient way to detect an incoming frame without resorting to a power-hungry continuous receiving mode._

Since we are in continuous receive mode already, there is no benefit.

Maybe tuning int.thresh could be useful, I don't know. I think it was changed to disabled by default because it was deemed to be causing more issues than it solved shortly after it was introduced. That was a long time ago now 😅

oltaco:
I should add, I agree maybe there is more that can be done but I think CAD is maybe not it.

ViezeVingertjes:
> I should add, I agree maybe there is more that can be done but I think CAD is maybe not it.

Hmm, interesting... ill close it for now, it did seem to help, but that might be a by effect of some kind.

oltaco:
That's interesting, I wonder if there is some unknown way it can help?

It would be great if we had a way to test these things without testing them on the users.

I wonder if a desktop test could be set up with a handful of nodes purposely made to cause collisions (on a separate unused frequency obviously).

ViezeVingertjes:
> That's interesting, I wonder if there is some unknown way it can help?
> 
> It would be great if we had a way to test these things without testing them on the users.
> 
> I wonder if a desktop test could be set up with a handful of nodes purposely made to cause collisions (on a separate unused frequency obviously).

We will be running it for now so perhaps that gives more test data, indeed on the desk its pretty easy to see them overlap with an SDR and they will not make it through (packet lost). the only thing i could think of now is; continuous receive works in terms of, it will see a packet start and capture it, but once the packet is going to be transmitted, its processing before transmitting etc and that could be the timeframe a new packet starts with 60+ neighbors.

PR #1373, opened 2026-01-11
DISABLE_DEBUG=1 env variable to build.sh

liquidraver:
Hopefully I got all of them from the platformio ini-s

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -29,6 +29,20 @@ $ sh build.sh build-repeater-firmwares
 
 Build all chat room server firmwares
 $ sh build.sh build-room-server-firmwares
+
+Environment Variables:
+  DISABLE_DEBUG=1: Disables all debug logging flags (MESH_DEBUG, MESH_PACKET_LOGGING, etc.)
+                   If not set, debug flags from variant platformio.ini files are used.
+
+Examples:
+Build without debug logging:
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -68,6 +82,13 @@ get_pio_envs_ending_with_string() {
   done
 }
 
+# disable all debug logging flags if DISABLE_DEBUG=1 is set
+disable_debug_flags() {
+  if [ "$DISABLE_DEBUG" = "1" ]; then
```
liamcottle: Can we change this to use `==` instead of `=` please. Keeps it consistent with the other string checks used in the build script, and helps to avoid confusion of "is this assigning the value?" even though in bash it actually does check if it's equal...

liamcottle:
Looks good, have left one comment :)

liquidraver:
I took out MESH_PACKET_LOGGING removal because that is needed for logging repeater releases

liamcottle:
Turns out I didn't submit the review... Sorry!

build.sh
liquidraver: done!

liquidraver:
I've added -UCFG_DEBUG -URADIOLIB_DEBUG_BASIC and -URADIOLIB_DEBUG_PROTOCOL too by @recrof 's recommendation

PR #1377, opened 2026-01-12
FIX: remove serial debug logging from t3s3 sx1276 companion usb

recrof:


nextgens:
That's confirmed working here.

PR #1378, opened 2026-01-12
cleanup ikoka variants and add all supported sensors

recrof:
* standardize platformio.ini definitions
* add all supported sensors to all ikoka variants


oltaco:
Nice work, this is much neater 😎👌

oltaco (approved):

PR #1380, opened 2026-01-12
Implement LRU cache for storing hashes to filter out for flood repeating

weebl2000:
Replace the FIFO (cyclic index) hash table with an LRU cache using `uint32_t millis()` timestamps. The oldest entry is evicted instead of the next sequential slot, so frequently-seen hashes stay in the table longer.

**Changes:**
- `_next_idx` replaced by `_last_seen[MAX_PACKET_HASHES]` (`uint32_t` timestamps)
- On duplicate hit: refresh timestamp (LRU touch)
- On insert: evict the entry with the oldest timestamp
- `restoreFrom()`: restored hashes get fresh timestamps; empty slots stay at 0
- `saveTo()`: writes a dummy `_next_idx` for backward-compatible file format
- No timeout-based eviction — purely LRU

**Why this helps:**

Some busy nodes see more than 128 packets before a flooded packet returns. With FIFO, a hash gets evicted after 128 new packets regardless of how recently it was seen. With LRU, actively-seen hashes stay in the table while stale ones get evicted first.

For example, the [Utrecht repeater](https://analyzer.letsmesh.net/node/CCCC2F1EFE5EE0D1A8FAD40DAC12E4E0F25A37D7B58B72D63A7B0A09BF6F6EF5) sees ~658 packets/20 min with 72 neighbors. With FIFO:

- Time to cycle 128 slots: 128 / 0.548 pkt/s ≈ 234 seconds
- Any packet returning after ~4 minutes won't be filtered → loop

With LRU, recently-seen hashes survive even under high packet volume, so only truly stale entries get evicted.

**Memory cost:** +504 bytes (`uint32_t[128]` replaces `int _next_idx`)

---
**Build firmware:** [Build from this branch](https://mcimages.weebl.me/?commitId=LRU_cache_filterflood)

### DIFF
--- a/src/helpers/SimpleMeshTables.h
+++ b/src/helpers/SimpleMeshTables.h
@@ -10,34 +10,50 @@
 
 class SimpleMeshTables : public mesh::MeshTables {
   uint8_t _hashes[MAX_PACKET_HASHES*MAX_HASH_SIZE];
-  int _next_idx;
+  uint32_t _last_seen[MAX_PACKET_HASHES];  // timestamp for LRU eviction
   uint32_t _direct_dups, _flood_dups;
 
 public:
-  SimpleMeshTables() { 
+  SimpleMeshTables() {
     memset(_hashes, 0, sizeof(_hashes));
-    _next_idx = 0;
+    memset(_last_seen, 0, sizeof(_last_seen));
     _direct_dups = _flood_dups = 0;
   }
 
 #ifdef ESP32
   void restoreFrom(File f) {
     f.read(_hashes, sizeof(_hashes));
-    f.read((uint8_t *) &_next_idx, sizeof(_next_idx));
+    int dummy_idx;
+    f.read((uint8_t *) &dummy_idx, sizeof(dummy_idx));  // legacy _next_idx, ignore
+    // Treat restored hashes as just seen - give them fresh timestamps
+    uint32_t now = millis();
+    const uint8_t* sp = _hashes;
+    for (int i = 0; i < MAX_PACKET_HASHES; i++, sp += MAX_HASH_SIZE) {
+      // Check if slot has data (not all zeros)
+      bool empty = true;
+      for (int j = 0; j < MAX_HASH_SIZE && empty; j++) {
+        if (sp[j] != 0) empty = false;
+      }
+      _last_seen[i] = empty ? 0 : now;
+    }
   }
   void saveTo(File f) {
     f.write(_hashes, sizeof(_hashes));
-    f.write((const uint8_t *) &_next_idx, sizeof(_next_idx));
+    int dummy_idx = 0;
+    f.write((const uint8_t *) &dummy_idx, sizeof(dummy_idx));  // legacy format
   }
 #endif
 
   bool wasSeen(const mesh::Packet* packet) override {
+    uint32_t now = millis();
     uint8_t hash[MAX_HASH_SIZE];
     packet->calculatePacketHash(hash);
 
     const uint8_t* sp = _hashes;
     for (int i = 0; i < MAX_PACKET_HASHES; i++, sp += MAX_HASH_SIZE) {
-      if (memcmp(hash, sp, MAX_HASH_SIZE) == 0) {
+      if (memcmp(hash, sp, MAX_HASH_SIZE) == 0 && _last_seen[i] != 0) {
+        // Match found - refresh timestamp (LRU touch) and return true
+        _last_seen[i] = now;
         if (packet->isRouteDirect()) {
           _direct_dups++;
         } else {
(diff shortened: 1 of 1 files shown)


weebl2000:
image

I have no clue how this is happening.

oltaco:
There's no reason they are both the same repeater, they are just two repeaters which share the same first byte.

weebl2000:
> There's no reason they are both the same repeater, they are just two repeaters which share the same first byte.

I would like to think so too, but how likely is the path 2b,d0,cc,2b,d0,cc,cc and then more d0 showing up later?

I've even observed this packet:

```21320a837fe3e9cc2778c1d87706d0cc634dd0cc88a8163ed0cdd0acd079cccdd02b4dd079d0cc1d18d0cc88dd3ed03ad0cc16ee158936f4250eb42568758c1b6d4e9e7a7318ef2db20f30b2fa86b4b6ed1e8dc02e0eb9e47e1b2d573b3bcaab51f45196aae3c4e53b58c5c2d8c9c9c7366e853025be7adc```

d0 occurs a staggering 11 times.

sequence d0cc occurs 5 times!

oltaco:
> I would like to think so too, but how likely is the path 2b,d0,cc,2b,d0,cc,cc and then more d0 showing up later?

Sorry, I missed that part and only saw the highlighted pair of e5. I agree that does look odd, I would even say it looks deliberate. Is it possible that the entire beginning of the path is faked, and the packet was custom built and injected somewhere closer to the receiving node?

weebl2000:
> > I would like to think so too, but how likely is the path 2b,d0,cc,2b,d0,cc,cc and then more d0 showing up later?
> 
> Sorry, I missed that part and only saw the highlighted pair of e5. I agree that does look odd, I would even say it looks deliberate. Is it possible that the entire beginning of the path is faked, and the packet was custom built and injected somewhere closer to the receiving node?

Somehow I doubt that. Hash is still calculated on the non changing parts of a message and that should cause the message to not be repeated more than once.

If this is malicious it's more likely that someone is spamming d0 with dozens of unique packets causing its hasSeen array to be evicting hashes very quickly.

The fact that's it's occurring with cc too makes me wonder if it's just a really busy network. 👀

Ways to trigger this would be to send repeater discovery requests near those nodes since that would yield a lot of unique packets.

If the packet is fresh close to d0 or cc I would expect it to live in their hasSeen array for a while before they evict it.

Something that is strange is that packets are showing up on mctomqtt analyzer with only long paths. I.e below 
Screenshot_20260113-024101

This either means the packets traversed the mesh a bunch of times before an observer picked them up, or it's a packet that has already been evicted from the aggregation on mctomqtt because its been too long since it was last seen. (I think you can only see the last 2000 packets). Would need to do some local packet analyzing to get to the bottom of this

ViezeVingertjes:
Hmm, i thought loops occur in a much shorter timeframe too, like with only 1-2 hops before it happens, my conclusion was before that its not possible to have 128 packets in that time (due to airtime etc).

[Edit]
We also seen it happen with the hashtable not even filled up to 128*

awhite2000:
@weebl2000 see also https://github.com/meshcore-dev/MeshCore/issues/1386 , we're seeing such loops daily. There's definitely something fishy here because I don't think we have that much traffic to fill the hashtable.

weebl2000:
> @weebl2000 see also #1386 , we're seeing such loops daily. There's definitely something fishy here because I don't think we have that much traffic to fill the hashtable.

It would be interesting to see if some of the more central nodes in the network could run with LRU eviction and see if anything changes.

I'm running this PR on two of my repeaters - they are behaving like normal - but they are not really central nodes.

pon-wessel:
@awhite2000 also: question - do you know if and how many nodes have **rxdelay** set? (anything other than 0)

awhite2000:
@pon-wessel good question, I'll ask around.

awhite2000:
@pon-wessel I did a survey among the operators in the WAW area, on a sample of 22 only one rxdelay was longer than default. On a sample of 25 just one txdelay was shorter, 5 was longer and 19 remained at default.

Since I have two repeaters very close to each other, one with directional and the other with omni antenna, one of them uses a longer txdelay, but I didn't observe any loops originating at one of my repeaters.

@rightup I also received one report from an operator who used pyMC_repeater and there may be a bug there related to packet dedupe, where the hash used for deduplication is reportedly calculated over the whole packet so the same packet arriving again with a different path would not be considered a duplicate and retransmitted.

rightup:
Thanks for the heads up, do we know what version this user is running? 

Thanks

itk80:
> Thanks for the heads up, do we know what version this user is running?
> 
> Thanks

@rightup  pyMC_Repeater v1.0.5. 

I added an additional condition to the test in the def flood_forward function that rejects the packet if my node has already appeared in the path:
    if packet.path and self.local_hash in packet.path:
        packet.drop_reason = "Loop detected (already in path)"
        return None

weebl2000:
> > Thanks for the heads up, do we know what version this user is running?
> > Thanks
> 
> @rightup pyMC_Repeater v1.0.5.
> 
> I added an additional condition to the test in the def flood_forward function that rejects the packet if my node has already appeared in the path: if packet.path and self.local_hash in packet.path: packet.drop_reason = "Loop detected (already in path)" return None

This will lead to dropping legit packets, because repeater hash collissions are very common nowadays. Not sure if this is the right approach.

weebl2000:
This feels like a no-brainer. LRU makes way more sense than FIFO for packets seen! I don't think there's a valid situation where packet seen hash before shouldn't be evicted LRU over FIFO @liamcottle please weigh in here.

liamcottle:
> @liamcottle please weigh in here.

My thoughts are, you've opened about 10,000 PRs, so I haven't reviewed many of them 😅

weebl2000:
> > @liamcottle please weigh in here.
> 
> My thoughts are, you've opened about 10,000 PRs, so I haven't reviewed many of them 😅

Heh, no worries. Feel free to DM me on github/discord if any of the PRs I've opened seem to make sense but you still have questions.

JDat:
> > @liamcottle please weigh in here.
> 
> My thoughts are, you've opened about 10,000 PRs, so I haven't reviewed many of them 😅

Agree! Too much PRs from one person...

liamcottle:
Firmware now has a `set loop.detect` command, that could help with dropping packets from the same repeater IDs. You'd have to convince Scott about using LRU cache for the hashes. This isn't something I'll touch right now.

weebl2000:
> Firmware now has a `set loop.detect` command, that could help with dropping packets from the same repeater IDs. You'd have to convince Scott about using LRU cache for the hashes. This isn't something I'll touch right now.

@ripplebiz How can I convince you of using LRU over FIFO-based evicition? I don't see any real benefit from FIFO eviction. LRU eviction is better in every way and the memory cost is only 508 bytes extra for 128 packets.

PR #1388, opened 2026-01-13
Fix incorrect INA260 address in debug message

piotrmalek:
Fixes a typo where the INA260 error log printed INA219 address macro.
This improves debugging when the INA260 device is not detected.


ripplebiz:
PR's should have 'dev' as base branch

PR #1396, opened 2026-01-14
Heltec v4: Add cli command to toggle GPIO

mattzzw:
This PR adds the capability to control a gpio pin defined in the target's `platformio.ini` by `-D PIN_GPIO` via command line.

`gpio 1` will set the pin HIGH
`gpio 0` will set the pin LOW

States are not persisted, rebooting resets the pin to LOW.
This could be useful to control other hardware next to the repeater, like e.g. starting a RPi next to the repeater.
Implementation should be straightforward for other targets.

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -264,6 +264,15 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
     } else if (memcmp(command, "clear stats", 11) == 0) {
       _callbacks->clearStats();
       strcpy(reply, "(OK - stats reset)");
+    #ifdef PIN_GPIO
+    } else if (memcmp(command, "gpio 1", 6) == 0) {
+        digitalWrite(PIN_GPIO, HIGH);
+        strcpy(reply, "(OK - gpio HIGH)");
+    } else if (memcmp(command, "gpio 0", 6) == 0) {
+        digitalWrite(PIN_GPIO, LOW);   
+        strcpy(reply, "(OK - gpio LOW)");
+
+    #endif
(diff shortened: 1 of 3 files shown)


IoTThinks:
To turn on/off a fan in repeater?

Then why not "set gpio 1 on". 
We can use for all available GPIOs.

So we do not need to set -D GPIO=1

mattzzw:
> To turn on/off a fan in repeater?

As I wrote above:
This could be useful to control other hardware next to the repeater, like e.g. starting a RPi next to the repeater.

> Then why not "set gpio 1 on".

This PR expects one GPIO to be defined. There's probably a thousand ways to implement this. 
I needed to be able to control one GPIO. Expanding this to multiple GPIOs for multiple targets was a bit beyond my scope and capabilities ;-)

JDat:
What if I want to control 5 GPIOs? Code and config must be more generic.

mattzzw:
Feel free to expand on this PR. Should be more or less straight forward, depending on the availability of GPIOs on different targets.

JDat:
Can you make it already with multiple GPIO in one pull request, so no need to expand or submit code in future?

ripplebiz:
Have a look at the sensor code:
https://github.com/meshcore-dev/MeshCore/blob/main/examples/simple_sensor/SensorMesh.cpp#L426

So, we already have a template for get/set/toggle digital GPIOs, but is not currently in CommonCLI. They are 'extra' commands the sensor nodes understand. Try refactoring this PR to move it to CommonCLI?

liamcottle:
Thanks for the PR! Closing out for now due to merge conflicts and no follow up to requested changes.

PR #1400, opened 2026-01-15
Add support for esp32s3-wroom1-n8r2 and  esp32s3-wroom2-n32r16v with sx1276 lora radio

maciek01:
This adds basic versions of the firmware for

esp32s3-wroom1-n8r2 and  esp32s3-wroom2-n32r16v

connected to sx1276 lora radio

THis change was positively verified in the field


oltaco:
Thanks for your efforts. I think this PR is maybe out of scope for the project though because this device appears to be a custom build so it wouldn't be relevant to most users and would be impossible for us to maintain.

PR #1412, opened 2026-01-16
RP2040 internal temp as telemetry, fix TCXO voltage. Create New variant Pico Hat for SX1262

### LINKED ISSUE #1109: Waveshare RP2040 Lora HF
Not working with the newer usb companion i can open it with the old client but not the newer

repeater / room server works fine

thomasabbott:
Two in one:

1a) Adding telemetry for the RP2040 internal temperature sensor on the waveshare RP2040 lora board.
1b) Fixing the TCXO voltage which can't be zero in RadioLib 7.3.0. Left in some MESH_DEBUG prints used to diagnose this. This may fix #1109.
Not tested on hardware.

2) New variant for the [Waveshare SX1262 LoRa Node Module for Raspberry Pi Pico](https://www.waveshare.com/pico-lora-sx1262-868m.htm) cloning the Waveshare RP2040 Lora and changing the GPIO pins.

Tested pico_hat_sx1262_companion_radio_usb and pico_hat_sx1262_repeater on hardware:

- New temperature telemetry is reported correctly (within the limits of the sensor, so +- 5 degrees)
- on-board LED flashes for LORA Tx

### DIFF
--- a/variants/pico_hat_sx1262/WaveshareBoard.cpp
+++ b/variants/pico_hat_sx1262/WaveshareBoard.cpp
@@ -11,8 +11,8 @@ void WaveshareBoard::begin() {
   pinMode(P_LORA_TX_LED, OUTPUT);
 #endif
 
-#ifdef PIN_VBAT_READ
-  pinMode(PIN_VBAT_READ, INPUT);
+#ifdef P_VBAT_READ
+  pinMode(P_VBAT_READ, INPUT);
 #endif
 
 #if defined(PIN_BOARD_SDA) && defined(PIN_BOARD_SCL)
--- a/variants/pico_hat_sx1262/WaveshareBoard.h
+++ b/variants/pico_hat_sx1262/WaveshareBoard.h
@@ -3,26 +3,8 @@
 #include <Arduino.h>
 #include <MeshCore.h>
 
-// LoRa radio module pins for Waveshare RP2040-LoRa-HF/LF
-// https://files.waveshare.com/wiki/RP2040-LoRa/Rp2040-lora-sch.pdf
-
-/*
- * This board has no built-in way to read battery voltage.
- * Nevertheless it's very easy to make it work, you only require two 1% resistors.
- *
- *    BAT+ -----+
- *              |
- *       VSYS --+ -/\/\/\/\- --+
- *                   200k      |
- *                             +-- GPIO28
- *                             |
- *        GND --+ -/\/\/\/\- --+
- *              |    100k
- *    BAT- -----+
- */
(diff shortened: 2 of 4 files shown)


recrof:
we already have rpi pico w and waveshare rp2040 in variants, is this something different?

also - we use internal temperature sensors for esp32 and nrf52, would be good to use same way of doing things across platforms: https://github.com/meshcore-dev/MeshCore/pull/1206

JDat:
Not sure about new variant, but other thing looks OK. I am also working with RP2040. I want to add SPI selection/config for RP2040 and powersaving by downclocking RP2040 to 12 MHz. Next thing is to experiment with RP2040 sleep modes. All these changes/additions can be done on PicoW variant for now.

thomasabbott:
@recrof thanks for reading. In order of priority:

Moving to the new radiolib exposed a mistake in the waveshare RP2040 lora variant - the sx1262 TCXO voltage can't be zero, then the radio fails. This fixes #1109 - someone's tested it. Should I rather submit this as a stand-alone PR?

> we already have rpi pico w and waveshare rp2040 in variants, is this something different?

The new variant is different to the all-in-one waveshare board - it's a hat for the regular Pico, with the SX1262 connected to different pins, and a battery voltage monitor. It's not currently a variant (though there is working community firmware on the web flasher).
image  vs   image
Is there a more elegant way to do this? I think it's a fully qualified variant, but it could be crammed into the same platformio.ini as the original waveshare all-in-one. What's best here?


> also - we use internal temperature sensors for esp32 and nrf52, would be good to use same way of doing things across platforms: #1206

OK I see what was done in #1206, my way is not as general. Shall I rewrite my sensor to fit into this style, so mymesh.cpp can just call   float temperature = board.getMCUTemperature();   for RP2040 boards too?

recrof:
> The new variant is different to the all-in-one waveshare board - it's a hat for the regular Pico, with the SX1262 connected to different pins, and a battery voltage monitor. It's not currently a variant (though there is working community firmware on the web flasher).

we already support waveshare board with plain old picow - https://github.com/meshcore-dev/MeshCore/blob/main/variants/rpi_picow/platformio.ini

> OK I see what was done in https://github.com/meshcore-dev/MeshCore/pull/1206, my way is not as general. Shall I rewrite my sensor to fit into this style, so mymesh.cpp can just call float temperature = board.getMCUTemperature(); for RP2040 boards too?

Yes

thomasabbott:
Thanks @recrof . Closing this one now.

JDat:
@thomasabbott can you make internal temperature as separate thing for PicoW ?

PR #1417, opened 2026-01-18
nRF52840 Watchdog

entr0p1:
Implement the nRF52840's built-in watchdog to reset the board if the firmware locks up for more than 30 seconds.


entr0p1:
Happy to report this has worked perfectly. No erroneous resets overnight, and I was able to force the board to lock up by adding the below into the CLI:
wdt-test

I then called `set wdt.test` from the CLI, which caused the board to lock up completely:
IMG_20260119_092533

I confirmed the board locked up completely, unable to log in over the mesh and the physical user button and UI stopped working:
IMG_20260119_092315

After the watchdog timed out, the board reset as expected and booted up normally. Happy to mark this ready for review once I've added in the supported boards.

4np:
What will happen with the `tick()` calls when the CPU is sleeping after #1353 is merged? Or does the NRF watchdog account for sleeping internally?

fschrempf:
> What will happen with the tick() calls when the CPU is sleeping after https://github.com/meshcore-dev/MeshCore/pull/1353 is merged? Or does the NRF watchdog account for sleeping internally?

I haven't checked, but I guess we will need to feed the watchdog in the wait for event loop.

entr0p1:
> What will happen with the `tick()` calls when the CPU is sleeping after #1353 is merged? Or does the NRF watchdog account for sleeping internally?

It does - the watchdog gets paused during sleep and halt states: https://github.com/meshcore-dev/MeshCore/blob/2fc3e7878744be7a69ddf96ae9d2a851ece1ca94/src/helpers/nrf52/Watchdog.cpp#L22-L24

entr0p1:
I've refactored this code and made it a lot tidier, plus rebased on dev (which I'll need to do again - meshcore moves fast!). 

There's a compile flag to include WDT in the build, and a preference (disabled by default) with CLI command to enable WDT. The runtime code only exists in the repeater firmware but is simple to expand to others.

446564:
if you can resolve the conflicts I'm happy to try this out and review.

entr0p1:
> if you can resolve the conflicts I'm happy to try this out and review.

I've updated it for 1.15.0, I'll push the updated code shortly.

PR #1428, opened 2026-01-19
Update T114 I2C pins for external RTC

etienn01:
While adding an external DS3231 RTC on the Heltec T114, I found that pins P0.26 and P0.27 are not exposed to headers. They are wired to a non-populated footprint for an onboard PCF8563TS RTC (see https://github.com/meshtastic/firmware/issues/4605 for details).

Given this, replacing the default I2C pins would work for external RTCs, but may break backward compatibility if any T114 revision exists (or appears in the future) with the PCF8563TS populated.

An alternative approach would be to keep the existing default pins and add support for a second I2C bus, calling `AutoDiscoverRTCClock.begin()` on both `Wire` and `Wire1`. This would allow the firmware to pick-up the onboard RTC on `Wire` if present, and fall back to discovering an external RTC on `Wire1` otherwise.

This was tested with a DS3231 on a T114 V2 using pins P0.16 and P0.13, and works correctly with this configuration.

image


oltaco:
This seems like a relatively benign change to me, since those pins are indeed not easily accessible and not likely for most people to use. Having said that I do know of some people that have successfully populated the optional flash chip so I wonder if anyone has done the same for the RTC?

> An alternative approach would be to keep the existing default pins and add support for a second I2C bus, calling `AutoDiscoverRTCClock.begin()` on both `Wire` and `Wire1`. This would allow the firmware to pick-up the onboard RTC on `Wire` if present, and fall back to discovering an external RTC on `Wire1` otherwise.

This is worth considering, but I'm not really across the AutoDiscoverRTCClock code or any potential pitfalls.

IoTThinks:
I want this change too as I want to support sensors for T114 repeaters.
I believe T114 never has sensor working yet since P0.26 and P0.27 do not point to any visible pinouts.
T114 also does not have sensor flags in the platformio.ini currently.

image

I feel this P0.26 and P0.27 have been never used and a copy-paste issue when creating this variant.
So quite safe to use 0.16 and 0.13.

IoTThinks:
Yeah, it is merged.
Thanks a lot.

PR #1432, opened 2026-01-20
Add 0-hop advert CLI command

mesher-de:
Add CLI-command to manually send 0-hop adverts
Reduces mesh-load if owners can send 0-hop adverts via cli instead of always flooding


mesher-de:
closed as this depends on #1199 being implemented first

PR #1435, opened 2026-01-20
Use atomic writes to prevent file corruption

ViezeVingertjes:
I have had a few instances within a span of a couple of hours, where my flash got wiped after doing some intense BLE connect/disconnect/request data etc. Only a full erase and reflash seemed to fix it, although i could see it lost its identity etc already after a simple reboot.

Not the prettiest fix, but it seems it fixed it for me, so will leave the PR here in case people notice the same behavior and want to test this fix also.

### DIFF
--- a/examples/companion_radio/DataStore.cpp
+++ b/examples/companion_radio/DataStore.cpp
@@ -31,15 +31,28 @@ DataStore::DataStore(FILESYSTEM& fs, FILESYSTEM& fsExtra, mesh::RTCClock& clock)
 }
 #endif
 
+#define MAX_PATH_LEN 64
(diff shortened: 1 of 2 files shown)


liamcottle:
Please see this previous PR #1363 regarding the use of temp files

ViezeVingertjes:
> Please see this previous PR #1363 regarding the use of temp files

I see, not sure what the right way could be then, i lose my contacts multiple times per day when testing things, meaning connecting/disconnecting/requesting contacts etc a lot and need to do a full erase + reflash after it without this change. I do use it on nRF but left it in draft for now, maybe there is a smarter way to prevent these losses.

Not sure how often this happens with 'normal' usage though.

PR #1437, opened 2026-01-21
Remove _serial->isConnected() logic from buzzer notifications

nakoeppen:
Some may have serial connections active and still want to hear notifications using the built-in buzzer on their companion node. For example, if the node is actively connected to a desktop PC that does not have speakers. This will also prevent confusion about the pre-existing buzzer toggle functionality.

The new logic makes it so that the only conditional which decides whether a buzzer notification is played or not is whether the buzzer_quiet preference in NodePrefs.h is enabled or not. This can still be toggled by clicking the user button three times, which will be retained in the event of a companion node reboot.

This has been tested successfully on both a Seeed T1000-E and a RAK WisMesh Tag.

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -330,11 +330,10 @@ void MyMesh::onDiscoveredContact(ContactInfo &contact, bool is_new, uint8_t path
       memcpy(&out_frame[1], contact.id.pub_key, PUB_KEY_SIZE);
       _serial->writeFrame(out_frame, 1 + PUB_KEY_SIZE);
     }
-  } else {
+  } 
 #ifdef DISPLAY_CLASS
-    if (_ui) _ui->notify(UIEventType::newContactMessage);
+  if (_ui && !_prefs.buzzer_quiet) _ui->notify(UIEventType::newContactMessage); //buzz if enabled
 #endif
-  }
 
   // add inbound-path to mem cache
   if (path && path_len <= sizeof(AdvertPath::path)) {  // check path is valid
@@ -441,9 +440,7 @@ void MyMesh::queueMessage(const ContactInfo &from, uint8_t txt_type, mesh::Packe
   bool should_display = txt_type == TXT_TYPE_PLAIN || txt_type == TXT_TYPE_SIGNED_PLAIN;
   if (should_display && _ui) {
     _ui->newMsg(path_len, from.name, text, offline_queue_len);
-    if (!_serial->isConnected()) {
-      _ui->notify(UIEventType::contactMessage);
-    }
+    if (!_prefs.buzzer_quiet) _ui->notify(UIEventType::contactMessage); //buzz if enabled
   }
 #endif
 }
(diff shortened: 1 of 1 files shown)


oltaco:
I can see the merit to this but it also would be a change that some users will not expect.

Maybe it is better to make the buzzer prefs more granular? Always On/Always Off/On When Connected/Off When Connected ?

nakoeppen:
@oltaco I initially thought of a similar approach (via a buzzer_quietOnSerial variable in NodePrefs.h), and truthfully I really am open to implementing that way. However, there are some considerations I had which ultimately made me choose this path. 

1. There is already a lot of confusion about how/if buzzer functionality is implemented in MeshCore (I only found out the _serial->isConnected() logic and the three button presses to toggle the buzzer by looking at the source code). Parameters which are not strictly necessary could add to that confusion for less programmatically-inclined users.
2. We would then rely on MeshCore client app settings to control such a setting, as we are pretty much maxed out on User Button preference mappings. This is an inconvenience if your companion node is on your person, but you are ultimately away from the device it normally is connected to. On the other hand, the currently proposed solution allows for a simple notification toggle on the same device that the notifications would come from, and work independently of any serial connection.

If other people wish to see another parameter added, I'd be happy to do so. However I do appreciate the granularity and simplicity that this currently-proposed solution provides. If we were to add a secondary parameter to NodePrefs.h, such as buzzer_quietOnSerial, I would strongly urge that we default that to false so as to make it at least a little bit more intuitive for the end user.

oltaco:
@nakoeppen You raise some good points, and I agree that it's probably not good to add another new preference since we currently don't have a good way to set it in the node UI. 👍

I must admit I do like that the buzzer is silent when I'm connected but I can see that it's a subjective thing and I don't have a strong opinion either way, so hopefully other people can give some feedback on this PR.

liamcottle:
Just a heads up, this PR has been reverted in dev branch. We will look to add a new preference, or some other way for the firmware to play a custom buzzer under different circumstances.

Unfortunately with this change it wasn't possible to have the buzzer enabled when disconnected only, which means it was all or nothing.

Most users were expecting to receive a buzzer alert when app is disconnected, which tells them a new message has arrived, and no buzzer alert when app is connected as the app will play the alert instead.

If the user disabled the buzzer (with this PR change) so it wouldn't buzz while app was connected, they also no longer received a buzz when disconnected. Potentially meaning they miss received message alerts.

We will look into some changes for a future update :)

nakoeppen:
@liamcottle Thank you for the heads up. While I can appreciate the convenience of giving the users these options, as I mentioned in a comment above, I had to look through the source code to see why I was not consistently getting buzzer notifications. I would argue with the current lack of documentation on this subject, most users (at least those who are not as technically-inclined as we are) actually do not expect it to function this way.

Nevertheless, if it is alright with you, I'd love to contribute to this project with a revised PR implementing this functionality as an option (at least on the firmware-side, I'd certainly need to rely on you for the MeshCore Companion app). However, I'd like to request that we have such a 'quietOnSerial' setting default to false, as it can be quite confusing toggling the buzzer with three button presses and still not seeing notifications come through on the companion device. 

Edit: Taking a deeper look at the legacy code, as far as I can tell, the undoing of this merge means that you cannot disable the buzzer whatsoever. If serial is connected, it will never play notifications, and if it is disconnected it will always play them even if the user tries to disable the buzzer.

PR #1441, opened 2026-01-21
Document alt INA3221 addr, avoid INA219 conflict

tahnok:
This adds a support for defining alternate INA3221 addresses.

It also adds some special handling in the case you change the address to 0x40 and it conflicts with INA219. In that case, the INA3221 responds to INA219 commands, but with garbage data

I was very confused when the [INA3221](https://done.land/components/power/measuringcurrent/viashunt/ina3221/) generic board I got from Aliexpress wasn't making sense. I got an additional channel of data, but the voltage was halved and the current was related to channel 2 in a weird way. If you can solder, fixing the problem is a bridge away, but sometimes you need a software fix

AI disclosure: I used Claude Code when developing this, but I tested it on my hardware and have fully read through the diff

### DIFF
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -61,7 +61,9 @@ static SensirionI2cSht4x SHT4X;
 #endif
 
 #if ENV_INCLUDE_INA3221
-#define TELEM_INA3221_ADDRESS   0x42      // INA3221 3 channel current sensor I2C address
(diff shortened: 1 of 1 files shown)


andyshinn:
Can you add an override for `TELEM_INA3221_SHUNT_VALUE` as well in this PR? The Adafruit INA3221 board has 0.050 ohm shunt resistors on it so those need to be overridden when using that board.

pcmoore:
Related PR: https://github.com/meshcore-dev/MeshCore/pull/2075

Issue #1329, opened 2026-01-06
Clock-synchronization feature based on the average time from the last N adverts


liktor88:
In case of hidden and inaccessible devices on the roof I always add Arduino for restart device each 24hours to avoid complete system crash. But after reset, clock on RPT have wrong date (~2024) - and this is problem for adverts send from restarted RPT. 

Clock sync function based on average time provided from last N adverts received. Function should be enable/disable in the settings with information how often synchronization should be executed and how many times.

Required:
enable/disable in settings
configurable how often synchronization runs (interval)
configurable how many samples to average (5 or 10)
configurable how many attempts to perform per scheduled syncimmediate re-sync-on-boot behavior if system clock is clearly wrong (after received N required adverts)

IoTThinks:
I think we can reference to how NTP works.

Generally, it is similar to what you are proposing.

However, I still believe we should remove time in repeaters.
So we can avoid this problem totally.

4np:
Average would be unreliable as time from other repeaters cannot be trusted either. Probably it's best to take the last n packets from different -low hopcount- sources that have a time, and determine the min/max variation. If the min / max exceeds a number of seconds, it can be deemed unreliable and ignored / discarded. Ideally you would only consider packets which devices contain an RTC, that's probably not advertised so it could be useful if that's sent along with the time.

There will always be some invalid time as the transmission duration is not taken into account and can be many seconds. But it could help to get rid of those dates that are a year off (or more).

tonurics:
I was coming here to suggest something similar. But rather than the adjustment always happening: make it an optional setting for hardware with known clock issues.

We should also use the median value of the observed packets for the datetime; which would be less affected by any outliers than the mean (average).

pelgraine:
Created an alternative solution for consideration which syncs time from adverts - tested with flood advert from companion BLE, which I know has the correct time because it's synced to my automatic network-time-synced phone: https://github.com/meshcore-dev/MeshCore/pull/2423

mEDI-S:
Also keep in mind that each hop takes at least ~600 ms; with 3 hops, a drift of 12 seconds is quite possible. Therefore, you should only account for ~2 hops and add the 600 ms for each hop.

4np:
[...]

PR #1443, opened 2026-01-21
Add peer-based time synchronization for nodes without hardware RTC

4np:
# Context

## Problem statement

After cold start, repeaters without Real Time Clock (RTC) module will by default be set to **May 15, 2024 10:52:31 UTC** (the initial base date that's used throughout the firmware, e.g. [`1715770351`](https://github.com/search?q=repo%3Ameshcore-dev%2FMeshCore%201715770351&type=code)). Often, the time on these repeaters is then not synchronized properly and the time stays invalid for long periods of time.

> Note: While working on this pull request, I noticed that quite some repeaters were sending adverts that were 1h off. I am in Europe, in the CET TimeZone, which is UTC+1. So it looks like quite some people are setting the clock on their repeater to CET rather than normalizing to UTC. I am unsure how that happens, perhaps through the [Python CLI](https://pypi.org/project/meshcore-cli/)?

## In more detail

> **TLDR**: This PR adds automatic peer-based time synchronization for MeshCore nodes without hardware RTC modules, using airtime-compensated timestamps from nearby peers with statistical outlier filtering and hop-count-weighted consensus via the `SETUP_RTC_WITH_PEER_SYNC` macro.

PeerSyncRTCClock is a time-sync wrapper that lets MeshCore nodes without hardware RTCs maintain accurate time by syncing to nearby peers’ advertisement timestamps. This is especially important for repeaters and low-cost nodes that still need reliable timestamps for messages, telemetry, and coordination. If a hardware RTC is auto-detected over I2C at startup, it takes full priority and peer sync is disabled; otherwise peer-based sync is enabled automatically.

The system gathers timestamps from received advertisements and compensates for multi-hop delay by adding estimated airtime: **(hop_count + 1) × airtime_per_hop_ms**, where airtime is derived from current radio settings (SF, bandwidth, packet length). This avoids relying on the local clock while correcting it. Timestamps are then filtered using MAD ([Median Absolute Deviation](https://en.wikipedia.org/wiki/Median_absolute_deviation))) with a 3×MAD threshold, requiring ≥70% agreement (15 of 21 samples by default) to be accepted.

After filtering, a weighted median is computed with exponentially higher weight for closer peers (e.g. weight 20 at 1 hop, down to 1 at 20 hops), ensuring nearby sources dominate. Validation is adaptive: permissive date bounds (May 2024–May 2034) before the first sync, then strict (±24 h) to prevent drift. Once the offset is under 2 minutes, peer sync pauses for a configurable period (default 24 h) and resumes periodically to verify accuracy. All timing uses RTC timestamps, not millis(), so it works correctly with deep sleep (for example, see #1353).

This is an upgrade over the previous behavior, where nodes without hardware RTCs drifted indefinitely. PeerSyncRTCClock enables these devices to maintain time accuracy within seconds across the mesh, with resilience against bad data and time-spoofing via statistical filtering and proximity-based trust.

## Features

**Priority System:**
  - Hardware RTC always takes priority when present (detected via I2C auto-discovery)
  - Peer sync only activates as fallback when no hardware RTC is found

**Timestamp Collection:**
  - Collects timestamps from received advertisement packets
  - Filters by hop count (default: max 20 hops for reliability)
  - Compensates for multi-hop transmission delay: timestamp + ((hop_count + 1) × airtime_per_hop_ms)
  - Records collection time using RTC timestamps (sleep-safe, not millis())
  - Adjusts timestamps for elapsed time before using them (accounts for clock ticking)
  - Adaptive validation:
    - Before first sync: Lenient validation (May 2024 to ~May 2034 range) to bootstrap from cold start
    - After successful sync: Strict 24-hour validation to prevent drift
  - Maintains rolling buffer of recent timestamps

**Statistical Outlier Filtering + Weighted Median:**
  - Collects minimum sample size (default: 21 samples) before filtering
  - Applies MAD (Median Absolute Deviation) to detect outliers:
    - Calculates initial median of all samples
    - Computes MAD = median(|timestamp - median|)
    - Filters timestamps beyond median ± 3×MAD (minimum 60 seconds threshold)
    - Removes detected outliers from buffer permanently
  - Requires minimum consensus (default: 15 of 21 = ~70% good samples)
  - Calculates weighted median of filtered samples
  - Weights by hop count: closer peers = higher influence
    - Weight formula: (`MAX_HOP_COUNT` + 1 - actual_hop_count)
    - Example: 1-hop gets weight 20, 20-hop gets weight 1

**Clock Update Logic:**
  - Only updates if median is ahead by minimum threshold (default: 2 minutes)
  - Rate-limited to prevent frequent updates (minimum 5 minutes between syncs)
  - Clears sample buffer after successful sync
  - Automatically pauses peer sync when clock is accurate (default: 24 hours) to reduce CPU overhead
  - Resumes peer sync after pause period expires or when clock needs adjustment

**Configuration Options:**
  - `PEER_SYNC_MAX_HOP_COUNT`: Maximum hops for time sync (default: 20)
  - `PEER_SYNC_MIN_OFFSET_SECONDS`: Minimum offset to trigger sync (default: 120s / 2 min)
  - `PEER_SYNC_SAMPLE_SIZE:` Timestamps to collect before outlier filtering and median timestamp determination (default: 21)
  - `PEER_SYNC_MIN_SAMPLES_AFTER_FILTERING`: Minimum good samples after outlier filtering (default: 15)
  - `PEER_SYNC_MIN_SYNCS_BEFORE_STRICT_VALIDATION`: Successful syncs before 24h validation (default: 1)
  - `PEER_SYNC_PAUSE_DURATION_SECONDS`: Pause duration when clock is accurate (default: 86400s / 24h)

## Scope

I have implemented the changes for the RAK4631 variant to make it easier to review. I'll add the changes for [the other variants](https://github.com/4np/MeshCore/pull/1) when this has been reviewed.

## Relevant issues

- #1329

## Disclaimer

I have used Claude Code to assist with this implementation.

### Cold start (the 1st clock sync)

After initial cold start, and an initial advert is heard:

```log
DEBUG: Mesh::collectPeerTimestamp: timestamp=1715878353, hop_count=5, airtime_per_hop=1229ms
DEBUG: PeerSyncRTCClock::addPeerTimestamp called: timestamp=1715878353, hop_count=5, airtime_per_hop=1229ms
DEBUG: PeerSyncRTCClock: Current time: 1715770445, Peer timestamp: 1715878353, Diff: 107908 seconds
DEBUG: PeerSyncRTCClock: Timestamp validation passed (initial mode, syncs=0/1)
DEBUG: PeerSyncRTCClock: Compensating for airtime: 5 hops (+1 transmission) × 1229ms = 7s, adjusted 1715878353 → 1715878360 (2024-05-16 16:52:40 UTC)
DEBUG: PeerSyncRTCClock: Timestamp accepted, sample_count=1
DEBUG: PeerSyncRTCClock: Weighted median from 1 samples (total weight: 16, position: 0, hops: 5)
DEBUG: PeerSyncRTCClock: Collecting timestamps (1 of 21 needed samples)
DEBUG:   Weighted median estimate: 1715878360 (2024-05-16 16:52:40 UTC), drift: +1 days, 5 hours
```

After we have 20 samples since cold start:

```log
DEBUG: Mesh::collectPeerTimestamp: timestamp=1715843095, hop_count=12, airtime_per_hop=1295ms
DEBUG: PeerSyncRTCClock::addPeerTimestamp called: timestamp=1715843095, hop_count=12, airtime_per_hop=1295ms
DEBUG: PeerSyncRTCClock: Current time: 1715772367, Peer timestamp: 1715843095, Diff: 70728 seconds
DEBUG: PeerSyncRTCClock: Timestamp validation passed (initial mode, syncs=0/1)
DEBUG: PeerSyncRTCClock: Compensating for airtime: 12 hops (+1 transmission) × 1295ms = 16s, adjusted 1715843095 → 1715843111 (2024-05-16 07:05:11 UTC)
DEBUG: PeerSyncRTCClock: Timestamp accepted, sample_count=20
DEBUG: PeerSyncRTCClock: Weighted median from 20 samples (total weight: 230, position: 10, hops: 7)
DEBUG: PeerSyncRTCClock: Collecting timestamps (20 of 21 needed samples)
DEBUG:   Weighted median estimate: 1769000785 (2026-01-21 13:06:25 UTC), drift: +616 days, 1 hours
```

> Note: samples with hops exceeding `PEER_SYNC_MAX_HOP_COUNT` will be ignore / discarded.

After the 21st sample (e.g. `PEER_SYNC_SAMPLE_SIZE`), when there is a pool of timestamps that's large enough to reliably determine a median timestamp, outlier detection will discard the samples that deviate too much:

```log
DEBUG: Mesh::collectPeerTimestamp: timestamp=1715770353, hop_count=18, airtime_per_hop=1295ms
DEBUG: PeerSyncRTCClock::addPeerTimestamp called: timestamp=1715770353, hop_count=18, airtime_per_hop=1295ms
DEBUG: PeerSyncRTCClock: Current time: 1715772406, Peer timestamp: 1715770353, Diff: -2053 seconds
DEBUG: PeerSyncRTCClock: Timestamp validation passed (initial mode, syncs=0/1)
DEBUG: PeerSyncRTCClock: Compensating for airtime: 18 hops (+1 transmission) × 1295ms = 24s, adjusted 1715770353 → 1715770377 (2024-05-15 10:52:57 UTC)
DEBUG: PeerSyncRTCClock: Timestamp accepted, sample_count=21
DEBUG: PeerSyncRTCClock: Removed 10 persistent outliers from buffer, 11 samples remain
DEBUG: PeerSyncRTCClock: Only 11 of 21 samples remain after filtering (need 15 minimum)
```

In this example, MAD ([Median Absolute Deviation](https://en.wikipedia.org/wiki/Median_absolute_deviation)) based outlier detection discarded 10 outliers, resulting in a remaining pool of 11 samples. As this is lower than 15 (e.g. `PEER_SYNC_MIN_SAMPLES_AFTER_FILTERING`) collection continues until we again have a pool of 21 samples (e.g. `PEER_SYNC_SAMPLE_SIZE`):

```log
DEBUG: PeerSyncRTCClock::addPeerTimestamp called: timestamp=1769001870, hop_count=7, airtime_per_hop=1229ms
DEBUG: PeerSyncRTCClock: Current time: 1715773229, Peer timestamp: 1769001870, Diff: 53228641 seconds
DEBUG: PeerSyncRTCClock: Timestamp validation passed (initial mode, syncs=0/1)
DEBUG: PeerSyncRTCClock: Compensating for airtime: 7 hops (+1 transmission) × 1229ms = 9s, adjusted 1769001870 → 1769001879 (2026-01-21 13:24:39 UTC)
DEBUG: PeerSyncRTCClock: Timestamp accepted, sample_count=21
DEBUG: PeerSyncRTCClock: Removed 1 persistent outliers from buffer, 20 samples remain
DEBUG: PeerSyncRTCClock::attemptClockSync called, sample_count=20
DEBUG: PeerSyncRTCClock: Collected samples:
DEBUG:   Sample 1: timestamp=1769001667 (2026-01-21 13:21:07 UTC), hops=10, weight=11, age=2689s
DEBUG:   Sample 2: timestamp=1769001647 (2026-01-21 13:20:47 UTC), hops=7, weight=14, age=2564s
DEBUG:   Sample 3: timestamp=1769001626 (2026-01-21 13:20:26 UTC), hops=13, weight=8, age=2513s
DEBUG:   Sample 4: timestamp=1769001628 (2026-01-21 13:20:28 UTC), hops=4, weight=17, age=2383s
DEBUG:   Sample 5: timestamp=1769001684 (2026-01-21 13:21:24 UTC), hops=16, weight=5, age=1791s
DEBUG:   Sample 6: timestamp=1769001684 (2026-01-21 13:21:24 UTC), hops=0, weight=21, age=1600s
DEBUG:   Sample 7: timestamp=1769001686 (2026-01-21 13:21:26 UTC), hops=13, weight=8, age=1429s
DEBUG:   Sample 8: timestamp=1769001648 (2026-01-21 13:20:48 UTC), hops=8, weight=13, age=1318s
DEBUG:   Sample 9: timestamp=1769001682 (2026-01-21 13:21:22 UTC), hops=3, weight=18, age=1228s
DEBUG:   Sample 10: timestamp=1769001653 (2026-01-21 13:20:53 UTC), hops=10, weight=11, age=912s
DEBUG:   Sample 11: timestamp=1769001664 (2026-01-21 13:21:04 UTC), hops=8, weight=13, age=891s
DEBUG:   Sample 12: timestamp=1769001647 (2026-01-21 13:20:47 UTC), hops=10, weight=11, age=810s
DEBUG:   Sample 13: timestamp=1769001677 (2026-01-21 13:21:17 UTC), hops=2, weight=19, age=646s
DEBUG:   Sample 14: timestamp=1769001644 (2026-01-21 13:20:44 UTC), hops=8, weight=13, age=605s
DEBUG:   Sample 15: timestamp=1769001612 (2026-01-21 13:20:12 UTC), hops=10, weight=11, age=557s
DEBUG:   Sample 16: timestamp=1769001683 (2026-01-21 13:21:23 UTC), hops=0, weight=21, age=514s
DEBUG:   Sample 17: timestamp=1769001662 (2026-01-21 13:21:02 UTC), hops=10, weight=11, age=234s
DEBUG:   Sample 18: timestamp=1769001660 (2026-01-21 13:21:00 UTC), hops=8, weight=13, age=205s
DEBUG:   Sample 19: timestamp=1769001658 (2026-01-21 13:20:58 UTC), hops=6, weight=15, age=177s
DEBUG:   Sample 20: timestamp=1769001714 (2026-01-21 13:21:54 UTC), hops=8, weight=13, age=44s
DEBUG:   Raw statistics: min=1769001612 (2026-01-21 13:20:12 UTC), max=1769001714 (2026-01-21 13:21:54 UTC), spread=102 seconds, average=51014742
DEBUG: PeerSyncRTCClock: Weighted median from 20 samples (total weight: 266, position: 10, hops: 10)
DEBUG: PeerSyncRTCClock: Offset check: offset=53228433, min_required=120
DEBUG: PeerSyncRTCClock: Weighted median: 1769001662 (2026-01-21 13:21:02 UTC)
DEBUG: PeerSyncRTCClock: *** SYNCING CLOCK ***
DEBUG:   First clock sync
DEBUG:   Syncing clock from 20 peers
DEBUG:   Old time: 1715773229 (2024-05-15 11:40:29 UTC)
DEBUG:   New time: 1769001662 (2026-01-21 13:21:02 UTC)
DEBUG:   Offset: +53228433 seconds (+887140 minutes)
DEBUG: PeerSyncRTCClock: Clock sync complete, total_syncs=1
DEBUG: PeerSyncRTCClock: Strict 24h validation will be ENABLED on next sync
```

This time, the outlier detection only removed a single outlier, resulting in a remaining pool of 20 timestamps. As this is higher than 15 (e.g. `PEER_SYNC_MIN_SAMPLES_AFTER_FILTERING`), the pool will be used to determine the weighted median timestamp and synchronize the clock.

In this particular case the peer-synced clock was about 40 seconds off compared to the actual UTC time.

### After the first clock sync

The first clock sync after cold start is quite lenient as it will accept a time between **May 15, 2024 10:52:31 UTC** (the initial base date that's used throughout the firmware: [`1715770351`](https://github.com/search?q=repo%3Ameshcore-dev%2FMeshCore%201715770351&type=code)) and **~May 2034** (base + 10 years).

After the first clock sync, the Peer based synchronization logic becomes more strict and will only accept timestamp that lie within a 24h difference.

```log
DEBUG: PeerSyncRTCClock::addPeerTimestamp called: timestamp=1769001735, hop_count=12, airtime_per_hop=1229ms
DEBUG: PeerSyncRTCClock: Current time: 1769001744, Peer timestamp: 1769001735, Diff: -9 seconds
DEBUG: PeerSyncRTCClock: Timestamp validation passed (strict 24h mode)
DEBUG: PeerSyncRTCClock: Compensating for airtime: 12 hops (+1 transmission) × 1229ms = 15s, adjusted 1769001735 → 1769001750 (2026-01-21 13:22:30 UTC)
DEBUG: PeerSyncRTCClock: Timestamp accepted, sample_count=1
DEBUG: PeerSyncRTCClock: Weighted median from 1 samples (total weight: 9, position: 0, hops: 12)
DEBUG: PeerSyncRTCClock: Collecting timestamps (1 of 21 needed samples)
DEBUG:   Weighted median estimate: 1769001750 (2026-01-21 13:22:30 UTC), drift: +6 seconds
```

After again 21 samples are collected (e.g. `PEER_SYNC_SAMPLE_SIZE`), the computed drift is now much smaller:

```log
DEBUG: Mesh::collectPeerTimestamp: timestamp=1769005474, hop_count=13, airtime_per_hop=1164ms
DEBUG: PeerSyncRTCClock::addPeerTimestamp called: timestamp=1769005474, hop_count=13, airtime_per_hop=1164ms
DEBUG: PeerSyncRTCClock: Current time: 1769005495, Peer timestamp: 1769005474, Diff: -21 seconds
DEBUG: PeerSyncRTCClock: Timestamp validation passed (strict 24h mode)
DEBUG: PeerSyncRTCClock: Compensating for airtime: 13 hops (+1 transmission) × 1164ms = 16s, adjusted 1769005474 → 1769005490 (2026-01-21 14:24:50 UTC)
DEBUG: PeerSyncRTCClock: Timestamp accepted, sample_count=21
DEBUG: PeerSyncRTCClock: Removed 2 persistent outliers from buffer, 19 samples remain
DEBUG: PeerSyncRTCClock::attemptClockSync called, sample_count=19
DEBUG: PeerSyncRTCClock: Collected samples:
DEBUG:   Sample 1: timestamp=1769005501 (2026-01-21 14:25:01 UTC), hops=12, weight=9, age=3751s
DEBUG:   Sample 2: timestamp=1769005493 (2026-01-21 14:24:53 UTC), hops=10, weight=11, age=3689s
DEBUG:   Sample 3: timestamp=1769005485 (2026-01-21 14:24:45 UTC), hops=13, weight=8, age=3525s
DEBUG:   Sample 4: timestamp=1769005471 (2026-01-21 14:24:31 UTC), hops=10, weight=11, age=2523s
DEBUG:   Sample 5: timestamp=1769005482 (2026-01-21 14:24:42 UTC), hops=9, weight=12, age=2435s
DEBUG:   Sample 6: timestamp=1769005520 (2026-01-21 14:25:20 UTC), hops=5, weight=16, age=2163s
DEBUG:   Sample 7: timestamp=1769005511 (2026-01-21 14:25:11 UTC), hops=3, weight=18, age=2019s
DEBUG:   Sample 8: timestamp=1769005544 (2026-01-21 14:25:44 UTC), hops=11, weight=10, age=1620s
DEBUG:   Sample 9: timestamp=1769005488 (2026-01-21 14:24:48 UTC), hops=12, weight=9, age=1443s
DEBUG:   Sample 10: timestamp=1769005505 (2026-01-21 14:25:05 UTC), hops=6, weight=15, age=1343s
DEBUG:   Sample 11: timestamp=1769005498 (2026-01-21 14:24:58 UTC), hops=7, weight=14, age=952s
DEBUG:   Sample 12: timestamp=1769005502 (2026-01-21 14:25:02 UTC), hops=7, weight=14, age=750s
DEBUG:   Sample 13: timestamp=1769005472 (2026-01-21 14:24:32 UTC), hops=7, weight=14, age=680s
DEBUG:   Sample 14: timestamp=1769005502 (2026-01-21 14:25:02 UTC), hops=6, weight=15, age=645s
DEBUG:   Sample 15: timestamp=1769005479 (2026-01-21 14:24:39 UTC), hops=9, weight=12, age=610s
DEBUG:   Sample 16: timestamp=1769005469 (2026-01-21 14:24:29 UTC), hops=16, weight=5, age=463s
DEBUG:   Sample 17: timestamp=1769005518 (2026-01-21 14:25:18 UTC), hops=0, weight=21, age=151s
DEBUG:   Sample 18: timestamp=1769005484 (2026-01-21 14:24:44 UTC), hops=8, weight=13, age=60s
DEBUG:   Sample 19: timestamp=1769005490 (2026-01-21 14:24:50 UTC), hops=13, weight=8, age=0s
DEBUG:   Raw statistics: min=1769005469 (2026-01-21 14:24:29 UTC), max=1769005544 (2026-01-21 14:25:44 UTC), spread=75 seconds, average=186649123
DEBUG: PeerSyncRTCClock: Weighted median from 19 samples (total weight: 235, position: 10, hops: 7)
DEBUG: PeerSyncRTCClock: Offset check: offset=3, min_required=120
DEBUG: PeerSyncRTCClock: Weighted median: 1769005498 (2026-01-21 14:24:58 UTC)
DEBUG: PeerSyncRTCClock: Offset 3 is less than minimum 120, not syncing
DEBUG: PeerSyncRTCClock: Clock accurate, pausing peer sync for 24 hours to reduce CPU overhead
```

The weighted median for this run determined that the clock offset is now 3 seconds. As this is lower than the threshold of 120 seconds (e.g. `PEER_SYNC_MIN_OFFSET_SECONDS`), the clock is not updated.

To reduce CPU cycles, peer based clock synchronization is now paused for 24 hours (e.g. `PEER_SYNC_PAUSE_DURATION_SECONDS`).

> Note: as the peer-based clock synchronization will never be as accurate as a true RTC module, it will always deviate to some extent from the actual UTC time. In my testing it deviated about 30 seconds, which IMHO is pretty accurate.

### DIFF
--- a/src/helpers/PeerSyncRTCClock.cpp
+++ b/src/helpers/PeerSyncRTCClock.cpp
@@ -1,0 +1,524 @@
+/*
+ * PEER-BASED TIME SYNCHRONIZATION SYSTEM
+ *
+ * This module implements automatic time synchronization for mesh network nodes without
+ * hardware RTC, using timestamps from nearby nodes' advertisement packets.
+ *
+ * == HOW IT WORKS ==
+ *
+ * 1. Priority System:
+ *    - If hardware RTC is present: Use hardware RTC only, no peer sync
+ *    - If no hardware RTC: Sync using median time from nearby nodes
+ *
+ * 2. Timestamp Collection:
+ *    - Collects timestamps from received advertisement packets
+ *    - Filters by hop count (default: max 20 hops for reliability)
+ *    - Records collection time (RTC time) for each timestamp (sleep-safe)
+ *    - Adjusts timestamps by elapsed time before using them (accounts for clock ticking)
+ *    - Adjusts timestamps by hop count and estimated mesh packet air time (e.g. total estimated air time)
+ *      * Adjustment: timestamp + seconds((hop count + 1) * estimated mesh packet air time)
+ *    - Adaptive validation:
+ *      * BEFORE first successful sync: accepts timestamps from May 2024 to ~May 2034 range
+ *        Trusts that if multiple peers agree, they're correct (even if years ahead)
+ *      * AFTER N successful syncs (default N=1): strict 24-hour validation
+ *        Rejects timestamps more than 24 hours from current time
+ *    - Maintains a rolling buffer of recent timestamps
+ *
+ * 3. Statistical Outlier Filtering + Weighted Median:
+ *    - Collects PEER_SYNC_SAMPLE_SIZE samples (default: 21) without filtering
+ *    - Once buffer is full, applies MAD (Median Absolute Deviation) to detect outliers:
+ *      * Adjusts each timestamp by adding elapsed seconds since collection
+ *      * Calculate initial median of all samples
+ *      * Calculate MAD = median(|timestamp - median|)
+ *      * Filter out timestamps beyond median ± 3×MAD (robust threshold)
+ *      * Minimum threshold of 60 seconds to handle legitimate variation
+ *      * Removes detected outliers from buffer permanently
+ *    - Checks if enough good samples remain (default: at least 15 of 21 = ~70%)
+ *    - If enough good samples, calculates WEIGHTED median of the filtered samples
+ *    - Weights by hop count: closer peers = more influence
+ *    - Weight formula: (MAX_HOP_COUNT + 1 - actual_hop_count)
+ *    - Example with MAX=20: 1-hop gets weight 20, 2-hop gets weight 19, ..., 20-hop gets weight 1
+ *    - This approach is robust to malicious/incorrect timestamps while favoring nearby sources
+ *
+ * 4. Clock Update Logic:
+ *    - Only updates if median is AHEAD by at least threshold (default: 10 minutes)
+ *    - Prevents frequent updates (minimum 5 minutes between syncs)
+ *    - Clears sample buffer after successful sync
+ *
+ * == CONFIGURATION ==
+ *
+ * Constants can be defined before including this header:
+ * - PEER_SYNC_MAX_HOP_COUNT: Maximum hops for time sync (default: 20)
+ *   Lower = more reliable but less coverage, Higher = more coverage but less reliable
+ * - PEER_SYNC_MIN_OFFSET_SECONDS: Minimum offset to trigger sync (default: 300s / 5 min)
+ * - PEER_SYNC_SAMPLE_SIZE: Timestamps to collect before filtering (default: 21, odd for median)
+ * - PEER_SYNC_MIN_SAMPLES_AFTER_FILTERING: Minimum good samples after filtering (default: 15)
+ *   Ensures at least ~70% of samples are good before trusting the result
+ * - PEER_SYNC_MIN_SYNCS_BEFORE_STRICT_VALIDATION: Successful syncs before 24h rule (default: 1)
+ * - PEER_SYNC_PAUSE_DURATION_SECONDS: Pause peer sync when clock accurate (default: 86400s / 24h)
+ *   When clock is already accurate, pauses timestamp collection to reduce CPU overhead
+ *
+ * == USAGE ==
+ *
+ * Example initialization:
+ *   VolatileRTCClock volatile_rtc;
+ *   AutoDiscoverRTCClock auto_rtc(volatile_rtc);
+ *   PeerSyncRTCClock peer_sync_rtc(auto_rtc);
+ *
+ *   // After I2C init (hardware RTC detection happens here)
+ *   auto_rtc.begin(Wire);
+ *
+ *   // Pass to mesh
+ *   mesh::Mesh mesh(..., peer_sync_rtc, ...);
+ *
+ * The mesh network automatically:
+ * - Collects timestamps from advertisements
+ * - Checks if hardware RTC present via hasHardwareRTC()
+ * - Only syncs from peers if no hardware RTC detected
+ */
+
+#include "PeerSyncRTCClock.h"
+#include <time.h>
+
+// Timestamp validation bounds for lenient mode (before first successful sync)
+// Lower bound: Base time when firmware was developed (May 15, 2024)
+// Upper bound: ~10 years from base time (around 2034)
+#define MIN_VALID_TIMESTAMP 1715770351  // May 15, 2024 10:52:31 UTC
+#define MAX_VALID_TIMESTAMP 2031346351  // ~May 2034 (base + 10 years)
+
+// Helper function to format timestamp as UTC string
+static void formatUTC(uint32_t timestamp, char* buffer, size_t size) {
+  time_t t = timestamp;
+  struct tm* tm_info = gmtime(&t);
+  strftime(buffer, size, "%Y-%m-%d %H:%M:%S UTC", tm_info);
+}
+
+bool PeerSyncRTCClock::validateTimestamp(uint32_t timestamp, uint32_t current_time) {
+  int32_t diff = (int32_t)(timestamp - current_time);
+
+  // Apply strict 24-hour validation AFTER clock has been successfully synced at least once
+  // Before that, use lenient validation (trust that if multiple peers agree, they're right)
+  if (_successful_sync_count >= PEER_SYNC_MIN_SYNCS_BEFORE_STRICT_VALIDATION) {
+    // Strict validation: reject timestamps more than 24 hours away
+    if (diff < -86400 || diff > 86400) {
+      MESH_DEBUG_PRINTLN("PeerSyncRTCClock: Timestamp rejected (strict mode, out of 24h range): diff=%d",
+                         diff);
+      return false;
+    }
+    MESH_DEBUG_PRINTLN("PeerSyncRTCClock: Timestamp validation passed (strict 24h mode)");
+    return true;
+  } else {
+    // Lenient validation for initial sync: only reject obviously invalid timestamps
+    if (timestamp < MIN_VALID_TIMESTAMP || timestamp > MAX_VALID_TIMESTAMP) {
+      char utc_buffer[32];
+      formatUTC(timestamp, utc_buffer, sizeof(utc_buffer));
+      MESH_DEBUG_PRINTLN("PeerSyncRTCClock: Timestamp rejected (initial mode, out of valid range): timestamp=%u (%s)",
+                         timestamp, utc_buffer);
+      return false;
+    }
+    MESH_DEBUG_PRINTLN("PeerSyncRTCClock: Timestamp validation passed (initial mode, syncs=%d/%d)",
+                       _successful_sync_count, PEER_SYNC_MIN_SYNCS_BEFORE_STRICT_VALIDATION);
+    return true;
+  }
+}
+
+void PeerSyncRTCClock::addPeerTimestamp(uint32_t timestamp, uint8_t hop_count, uint32_t estimated_airtime_per_hop_ms) {
+  MESH_DEBUG_PRINTLN("PeerSyncRTCClock::addPeerTimestamp called: timestamp=%u, hop_count=%d, airtime_per_hop=%ums",
+                     timestamp, hop_count, estimated_airtime_per_hop_ms);
+
+  // If we have a hardware RTC, don't sync from peers
+  if (hasHardwareRTC()) {
+    MESH_DEBUG_PRINTLN("PeerSyncRTCClock: Hardware RTC detected, ignoring peer timestamp");
+    return;
+  }
+
+  // If peer sync is paused (clock is already accurate), don't collect timestamps
+  uint32_t current_time = getCurrentTime();
+  if (_paused_until > 0 && current_time < _paused_until) {
+    uint32_t remaining_seconds = _paused_until - current_time;
+    MESH_DEBUG_PRINTLN("PeerSyncRTCClock: Peer sync paused for %u more seconds (clock accurate)", remaining_seconds);
+    return;
+  }
+
+  // Only accept timestamps from packets within hop count threshold
+  if (hop_count > PEER_SYNC_MAX_HOP_COUNT) {
+    MESH_DEBUG_PRINTLN("PeerSyncRTCClock: Hop count %d exceeds max %d, rejecting", hop_count, PEER_SYNC_MAX_HOP_COUNT);
+    return;
+  }
+
+  int32_t diff = (int32_t)(timestamp - current_time);
+
+  MESH_DEBUG_PRINTLN("PeerSyncRTCClock: Current time: %u, Peer timestamp: %u, Diff: %d seconds",
+                     current_time, timestamp, diff);
+
+  // Validate timestamp using adaptive validation (strict after first sync, lenient before)
+  if (!validateTimestamp(timestamp, current_time)) {
+    return;
+  }
+
+  // Compensate for transmission airtime: each hop requires time to transmit
+  // The timestamp represents when the packet was SENT by the originator, but it took time to reach us
+  // estimated_airtime_per_hop_ms is the airtime per hop in milliseconds (calculated from radio settings)
+  // Note: hop_count + 1 because even 0 hops means 1 transmission (sender → us directly)
+  uint32_t total_airtime_seconds = ((hop_count + 1) * estimated_airtime_per_hop_ms) / 1000;
+  uint32_t adjusted_timestamp = timestamp + total_airtime_seconds;
+
+  char adjusted_utc[32];
+  formatUTC(adjusted_timestamp, adjusted_utc, sizeof(adjusted_utc));
+  MESH_DEBUG_PRINTLN("PeerSyncRTCClock: Compensating for airtime: %d hops (+1 transmission) × %ums = %us, adjusted %u → %u (%s)",
+                     hop_count, estimated_airtime_per_hop_ms, total_airtime_seconds, timestamp, adjusted_timestamp, adjusted_utc);
+
+  // Add adjusted timestamp to buffer with collection time (RTC time, not millis, for sleep-safe operation)
+  uint32_t collection_time = current_time;  // Use RTC time so age adjustment works even after CPU sleep
+  if (_sample_count < PEER_SYNC_SAMPLE_SIZE) {
+    _timestamps[_sample_count] = adjusted_timestamp;  // Store adjusted, not original
+    _hop_counts[_sample_count] = hop_count;
+    _sample_times[_sample_count] = collection_time;
+    _sample_count++;
+  } else {
+    // Buffer full, shift and add new sample
+    for (int i = 0; i < PEER_SYNC_SAMPLE_SIZE - 1; i++) {
+      _timestamps[i] = _timestamps[i + 1];
+      _hop_counts[i] = _hop_counts[i + 1];
+      _sample_times[i] = _sample_times[i + 1];
+    }
+    _timestamps[PEER_SYNC_SAMPLE_SIZE - 1] = adjusted_timestamp;  // Store adjusted, not original
+    _hop_counts[PEER_SYNC_SAMPLE_SIZE - 1] = hop_count;
+    _sample_times[PEER_SYNC_SAMPLE_SIZE - 1] = collection_time;
+  }
+
+  MESH_DEBUG_PRINTLN("PeerSyncRTCClock: Timestamp accepted, sample_count=%d", _sample_count);
+
+  #if MESH_DEBUG
+  // Log current status and weighted median estimate (simple estimate during collection)
+  // Wrapped in MESH_DEBUG to avoid expensive computation in release builds
+  if (_sample_count < PEER_SYNC_SAMPLE_SIZE) {
+    // During collection: show simple weighted median without MAD filtering
+    uint32_t median_estimate = calculateSimpleWeightedMedian();
+    int32_t drift = (int32_t)(median_estimate - current_time);
+
+    char utc_buffer[32];
+    formatUTC(median_estimate, utc_buffer, sizeof(utc_buffer));
+
+    // Format drift in human-readable format
+    char drift_str[64];
+    uint32_t abs_drift = (drift < 0) ? -drift : drift;
+    if (abs_drift < 60) {
+      snprintf(drift_str, sizeof(drift_str), "%+d seconds", drift);
+    } else if (abs_drift < 3600) {
+      snprintf(drift_str, sizeof(drift_str), "%+d minutes, %d seconds",
+               drift / 60, (int)(abs_drift % 60));
+    } else if (abs_drift < 86400) {
+      snprintf(drift_str, sizeof(drift_str), "%+d hours, %d minutes",
+               drift / 3600, (int)((abs_drift % 3600) / 60));
+    } else {
+      snprintf(drift_str, sizeof(drift_str), "%+d days, %d hours",
+               drift / 86400, (int)((abs_drift % 86400) / 3600));
+    }
+
+    MESH_DEBUG_PRINTLN("PeerSyncRTCClock: Collecting timestamps (%d of %d needed samples)",
+                       _sample_count, PEER_SYNC_SAMPLE_SIZE);
+    MESH_DEBUG_PRINTLN("  Weighted median estimate: %u (%s), drift: %s",
+                       median_estimate, utc_buffer, drift_str);
+  }
+  #endif
+
+  // Try to sync if we have enough samples
+  if (_sample_count >= PEER_SYNC_SAMPLE_SIZE) {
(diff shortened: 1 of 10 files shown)


src/helpers/radiolib/RadioLibWrappers.cpp
```diff
@@ -72,13 +72,18 @@ void RadioLibWrapper::loop() {
...
+    int16_t old_noise_floor = _noise_floor;
     _noise_floor = _floor_sample_sum / NUM_NOISE_FLOOR_SAMPLES;
     if (_noise_floor < -120) {
       _noise_floor = -120;    // clamp to lower bound of -120dBi
     }
     _floor_sample_sum = 0;
 
-    MESH_DEBUG_PRINTLN("RadioLibWrapper: noise_floor = %d", (int)_noise_floor);
+    // Only log if noise floor changed
+    if (_noise_floor != old_noise_floor) {
```
4np, line 83: This log entry is rather chatty, this will now only log whenever the noise floor changes.

src/helpers/PeerSyncRTCClock.cpp
```diff
@@ -0,0 +1,524 @@
+/*
+ * PEER-BASED TIME SYNCHRONIZATION SYSTEM
```
4np, line 2: The peer based time synchronization logic is also explained here in detail in in-line documentation.

src/helpers/RTCClockHelper.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <helpers/PeerSyncRTCClock.h>
+#include <helpers/AutoDiscoverRTCClock.h>
+
+/**
+ * \brief Helper macro to setup RTC clock with automatic peer synchronization
```
4np, line 7: This is the new helper macro to simplify RTC use in the various hardware variants.

oltaco:
Thanks for trying to solve clock sync. I’m not convinced that this relatively minor issue warrants the additional complexity introduced by this PR. A simpler and safer approach might be to allow repeater owners to configure a whitelist of nodes whose adverts they trust?

There’s also a non-trivial risk of clock poisoning when relying on anonymous advert timestamps, and a future dated clock can be a more annoying problem for ESP32 repeaters which can't just restart to reset their clock.

liktor88:
I hope that this change will be added to main release soon. Some of you have doubts how to rely on time from anonymous nodes but in my opinion everything is better than 15 May 2024 after restart :)

4np:
> Thanks for trying to solve clock sync. I’m not convinced that this relatively minor issue warrants the additional complexity introduced by this PR. 

@oltaco , fair points. Note though this PR is basically just adding an abstraction layer for the clock, so from the perspective of a hardware variant not much has changed.

> ..and a future dated clock can be a more annoying problem for ESP32 repeaters which can't just restart to reset their clock.

This is probably a non-issue as devices with a dedicated RTC (such as ESP32) [will bypass](https://github.com/meshcore-dev/MeshCore/pull/1443/files#diff-c8b20e5dd4dd02afb3f6160d22f46082fe364c588fa04b64b0a567922a70c8b5R131) this peer-based synchronization. Peer-based synchronization only applies to devices that don't [have](https://github.com/meshcore-dev/MeshCore/pull/1443/files#diff-f6ea1882edd73c40affbf28f5689b2fe8734fee1a6fa3975829679bb5c22b575R112) an [RTC module](https://github.com/meshcore-dev/MeshCore/pull/1443/files#diff-8bebb3b7be71662549cecf7a66b089b53b814bd997790667a71beb8b6d5f2899R74) present which always reset to **May 15, 2024 10:52:31 UTC** after cold start.

> A simpler and safer approach might be to allow repeater owners to configure a whitelist of nodes whose adverts they trust?
> ...
> There’s also a non-trivial risk of clock poisoning when relying on anonymous advert timestamps

Another option I considered is extending the advert packet to include 0/1 whether or not the repeater contains a dedicated RTC module. That would make things much simpler as you can just take the timestamp of the -preferably low hop count- repeater that has a dedicated RTC module, and account for air time. However, as that meant I had to use the reserved space in the header I didn't go for that option as I don't know if there are already plans to use that reserved space.

Perhaps, as you suggested, there could be an option to pick a trusted repeater from a list of repeaters of -say-  max 10 hops away that advert they have an RTC module (or maybe even a GPS module). However, that also involves client work an ideally a repeater would just work and auto-correct transparently.

For what it's worth, my test repeater has been up for almost 5 days (with `PEER_SYNC_PAUSE_DURATION_SECONDS` set to 20 seconds so it doesn't really pause), and its clock is 22 seconds behind CET / UTC+1. I don't see additional battery drain either (objectively, I don't have a power meter).

> Note: While having an RTC module still does not guarantee a repeater's clock is set correctly, it will have defaulted to May 15, 2024 10:52:31 UTC so that's easy to reject.

4np:
@oltaco, after reading up on ESP32's internal RTC, setting time ahead of UTC indeed sounds problematic. The peer synchronized time logic will be bypassed for I2C discovered RTC modules, which is not the case for ESP32.

Another challenge which may arise with peer-synchronized time is the risk of consensus around an incorrect clock. Once most repeaters rely exclusively on peer synchronization, the network’s time effectively becomes whatever the majority believes to be correct. If the majority of repeaters has an incorrect time (due to manual misconfiguration, drift or poor initial synchronization) then the few repeaters equipped with accurate time sources such as hardware RTCs or GPS may appear as statistical outliers. In that case, median- or consensus-based algorithms can converge on a time that is internally consistent across the mesh, yet still wrong relative to true UTC.

An option could be to make peer synchronized time opt-in via a console command.

vmatt:
Hey @4np, thanks for your work on this!

I agree with @oltaco that a manual opt-in should be necessary here. Additionally, as you said before repeaters with RTCs can have significant time drift if they were set incorrectly initially or haven't been synced in a while. GPS-equipped nodes are really the only ones with a reliably accurate time source.

To address this, it might be worth offering a tiered trust configuration during opt-in:

- Trust all peers – Sync from any node (current behavior, least secure)
- Trust RTC/GPS nodes only – Only sync from nodes advertising a hardware RTC or GPS module
- Trust GPS nodes only – Only sync from nodes with GPS (most accurate/secure)

This would require extending the advert packet to include a flag indicating the node's time source capability (none/RTC/GPS), which you mentioned considering earlier. I think using that reserved header space for this purpose would be justified given the security implications.

This approach gives users explicit control over the trust model while enabling the feature for those who want it.

What do you think?

vmatt:
Hey @4np, thanks for your work on this!

I agree with @oltaco that a manual opt-in should be necessary here. Additionally, as you mentioned repeaters with RTCs can still have significant time drift if set incorrectly initially, so GPS-equipped nodes are really the only reliably accurate time source.

A few suggestions:

Tiered trust configuration during opt-in:

Trust all peers (current behavior, least secure)
Trust RTC/GPS nodes only
Trust GPS nodes only (most accurate/secure)
This would require extending the advert packet to include a flag indicating the node's time source capability (none/RTC/GPS).

Consensus among trusted sources:
The current statistical consensus works across all anonymous peers. Instead, we should require multiple trusted sources (e.g., 2-3 GPS nodes) to agree within a tolerance before accepting sync. A single spoofed node shouldn't be able to poison the clock.

Signed adverts:
Even with a whitelist, bad actors could spoof a trusted repeater's ID since adverts aren't cryptographically signed currently. Long-term, signed timestamps would be the cleanest solution to prevent forgery.

What do you think?

vmatt:
Also take a look at https://github.com/meshcore-dev/MeshCore/pull/1374, if that PR is merged, in case pull based advert, the response will be signed, that should increase trust for white list scenarios.

4np:
With the recent changes related to regions, I think the usefulness of this PR will start to decline as it relies on adverts. When regions start to be implemented, and multi-hop adverts may start getting dropped, the reliability of the MAD algorithm will drop along with it as the pool of considered devices will become smaller. As such, I think it's better to rethink the way to do time synchronization. 

I still think it would be valuable to have devices not be stuck on **May 15, 2024 10:52:31 UTC** though. An alternative approach using dedicated mesh packets:

- repeaters that keep accurate time will be considered as 'time keepers', these are:
  - repeaters with a hardware and battery powered **Real Time Clock** (RTC) (such as [RAK12002](https://docs.rakwireless.com/product-categories/wisblock/rak12002/overview/)), or 
  - repeaters with a GPS module that has a reliable GPS fix (≧ 4 satellites, although reliability depends on GNSS/GPS chip)
- repeaters that don't keep accurate time:
  - will be able to send a 'discovery' mesh packet which _requires_ a region
- intermediate repeaters will apply region ruleset to determine if the packet should be forwarded or dropped
- 'time keeper' repeaters which keep accurate time and that receive the 'discovery packet'
  - will not forward the discovery packet
  - will respond with their name and full public key
- repeaters that don't keep accurate time
  - will be able to discover repeaters that keep accurate time
  - will support configuring an allow list of repeater's public keys
  - can be configured to to sync the time (on cold start? at set intervals?)
- when the allow list is configured, the repeater that does not keep accurate time:
  - can send a 'request for time' mesh packet for each of 'time keeper' repeaters in its allow list
  - include the (part of the) public key of the 'time keeper' repeater
- the 'time keeper' repeater that matches the public key will respond with a signed 'time mesh packet'
  - signed using its private key
  - includes the actual UTC time
  - as such has a relatively small payload
- the repeater that sent out the 'request for time' will check the received packet against the public keys in its allow list
  - if it matches, updates its local clock by taking the approximate air-time into account.

> Note: the discovery and discovery response could be optional or added later, as just configuring the allow list could cover the primary use case of time synchronization.

PR #1446, opened 2026-01-22
Add an external watchdog module

Quency-D:
1. It has already been tested on heltec-mesh-solar and heltec-v4.
2. The newly added watchdog module does not conflict with the existing meshsolar.


0x156:
Found this from here: http://community.heltec.cn/t/heltec-meshtower-resets-its-configuration-itself/20570/36?u=brikots
I have flashed it to my Heltec mesh solar/tower, and I can confirm it works properly: it does not reboot every 9 minutes.

src/helpers/ExWatchdogManager.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+class ExWatchdogManager {
+public:
+  unsigned long next_feed_watchdog;
+  ExWatchdogManager() { next_feed_watchdog = 0; }
```
weebl2000: `next_feed_watchdog` is exposed as public but should be an implementation detail:

```suggestion
protected:
  unsigned long next_feed_watchdog;
public:
  ExWatchdogManager() { next_feed_watchdog = 0; }
```

variants/heltec_mesh_solar/target.cpp
```diff
@@ -121,3 +122,30 @@ bool SolarSensorManager::setSettingValue(const char* name, const char* value) {
...
+void SolarExWatchdog::loop() {
+  if (millis() > next_feed_watchdog) {
+    feed();
+    next_feed_watchdog = millis() + EX_WATCHDOG_TIMEOUT_MS;
+  }
+}
+
+unsigned long SolarExWatchdog::getIntervalMs() const {
+    return next_feed_watchdog - millis();
+}
```
weebl2000: `next_feed_watchdog - millis()` underflows to ~4 billion when `millis()` has passed the deadline:

```suggestion
unsigned long SolarExWatchdog::getIntervalMs() const {
    unsigned long now = millis();
    return (now < next_feed_watchdog) ? (next_feed_watchdog - now) : 0;
}
```

examples/simple_repeater/main.cpp
```diff
@@ -124,11 +128,17 @@ void loop() {
...
-
+#ifdef HAS_EX_WATCHDOG
+  ex_watchdog.loop();
+#endif
   if (the_mesh.getNodePrefs()->powersaving_enabled &&                     // To check if power saving is enabled
       the_mesh.millisHasNowPassed(lastActive + nextSleepinSecs * 1000)) { // To check if it is time to sleep
     if (!the_mesh.hasPendingWork()) { // No pending work. Safe to sleep
+#ifdef HAS_EX_WATCHDOG
+      board.sleep(ex_watchdog.getIntervalMs()>1800?1800:ex_watchdog.getIntervalMs());             // To sleep. Wake up after 30 minutes or when receiving a LoRa packet
+#else
```
weebl2000: Three issues here:
1. **Units mismatch**: `getIntervalMs()` returns milliseconds but is compared to `1800` (seconds). `board.sleep()` takes seconds.
2. **TOCTOU**: `getIntervalMs()` is called twice — `millis()` advances between calls so the two values differ.
3. **Missing feed**: The watchdog should be fed right before sleeping, otherwise the remaining time may be too short to survive the sleep + wake cycle.

```suggestion
#ifdef HAS_EX_WATCHDOG
      ex_watchdog.feed();  // reset watchdog before sleeping
      { unsigned long wdSec = ex_watchdog.getIntervalMs() / 1000;
        board.sleep(wdSec < 1800 ? wdSec : 1800); }
#else
```

variants/heltec_mesh_solar/target.cpp
```diff
@@ -121,3 +122,30 @@ bool SolarSensorManager::setSettingValue(const char* name, const char* value) {
...
+
+void SolarExWatchdog::feed() {
+    digitalWrite(EX_WATCHDOG_DONE_PIN, HIGH);
+    delay(1);
+    digitalWrite(EX_WATCHDOG_DONE_PIN, LOW);
+}
```
weebl2000: Missing newline at end of file:

```suggestion
}
```

weebl2000 (commented):
A few suggestions — the main one is a units mismatch (ms vs seconds) in the repeater sleep guard.

Quency-D:
@weebl2000 Thank you for your suggestions. All have been revised.

liamcottle:
Please refactor all instances of `ex` to `external`, so the language is more explicit.

E.g:
- `HAS_EX_WATCHDOG` -> `HAS_EXTERNAL_WATCHDOG`
- `EX_WATCHDOG_DONE_PIN` -> `EXTERNAL_WATCHDOG_DONE_PIN`
- `EX_WATCHDOG_WAKE_PIN` -> `EXTERNAL_WATCHDOG_WAKE_PIN`
- `EX_WATCHDOG_TIMEOUT_MS` -> `EXTERNAL_WATCHDOG_TIMEOUT_MS`
- `ExWatchdogManager` -> `ExternalWatchdogManager`
- `SolarExWatchdog` -> `SolarExternalWatchdog`
- `ex_watchdog` -> `external_watchdog`

weebl2000:
@Quency-D addressed the review comments and opened a PR to your branch https://github.com/Quency-D/MeshCore/pull/4

weebl2000:
@liamcottle addressed the comments, variables are clearly named now!

Quency-D:
> @Quency-D addressed the review comments and opened a PR to your branch [Quency-D#4](https://github.com/Quency-D/MeshCore/pull/4)

Thank you for your contribution; the connection has been merged.

IoTThinks:
ESP32 has various watchdogs.
May I know the context or usecases why we need this extra watchdog?

Quency-D:
> ESP32 has various watchdogs. May I know the context or usecases why we need this extra watchdog?

An external watchdog is more reliable than an internal one. It operates independently, providing an extra layer of protection.

liamcottle (approved):

variants/heltec_mesh_solar/target.cpp
```diff
@@ -121,3 +122,35 @@ bool SolarSensorManager::setSettingValue(const char* name, const char* value) {
...
+  pinMode(EXTERNAL_WATCHDOG_WAKE_PIN, INPUT);
+  pinMode(EXTERNAL_WATCHDOG_DONE_PIN, OUTPUT);
+  delay(1);
+  digitalWrite(EXTERNAL_WATCHDOG_DONE_PIN, LOW);
+  delay(1);
+  feed();
+  return true;
+}
+void SolarExternalWatchdog::loop() {
+  if (millis() > next_feed_watchdog) {
```
ripplebiz: Just wondering about doing the 2's complement trick here?  So that the 49-day glitch doesn't happen.

IoTThinks:
@Quency-D What are the real usecases or problems this watchdog is trying to address / fix?

variants/heltec_mesh_solar/target.cpp
Quency-D: This is indeed a problem; let me think of a good solution.

Quency-D:
> @Quency-D What are the real usecases or problems this watchdog is trying to address / fix?

Outdoor solar energy equipment can guarantee continuous normal operation.

variants/heltec_mesh_solar/target.cpp
```diff
@@ -121,3 +122,35 @@ bool SolarSensorManager::setSettingValue(const char* name, const char* value) {
...
+    feed();
+    next_feed_watchdog = millis() + EXTERNAL_WATCHDOG_TIMEOUT_MS;
+  }
+}
+
+unsigned long SolarExternalWatchdog::getIntervalMs() const {
```
ripplebiz, line 127: If I'm reading this right, this is the _remaining_ time left until next feed()  ?
Or should this be whatever the fixed interval is?

variants/heltec_mesh_solar/target.cpp
ripplebiz: Just see Dispatcher::millisHasNowPassed() for reference.  Usually, you just keep a _last_ (ie in the past) timestamp, and compare to millis() for some _duration_ value. Rather than holding a _future_ timestamp

variants/heltec_mesh_solar/target.cpp
```diff
@@ -121,3 +122,35 @@ bool SolarSensorManager::setSettingValue(const char* name, const char* value) {
...
+  delay(1);
+  digitalWrite(EXTERNAL_WATCHDOG_DONE_PIN, LOW);
+  delay(1);
+  feed();
+  return true;
+}
+void SolarExternalWatchdog::loop() {
+  if (millis() > next_feed_watchdog) {
+    feed();
+    next_feed_watchdog = millis() + EXTERNAL_WATCHDOG_TIMEOUT_MS;
```
ripplebiz: I don't know the specifics of the Mesh Solar's watchdog timer, but does it just have a default timeout of some known value?  And is the EXTERNAL_WATCHDOG_TIMEOUT_MS some _safe_ interval INSIDE this default timeout interval?  (ie. you don't want to two to be exactly the same, as there will be slight delays until feed() will be called)

Quency-D: The hardware is configured to run for about 10 minutes, while the code is configured for 8 minutes, leaving a 2-minute margin, which should be more than enough.

variants/heltec_mesh_solar/target.cpp
Quency-D: This is the time remaining until the next feed(), not a fixed time interval.

Quency-D:
> Just see Dispatcher::millisHasNowPassed() for reference. Usually, you just keep a last (ie in the past) timestamp, and compare to millis() for some duration value. Rather than holding a future timestamp

Hi @ripplebiz, I made some changes. I'm now using the last feeding time as the criterion instead of a future time. I'm not sure if my understanding is correct.

Quency-D:
Hi @liamcottle @ripplebiz, this PR looks good now. Could you please help review and merge it when you have a chance?

beala:
This was never merged and is now breaking the Mesh Tower V2 https://github.com/meshcore-dev/MeshCore/pull/2904

PR #1467, opened 2026-01-23
Add SPI support for SSD1306 display and boot status UI

kotovasia5120:
Introduces optional SPI support for SSD1306 displays via USE_SPI_SSD1306 (Note: current display implementation shares the same SPI bus as the radio.), including relevant pin and SPI class configuration. Also adds optional on-screen boot status messages in main.cpp for boards with a display, improving user feedback during initialization.

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -2,6 +2,16 @@
 #include <Mesh.h>
 #include "MyMesh.h"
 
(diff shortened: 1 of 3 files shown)


liamcottle:
Hey, thanks for the PR. It looks like this one didn't get any traction. Will close out due to old age and merge conflicts.

PR #1476, opened 2026-01-24
Add cli command `region list {allowed|denied}`, enable output of region cmd via remote cli

mattzzw:
As of today the `region` command is limited to serial output only. [1]

This PR adds the ability to also access the output of the `region` command via the remote cli.
Many repeaters are in remote and/or inaccessible locations. The need for a physical USB connection to debug region settings of already installed repeaters is far from ideal, IMHO.
As region configuration is unlikely to change frequently this is a trade-off between usability and LoRa data minimalism.

The output is limited to 160 chars for remote cli limiting the data to one LoRa packet. As a consequence more complex region setups might not be printed out completely.

[1] https://github.com/meshcore-dev/MeshCore/wiki/Repeater-&-Room-Server-CLI-Reference

### DIFF
--- a/src/helpers/RegionMap.cpp
+++ b/src/helpers/RegionMap.cpp
@@ -2,6 +2,45 @@
 #include <helpers/TxtDataHelpers.h>
 #include <SHA256.h>
 
+// helper class for region map exporter, we emulate Stream with a safe buffer writer.
+
+class BufStream : public Stream {
+public:
+  BufStream(char *buf, size_t max)
+    : _buf(buf), _max(max), _pos(0) {
+    if (_max > 0) _buf[0] = 0;
+  }
+
+  size_t write(uint8_t c) override {
+    if (_pos + 1 >= _max) return 0;
+    _buf[_pos++] = c;
+    _buf[_pos] = 0;
+    return 1;
+  }
+
+  size_t write(const uint8_t *buffer, size_t size) override {
+    size_t written = 0;
+    while (written < size) {
+      if (!write(buffer[written])) break;
+      written++;
+    }
+    return written;
+  }
+
+  int available() override { return 0; }
+  int read() override { return -1; }
+  int peek() override { return -1; }
+  void flush() override {}
+
+  size_t length() const { return _pos; }
+
+private:
+  char *_buf;
+  size_t _max;
+  size_t _pos;
+};
+
+
 RegionMap::RegionMap(TransportKeyStore& store) : _store(&store) {
   next_id = 1; num_regions = 0; home_id = 0;
   wildcard.id = wildcard.parent = 0;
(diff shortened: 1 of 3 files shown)


ripplebiz:
I had been toying with possibly adding a CLI command which uses the existing exportNamesTo() method, and just filtering on the FLOOD permission. ie just list comma sep region names of ones that are currently allowed. Can you re-work this PR to do that?

ripplebiz:
Was just chatting with Liam, and we think this CLI command would be good: "region list {allowed|denied}"

mattzzw:
Got it. Would the new commands be in addition to the `region` command or replacing it?

mattzzw:
Added `region list {allowed|denied}` command

(Don't know if this example configuration actually makes sense)

```
region
  -> * F
 de
  de-hh^ F
  de-ni F
  de-sh F
  de-mv
  de-sa

region list allowed
  -> *,de-hh,de-ni,de-sh
region list denied
  -> de,de-mv,de-sa
>
```

mesher-de:
@mattzzw 
Your logfile ‎`meshcore-evo-fw.log‎` is part of this PR.
Is that intended?

mattzzw:
> Is that intended?

Nope, thanks for the heads up!

oltaco:
It would be good to tidy this up with a rebase on dev instead of merge, and then a squash so that it's just one commit. It makes the git history so much cleaner to look through.

src/helpers/RegionMap.cpp
```diff
@@ -249,25 +288,40 @@ void RegionMap::exportTo(Stream& out) const {
...
-      int len = strlen(name);
+    
+    // Check if region matches the filter criteria
+    bool region_matches = invert ? (region->flags & mask) : !(region->flags & mask);
+    
+    if (region_matches) {
+      int len = strlen(region->name);
       if ((dp - dest) + len + 2 < max_len) {   // only append if name will fit
-        memcpy(dp, name, len);
+        memcpy(dp, region->name, len);
```
ripplebiz: need to add skip_hash() here

src/helpers/RegionMap.cpp
```diff
@@ -249,25 +288,40 @@ void RegionMap::exportTo(Stream& out) const {
...
     auto region = &regions[i];
-    if ((region->flags & mask) == 0) {   // region allowed? (per 'mask' param)
-      const char* name = skip_hash(region->name);
-      int len = strlen(name);
+    
+    // Check if region matches the filter criteria
+    bool region_matches = invert ? (region->flags & mask) : !(region->flags & mask);
+    
+    if (region_matches) {
+      int len = strlen(region->name);
```
ripplebiz: need to trim the optional '#' now, with skip_hash()

PR #1477, opened 2026-01-25
Expose a counter to track RadioLib receive errors

Cisien:
This change counts when `readData` returns an err code other than `RADIOLIB_ERR_NONE`. This will be a CRC error in most cases, with edge cases such as bad radio settings also incrementing this. A CRC error would be caused by conditions such as collisions, interference, and weak signals.

This counter is exposed in the `stats-packets` command, and in the repeater stats payload as 4 additional bytes added to the end of the structure. The app handles this gracefully.

My incompetent robot claims the total payload size is 96 bytes (unverified but probably close).

I intend to use this field in the LetsMesh analyzer to get a better idea of how much impact my repeaters are seeing from collisions/interference.

### DIFF
--- a/src/helpers/radiolib/RadioLibWrappers.h
+++ b/src/helpers/radiolib/RadioLibWrappers.h
@@ -7,7 +7,7 @@ class RadioLibWrapper : public mesh::Radio {
 protected:
   PhysicalLayer* _radio;
   mesh::MainBoard* _board;
-  uint32_t n_recv, n_sent;
+  uint32_t n_recv, n_sent, n_recv_errors;
   int16_t _noise_floor, _threshold;
   uint16_t _num_floor_samples;
   int32_t _floor_sample_sum;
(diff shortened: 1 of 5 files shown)


src/helpers/radiolib/RadioLibWrappers.h
```diff
@@ -45,8 +45,9 @@ class RadioLibWrapper : public mesh::Radio {
   void loop() override;
 
   uint32_t getPacketsRecv() const { return n_recv; }
+  uint32_t getPacketsRecvErrors() const { return n_recv_errors; }
```
oltaco, line 48: Minor nitpick, should this be getPacketRecvErrors() ?

oltaco:
Nice, I've been wondering about adding this for a while now. Since our local mesh has grown I'm now seeing a non-negligible amount of -707 in the serial log while doing testing and this would be a great addition to track.

PR #1483, opened 2026-01-26
fix: avoid redundant redefinition of SX126X_DIO3_TCXO_VOLTAGE

ssp97:
SX126X_DIO3_TCXO_VOLTAGE was repeatedly defined in header files when
multiple headers were included, leading to unintended TCXO voltage
override.

Ensure the macro is defined only once.

```
#include <helpers/radiolib/CustomSX1262Wrapper.h>   <- ok
#include <helpers/radiolib/CustomSX1268Wrapper.h>   <- ng
```


weebl2000:
@liamcottle Simple, clean tidying up, I think it's good to merge.

PR #1485, opened 2026-01-26
Fix onAnonDataReq() always sending response as flood

oltaco:
This PR fixes onAnonDataReq() always sending response as flood, even when repeater login etc was sent with a path or direct.

Confirmed that now direct and pathed logins will respond via path instead of flood, and flood logins still reply with flood.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -128,8 +128,11 @@ uint8_t MyMesh::handleLoginReq(const mesh::Identity& sender, const uint8_t* secr
     }
   }
 
-  if (is_flood) {
-    client->out_path_len = -1;  // need to rediscover out_path
+  if (!is_flood && client->out_path_len >= 0) { // login via path and we have an out_path so use it
+    reply_path_len = client->out_path_len;
+    memcpy(reply_path, client->out_path, client->out_path_len);
+  } else {
+    client->out_path_len = -1; // need to rediscover out_path
   }
 
   uint32_t now = getRTCClock()->getCurrentTimeUnique();
--- a/examples/simple_sensor/SensorMesh.cpp
+++ b/examples/simple_sensor/SensorMesh.cpp
@@ -359,8 +359,11 @@ uint8_t SensorMesh::handleLoginReq(const mesh::Identity& sender, const uint8_t* 
     dirty_contacts_expiry = futureMillis(LAZY_CONTACTS_WRITE_DELAY);
   }
 
-  if (is_flood) {
-    client->out_path_len = -1;  // need to rediscover out_path
+  if (!is_flood && client->out_path_len >= 0) { // login via path and we have an out_path so use it
+    reply_path_len = client->out_path_len;
+    memcpy(reply_path, client->out_path, client->out_path_len);
+  } else {
+    client->out_path_len = -1; // need to rediscover out_path
   }
 
   uint32_t now = getRTCClock()->getCurrentTimeUnique();
@@ -469,9 +472,12 @@ void SensorMesh::onAnonDataRecv(mesh::Packet* packet, const uint8_t* secret, con
       mesh::Packet* path = createPathReturn(sender, secret, packet->path, packet->path_len,
                                             PAYLOAD_TYPE_RESPONSE, reply_data, reply_len);
       if (path) sendFlood(path, SERVER_RESPONSE_DELAY);
-    } else {
+    } else if (reply_path_len < 0) {
       mesh::Packet* reply = createDatagram(PAYLOAD_TYPE_RESPONSE, sender, secret, reply_data, reply_len);
       if (reply) sendFlood(reply, SERVER_RESPONSE_DELAY);
+    } else {
+      mesh::Packet* reply = createDatagram(PAYLOAD_TYPE_RESPONSE, sender, secret, reply_data, reply_len);
+      if (reply) sendDirect(reply, reply_path, reply_path_len, SERVER_RESPONSE_DELAY);
     }
   }
 }
--- a/examples/simple_sensor/SensorMesh.h
+++ b/examples/simple_sensor/SensorMesh.h
@@ -136,6 +136,8 @@ private:
   ClientACL  acl;
   CommonCLI _cli;
   uint8_t reply_data[MAX_PACKET_PAYLOAD];
+  uint8_t reply_path[MAX_PATH_SIZE];
+  int8_t  reply_path_len;
   unsigned long dirty_contacts_expiry;
   CayenneLPP telemetry;
   uint32_t last_read_time;


oltaco:
Converted to draft, @ripplebiz pointed out that the handlers are meant to do this so I'll investigate further.

I'll also remember to include room server and sensor next time 🤦‍♂️

fizzyfuzzle:
@oltaco I am currently running this PR and its indeed fixing the login response flood (its now a nice direct packet).

I was wondering if this is still waiting for something, because in NL the mesh is very (very) busy and now we see a lot of additional unnecessary floods from the remote admin, which will be fixed with this PR ;)

Thanks in advance

oltaco:
Amazing timing! I literally just finished testing this and confirm it works as expected. I just need to add the fix to room server firmware and then I'll mark it ready for review 👍

oltaco:
It turns out that room server login responses were already sent direct if the login request arrived via direct path and a path is available. This PR just brings Repeater and Sensor firmwares in line with that.

I've tested it and it works as intended, flood logins will generate path packet and encode the response in that, otherwise the login response is sent via direct path.

@ripplebiz What do you think?

fizzyfuzzle:
@ripplebiz with around 1800 Repeaters in NL, this PR would help to reduce the load on flooding when using Remote Admin, reducing  the Airtime RX a bit for the whole mesh.
Is this waiting for something, or is it fallen of the radar ;)

Thanks in advance!

fizzyfuzzle:
@oltaco @liamcottle please don't forget this (relatively) simple PR :sweat_smile:

examples/simple_repeater/MyMesh.cpp
```diff
@@ -128,8 +128,11 @@ uint8_t MyMesh::handleLoginReq(const mesh::Identity& sender, const uint8_t* secr
     }
   }
 
-  if (is_flood) {
-    client->out_path_len = -1;  // need to rediscover out_path
+  if (!is_flood && client->out_path_len >= 0) { // login via path and we have an out_path so use it
+    reply_path_len = client->out_path_len;
+    memcpy(reply_path, client->out_path, client->out_path_len);
```
weebl2000, line 133: - client->out_path_len >= 0 is always true because it's uint8_t, the first-time direct login out_path_len is OUT_PATH_UNKNOWN (0xFF) and memcpy runs unconditionally, then tries to copy 255 bytes from a 64-byte buffer? We need to check if we have a known path.

```suggestion
  if (!is_flood && client->out_path_len != OUT_PATH_UNKNOWN) { // login via path and we have an out_path so use it
    reply_path_hash_size = (client->out_path_len >> 6) + 1;
    reply_path_len = client->out_path_len & 63;
    mesh::Packet::copyPath(reply_path, client->out_path, client->out_path_len);
```

examples/simple_repeater/MyMesh.cpp
```diff
@@ -128,8 +128,11 @@ uint8_t MyMesh::handleLoginReq(const mesh::Identity& sender, const uint8_t* secr
...
+  } else {
+    client->out_path_len = -1; // need to rediscover out_path
```
weebl2000, line 135: Style

- client->out_path_len = -1 works (wraps to 0xFF = OUT_PATH_UNKNOWN) but is misleading since the field is uint8_t.


```suggestion
    client->out_path_len = OUT_PATH_UNKNOWN; // need to rediscover out_path
```

examples/simple_sensor/SensorMesh.cpp
```diff
@@ -359,8 +359,11 @@ uint8_t SensorMesh::handleLoginReq(const mesh::Identity& sender, const uint8_t*
     dirty_contacts_expiry = futureMillis(LAZY_CONTACTS_WRITE_DELAY);
   }
 
-  if (is_flood) {
-    client->out_path_len = -1;  // need to rediscover out_path
+  if (!is_flood && client->out_path_len >= 0) { // login via path and we have an out_path so use it
+    reply_path_len = client->out_path_len;
+    memcpy(reply_path, client->out_path, client->out_path_len);
```
weebl2000, line 364: same, uint8_t is always >= 0 👀 

```suggestion
  if (!is_flood && client->out_path_len != OUT_PATH_UNKNOWN) { // login via path and we have an out_path so use it
    reply_path_len = client->out_path_len;
    mesh::Packet::copyPath(reply_path, client->out_path, client->out_path_len);
```

examples/simple_sensor/SensorMesh.cpp
```diff
@@ -359,8 +359,11 @@ uint8_t SensorMesh::handleLoginReq(const mesh::Identity& sender, const uint8_t*
...
+  } else {
+    client->out_path_len = -1; // need to rediscover out_path
```
weebl2000, line 366: Style, same as above.

```suggestion
    client->out_path_len = OUT_PATH_UNKNOWN; // need to rediscover out_path
```

examples/simple_sensor/SensorMesh.h
```diff
@@ -136,6 +136,8 @@ class SensorMesh : public mesh::Mesh, public CommonCLICallbacks {
   ClientACL  acl;
   CommonCLI _cli;
   uint8_t reply_data[MAX_PACKET_PAYLOAD];
+  uint8_t reply_path[MAX_PATH_SIZE];
+  int8_t  reply_path_len;
```
weebl2000, line 140: `int8_t` truncates encoded path_len values > 127 (e.g. 2-byte hash paths). Use `uint8_t` and `OUT_PATH_UNKNOWN` as sentinel instead of negative check.

```suggestion
  uint8_t reply_path_len;
```

examples/simple_sensor/SensorMesh.cpp
```diff
@@ -469,9 +472,12 @@ void SensorMesh::onAnonDataRecv(mesh::Packet* packet, const uint8_t* secret, con
       mesh::Packet* path = createPathReturn(sender, secret, packet->path, packet->path_len,
                                             PAYLOAD_TYPE_RESPONSE, reply_data, reply_len);
       if (path) sendFlood(path, SERVER_RESPONSE_DELAY);
-    } else {
+    } else if (reply_path_len < 0) {
```
weebl2000, line 475: - `reply_path_len` is a class member but never initialized in `onAnonDataRecv` before `handleLoginReq`. Stale value from a previous call could route to the wrong peer. Initialize at the top of `onAnonDataRecv` and adjust condition to match `uint8_t` type.

```suggestion
    } else if (reply_path_len == OUT_PATH_UNKNOWN) {
```

weebl2000 (changes requested):

446564:
:eyes: still want to get this one in @oltaco ?

oltaco:
> 👀 still want to get this one in @oltaco ?

I don't think it's out of the question but we don't want to risk making logins to remote repeaters worse.

It can be sometimes be difficult to get a login response from a flood login to a remote repeater (presumably due to collisions) where a direct login is OK. With this PR there's a risk that it forces people to send a flood login first so that they can get a new path, instead of just sending a direct login and getting a flooded login response.

Login response is at least a fairly small packet compared to a login request which includes your pubkey.

446564:
very good point, maybe we move to draft until ready to review again? (might be a good habit to get into anyway?) What do you think?

Issue #1482, opened 2026-01-25
Last Message on Display


mirkosw:
Hello , 
Currently, I always see the first message I receive after disconnecting from my phone. It would certainly be better if the last received message was always displayed on the device's screen.

thanks

oltaco:
I agree with this, I find it annoying that every time a new message comes through the screen lights up and it just shows the first message that arrived when the node wasn't connected.

VID-PRO:
i am totally with you.
would be a great update

PR #1487, opened 2026-01-26
Show most recent message on display when no client is connected

oltaco:
This PR changes the UI so that when no client is connected the display shows the most recent message instead of the first unread one. Incoming messages evict the oldest messages, so the UI always shows the latest 32 messages

Closes https://github.com/meshcore-dev/MeshCore/issues/1482

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -458,15 +458,17 @@ class MsgPreviewScreen : public UIScreen {
   };
   #define MAX_UNREAD_MSGS   32
   int num_unread;
+  int head = MAX_UNREAD_MSGS - 1; // index of latest unread message
   MsgEntry unread[MAX_UNREAD_MSGS];
 
 public:
   MsgPreviewScreen(UITask* task, mesh::RTCClock* rtc) : _task(task), _rtc(rtc) { num_unread = 0; }
 
   void addPreview(uint8_t path_len, const char* from_name, const char* msg) {
-    if (num_unread >= MAX_UNREAD_MSGS) return;  // full
+    head = (head + 1) % MAX_UNREAD_MSGS;
+    if (num_unread < MAX_UNREAD_MSGS) num_unread++;
 
-    auto p = &unread[num_unread++];
+    auto p = &unread[head];
     p->timestamp = _rtc->getCurrentTime();
     if (path_len == 0xFF) {
       sprintf(p->origin, "(D) %s:", from_name);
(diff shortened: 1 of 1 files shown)


oltaco:
@ripplebiz @liamcottle Any thoughts on this one?

liamcottle:
I haven't had a chance to test this, but a quick comment, maybe it already does it ;)

Being able to use the left/right keys to go back and forward between unread messages would be neat, and a way to recall those unread messages if they were dismissed and not retrieved by phone yet.

I found when clicking the wrong button, all messages would disappear, and it wasn't possible to view again, even though UI said 5 MSGS. Had to use phone app to view them.

oltaco:
> Being able to use the left/right keys to go back and forward between unread messages would be neat, and a way to recall those unread messages if they were dismissed and not retrieved by phone yet.
> 
> I found when clicking the wrong button, all messages would disappear, and it wasn't possible to view again, even though UI said 5 MSGS. Had to use phone app to view them.

Yeah that is annoying. I did consider a change to allow devices with "joystick" to scroll through messages without clearing them, but I decided to go for something a bit more simple to begin with.

In the meantime this PR means not having the screen light up with the same ancient message every time a new one comes in which is a much better user experience in my opinion.

We should definitely think about how to make messages on the node UI better. Maybe clearing all messages shouldn't be a single click in joystick mode? Or it should be moved to the other button?

I didn't look at bringing messages back, but I think that would require a bit of refactoring. Maybe the UI should just have access to the full unread message queue instead of new messages getting pushed to a separate UI queue?

Food for thought :)

liamcottle:
Happy to merge as is for now, I agree showing latest message is a better experience. Just tested on my WioTracker L1 Pro.

Potentially for a future PR, it would be nice if the button press went to next message instead of clearing all immediately. It's more natural to press the joystick enter button to dismiss/mark read, rather than arrow right. I find myself clearing all unread messages by default quite often.

Thoughts for future PR:
- pressing joystick enter marks current viewed message as read (clears it)
- pressing left/right joystick button navigates back and forward through unread messages
- left/right should not clear the unread message, it should stay until enter is pressed to mark read.
- pressing joystick enter when it shows MSG: 9 on screen should bring them all back up to view again
- maybe that secondary, small orange button could be classified as back, to exit out of view messages mode?

Something to think about anyway. As if you don't have your phone it's impossible to read the pending messages that may have been dismissed.

oltaco:
> Thoughts for future PR:
> 
> * pressing joystick enter marks current viewed message as read (clears it)
> * pressing left/right joystick button navigates back and forward through unread messages
> * left/right should not clear the unread message, it should stay until enter is pressed to mark read.
> * pressing joystick enter when it shows MSG: 9 on screen should bring them all back up to view again
> * maybe that secondary, small orange button could be classified as back, to exit out of view messages mode?
> 
> Something to think about anyway. As if you don't have your phone it's impossible to read the pending messages that may have been dismissed.

Yeah this sounds much nicer. I've made a note to come back and look at this stuff soon 👍

PR #1490, opened 2026-01-26
Add CLI-command for zerohop advert

mesher-de:
Add a command to CLI to allow repeater owner to send a 0-hop advert manually as this can help reduce load on the mesh.

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -200,6 +200,10 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
       // Reset clock
       getRTCClock()->setCurrentTime(1715770351);  // 15 May 2024, 8:50pm
       _board->reboot();  // doesn't return
+     } else if (memcmp(command, "advert.zerohop", 14) == 0 && (command[14] == 0 || command[14] == ' ')) {
+      // send zerohop advert
+      _callbacks->sendSelfAdvertisement(1500, false);  // longer delay, give CLI response time to be sent first
+      strcpy(reply, "OK - zerohop advert sent");
     } else if (memcmp(command, "advert", 6) == 0) {
       // send flood advert
       _callbacks->sendSelfAdvertisement(1500, true);  // longer delay, give CLI response time to be sent first


nakoeppen:
I love this feature being added to the CLI, but to clean this up for the sake of the end user, it might be nice to append "flood" to the preexisting advert command and to spell out "zerohop" for the zero hop variant so as to match the MeshCore Client App UI. Most certainly not a requirement, but just a thought to avoid having to clean this up later.

mesher-de:
@nakoeppen Thanks for the feedback!
Happy to change the spelling to "zerohop", Thanks for the heads-up.

Please let me check regarding appending flood to the legacy command, as this should not interfere with legacy features

mesher-de:
Changed the spelling to zerohop

If I modify the "old" flood-command, it would break the "Advert"-button in the UI

Any suggestions?

nakoeppen:
> Changed the spelling to zerohop
> 
> 
> 
> If I modify the "old" flood-command, it would break the "Advert"-button in the UI
> 
> 
> 
> Any suggestions?

@liamcottle any thoughts here? If we need to keep it as just "advert" that's fine as well. I just thought it might be a low-hanging fruit for improving CLI usability, particularly to avoid people from defaulting to the advert command and flooding the network unnecessarily as a result.

mesher-de:
> > Changed the spelling to zerohop
> > If I modify the "old" flood-command, it would break the "Advert"-button in the UI
> > Any suggestions?
> 
> @liamcottle any thoughts here? If we need to keep it as just "advert" that's fine as well. I just thought it might be a low-hanging fruit for improving CLI usability, particularly to avoid people from defaulting to the advert command and flooding the network unnecessarily as a result.

I like that thought!
We also could keep the default "advert" command, modifying it to only trigger a 0-hop-advert and add "advert flood" as a new command instead.

Just thinking...

nakoeppen:
> I like that thought!
> 
> We also could keep the default "advert" command, modifying it to only trigger a 0-hop-advert and add "advert flood" as a new command instead.
> 
> 
> 
> Just thinking...

I certainly think from a network integrity standpoint that this would be the best route, but input from project admins would also be appreciated!

mesher-de:
Updated syntax to `advert.zerohop`

weebl2000 (approved):

weebl2000:
Simple, correct. @liamcottle @ripplebiz looking at reducing mesh traffic and especially flood averts I think this is a command worth having.

liamcottle:
> Simple, correct. @liamcottle @ripplebiz looking at reducing mesh traffic and especially flood averts I think this is a command worth having.

Yep, I commented on the other PR, but got sidetracked and never came back to this one.
https://github.com/meshcore-dev/MeshCore/pull/1714

Will merge. Keep in mind, sending `advert.zerohop` to older firmware would still send flood advert.

PR #1492, opened 2026-01-27
Support for Meshtiny

oltaco:
Support for meshtiny. All functions working. Used the 3-way "encoder" switch on top to act like the "joystick" on the WioTrackerL1. Hibernate works with wake from the side button.

I left the annoying Bluefruit flashy LED and the status LED disabled to try and save some juice, the tiny little thing only has 250mah :)

image

### DIFF
--- a/boards/meshtiny.json
+++ b/boards/meshtiny.json
@@ -1,0 +1,74 @@
+{
+  "build": {
+    "arduino": {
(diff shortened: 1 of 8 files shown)


ripplebiz:
haha ... "tiny taco" :-)

PR #1493, opened 2026-01-27
Add a cli command reference document

Cisien:
This list was largely LLM generated from the codebase. I've tweaked and adjusted things quite a bit as well as confirmed many of these commands based on my own knowledge and referencing the code "the old fashioned way", but I may have (likely) missed something, or my layout/summary/descriptions are confusing or lacking details.

PLEASE suggest changes! I want to make it easier for people to understand and use these commands!

The intent for this document is to replace the hard to maintain wiki page that we currently use.

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -1,0 +1,881 @@
+# MeshCore Repeater & Room Server CLI Commands
+
+## Navigation
+
+- [Operational](#operational)
+- [Neighbors](#neighbors-repeater-only)
+- [Statistics](#statistics)
+- [Logging](#logging)
+- [Information](#info)
+- [Configuration](#configuration)
+  - [Radio](#radio)
+  - [System](#system)
+  - [Routing](#routing)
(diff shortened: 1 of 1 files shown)


Cisien:
The github "files changed" view of the markdown preview is ... weird. It's easier to review the document in my repo: https://github.com/Cisien/MeshCore/blob/dev/docs/cli_commands.md

ripplebiz:
Just some edits needed:
* flood.max can be 0
* region list   can be over LoRa
* Various region load examples have missing newline after ‘load’

timdj:
Would be nice to add the default values as well?

Cisien:
> Just some edits needed:
> 
> * flood.max can be 0
> * region list   can be over LoRa
> * Various region load examples have missing newline after ‘load’

Fixed

> Would be nice to add the default values as well?

Added Defaults

Cisien:
The latest push added the build flag(s) that set the initial value, where applicable

PR #1495, opened 2026-01-27
Fix unlimited advert blob growth on ESP32

oltaco:
This PR corrects ESP32 (and RP2040?) devices never removing old advert blobs, leading to the filesystem getting filled.

Changes:
- new DataStore::deleteBlobByKey() function
- tweaked onAdvertRecv() logic to only store advert blobs when adding/updating contacts.
- new makeBlobPath inline helper for ESP32

Tested on Heltec V3 and confirmed that advert blobs are now removed on overwrite and when removing contacts via the app.
Tested on RAK4631 and confirmed no change in behaviour.

### DIFF
--- a/examples/companion_radio/DataStore.cpp
+++ b/examples/companion_radio/DataStore.cpp
@@ -564,13 +564,16 @@ bool DataStore::deleteBlobByKey(const uint8_t key[], int key_len) {
   return true; // this is just a stub on NRF52/STM32 platforms
 }
 #else
-uint8_t DataStore::getBlobByKey(const uint8_t key[], int key_len, uint8_t dest_buf[]) {
-  char path[64];
+inline void makeBlobPath(const uint8_t key[], int key_len, char* path, size_t path_size) {
   char fname[18];
-
   if (key_len > 8) key_len = 8; // just use first 8 bytes (prefix)
   mesh::Utils::toHex(fname, key, key_len);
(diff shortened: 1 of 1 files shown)


oltaco:
Thankfully @liamcottle noticed that putBlobByKey() wasn't being called in the update contact path. Fixed now.

oltaco:
Now with added helper 👍

liamcottle:
Looks good thanks! Removed some duped code, and advert blobs shouldn't fill the filesystem anymore :)

src/helpers/BaseChatMesh.h
```diff
@@ -120,6 +120,7 @@ class BaseChatMesh : public mesh::Mesh {
   // storage concepts, for sub-classes to override/implement
   virtual int  getBlobByKey(const uint8_t key[], int key_len, uint8_t dest_buf[]) { return 0; }  // not implemented
   virtual bool putBlobByKey(const uint8_t key[], int key_len, const uint8_t src_buf[], int len) { return false; }
+  virtual bool deleteBlobByKey(const uint8_t key[], int key_len) { return true; }  // returns true even if blob did not exist
```
ripplebiz: I'm not sure this is ever called?

oltaco: Good catch, I've removed that and force pushed 👍

PR #1500, opened 2026-01-27
Allow usage of "/" in radio names

jbrazio:
In the amateur radio (ham radio) world, the forward slash "/" is used as a suffix to callsigns to indicate various operational conditions and locations thus I don't see a reason why "/" should be be considered a invalid char on the radio's name.

IMO it would be far more helpful to block emojis.. but that's outside the scope of this PR.


jbrazio:
Latest v1.12 is creating a lot of issues for us, we have a well defined naming convention https://www.meshcore.pt/en/docs/articles/nomenclature that is wildly adopted in our community which uses "/" to mark the operating mode of the repeater.

PR #1501, opened 2026-01-27
Acoustic feedback when Buzzer is toggled, current buzzer state is reflected in status LED

MisterCodeRalf:
I have added acoustic feedback when the buzzer is toggled by the triple button press. Additionally: If the buzzer is disabled the status LED does a continous double blink instead of the single blink. For me this second feature is also very important since this ensures the operator that the device remains silent. Especially if one has more that one T1000e. (will save hours of sleep).


446564:
I just tested quick on a nano g2 ultra and the sound does not emit when the buzzer turns off. I will try my t1000e tomorrow.

PR #1505, opened 2026-01-28
Fix Wio Tracker L1 Enter Button Mapping

cvaisnor:
# Fix Wio Tracker L1 Enter Button Mapping
### Problem
The enter/select button on the Wio Tracker L1 was not responding to input. The joystick directions (up, down, left, right) worked correctly, but pressing the physical Menu/User button had no effect.

### Cause
`PIN_USER_BTN` was mapped to `PIN_BUTTON6` (Joystick Press), but the physical enter button on the Wio Tracker L1 is actually the Menu/User button `PIN_BUTTON1`.

### Fix
Changed `PIN_USER_BTN` from `PIN_BUTTON6` to `PIN_BUTTON1` in `variant.h`.

### Testing
Tested on physical Wio Tracker L1 hardware. The Menu button now correctly triggers enter/select actions in the UI.

### DIFF
--- a/variants/wio-tracker-l1/variant.h
+++ b/variants/wio-tracker-l1/variant.h
@@ -37,7 +37,7 @@
 #define JOYSTICK_LEFT           PIN_BUTTON4
 #define JOYSTICK_RIGHT          PIN_BUTTON5
 #define JOYSTICK_PRESS          PIN_BUTTON6
-#define PIN_USER_BTN            PIN_BUTTON6
+#define PIN_USER_BTN            PIN_BUTTON1
 
 // Buzzer
 // #define PIN_BUZZER           (12) // Buzzer pin (defined per firmware type)


towerviewcams:
Yes mine works now! awesome!

cvaisnor:
> Yes mine works now! awesome!

Glad it works! I had tried the community version and Ripple version and neither firmware had the button working. I wasn't able to get past the startup screens on the Ripple version because of it. I'm not sure where the Ripple version is compiled or how it's setup, but it will need this change as well. But hopefully this can get merged in and everyone can working with this device. It's a popular board.

oltaco:
I thought this was intentional, that joystick press was considered enter and the other button was considered back?

liamcottle:
> I thought this was intentional, that joystick press was considered enter and the other button was considered back?

Yeah, I probably wouldn't be interested in making this change sorry @cvaisnor.
The joystick button can be pressed to perform the action.

The smaller orange button is used for triple click to mute/unmute buzzer. But it's not meant to be the main enter key.

Moving the joystick left and right, and then pressing the joystick for enter feels more natural than this change, whereas with this PR you now have to press a secondary button to perform the action.

If you are running latest firmware v1.11.0, or dev branch, and you navigate to the "Advert: press enter" screen, if pressing the joystick button doesn't send an advert, you might have damaged hardware?

Pressing the joystick should send an advert from that screen.

cvaisnor:
> and then pressing the joystick for enter feels more natural than this change, whereas with this PR you now have to press a secondary button to perform the action

Hmm, I am nearly certain that the joystick on my board does not have a "center-click" button. 

If it is true that there is one, then I think that totally makes sense as the selection behavior. But I'm almost certain that the board does not have it, or at least my version doesn't. 

I will open up my client node and take a look more thoroughly to confirm. This is the case I use: 
meshnology-n37-black-1pcs png

Will report back soon!

cvaisnor:
Woah! Just found this: https://meshnology.com/products/meshnology-n37-wio-tracker-l1-lora-meshtastic-gps-dev-board-with-1-3-oled#:~:text=Important%20Note%20about,projects%20you%20create

Will give it a shot!

liamcottle:
> Woah! Just found this: https://meshnology.com/products/meshnology-n37-wio-tracker-l1-lora-meshtastic-gps-dev-board-with-1-3-oled#:~:text=Important%20Note%20about,projects%20you%20create
> 
> Will give it a shot!

Cool, thanks for sharing. Looks to be a hardware issue, mostly with the 3d printed case. I'll close this one out as the firmware is behaving as expected for the Wio Tracker L1 :)

Issue #1442, opened 2026-01-21
Request support for the LilyGo T-Beam-1W


teffcz:
Hello there, i would like to request a support for LilyGo T-Beam-1W

Documentation can be found at https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/docs/en/t_beam_1w/t_beam_1w_hw.md

I can test in Europe (Prague) if everything work

recrof:
> I can test in Europe (Prague) if everything work

we need the device to develop on first.

stevenlafl:
@ripplebiz Good to close for issue cleanup, or do you wait until the newest release adds support?

recrof:
closed by [#1509](https://github.com/meshcore-dev/MeshCore/pull/1509)

PR #1509, opened 2026-01-28
Add LilyGO T-Beam 1W Support

stevenlafl:
Brings support for the LilyGO T-Beam 1W model. Apologies if something is not ideal, I am merely trying to help. In trying to follow examples, various variants were laid out differently, making it more difficult to determine the standard.

This one is based off of a combination of documentation from: https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/docs/en/t_beam_1w/t_beam_1w_hw.md and reading through https://github.com/meshtastic/firmware/pull/8967.

I have verified:
1. Transmit
2. Receive
3. GPS
4. Bluetooth

Couple things I would need help with:
1. I don't have the equipment to verify that TX is at the full 1W but the code is there
2. I have not tested ALL firmware, just the BLE companion
3. The battery level reporting - though MeshTastic's implementation doesn't work correctly. I calculated the ADC multiplier from the schematic instead - not with a voltmeter. The difference was 0.0667 from theirs.

Notes:
1. Fan is always-on (like MeshTastic), because I don't see a way to define a callback for a reasonable timer/clock where I could perform a temperature check.
2. GPS flags needed: PERSISTANT_GPS=1 and ENV_SKIP_GPS_DETECT=1 for all environments because (I'm guessing) initialization takes longer than the 1 second sleep. This is consistent with ThinkNode M5 support as well.
3. This model uses a hybrid internal DIO2 + external GPIO21 control to switch between TX/RX modes. I used tiny_relay (the STM32WLx) as an example.

    **Warning/disclaimer from LilyGO**:
    > During TX DATA, if the switch is not switched to the correct position in advance, the PA chip inside the module is likely to be damaged.
    
    Tracing through RadioLib, it appears that this happens automatically when you configure the pins to use, as I have here. I have not experienced any issues running this for 4 days and communicating frequently with the mesh.  _Additional community/MeshCore maintainer verification is desired._

Resolves https://github.com/meshcore-dev/MeshCore/issues/1442


stevenlafl:
It appears the frontend app needs to have these battery values communicated back rather than just the voltage. It may be less bandwidth to report the percentage. 3000-4200 was hardcoded as the default, showing always 100% on the device itself. That's fixed, but the app shows 100% regardless of voltage level.

ripplebiz:
Please change the base branch to 'dev'

stevenlafl:
> Please change the base branch to 'dev'

Done! Couldn't rebase cleanly, cherry-picked again.

PR #1517, opened 2026-01-28
Add LED_STATE_ON build flag for Xiao S3 WIO environments

smithpl2000:


recrof:
this is already fixed in dev: https://github.com/meshcore-dev/MeshCore/pull/1162

smithpl2000:
Good to know :D, Now I have to check why I can't see Xiao with firmware compilet as repeater/room-server or sensor - module is cold and is not showing in meshcore. As companion works well :(

PR #1525, opened 2026-01-29
Xiao nrf52 temperature

derekg612:
Added temperature telemetry for the built-in NRF52840 thermistor.


alexsporn:
This is already included in 1.12.0 btw.

Have a look at `NRF52Board::getMCUTemperature()` which gets added to the response for a `REQ_TYPE_GET_TELEMETRY_DATA` request.

derekg612:
totally missed it. thanks!

Issue #1526, opened 2026-01-29
New version 1.12 does not compile for T-Beam SX1262


Makoland:
Hello,

the new version 1.12 does not compile for the T-Beam SX1262 Companion with following message:
`Linking .pio\build\Tbeam_SX1262_companion_radio_ble\firmware.elf
./.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld.exe: .pio/build/Tbeam_SX1262_companion_radio_ble/firmware.elf section `.dram0.bss' will not fit in region `dram0_0_seg'
./.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld.exe: DRAM segment data does not fit.
./.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld.exe: DRAM segment data does not fit.
./.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld.exe: region `dram0_0_seg' overflowed by 512 bytes
collect2.exe: error: ld returned 1 exit status
`

Can someone give a hint where to look for fixing this?
Where is the size for the BSS segment set, or is this dynamically defined at compile time?

Sorry, my coding skill are a little rusty, especially with platformio.

Thanks,
DG1GAC Markus

oltaco:
Unfortunately it's dynamic, it looks like something will have to be pruned 😢

oltaco:
The commit that pushed it over by 504 bytes is https://github.com/meshcore-dev/MeshCore/commit/5c6c15942bda590f02193509c496b8e3c71f3e94

Probably because of wifi.h being included. I don't have time right now to figure out where/how to make some savings but at least it's a start.

IoTThinks:
@oltaco For ESP32, there is an existing OTA feature with WiFi.h. So another include WiFi.h should not be an issue.

BTW, this is my merged PR for PowerSaving.
https://github.com/meshcore-dev/MeshCore/pull/1266/changes#diff-bf2a85f9de86622aa1706e48181c52038e533cae13e3f52f5cd2a2a063b61dc8

[...]

PR #1532, opened 2026-01-30
Refactor ESP32Board.h to save ~500 bytes of DRAM

oltaco:
Instead of calling esp_wifi_get_mode() we can set a flag to inhibit sleep when calling wifi.begin(), and check it before going to sleep.

This saves ~500 bytes of DRAM and allows Tbeam to compile again, fixes https://github.com/meshcore-dev/MeshCore/issues/1526

I'll mark this as draft until I can confirm it works as intended, but my reading of the docs for esp_wifi_get_mode() suggests it should be fine.

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -194,6 +194,7 @@ void setup() {
   );
 
 #ifdef WIFI_SSID
+  board.setInhibitSleep(true);   // prevent sleep when WiFi is active
   WiFi.begin(WIFI_SSID, WIFI_PWD);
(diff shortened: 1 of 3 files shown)


IoTThinks:
Great.
Let me test.

oltaco:
> Great. Let me test.

Thank you! I haven't had time 😞

oltaco:
This fixes the build for LilyGo_TLora_V2_1_1_6_companion_radio_ble as well. 👍

IoTThinks:
Merged already. Great.
Let me rebase to use the fix in this PR.

Thanks a lot.

PR #1534, opened 2026-01-30
Add weak custom CLI hook to extend CommonCLI without forking core commands

piotrmalek:
## Summary

This PR adds an optional, weakly-linked hook to `CommonCLI::handleCommand()`:

- `extern "C" bool meshcore_custom_cli(uint32_t sender_ts, const char* cmd, char* reply)`
- If the hook returns `true`, the command is considered handled and the standard CLI processing is skipped.
- Default implementation returns `false`, so existing behavior is unchanged.

## Motivation

Board / variant-specific builds sometimes need to add a few extra CLI commands without modifying upstream CLI code paths each time.
A weak symbol hook allows vendors and custom forks to ship additional commands as a separate translation unit, while keeping the upstream CLI intact.

## Backward compatibility

- No existing commands are modified.
- If no custom implementation is provided, the weak default returns `false` and the standard CLI behaves exactly as before.

## Example usage

Create e.g. `customCli.cpp` in your board/variant build and provide your own implementation:

```cpp
#include <cstring>
#include <cstdio>
#include <cstdlib>

extern "C" bool meshcore_custom_cli(uint32_t sender_ts, const char* cmd, char* reply) {
  (void)sender_ts;

  if (strcmp(cmd, "pingx") == 0) {
    std::snprintf(reply, 64, "PONGx!");
    return true;
  }

  return false; // let it handle the standard CLI
}

### CODE REVIEW

File: src/helpers/CommonCLI.cpp [hunk 1]
```diff
@@ -4,6 +4,11 @@
 #include "AdvertDataHelpers.h"
 #include <RTClib.h>
 
+extern "C" bool __attribute__((weak))
+meshcore_custom_cli(uint32_t sender_ts, const char* cmd, char* reply) {
+  return false;
+}
```
weebl2000, line 14: Custom implementations have no way to know the reply buffer size (160–161 bytes depending on caller). Pass it through so they can `snprintf` safely.

```suggestion
extern "C" bool __attribute__((weak))
meshcore_custom_cli(uint32_t sender_ts, const char* cmd, char* reply, size_t reply_size) {
  return false;
}
```

src/helpers/CommonCLI.cpp
```diff
@@ -194,6 +199,7 @@ uint8_t CommonCLI::buildAdvertData(uint8_t node_type, uint8_t* app_data) {
 }
 
 void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, char* reply) {
+    if (meshcore_custom_cli(sender_timestamp, command, reply)) return;
```
weebl2000: And pass the size at the call site (160 matches the smallest caller allocation):

```suggestion
    if (meshcore_custom_cli(sender_timestamp, command, reply, 160)) return;
```

weebl2000 (commented):
Nice minimal extensibility hook. One suggestion: pass the reply buffer size so custom implementations can write safely.

446564:
can you resolve conflicts here and I will review :)

PR #1539, opened 2026-01-30
power saving: enable DCDC, reduce clock speeds and voltage

thomasabbott:
Power saving for the waveshare Pico hat. In my tests this reduces the Rx current into the pico Vsys from 32 mA to about 10 mA, and Tx current from 99 mA to 85 mA. (Measured at Vsys = 5 V. At 3.6 V all the currents are higher)

1) use the lower power PFM mode of the pico SMPS.

2) enable the SX1262 switching DC-DC converter. The board has the necessary inductor fitted. ([rpi_picow circuit](https://files.waveshare.com/upload/d/d8/Pico-LoRa-SX1262_Sch.pdf))

3) detect USB power present and lower clock to 48 MHz.
Otherwise, if on battery:
  system clock to 18 MHz
  move ADC to system PLL
  disable USB clock and PLL
  lower core voltage to 0.95 V

Futher possible savings:
* Physically disabling the red power LED saves another 1.2 mA.
* Adding __asm volatile ("wfi"); at the end of Dispatcher::loop() lets it sleep for a short time, but only saves another 0.5 mA. I've left this out as it requires a processor-specific change to Dispatcher.cpp.

The waveshare_rp2040_lora variant would also benefit from some of these changes, but it lacks USB power detection, so we couldn't go below 48 MHz.  ([waveshare_rp2040_lora circuit](https://files.waveshare.com/wiki/RP2040-LoRa/Rp2040-lora-sch.pdf))


fschrempf:
@thomasabbott Just a quick question: Where does `SX126X_REGULATOR` get evaluated? I don't see it either in the Meshcore code nor in the RadioLib code. What am I missing?

thomasabbott:
I'm sorry, it seems to be completely unused. DCDC is the default in Radiolib and this parameter isn't passed to it.

I thought I saw a modest power reduction when adding this, but I was in the middle of fighting a different (hardware) problem, and the radio wasn't starting successfully, and I didn't isolate this change.

* Radiolib already defaults to using the DCDC converter. in Sx126x.h:
```
/*!  \param useRegulatorLDO Whether to use only LDO regulator (true) or DC-DC regulator (false). Defaults to false. */
int16_t begin(uint8_t cr, uint8_t syncWord, uint16_t preambleLength, float tcxoVoltage, bool useRegulatorLDO = false);
```
* Meshcore doesn't pass the 9th parameter to radiolib. CustomSX1262.h:
```
int status = begin(LORA_FREQ, LORA_BW, LORA_SF, cr, RADIOLIB_SX126X_SYNC_WORD_PRIVATE, LORA_TX_POWER, 16, tcxo);
```

Now measuring just this change by forcing useLDO=true: the board uses 23.6 mA before clocks are adjusted, 5.8 mA afterwards. As the radio starts it jumps up to 9.4 mA (DCDC) or 13 mA (LDO).

So Meshcore's wrapper doesn't support the parameter, and Radiolib defaults to using DCDC. I'll remove it.

JDat:
Looks good as experimental/sample code, but not ready to push into production/dev branch yet. There should be configuration options in .ini file to disable this and run normally without power saving or other detection stuff.

IoTThinks:
@JDat We can share the same CLI "powersaving on/off" among MCUs.

JDat:
> @JDat We can share the same CLI "powersaving on/off" among MCUs.

Exactly! There are common things for all platforms/variants and at the same time totally different hardware details (display/no display; BT/no BT etc). Things. I would like to see that hardware/platform/variant code stay in variants and meshcore specific code stay in src folder.

Power saving command in example repeater or companion, that commands call functions/classes for variants and detailed implementations stay in variants or platforms.
I am already considering to add things like Board.loop() into main loop, refactor "file system/config" stuff so no platform dependent code (for example #ifdef ES32 etc) exist in src and examples folder or at least it it exist in separate helper files. I experienced with this when I wanted to move from SPI1 to SPI0 and from Wire1 to Wite0 hardware on my MCU.

thomasabbott:
The changes are pretty specific to this variant.
* While they'd work on the other RP2040 it lacks the USB detection pin, so it would be risky to drop the clock down below 48 MHz.
* though I think it's possible to detect whether there's an active USB connection, but this would need lots of careful testing.
* My code only checks for USB once, so if you plug in USB after powerup, you have to reboot the pico to see it.
* The savings don't matter when USB is connected, so this is really only useful for a battery powered repeater.
* I don't think there's any performance impact even at 12 MHz, and no real benefit going slower than that.

What do you suggest is the next step here?

1) I could wrap the clock changes in an #ifdef and add a flag to the .ini file. Then at least it's available for experiments in an easy-to-use form.

2) it could become an OTA console command for both RP2040 variants, switching to slow clocks (and back). Needs persistent storage. possible to "brick" your device, unless it has a recovery mode or USB detection or a startup delay or something. This would be nice but it's beyond me to implement.

(2) would be a continuation of (1) - shall I just do (1) to get started?

JDat:
Hello!

I am experimenting with WaveShare Pico2040-Zero https://www.waveshare.com/wiki/RP2040-Zero . Actually with clone from Aliexpress.

No DC/DC control pin here. It uses Linear regulator (KindaLM1117 clone). So, DC/DC control functionality must be configured in platformio.ini file depending on user needs/used board.

Regarding USB, isn't it possible to configure USB clock using PLL in a way that CPU clock stay at 12 MHz and USB work on 48 MHz. I am still thinking about USB cable detection both in Original Pico and WaveShare-Zero context.

Because one of reasons for power saving are solar repeaters, for me it is important to shave every mA while using.

Also ADC voltage measurement can be in .ini config, including, ADC pin, ADC reference voltage and resistors values. Of, if possible, move these configs into web/serial config. But this can't happen soon, because need to change meshcore core sources and user applications/web interface.

OTA is nice idea, but not priority for me right now.


Offtopic:
Got into problems with I2C (wire) stuff and solving them now. Some kinda conflicts between SPI0 and I2C pins...

TheMalkavien:
Hi,

Just to share my own experience with Waveshare rp2040-lora. We use theses boards a lot in my area for our solar nodes, because its cheap and effective. And I'm using pretty much the same powersaving tips as this PR. 

Most of the changes are made by default for my "lowpower" build (macro definition in platform.ini). Except for the USB. When powersaving mode is enabled, instead of sleeping (which does not give any advantage on this borad) it disable the USB completely and put the CPU to 18mhz. It saves about 6ma.
If you want USB back, you can reboot the board, or disable powersaving and use "usb on" command, which gives back USB + 48mhz CPU. But this last part is not really useful on a solar node when put in his definitive location.
When the USB is off, the current is about 10.5ma which is largely sufficient for our solar builds (multiple 18650).

P.S. : We also use a custom ADC pin, and custom ADC multiplier, but only because we reused our old boards, already soldered.
PPS. : Some nodes have OTA. To do so, we plugged a tiny esp32, sleeping most of the time, and waking up when needed, WIFI connexion, push firmware.bin, custom rp2040 bootloader, etc... nothing fancy, you know the drill. Not as easy as an integrated rp2040 OTA, but works well.

JDat:
Make it configurable via platformio.ini file instead of simply adding code. What if I want to disable power saving during debugging or tests other reasons? How to disable your sleep implementation? By roll back code and removing your changes?

Real life situation:
Comparing V11 and V12  repeater firmware I already have problems because scriptkiddies inserting pin wiggling code into their Healtec's and it conflict with my RP2040. For example in V11 no problems with alert pins from INA3221 sensor, but when upload V12 firmware, for unknown reason "warning LED" turns on on INA3221 boars. What kind bull*** code is written by that pull-monkeys?

Hope you got an idea what I am talking about. Do not push code, what need to be edited later to get it working for other users. Instead make it configurable. Remember that there different people using same variant/platform/code for they own needs in different real life cases and sleeping isn't the only case/problem on RP2040.

thomasabbott:
Closing as a) Dev has moved on a lot since I made this  b) no time to refine this to fit perfectly and configurably into the codebase.

Note for future users of the Waveshare Pico hat: it's still quite easy to add this 3x power saving to any new version of Meshcore, pasting the summary in my first post into your prompt. At 60mW the RP2040 is a pretty good platform for a battery or solar powered repeater.

PR #1543, opened 2026-01-31
More Granular Buzzer Controls

nakoeppen:
This PR vastly improves buzzer controls for devices like the RAK WisMesh Tag and the T1000-E. Initially implemented as #1437, this expands on that to allow for the previously used master buzzer disable boolean AND a disable-on-serial boolean, both packed into the preexisting buzzer_quiet integer variable. Currently, I kept it so that three button clicks will only toggle the master buzzer disable, meaning that the disable-on-serial functionality will rely solely on the client app side of things. 

On the technical side of things, Bit 0 is the master disable and Bit 1 is the serial disable. This means that:
- If buzzer_quiet = 0 (00 in binary), the speaker will never be disabled
- If buzzer_quiet = 1 (01 in binary), the speaker will always be disabled
- If buzzer_quiet = 2 (10 in binary), the speaker will only be disabled when connected to serial
- If buzzer_quiet = 3 (11 in binary), the speaker will always be disabled, with the serial disable also enabled

I have a different toggle mechanism which I have implemented but left commented out for now within UITask.cpp in the ui-new folder. This helps with testing since the client app has not implemented this functionality yet (buzzer will briefly beep when buzzer_quiet is set to zero, which helps to determine which setting you are on). If the community prefers the commented out toggle method as a means to swap, I would be happy to build on that. 

@liamcottle @oltaco, I appreciated your opinions on the inital PR I had. Please feel free to let me know if anything needs to be adjusted here to maximize user accessibility. Pretty new to contributing to this project (and frankly open-source projects overall), but I certainly hope to be able to contribute more! Otherwise, this was tested successfully using the aforementioned commented out method on my WisMesh Tag.


nakoeppen:
@liamcottle I am very eager to help add this feature so that devices like the T1000-E or WisMesh Tag are not as reliant on notifications on external devices. Could you please give me any insight on how to adjust this PR so that we can get it merged in for 1.14? Will also fix the merge conflicts at the same time.

446564:
I will take a look and test today

446564 (commented):
I feel like the 4th state is redundant:

- If buzzer_quiet = 0 (00 in binary), the speaker will never be disabled
- If buzzer_quiet = 1 (01 in binary), the speaker will always be disabled
- If buzzer_quiet = 2 (10 in binary), the speaker will only be disabled when connected to serial
- If buzzer_quiet = 3 (11 in binary), the speaker will always be disabled, with the serial disable also enabled

Here `buzzer_quiet = 3` has the same behavior in each scenario as when `buzzer_quiet = 1`. So we only need states 0, 1 and 2.

It would also be nice for the UI icon to reflect the state, maybe a speaker always (which would be new), with the x when disabled and something for the hybrid state.

PR #1545, opened 2026-01-31
Add KISS-like Modem example and protocol documentation

ViezeVingertjes:
Adds a new KISS-like modem firmware that acts as a serial bridge between host applications and MeshCore.

### Changes
- New `examples/kiss_modem/` - KISS-framed serial protocol for raw packet TX/RX and cryptographic operations (signing, encryption, key exchange)
- New `docs/kiss_modem_protocol.md` - Protocol specification
- Updated README.md to document KISS Modem (and Simple Sensor, as it was missing) examples

### Use Case
Enables host applications (Python, Node.js, etc.) to interact with the mesh network directly, offloading crypto operations to the modem's secure identity while handling mesh protocol logic on the host side.

### DIFF
--- a/examples/kiss_modem/KissModem.cpp
+++ b/examples/kiss_modem/KissModem.cpp
@@ -11,10 +11,9 @@ KissModem::KissModem(Stream& serial, mesh::LocalIdentity& identity, mesh::RNG& r
   _pending_tx_len = 0;
   _setRadioCallback = nullptr;
   _setTxPowerCallback = nullptr;
-  _setSyncWordCallback = nullptr;
   _getCurrentRssiCallback = nullptr;
   _getStatsCallback = nullptr;
-  _config = {0, 0, 0, 0, 0, 0x12};
+  _config = {0, 0, 0, 0, 0};
 }
 
 void KissModem::begin() {
@@ -133,17 +132,11 @@ void KissModem::processFrame() {
     case CMD_SET_TX_POWER:
       handleSetTxPower(data, data_len);
       break;
-    case CMD_SET_SYNC_WORD:
-      handleSetSyncWord(data, data_len);
-      break;
     case CMD_GET_RADIO:
       handleGetRadio();
       break;
     case CMD_GET_TX_POWER:
       handleGetTxPower();
-      break;
-    case CMD_GET_SYNC_WORD:
-      handleGetSyncWord();
       break;
     case CMD_GET_VERSION:
       handleGetVersion();
@@ -347,21 +340,6 @@ void KissModem::handleSetTxPower(const uint8_t* data, uint16_t len) {
   writeFrame(RESP_OK, nullptr, 0);
 }
 
-void KissModem::handleSetSyncWord(const uint8_t* data, uint16_t len) {
-  if (len < 1) {
-    writeErrorFrame(ERR_INVALID_LENGTH);
-    return;
-  }
-  if (!_setSyncWordCallback) {
-    writeErrorFrame(ERR_NO_CALLBACK);
-    return;
-  }
-  
-  _config.sync_word = data[0];
-  _setSyncWordCallback(data[0]);
-  writeFrame(RESP_OK, nullptr, 0);
-}
-
 void KissModem::handleGetRadio() {
   uint8_t buf[10];
   memcpy(buf, &_config.freq_hz, 4);
@@ -373,10 +351,6 @@ void KissModem::handleGetRadio() {
 
 void KissModem::handleGetTxPower() {
   writeFrame(RESP_TX_POWER, &_config.tx_power, 1);
-}
-
-void KissModem::handleGetSyncWord() {
-  writeFrame(RESP_SYNC_WORD, &_config.sync_word, 1);
 }
 
 void KissModem::handleGetVersion() {
--- a/examples/kiss_modem/KissModem.h
+++ b/examples/kiss_modem/KissModem.h
@@ -25,10 +25,8 @@
 #define CMD_HASH              0x08
 #define CMD_SET_RADIO         0x09
 #define CMD_SET_TX_POWER      0x0A
-#define CMD_SET_SYNC_WORD     0x0B
 #define CMD_GET_RADIO         0x0C
 #define CMD_GET_TX_POWER      0x0D
-#define CMD_GET_SYNC_WORD     0x0E
 #define CMD_GET_VERSION       0x0F
 #define CMD_GET_CURRENT_RSSI  0x10
 #define CMD_IS_CHANNEL_BUSY   0x11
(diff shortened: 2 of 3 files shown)


ripplebiz:
looks good.

ViezeVingertjes:
> looks good.

Thanks! Currently implementing a game in Rust that uses it. Idea is to only use paths there possible so it doesnt burden the mesh with floods.

Good test to see if this set is sufficient, so far it is. Fingers crossed.✌️

recrof:
this could also be really cool alternative for nodes that have very low resources if @liamcottle implements client portion of MC protocol in the app.

liamcottle:
Hey, looks great!

This could open up the opportunity for new apps that store _all data_ on the client side. For example, a desktop app could store unlimited contacts, and do all of the packet logic itself, and use the kiss modem firmware to transmit and receive raw packets.

I've tested on `Heltec v3` and `RAK4631` boards, from MacOS and Windows with a basic NodeJS script that transmit some raw data by passing `CMD_DATA` frames over KISS. I can also see advert packets coming through when I transmit those from other devices so packet receive is working as expected.

Below are the platformio target environments I added locally for testing.

```
[env:Heltec_v3_kiss_modem]
extends = Heltec_lora32_v3
build_flags =
  ${Heltec_lora32_v3.build_flags}
build_src_filter = ${Heltec_lora32_v3.build_src_filter}
  +<../examples/kiss_modem/>
lib_deps =
  ${Heltec_lora32_v3.lib_deps}
```

```
[env:RAK_4631_kiss_modem]
extends = rak4631
build_flags =
  ${rak4631.build_flags}
build_src_filter = ${rak4631.build_src_filter}
  +<../examples/kiss_modem/>
lib_deps =
  ${rak4631.lib_deps}
 ```

I also added default radio settings at the bottom of `setup()`, so I didn't need to send commands to configure this every time.

```
radio_set_params(917.375, 250, 11, 5);
radio_set_tx_power(22);
```

I wonder if we should add support for our standard build flags:

```
-D LORA_FREQ=869.525
-D LORA_BW=250
-D LORA_SF=11
  ```

If provided at compile time, set the radio to those defaults. It could help if using some other external program that just sends data over `CMD_DATA` and doesn't know how to configure the radio. For example, historic software that's out in the wild...

Other than that thought, looks good to me :)

hpux735:
> this could also be really cool alternative for nodes that have very low resources if @liamcottle implements client portion of MC protocol in the app.

Does this also implement the KISS protocol over some kind of bluetooth protocol?

liamcottle:
> > this could also be really cool alternative for nodes that have very low resources if @liamcottle implements client portion of MC protocol in the app.
> 
> Does this also implement the KISS protocol over some kind of bluetooth protocol?

Nah, this firmware is Serial only. If we added a `CMD_SEND_PACKET` or similar to the companion firmware, it could send raw packet data anyway. This new example firmware is just a minimal KISS implementation.

mepholic:
I just want to point out that this is not a proper KISS implementation. For it to be so, most of these hardware/firmware specific commands should be under the KISS `0xX6` command byte. This implementation redefines multiple KISS standard command codes such as `0xX1` (txdelay), `0xX2` (CSMA persistence parameter), and `0xX3` (slot time), and also overloads the KISS command byte in such a way as to break KISS port addressability. This implementation is likely to break and cause other undefined behaviors when used with many KISS compliant clients.

I'm not intending to disparage the work, but please do not call this KISS if it's not actually KISS compliant. This is not only likely to cause confusion to users who are expecting it to be a proper KISS implementation, but it also could have a negative impact on the standard as a whole if it becomes popular.

RNode makes similar mistakes, and is not a great model to copy.

If you don't intend on it being KISS complaint, I suggest using a different name such as MCMCP (Meshcore Modem Control Protocol)

ViezeVingertjes:
> I just want to point out that this is not a proper KISS implementation. For it to be so, most of these hardware/firmware specific commands should be under the KISS `0xX6` command byte. This implementation redefines multiple KISS standard command codes such as `0xX1` (txdelay), `0xX2` (CSMA persistence parameter), and `0xX3` (slot time), and also overloads the KISS command byte in such a way as to break KISS port addressability. This implementation is likely to break and cause other undefined behaviors when used with many KISS compliant clients.
> 
> I'm not intending to disparage the work, but please do not call this KISS if it's not actually KISS compliant. This is not only likely to cause confusion to users who are expecting it to be a proper KISS implementation, but it also could have a negative impact on the standard as a whole if it becomes popular.
> 
> RNode makes similar mistakes, and is not a great model to copy.
> 
> If you don't intend on it being KISS complaint, I suggest using a different name such as MCMCP (Meshcore Modem Control Protocol)

The interoperability concern is theoretical. If someone connects a KISS client to this modem, the command mismatch is the least of their problems because the data layer is incompatible anyway. KISS clients expect AX.25 frames, this modem speaks MeshCore's encrypted mesh protocol. No one's accidentally piping AX.25 into a LoRa mesh network.

The framing (FEND, FESC, byte-stuffing) follows the Karn/Chepponis spec exactly that you linked to earlier. The commands are MeshCore-specific because things like TXDELAY and Persistence exist for keying up analog radios with PTT relays. LoRa doesn't have that.

That said, it's open source and already in the wild. People so far have found what they expected from the name, but if you feel it's genuinely confusing, feel free to open a pull request with proposed changes.

mepholic:
Regarding your statement:
> KISS clients expect AX.25 frames

That is incorrect, see the K3MC/KA9Q reference spec here: https://web.archive.org/web/20220607102917/http://www.ka9q.net/papers/kiss.html

Which states: 
> The KISS TNC solves these problems by eliminating as much as possible from the TNC software, giving the attached host complete control over and access to the contents of the HDLC frames transmitted and received over the air. This is central to the KISS philosophy: the host software should have control over all TNC functions at the lowest possible level.

> The AX.25 protocol is removed entirely from the TNC, as are all command interpreters and the like. The TNC simply converts between synchronous HDLC, spoken on the full- or half-duplex radio channel, and a special asynchronous, full duplex frame format spoken on the host/TNC link. Every frame received on the HDLC link is passed intact to the host once it has been translated to the asynchronous format; likewise, asynchronous frames from the host are transmitted on the radio channel once they have been converted to HDLC format. 

Regarding your statement:

> things like TXDELAY and Persistence exist for keying up analog radios with PTT relays. LoRa doesn't have that

Also incorrect, see the Meshcore Dispatcher code here: https://github.com/meshcore-dev/MeshCore/blob/main/src/Dispatcher.cpp#L314

> if you feel it's genuinely confusing, feel free to open a pull request with proposed changes

If I can find time in my busy schedule to do so, I will. In the meantime, I believe that this PR was merged prematurely, and with a bit more work, this code could be _much_ more robust, and actually live up to the expectations it sets by calling itself a KISS modem.

ViezeVingertjes:
A proper LoRa KISS mode TNC would basically need to be standalone firmware that has nothing to do with MeshCore anymore and that already exists as other projects.

This is what I see as the sweet spot: KISS frame format (FEND, FESC, byte-stuffing exactly per Karn/Chepponis spec you linked) combined with MeshCore, not 'pure KISS' or '100% by spec' by any means. It's a format that people are familiar with and can easily implement, nothing more than that.

image

PR #1547, opened 2026-01-31
Heltec T114: Remove extra DCDC enable call

oltaco:
This is call is unnecessary, it must have been forgotten when the call was refactored into NRF52BoardDCDC.begin()


fschrempf (approved):
Thanks for the fixup!

PR #1548, opened 2026-01-31
Added Backlight toggle on new quad-click button event

saradonim:
This aligns companion radio firmware with the stock LilyGo T-Echo behavior, which also uses quad-click to toggle the backlight.

**Highlights**
- Added BUTTON_EVENT_QUAD_CLICK to MomentaryButton for detecting 4 consecutive clicks
- Quad-clicking the user button toggles a "forced backlight" mode, keeping the backlight always on
- Added showAlert() feedback to both toggleBuzzer() and toggleBacklight() to notify the user when the requested hardware is not available. This reduces confusion when triple/quad clicks appear to do nothing.

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -730,7 +730,7 @@ void UITask::loop() {
     c = handleLongPress(KEY_RIGHT);
   }
   ev = back_btn.check();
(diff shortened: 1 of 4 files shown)


446564:
x4 press is now toggle GPS

we would need to move to x5 press and also resolve merge conflicts please.

PR #1549, opened 2026-01-31
Revert "Merge pull request #1428 from etienn01/update-t114-i2c"

jbrazio:
This reverts commit 616eb57b163f2123727347ab0425e1ad4fbca564, reversing changes made to 537acd7ea144ee077595c1171cd96770eb924b67.

This patch needs to be reverted because it boot freezes t114 433Mhz variant.

> An alternative approach would be to keep the existing default pins and add support for a second I2C bus, calling AutoDiscoverRTCClock.begin() on both Wire and Wire1. This would allow the firmware to pick-up the onboard RTC on Wire if present, and fall back to discovering an external RTC on Wire1 otherwise. @oltaco


oltaco:
Indeed, it looks like the alternative approach that was mentioned in the discussion for https://github.com/meshcore-dev/MeshCore/pull/1428 will need to be implemented.

oltaco (approved):

PR #1562, opened 2026-02-01
Add Low-Power Sleep for nRF52 Boards

oltaco:
~~**Note: Draft PR, Testing required**  - Please test this PR on your NRF52 repeaters, it seems to work well from my initial testing and in theory shouldn't cause packets to be dropped but **thorough testing is still required**.~~
I'm no longer considering this a draft PR as it's working fine for me. It would still be nice if some others could test it out though.

This PR implements proper low-power sleep for repeater firmware on NRF52 boards, reducing idle power consumption:
Repeater firmware: 6-7mA down to 5mA at 3.6v (after first advert is sent).
Companion firmware: Not yet integrated (see below)

Added NRF52Board::sleep() which uses __WFE() (Wait For Event) to put the CPU into idle whenever possible. CPU will automatically wake on any interrupt so there should be no effect on serial/lora/clock etc. This drops the number of main loops during idle from ~25000+/sec to ~1000/loops per second, because it wakes for every rtc.tick().

Includes the [workaround for errata 87](https://docs.nordicsemi.com/bundle/errata_nRF52840_Rev3/page/ERR/nRF52840/Rev3/latest/anomaly_840_87.html) which probably isn't necessary for repeaters but seems to be required for companions to go into idle.

Note on companions: I haven't added ``board.sleep()`` to the main loop for companions, although I did test it briefly and found it to drop idle current draw from ~8-9ma to ~6ma at 3.6v. More testing would be welcomed for this as well.

### DIFF
--- a/examples/simple_repeater/main.cpp
+++ b/examples/simple_repeater/main.cpp
@@ -127,14 +127,17 @@ void loop() {
 #endif
   rtc_clock.tick();
 
-  if (the_mesh.getNodePrefs()->powersaving_enabled &&                     // To check if power saving is enabled
-      the_mesh.millisHasNowPassed(lastActive + nextSleepinSecs * 1000)) { // To check if it is time to sleep
-    if (!the_mesh.hasPendingWork()) { // No pending work. Safe to sleep
+  if (the_mesh.getNodePrefs()->powersaving_enabled && !the_mesh.hasPendingWork()) {
+    #if defined(NRF52_PLATFORM)
+    board.sleep(1800); // nrf ignores seconds param, sleeps whenever possible
+    #else
+    if (the_mesh.millisHasNowPassed(lastActive + nextSleepinSecs * 1000)) { // To check if it is time to sleep
       board.sleep(1800);             // To sleep. Wake up after 30 minutes or when receiving a LoRa packet
       lastActive = millis();
       nextSleepinSecs = 5;  // Default: To work for 5s and sleep again
     } else {
       nextSleepinSecs += 5; // When there is pending work, to work another 5s
     }
+    #endif
   }
 }
--- a/src/helpers/NRF52Board.cpp
+++ b/src/helpers/NRF52Board.cpp
@@ -251,6 +251,32 @@ void NRF52BoardDCDC::begin() {
   }
 }
 
+void NRF52Board::sleep(uint32_t secs) {
+  // Clear FPU interrupt flags to avoid insomnia
+  // see errata 87 for details https://docs.nordicsemi.com/bundle/errata_nRF52840_Rev3/page/ERR/nRF52840/Rev3/latest/anomaly_840_87.html
+  #if (__FPU_USED == 1)
+  __set_FPSCR(__get_FPSCR() & ~(0x0000009F)); 
+  (void) __get_FPSCR();
+  NVIC_ClearPendingIRQ(FPU_IRQn);
+  #endif
+
+  // On nRF52, we use event-driven sleep instead of timed sleep
+  // The 'secs' parameter is ignored - we wake on any interrupt
+  uint8_t sd_enabled = 0;
+  sd_softdevice_is_enabled(&sd_enabled);
+
+  if (sd_enabled) {
+    // first call processes pending softdevice events, second call sleeps.
+    sd_app_evt_wait();
+    sd_app_evt_wait();
+  } else {
+    // softdevice is disabled, use raw WFE
+    __SEV();
+    __WFE();
+    __WFE();
+  }
+}
+
 // Temperature from NRF52 MCU
 float NRF52Board::getMCUTemperature() {
   NRF_TEMP->TASKS_START = 1; // Start temperature measurement
--- a/src/helpers/NRF52Board.h
+++ b/src/helpers/NRF52Board.h
@@ -51,6 +51,7 @@ public:
   virtual float getMCUTemperature() override;
   virtual void reboot() override { NVIC_SystemReset(); }
   virtual bool startOTAUpdate(const char *id, char reply[]) override;
+  virtual void sleep(uint32_t secs) override;
 
 #ifdef NRF52_POWER_MANAGEMENT
   bool isExternalPowered() override;


IoTThinks:
You may want to add periodically wakeup as in the repeater needs to do periodically adverts.
No idea if we can keep the same architecture flow as ESP32.

So this is to use main loop to execeute these three commands infinitely.
`  __SEV();     __WFE();     __WFE();`

oltaco:
> You may want to add periodically wakeup as in the repeater needs to do periodically adverts. No idea if we can keep the same architecture flow as ESP32.

It's unnecessary, this will wake to do anything required. It wakes every second on tick as well. This is SYSTEMON "sleep", more accurately it is putting the CPU into low power idle state (not executing) when there is nothing to do, and there is no requirement to schedule wake or anything.

You could achieve slightly more current reduction by disabling "unused" peripherals but it's probably not worth the added complexity and possible trade off in stability.

liquidraver:
I'm running with firmware like this for months now on my companion, I just added an sd_app_evt_wait(); with guarded IF-s for nRF to the bottom of the main loop.

@oltaco told me two calls will work better so I experimented with it:

One time:
[2026-02-02 14:38:35.3038 DEBUG: Sleep stats (10s): slept=1258ms awake=8742ms count=1258 (12.6% sleep)
[2026-02-02 14:38:36.0080 DEBUG: Loops/sec: 20322.0 (20322 loops in 1000ms)

Two times:
[2026-02-02 14:40:36.2427 DEBUG: Sleep stats (10s): slept=2125ms awake=7875ms count=2125 (21.2% sleep)
[2026-02-02 14:40:36.9458 DEBUG: Loops/sec: 18354.0 (18354 loops in 1000ms)

Three times:
[2026-02-02 14:43:11.1596 DEBUG: Sleep stats (10s): slept=2697ms awake=7303ms count=2697 (27.0% sleep)
[2026-02-02 14:43:11.8646 DEBUG: Loops/sec: 16835.0 (16835 loops in 1000ms)

third time adds diminishing returns, so we are maybe taking away time from real work there

indeed two times is the sweet spot.

first call returns instantly maybe because the pending timer event so it adds a small amount of sleep only.

oltaco:
> first call returns instantly maybe because the pending timer event so it adds a small amount of sleep only.

I didn't have a chance to investigate but it could be that the reason sd_app_evt_wait() wakes so soon compared to SEV(); WFE(); WFE(); is that we are still polling for BLE RX instead of being event driven.

sd_app_evt_wait() checks the SoftDevice queue first and if there are pending jobs it does them instead of sleeping. So calling it twice lets it process the SoftDevice queue, and then the next call sleeps straight away which gives us more time asleep before the next SoftDevice event gets queued. At least that's my theory.

That's why I wonder if we could change SerialBLEInterface to be fully event driven for RX like it is for TX maybe we could achieve even more sleep.

I have been running the sleep code in this PR on my companions for a couple of days now and haven't noticed any issues. To do it I just added this to the companion main.cpp loop():
```
#if defined(NRF52_PLATFORM)
  board.sleep(1800);
#endif
```

oltaco:
> I'm running with firmware like this for months now on my companion, I just added an sd_app_evt_wait(); with guarded IF-s for nRF to the bottom of the main loop.

After testing this on my home repeater for a few days, I haven't noticed anything out of the ordinary with regards to missing packets. I wouldn't expect to either, given that all this does is causes the CPU to go idle when there's nothing to do.

If anyone else wants to test either repeater or companion with you can try this branch https://github.com/oltaco/MeshCore/tree/nrf52-lowpower

recrof:
do we know how much can be saved by this?

oltaco:
> do we know how much can be saved by this?

Idle power consumption is down by around 20-25% so it's pretty substantial!

recrof:
> > do we know how much can be saved by this?
> 
> Idle power consumption is down by around 20-25% so it's pretty substantial!

yeah, that can defo add-up on repeaters

oltaco:
I just checked my repeater which has been running without issue for 3 days now against a couple of other repeaters on letsmesh.net, and I would say everything appears to be as expected.

image

IoTThinks:
Wow, congratulation for the quick merge.
Now we have Low-Power Sleep for NRF52 boards.

IoTThinks:
This merge is incorrect.
It breaks the powersaving for ESP32.
If it is not time to sleep yet, the code will add another ...5s.

https://github.com/oltaco/MeshCore/blob/a342ab8437e19ea792e51e46cd14c333c1c8b609/examples/simple_repeater/main.cpp#L134

image

Let me fix it in the new PR.

oltaco:
> This merge is incorrect. It breaks the powersaving for ESP32. If it is not time to sleep yet, the code will add another ...5s.

Will ESP32 run code that is gated by ```#if defined(NRF52_PLATFORM)``` ?

IoTThinks:
> > This merge is incorrect. It breaks the powersaving for ESP32. If it is not time to sleep yet, the code will add another ...5s.
> 
> Will ESP32 run code that is gated by `#if defined(NRF52_PLATFORM)` ?

In the right panel, you can imagine if you remove the whole NRF52_PLATFORM, the new code logic for ESP32 is 
1. If time to sleep, sleep
2. Else not time to sleep yet, increase 5 more seconds. Then it will take infinite time to have chance to sleep.

It is different from the old logic for ESP32: Only increase 5 more seconds if there is pending work.

IoTThinks:
No problem.

I will fix to this in my PR.
Let me try to implement immediate sleep for ESP32.

```
  if (the_mesh.getNodePrefs()->powersaving_enabled && !the_mesh.hasPendingWork()) {
#if defined(NRF52_PLATFORM)
    board.sleep(1800); // nrf ignores seconds param, sleeps whenever possible
#else
    if (the_mesh.millisHasNowPassed(FIRSTSLEEP_SECS * 1000)) { // To check if it is time to sleep
      board.sleep(1800); // Sleep. Wake up after 30 minutes or when receiving a LoRa packet
    }
#endif
```

IoTThinks:
@recrof @fschrempf 
This PR has broken PowerSaving for Heltec v3/v4.

It has been merged to MeshCore v1.13.
Instead of going to sleep every 5s. It will take 30s to sleep in low traffic network and unable to sleep in high traffic networks.
Tested and confirmed bug in Heltec v4.

https://github.com/meshcore-dev/MeshCore/pull/1562#issuecomment-3896083942
https://github.com/meshcore-dev/MeshCore/blob/main/examples/simple_repeater/main.cpp#L141

PR #1563, opened 2026-02-01
chore: enable 'get prv.key' for remote

### LINKED ISSUE #1552: 1.12.0 release mentions unsupported remote 'get prv.key' command
The release notes for the v1.12.0 Room Server and Repeater firmware mention this in the Change log:
* remote “get/set prv.key” CLI

When I try a `get prv.key` on the remote CLI of a repeater running v1.12.0, it returns: `??: prv.key`

It looks like PR #1457 has added support for `set prv.key` via remote CLI, but not for `get prv.key` via remote CLI. The code in src/helpers/CommonCLI.cpp still limits that command to serial only:
```
      } else if (sender_timestamp == 0 && memcmp(config, "prv.key", 7) == 0) {  // from serial command line only
```

I suppose the Change log should be changed to:
* remote “set prv.key” CLI

thebino:
Fixes #1552

enable `set prv.key` for remotes

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -290,7 +290,7 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
         sprintf(reply, "> %d", ((uint32_t) _prefs->advert_interval) * 2);
       } else if (memcmp(config, "guest.password", 14) == 0) {
         sprintf(reply, "> %s", _prefs->guest_password);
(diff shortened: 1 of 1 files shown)


ThePlexus:
the proposed code enables "get prv.key" for remote (not set prv.key as the title implies), which I would question the utility of. Private key extraction via a remote client seems counter intuitive to me. What is the use case?

thebino:
@ThePlexus You're right, the change will enable the `get prv.key` for remote.

recrof:
1. you are opening PR against `main` (should be `dev`)
2. you just deleted a comment, and did not change any code.
3. we will not allow `get prv.key` via LoRa because of security.

PR #1564, opened 2026-02-01
On-Demand GPS Activation for Remote Telemetry Requests

thebino:
Allow a node to temporarily enable GPS in response to an authorized remote telemetry request when GPS is otherwise disabled, obtain a fix (or timeout), include coordinates in the telemetry response, and then disable GPS again.

This draft does not handle a timeout nor any additional setting in the companion and might only be used as a PoC or discussion.


liamcottle:
Hey, thanks for the PR! Closing out for now since you mentioned it's incomplete :)

PR #1565, opened 2026-02-01
Some small fixes and improvements extracted from PR1238

fschrempf:
While working on https://github.com/meshcore-dev/MeshCore/pull/1238 ~~I discovered a bug~~ I had a hard time understanding the logic of the state handling in `RadioLibWrappers.cpp`. Therefore this PR improves the readability of that part and a few other things which can be reviewed and merged separately.

* ~~Fix subtle bug in `RadioLibWrappers.cpp` state machine~~
* Improve readability of state machine macros in  in `RadioLibWrappers.cpp`
* Enable reset pin for RAK4631
* Repeater: improve `hasPendingWork()` which is used to decide when the system is idle and can enter powersaving mode
* Repeater: use `constexpr` values for powersaving timings instead of inline literals

### DIFF
--- a/src/helpers/radiolib/RadioLibWrappers.cpp
+++ b/src/helpers/radiolib/RadioLibWrappers.cpp
@@ -4,9 +4,10 @@
 
 #define STATE_IDLE       0
 #define STATE_RX         1
-#define STATE_TX_WAIT    3
-#define STATE_TX_DONE    4
-#define STATE_INT_READY 16
+#define STATE_TX_WAIT    2
+#define STATE_TX_DONE    3
+
+#define FLAG_INT_READY   (1 << 4)
 
 #define NUM_NOISE_FLOOR_SAMPLES  64
 #define SAMPLING_THRESHOLD  14
@@ -21,7 +22,7 @@ static
 #endif
 void setFlag(void) {
   // we sent a packet, set the flag
-  state |= STATE_INT_READY;
+  state |= FLAG_INT_READY;
 }
 
 void RadioLibWrapper::begin() {
@@ -59,7 +60,7 @@ void RadioLibWrapper::doResetAGC() {
 
 void RadioLibWrapper::resetAGC() {
   // make sure we're not mid-receive of packet!
-  if ((state & STATE_INT_READY) != 0 || isReceivingPacket()) return;
+  if ((state & FLAG_INT_READY) != 0 || isReceivingPacket()) return;
 
   doResetAGC();
   state = STATE_IDLE;   // trigger a startReceive()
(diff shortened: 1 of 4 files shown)


src/helpers/radiolib/RadioLibWrappers.cpp
```diff
@@ -4,7 +4,7 @@
 
 #define STATE_IDLE       0
 #define STATE_RX         1
-#define STATE_TX_WAIT    3
+#define STATE_TX_WAIT    2
```
ripplebiz, line 7: I don't follow what this is supposed to fix?

fschrempf: Did you see the commit message?

> `STATE_TX_WAIT` is meant to be a single bit but is defined as two bits,
therefore overlapping with `STATE_RX`. This causes incorrect results
when the code uses the bits to handle the state machine.

Unless I'm getting it all wrong (which totally could be the case) you meant to use the `STATE_*` defines as single-bit masks for the `static volatile uint8_t state` variable. But `3` is not a single bit but two bits. Therefore currently if `state` is set to `STATE_TX_WAIT`, then `state & STATE_RX` will return true, which seems unexpected and looks wrong to me.

I don't now if this has any negative impact with the current implementation, but it certainly caused problems with the code I was working at here: https://github.com/meshcore-dev/MeshCore/pull/1238/changes/a6ce03dae8f7516e989052bc1ca1a7ee7e9951c1.

And anyway, it seems like a good idea to fix this no matter if it is currently causing any issues or not as it has the potential to cause problems in the future.

fschrempf: It looks like there is no actual impact at the moment, as checks for `STATE_RX` are currently all done with `==` to check for `STATE_RX` exclusively.

ripplebiz: Only STATE_INT_READY is intended as a bit mask.  The others are enums.

fschrempf: Ok, got it. I still think it is very easy to get wrong the way it is currently written. Please see the updated PR for a suggestion to make it a bit easier to understand.

fschrempf:
@ripplebiz Is there still any interest in this PR?

liamcottle:
Closing out due to no activity and other unrelated changes intermingled into single PR.

fschrempf:
> Closing out due to no activity and other unrelated changes intermingled into single PR.

@liamcottle I don't mind closing this, but the reasons you give don't really make sense to me. I repeatedly rebased this PR to keep it in sync with `dev` and as far as I can see it would still apply cleanly.

As for the "multiple unrelated changes in one PR": yes these are seemingly unrelated as they were extracted from a larger PR as explained.

Anyway, I get the feeling that there is no real interest in these changes so let's just abandon them.

PR #1277, opened 2025-12-28
Unify nRF & ESP BLE interface - Phase 1

liquidraver:
Core Stuff
- Unified common code into SerialBLECommon.h base class
(it should reside in a helpers/bluetooth folder with all the nrf/esp files, but this PR is not for tidying up)

<strike>- ESP32: Migrated from ESP32 BLE library to NimBLE - Re-pair could be needed! Full erase would be preferable for updating to clear out old BT library stuff from the flash. Factory reset does not clear that.
	- this also resolves xiao c3 bootloops because of too much sensors (-~500k from flash)</strike>
UPDATE: NimBLE migration will need much more testing to be seamless, so only the already used Bluedroid code is unified with nRF interface!

Connection
- Reject second connection attempts (single connection only) (edge case!)
- Improved disconnect handling with reason codes and initiator tracking (for debugging who is to blame for disconnects)

Dynamic connection parameters
- Added sync mode: switches to low-latency params (0 latency, 12-24ms interval) for high throughput
- Sync mode when the fw auto-detects large frame activity (40+ bytes, 3+ frames in 1.5s window)
- Falls back to default mode (3 latency, 12-36ms interval) after 5s inactivity
- Tracks connection parameter updates via BLE events

Queue & Transfer Improvements
- Circular queue implementation replaces linear array shifts - less RAM/CPU
- Added send interval throttling (8ms minimum) - undetectable by user, but helps stabilize the ble stack, on ESP this is mandatory!
- Improved retry logic with throttle window (250ms) - throttle sending when we detect spotty BT connection

ESP32 Specific:
Well... everything that nRF had already are now in ESP's interface :) even "turbo mode"

The two platform's BLE interface is now more or less the same (with library dependent differences of course)

PR #1568, opened 2026-02-01
Ported SerialBLEInterface to NimBLE from Bluedroid to support C3 variants

### LINKED ISSUE #1471: Xiao_C3_companion_radio_ble causes boot loop
Don't even need an SX1262 connected. Just upload directly to the Xiao ESP32C3. If you comment out  -D BLE_PIN_CODE=123456 or use Xiao_C3_companion_radio_wifi it works just fine.

Serial (repeated):

Saved PC:0x403cf94c
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fcd5810,len:0x438
load:0x403cc710,len:0x90c
load:0x403ce710,len:0x2624
entry 0x403cc710
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0x3 (RTC_SW_SYS_RST),boot:0x8 (SPI_FAST_FLASH_BOOT)

QTechIndustries:
The Bluetooth companion firmware for the Xiao C3 crashes trying to create the BLE characteristics. (SerialBLEInterface.cpp:44)
This issue seems to be related to an incompatibility between the C3 and the Bluedroid stack.

This patch creates a variant of the SerialBLEInterface using the NimBLE stack for the C3 to use.

Since MeshCore only uses BLE, it may be worth switching other variants to the NimBLE stack to save on heap/flash usage. 

Fixes #1471

Tested with:
- `ESP32 C3 Super Mini` and `ebyte E22 900M30S`
- `LilyGo T3S3 sx1262`

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -42,8 +42,13 @@ static uint32_t _atoi(const char* sp) {
       #define TCP_PORT 5000
     #endif
   #elif defined(BLE_PIN_CODE)
-    #include <helpers/esp32/SerialBLEInterface.h>
-    SerialBLEInterface serial_interface;
+    #ifdef BLE_USE_NIMBLE
+      #include <helpers/esp32/SerialNimBLEInterface.h>
+      SerialNimBLEInterface serial_interface;
+    #else
+      #include <helpers/esp32/SerialBLEInterface.h>
+      SerialBLEInterface serial_interface;
+    #endif
   #elif defined(SERIAL_RX)
     #include <helpers/ArduinoSerialInterface.h>
     ArduinoSerialInterface serial_interface;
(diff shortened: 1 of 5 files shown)


dt267:
NimBLE is on the way #1277

liquidraver:
> NimBLE is on the way #1277

NimBLE is fully functional on one of my branches, but read the whole PR, the seamless migration remains to be work on.
So there is a fully functional nimBLE ready and waiting :)

derekg612:
> > NimBLE is on the way #1277
> 
> NimBLE is fully functional on one of my branches, but read the whole PR, the seamless migration remains to be work on. So there is a fully functional nimBLE ready and waiting :)

which branch? I tested liquidraver:unifiedble and it has the same boot loop issue as the current main and dev. I tested @QTechIndustries branch c3-nimble-support and it's working as expected.

IoTThinks:
@QTechIndustries MeshCore v1.9 works fine for Xiao C3. I have just tried.
Bluedroid may not be the root cause to make Xiao C3 crash as MC 1.9 works fine.

PR #1569, opened 2026-02-02
Fixed RefCountedDigitalPin.h and SSD1306Display for Heltec v4

IoTThinks:
Hi friends,
This is to fix RefCountedDigitalPin.h to release claim correctly and to ensure no negative "claims" number.

Without this, Heltec v4 comsumes extra 26mA in normal mode and in power saving mode when the power to OLED does not cut off in turnOff correctly.
Tested working fine on Heltec v3 and v4.

It should be fine for other boards too.

### DIFF
--- a/variants/heltec_v4/platformio.ini
+++ b/variants/heltec_v4/platformio.ini
@@ -54,8 +54,6 @@ build_flags =
   -D PIN_BOARD_SDA=17
   -D PIN_BOARD_SCL=18
   -D PIN_OLED_RESET=21
-  -D ENV_PIN_SDA=4
-  -D ENV_PIN_SCL=3
 build_src_filter= ${Heltec_lora32_v4.build_src_filter}
 lib_deps = ${Heltec_lora32_v4.lib_deps}
 


Quency-D:
Hi @IoTThinks ,Thank you for your hard work, but we believe that https://github.com/meshcore-dev/MeshCore/pull/1570 is the correct reason.

IoTThinks:
Ok, let me try yours.

If your PR works then I will close this PR as you have clearer picture than I.

JDat:
Ho this is working? Will this code stay on on Heltec v3 +v4 platforms or this code also will impact my PicoW variant? I don't have OLED display and possibly use this GPIO pin for things...

IoTThinks:
> Ho this is working? Will this code stay on on Heltec v3 +v4 platforms or this code also will impact my PicoW variant? I don't have OLED display and possibly use this GPIO pin for things...

@JDat If you don't have OLED, this issue will not happen to you.
RefCountedDigitalPin is intended to control the **shared** vEXT (external power to OLED and sensors).
It should ensure that vExt will be on if there is at least one OLED/sensor is using vext.
And vExt will be off if no OLED/sensor uses vext.

However, the OLED and sensors need to use claim() and release().
So far, only OLEDs are using it.

So no impact for you.
I would like to fix this so the OLED can off correctly and sensors in future can on/off correctly.

IoTThinks:
I will keep this PR for review.
I believe this fix is good for not only OLEDs but also for sensors to activate/deactivate vExt correctly.

The release() function should work correctly on its own.
Thanks a lot.

Socalix:
LGTM. Nice and clean and a generic fix for other use cases as well.

src/helpers/RefCountedDigitalPin.h
```diff
@@ -20,10 +20,12 @@ class RefCountedDigitalPin {
       digitalWrite(_pin, _active);
     }
   }
+
   void release() {
-    _claims--;
     if (_claims == 0) {
       digitalWrite(_pin, !_active);
+    } else {
+      _claims--;
```
oltaco: Won't this mean that if claims == 1 and you call release(), claims is will end up as 0 but you the pin won't be deactivated until you call release an extra time?

IoTThinks: Haha, true.
Let me adjust.

IoTThinks:
@oltaco : When I fixed it properly, it does not work anymore.
So I explore more and found something more.

This is in powersaving off. So there is **no sleep** yet.
Expected to have **5x mA** after the screen is off. It should not stay at 8x mA.

Don't off the vExt at all and ...save power
```
 void release() {
    // Do nothing and it uses normal consumption at 5x mA
  }
```

So we aggressive "off" the vExt, the power is increased at 8x mA
```
  void release() {
    // This one will have 5x mA at first, then 8x mA
    digitalWrite(_pin, HIGH); // LOW is on, HIGH is off according to the latest platformio.ini
  }
```

So I can guess LOW = vExt off. HIGH = vExt ON, this is correct based on my previously found docs and same to other Heltec boards.
Now Heltec updated from HIGH to LOW (vExt on = LOW)
https://resource.heltec.cn/download/WiFi_LoRa_32_V4/datasheet/WiFi_LoRa_32_V4.2.0.pdf
image


```
  void release() {
    // This one will have 5x mA at first, then 5x mA
    digitalWrite(_pin, LOW);
  }
```

To test further, I guess OLED SSD1306 has nothing related to the vExt at all.
Well, I removed all digitalWrite(_pin,...) and all stuffs inside claim() and release(),
The screen is still on to see "Loading..." and "Heltec v4 repeater..."
And the power is at 5x mA as expected.

```
class RefCountedDigitalPin {
  uint8_t _pin;
  int8_t _claims = 0;
  uint8_t _active = 0;
public:
  RefCountedDigitalPin(uint8_t pin,uint8_t active=HIGH): _pin(pin), _active(active) { }

  void begin() {
    pinMode(_pin, OUTPUT);
    // digitalWrite(_pin, !_active);  // initial state
  }

  void claim() {
  }

  void release() {
  }
```

I can conclude SSD1306Display is not powered by vExt at all.
And vExt=LOW means Off instead.

That also explains why MeshCore 1.11.0 worked fine.
As MC 1.11.0 does not need to turn on vExt to use SSD1306Display.
SSD1306Display does not need vExt.

In MC 1.12.0, we turn on vExt unneccessarily.
And incorrectly turn on (instead of off) vExt at release() actually increases the power.

Please correct me if I'm incorrect somewhere.

src/helpers/RefCountedDigitalPin.h
IoTThinks: @oltaco We can change to this.
This is fix for future vExt usage.
However, it won't fix the OLED. We may need to fix or revert the SSD1306Display as it does not use vExt at all.

See my comment below. https://github.com/meshcore-dev/MeshCore/pull/1569#issuecomment-3852268333

```
  void release() {
    if (_claims == 0) return; // avoid negative _claims

    _claims--;
    if (_claims == 0) {
      digitalWrite(_pin, !_active);
    }
  }
```

IoTThinks:
So the fix this is one for future usage.
```
void release() {
    if (_claims == 0) return; // avoid negative _claims

    _claims--;
    if (_claims == 0) {
      digitalWrite(_pin, !_active);
    }
  }
```

Then revert the SSD1306Display.h as it does not use vExt at all.
Revert  -D PIN_VEXT_EN_ACTIVE=LOW to HIGH again.
And update the Heltec docs to the original vExt on = HIGH.

IoTThinks:
It works perfect now.
1. The fix for negative claims is for future usage. It should stop at 0.

2. I keep the vExt for SSD1306Display just in case other boards need vExt for SSD1306Display.
For Heltec v4, I passed a NULL periph_power to disable all claim() and release().
Heltec v4 does not use vExt to power on/off SSD1306Display.

Tested on Heltec v4 with screen.
- Able to show "Please wait..."
- Able to show "Heltec v4 repeater...."
- Able to use 5x mA in normal mode
- Able to auto off the screen after a short while.
- Able to achieve 13mA in PowerSaving on.

IoTThinks:
Reopen.
Heltec v4's current platformini.io breaks I2C sensors at pin 17, 18 and may cut off the sensors when the OLED is off.

weebl2000:
> Hi @IoTThinks ,Thank you for your hard work, but we believe that #1570 is the correct reason.

No, #1570 introduced the problem.

  1. Added digitalWrite(PIN_OLED_RESET, LOW) — masks the power issue by forcing hardware reset, but corrupts Wire (I2C bus lockup on GPIO17/18)                                               
  2. Added PIN_OLED_RESET=21 — enables the reset LOW code path                                                                                                                                
  3. Kept display(&(board.periph_power)) — unnecessarily toggles VEXT for an OLED that doesn't use it                                                                                         
  4. Kept PIN_VEXT_EN_ACTIVE=LOW — wrong polarity, so release() actually turns VEXT ON                                                                                                        

PR #1570 treated the symptom (power consumption) without understanding the root cause (OLED isn't on VEXT, polarity is wrong). It "fixed" power by holding the SSD1306 in reset, which happened to reduce current but broke I2C as a side effect. Then ENV_PIN_SDA=4/ENV_PIN_SCL=3 was added to work around the I2C breakage by moving sensors to a separate bus.

It's entirely possible to keep the other I2C bus operational without causing extra power usage.

The OLED is powered by VDD_3V3 (Always On). Toggling VEXT to "sleep" the display is functionally useless for the screen but actively harmful to external sensors sharing the I2C bus (which are powered by VEXT). Cutting their power while the ESP32 tries to talk to the always-on OLED is exactly what causes bus corruption.

weebl2000:
@IoTThinks see also https://github.com/HelTecAutomation/Heltec_ESP32/pull/225

IoTThinks:
It makes sense to have vext on by setting the gpio HIGH.

So when in deepsleep, we can keep the gpio LOW to cut the power of vext.
This is the correct hardware design, too.

It will consume more power in deepsleep if we need to pull the GPIO HIGH to cut off the vext.

I believe this is the same for Heltec v2,3,4.

IoTThinks:
I will rebase this PR and remove the below lines
  -D ENV_PIN_SDA=4
  -D ENV_PIN_SCL=3

IoTThinks:
Reverted to use GPIO 17, 18 as I2C for Heltec v4 repeater: https://github.com/meshcore-dev/MeshCore/pull/1569/commits/13d0dff9182bf7931b9b6c87b67d7e5120301099
Rebased and tested ok with `powersaving on`

BTW, this dev still has powersaving bug: only back to sleep after 30s.
But it is fixed in another [PR 1687 ](https://github.com/meshcore-dev/MeshCore/pull/1687/changes#diff-10a0092562d43e5bb73a25560e519d6106a0d09b143204ed36ae90d8a3ad24aa).

IoTThinks:
@oltaco Would you help to test/review this PR?
For the OLED RESET, I don't think it is neccessary to reset the OLED before turning off.
However, since we bypass the code with display(NULL), so we may not mind it.

The SDA and SCL for TFT seems weird to me. But I don't mind as it will not affect most of us.
image
https://github.com/IoTThinks/MeshCore/blob/13d0dff9182bf7931b9b6c87b67d7e5120301099/variants/heltec_v4/platformio.ini#L75

Thanks a lot.

weebl2000:
I think this is good to merge.

PR #1571, opened 2026-02-02
chore: Switch to EnvironmentSensorManager for re-usability

thebino:
Instead of duplicating custom logic for every hardware, it would be nice to re-use existing code where possible.
The already existing EnvironmentSensorManager is used by many variants and does a good job.
This PR added the hardware specifics where needed.

Baseline for #1564

### DIFF
--- a/variants/t1000-e/platformio.ini
+++ b/variants/t1000-e/platformio.ini
@@ -29,13 +29,14 @@ build_flags = ${nrf52_base.build_flags}
   -D LR11X0_DIO_AS_RF_SWITCH=true
   -D LR11X0_DIO3_TCXO_VOLTAGE=1.6
   -D ENV_INCLUDE_GPS=1
+  -D ENV_INCLUDE_BME680=0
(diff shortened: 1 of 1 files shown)


JDat:
Why different hardware need to be in environment manager? Make environment manager platform independent and fine tune your platform in a way you need. Common code must be platform independent as much as possible.

thebino:
Yes, thats what I did. Instead of 120 loc its now just 45 loc in the environment manager and no duplication anymore.

JDat:
Sorry if I overlooked, but... Who added T1000_GPS in EnvironmentSensors.cpp? Shouldn't it go to T1000 variant instead of Environment Sensors?

fschrempf:
I haven't looked at the details but I agree with @JDat. The code in `EnvironmentSensors.cpp` should be platform-/board-agnostic. Better add generic hooks in the board base class and call them from `EnvironmentSensors.cpp` if necessary.

PR #1580, opened 2026-02-03
Buzzer notification control per channel

lowmesh-dev:
Added the ability to silence or activate the buzzer notification per channel. This helps quiet down the buzzer notifications on the public channel for example but still have other channels provide buzzer feedback. The setting is controlled via a new SET_OTHER_PARAMS command on byte 6.

Tested on RAK4631 and works

### DIFF
--- a/docs/protocol_guide.md
+++ b/docs/protocol_guide.md
@@ -376,6 +376,44 @@ Byte 0: 0x14
 
 ---
 
+### 8. Set Other Params (device preferences, buzzer, notifications)
+
+**Purpose**: Set device preferences including manual-add-contacts, telemetry modes, advert policy, multi-acks, **buzzer on/off**, and **per-channel buzzer mask** (one bit per channel 0–7).
+
+**Command**: `SET_OTHER_PARAMS` = **0x26** (decimal 38)
+
+**Command Format**:
+```
+Byte 0: 0x26 (command)
+Byte 1: manual_add_contacts (0/1)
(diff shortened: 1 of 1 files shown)


liamcottle:
Hello, thanks for the contribution.

Unfortunately your PR is in a mangled state. Conflicts etc.
Also, we currently allow up to 40 channels, not just 7.

In the future, we might also increase this limit, so it needs to be able to support that.
It would probably be better suited to store this sort of flag on the channel struct instead.

From memory, there was already a historic PR that did something related to a "no store" flag for channel messages.
A similar approach is probably the way to go.

I'll close this one out for now :)

lowmesh-dev:
Thanks! The PR mangle I think was me pulling from main instead of dev and then trying to push to dev? I will look into that. 

Thanks for the clarifications on the channel count. I saw 8 channels on smaller builds but 40 is the full channel list, I missed that. I will revise and revisit it. If its a "no store" doesn't that mean it wont survive a reboot? Maybe at the very least we could toggle just the public vs private channels, that way we reduce the annoyance of the public channel, but at that point, maybe a system that could work for every channel could be good better

swannman:
@lowmesh-dev and @liamcottle - the "no store" flag is this PR, which I closed due to inactivity: https://github.com/meshcore-dev/MeshCore/pull/1052

Happy to resurrect it if we want to move forward with this design? I just didn't want to clutter the list of open PRs.

PR #1582, opened 2026-02-03
Update to support battery voltage on D3 pin for Xiao ESP32S3 repeater

pki791:
Added D3 pin as battery voltage input. This voltage divider is not present on a stock Xiao ESP32S3 board, but can be easy added, the D3 pin is not used and is near the battery pads.


pki791:
Is this a duplicate https://github.com/meshcore-dev/MeshCore/pull/1643 ?

PR #1589, opened 2026-02-03
Added gps support for Sensecap p1 solar

marnick-hartgers:
Added GPS to variant Sensecap p1 solar
Removed gps pin enable code, it is managed by MicroNMEALocationProvider


liamcottle:
Thanks, @recrof tested GPS module on Sensecap P1 solar and said it's working with this PR. Merged :)

PR #1591, opened 2026-02-04
fix(kiss): periodic noise floor calibration and AGC reset

agessaman:
In the previous implementation, the noise floor was calibrated once and the value was returned for the entire time the radio was active. This makes a change to:

- Trigger noise floor calibration every 2s and AGC reset every 30s in main loop.
- Reorder loop to match Dispatcher: calibrate + radio.loop() before AGC reset and recvRaw() so RSSI is never sampled right after startReceive().
- Update protocol doc with calibration intervals and typical noise floor range.
- Add limited variant platformio.ini configs for KISS modem (heltec_v3, rak4631).

### DIFF
--- a/examples/kiss_modem/main.cpp
+++ b/examples/kiss_modem/main.cpp
@@ -12,8 +12,8 @@
   #include <SPIFFS.h>
 #endif
 
-#define NOISE_FLOOR_CALIB_INTERVAL_MS  2000   // match Dispatcher default
-#define AGC_RESET_INTERVAL_MS         30000  // periodic RX restart so AGC doesn't drift (repeater uses same via prefs)
+#define NOISE_FLOOR_CALIB_INTERVAL_MS  2000
+#define AGC_RESET_INTERVAL_MS         30000
 
 StdRNG rng;
 mesh::LocalIdentity identity;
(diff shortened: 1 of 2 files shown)


agessaman:
Just a quick follow-up, this appears to be working well. I monitored the noise floor using pyMC_Repeater and the graph shows regular updates but a relatively consistent noise floor. The initial very flat lines were the result of the firmware with no noise floor recalibration logic.

image

examples/kiss_modem/main.cpp
```diff
@@ -94,7 +99,16 @@ void loop() {
...
+
+  // Match Dispatcher order: noise floor calib + loop() first, so we never sample RSSI in the same
+  // iteration as startReceive() (AGC reset -> recvRaw below). Sampling right after startReceive()
+  // can yield settling/cold RSSI and drive the floor toward -120.
+  if ((uint32_t)(millis() - next_noise_floor_calib_ms) >= NOISE_FLOOR_CALIB_INTERVAL_MS) {
+    radio_driver.triggerNoiseFloorCalibrate(0);   // 0 = no interference threshold (KISS has no prefs)
+    next_noise_floor_calib_ms = millis();
+  }
+  radio_driver.loop();
+
```
ripplebiz, line 109: Are there plans to add some Listen-Before-Talk logic here?

agessaman: That seems like a design preference. @ViezeVingertjes did the original work, I just added the necessary functions to get the KISS modem working properly with pyMC_core.

He has implemented `CMD_IS_CHANNEL_BUSY` (0x11) which returns `RESP_CHANNEL_BUSY` (0x31) with one byte: 0x00 = clear, 0x01 = busy. The KISS modem is kind of a dumb pipe and the user software can manage its own LBT like the repeater firmware calling _radio.isReceiving().

ViezeVingertjes:
Nice find. 👍🏼

ripplebiz:
> Nice find. 👍🏼



> Nice find. 👍🏼

If you're OK with this PR, I'll merge it

ViezeVingertjes:
> > Nice find. 👍🏼
> 
> If you're OK with this PR, I'll merge it

LGTM; only remark would be the comments hehe... claude loves to comment.
But technically that doesnt make much of a difference.

agessaman:
Yeah, sorry. I did a "hey, did I miss anything" pass and he was like, "here are more comments!"

I can reduce/eliminate them if you prefer.

ViezeVingertjes:
> Yeah, sorry. I did a "hey, did I miss anything" pass and he was like, "here are more comments!"
> 
> I can reduce/eliminate them if you prefer.

Only if you have time for it! otherwise, just let them be, for another time hehe.

agessaman:
I cleaned up or removed most of the comments my ever-helpful robot friend decided to add. Sorry about that. 😖

PR #1537, opened 2026-01-30
Enable I2C sensors and EnvironmentSensorManager for Heltec T114

pinkolin:
This is simple patch for enabling sensors on Heltec T114 (defacto copy of behavior from Heltec v3, T114 have support only for GPS and nothing more).

While standard I2C is occupied, i'm using PINs used by MT (i know, i know).
-D PIN_BOARD_SDA=16
-D PIN_BOARD_SCL=13

PR #1596, opened 2026-02-05
Refactor Heltec T114 sensor management

jbrazio:
This pull request builds upon the work initiated in #1537 by @pinkolin, taking it a step further. The refactoring implements native usage of `EnvironmentSensorManager` for both GPS and the I2C bus (Wire1).

We're deviating from the MT 16/13 pin definition for the second I2C bus due to compatibility issues encountered with T114 v1 boards in version 1.12. Instead, we enable `Wire1` in `variants.h` and define the bus pins as 8/7. This pin combination not only resolves the hardware compatibility problems but also the physical layout supports the standard `VCC GND SCL SDA` configuration commonly used for I2C connections.

As a side quest [bonus points], this PR fixed an issue where the T114 couldn't output early-boot debug messages to serial, adding a small boot delay when DEBUG is enabled."


kallanreed:
Where did these pin values come from? I've been working on a T114 build and pins defined here don't make sense generally. For example, Pin 21 isn't GPS enable, it's Vext Control. Pin 34 is actually the GPS "allow sleep" pin.
The Wire pins de are tied to a specific build too. Why were these added as general-purpose pin defs?
Likewise, why did you add all the sensor stuff for devices that don't have sensors?

heyitscosmin:
I have a DS3231 attached to my Heltec T114. When the I2C pins were set to 16&13, the RTC was working fine. After these changes, the RTC does not work anymore, but the strange thing is the DS3231 has a temperature sensor which is working fine.

jbrazio:
> I have a DS3231 attached to my Heltec T114. When the I2C pins were set to 16&13, the RTC was working fine. After these changes, the RTC does not work anymore, but the strange thing is the DS3231 has a temperature sensor which is working fine.

Please state what version of the board are you using.

> Where did these pin values come from? I've been working on a T114 build and pins defined here don't make sense generally. For example, Pin 21 isn't GPS enable, it's Vext Control. Pin 34 is actually the GPS "allow sleep" pin. The Wire pins de are tied to a specific build too. Why were these added as general-purpose pin defs? 

The GPS pinout was already part of the original target definition, I'm waiting for a GPS for the t114 the sort out the right pinout.

> Likewise, why did you add all the sensor stuff for devices that don't have sensors?

I would argue that having sensor enabled by default might not be the best option. But this is the current decision of upstream.

heyitscosmin:
> > I have a DS3231 attached to my Heltec T114. When the I2C pins were set to 16&13, the RTC was working fine. After these changes, the RTC does not work anymore, but the strange thing is the DS3231 has a temperature sensor which is working fine.
> 
> Please state what version of the board are you using.
> 

I am using Heltec T114 v2 without screen.

Issue #1556, opened 2026-01-31
Add LilyGo T-Beam 1Watt Support


mintylinux:
I have working firmware that me and another person has created and is at the current version of 1.12.0. It is here on my github page. https://github.com/mintylinux/Meshcore-T-beam-1W-Firmware 

Would you like me to submit a PR?

Thank you!

gjelsoe:
Seems like somebody also made one...
https://github.com/meshcore-dev/MeshCore/pull/1509

mintylinux:
> Seems like somebody also made one... [#1509](https://github.com/meshcore-dev/MeshCore/pull/1509)

Ah ok. Well my firmware is 100% working with fan control, GPS, Bluetooth, Battery Monitoring, etc etc. I am using it on meshcore right now.

mintylinux:
The webflasher needs to be updated to flash the T-Beam 1W binaries at offset 0x0 instead of 0x10000. The image flashes fine on with esptool, but when you go to put the binary in the webflasher, it flashes to the device, but then it will not boot afterward.

stevenlafl:
> Ah ok. Well my firmware is 100% working with fan control, GPS, Bluetooth, Battery Monitoring, etc etc. I am using it on meshcore right now.

I felt fan control needed more discussion. Didn't see a place it was done already, so wanted to talk about implementing support in the base class.

If you want to fill any gaps go for it, I'd love to see it.

mintylinux:
> > Ah ok. Well my firmware is 100% working with fan control, GPS, Bluetooth, Battery Monitoring, etc etc. I am using it on meshcore right now.
> 
> I felt fan control needed more discussion. Didn't see a place it was done already, so wanted to talk about implementing support in the base class.
> 
> If you want to fill any gaps go for it, I'd love to see it.

Basically how i have fan control setup is after a TX, it runs for 5 seconds, and then waits for the next TX.

stevenlafl:
> > > Ah ok. Well my firmware is 100% working with fan control, GPS, Bluetooth, Battery Monitoring, etc etc. I am using it on meshcore right now.
> > 
> > I felt fan control needed more discussion. Didn't see a place it was done already, so wanted to talk about implementing support in the base class.
> > 
> > If you want to fill any gaps go for it, I'd love to see it.
> 
> Basically how i have fan control setup is after a TX, it runs for 5 seconds, and then waits for the next TX.

[...]

PR #1597, opened 2026-02-05
Fix T-Beam 1W fan running continuously

ViezeVingertjes:
Only run fan during TX and for a short cooldown period after, instead of keeping it on all the time.


stevenlafl:
Great fallback if a temperature diode is not present. Discussion about temperature reading and potentially adding dedicated support for fans/temp control here https://github.com/meshcore-dev/MeshCore/issues/1556

PR #1398, opened 2026-01-14
Improve Heltec v4 RX reception with undocumented register patch

Socalix:
Following several rounds of testing Heltec v4 RX reception as part of PR #1249 I found out that the best solution was using the original v1.11.0 code, with undocumented register 0x8B5 patch (setting LSB=1) as desribed by @Quency-D (Heltec engineer) here: https://github.com/meshcore-dev/MeshCore/pull/1249#issuecomment-3717071840 ~and without the RX Boosted Gain feature~.

This PR is doing just that: turning on the 0x8B5 LSB ~and turning off RX Boosted Gain flag.~

UPDATE: After several rounds of testing here we determined that Boosted Gain flag should be turned on/off on a case-by-case basis, so this PR is now updated to only set the register value that has proven to consistently improve RX.

=== Additional Details ===

Test Process:
I put RAK4631 and Heltec v4, both with Alfa 915 antenna and repeater v1.11.0 firmware in the same location (attic, about 7 inches apart). I then took Heltec T114 with companion v1.11.0 about 1.25km (0.75mile) away and did ping every 3 seconds to the same repeater about 20-30 times. I then flashed the v4 with different firmwares and pinged again.

Test Sample:
| RAK4631 v1.11.0 | Heltec v4 PR1249 w/o RegPatch | Heltec v4 PR1249 w/RegPatch | Heltec v4 PR1398 w/RegPatch and no Boosted Gain |
| --- | --- | --- | --- |
| -4.25 | err | 4.75 | 5.5 |
| err | err | err | 6.25 |
| -2.75 | err | 2.75 | err |
| 3.75 | 3.5 | err | 4.25 |
| 1.75 | 5.5 | 2.0 | 5.5 |
| 3.0 | err | 6.0 | 4.75 |
| err | err | err | 1.25 |
| 1.75 | 4.75 | err | 7.75 |
| 4.0 | 3.25 | -0.25 | err |
| 4.75 | 3.0 | 1.75 | 2.5 |
| -7.5 | 4.25 | 5.0 | 2.0 |
| 6.75 | 4.25 | 2.25 | err |
| err | err | err | 5.5 |
| 6.25 | err | 4.0 | 0.25 |
| 3.75 | err | -0.75 | err |
| 20% err | 55% err | 33% err | 25% err |

PR #1249, opened 2025-12-19
Improve GC1109 FEM pin handling for Heltec v4 and Tracker v2

weebl2000:
Improves the GC1109 FEM pin management for Heltec v4 and Heltec Tracker v2. These are correctness and robustness improvements to how GPIO46 (CPS strapping pin) and deep sleep pins are handled - not a fix for the core TX/RX switching which was already working correctly via DIO2.

> **Note:** The register patch (`SX126X_REGISTER_PATCH=1`) that was the primary RX sensitivity improvement has already been merged via PR #1398.

### GPIO46 (CPS) strapping pin handling

GPIO46 is an ESP32-S3 strapping pin. Previously it was configured as OUTPUT at boot and left that way permanently. Now:
- **Boot**: GPIO46 is not configured - TX handlers are fully responsible for it
- **onBeforeTransmit**: Sets `pinMode(OUTPUT)` then drives HIGH (full PA mode)
- **onAfterTransmit**: Drives LOW then releases with `pinMode(INPUT)`

This avoids holding a strapping pin as OUTPUT when it's not needed (CPS is "don't care" during RX per the GC1109 datasheet).

### Deep sleep pin management

- Now holds `PA_POWER` (VFEM LDO) during sleep in addition to `PA_EN` (CSD), so the GC1109 remains powered for RX wake capability
- Properly releases RTC holds for `PA_POWER` and `PA_EN` on deep sleep wake
- Separates cold boot vs deep sleep wake paths for clarity

### Other

- 1ms delay after GC1109 power-up on cold boot to allow FEM to stabilise
- Moved `PIN_BOARD_SDA`/`PIN_BOARD_SCL` to the base `[Heltec_lora32_v4]` section
- Updated pin comments to reference GC1109 datasheet nomenclature (CSD, CPS, CTX, VFEM)
- Applied same changes to Heltec Tracker v2 (identical GC1109 FEM, CSD on GPIO4)

## GC1109 Control Logic Reference

| Mode            | CSD (GPIO2/4) | CTX (DIO2) | CPS (GPIO46) |
|-----------------|---------------|------------|--------------|
| Shutdown        | 0             | X          | X            |
| Receive LNA     | 1             | 0          | X (don't care) |
| Transmit bypass | 1             | 1          | 0            |
| Transmit PA     | 1             | 1          | 1            |

TX/RX path switching is handled automatically by SX1262 DIO2 → GC1109 CTX (`SX126X_DIO2_AS_RF_SWITCH=true`). This was already the case before this PR.

PR #1600, opened 2026-02-06
Hold GC1109 PA_POWER during deep sleep for LNA RX wake

weebl2000:
## Summary

- Hold the GC1109 VFEM_Ctrl (PA_POWER) pin HIGH during deep sleep so the FEM's LNA remains active for wake-on-packet RX
- Release the RTC hold on wake so normal GPIO operation resumes
- Add 1ms delay after GC1109 power-on during cold boot to allow the FEM to stabilize before use (skipped on deep sleep wake since the FEM was already powered)
- Applies to both Heltec V4 and Heltec Tracker V2

Without this, the GC1109 LNA could be unpowered during sleep, degrading RX wake sensitivity by ~17dB.

**Trade-off:** ~1.2mA additional deep sleep current (measured: 11.1mA → 12.3mA) for significantly improved wake-on-packet range.

## Test plan

- [x] Flash Heltec V4, verify wake-on-packet works at expected range
- [x] Verify TX still works correctly after wake
- [x] Measure deep sleep current (~1.2mA additional with LNA active, confirmed by @towerviewcams)

Supersedes #1249

Update 24 February 2026: someone pointed to me that[ Heltec does the same in their ESP32 library](https://github.com/HelTecAutomation/Heltec_ESP32/blame/7663bf84c4c54afddb181d64706c7ced77aea0a5/src/radio/radio.c#L889-L901C4), so I'm certain this is the right way of holding the GC1109 awake during sleep.

---
**Build firmware:** [Build from this branch](https://mcimages.weebl.me/?commitId=heltec_deep_sleep_lna)

### DIFF
--- a/variants/heltec_tracker_v2/HeltecTrackerV2Board.cpp
+++ b/variants/heltec_tracker_v2/HeltecTrackerV2Board.cpp
@@ -20,9 +20,12 @@ void HeltecTrackerV2Board::begin() {
     pinMode(P_LORA_PA_TX_EN, OUTPUT);
     digitalWrite(P_LORA_PA_TX_EN,LOW);
 
-    periph_power.begin();
-
     esp_reset_reason_t reason = esp_reset_reason();
+    if (reason != ESP_RST_DEEPSLEEP) {
+      delay(1);  // GC1109 startup time after cold power-on
+    }
+
+    periph_power.begin();
     if (reason == ESP_RST_DEEPSLEEP) {
       long wakeup_source = esp_sleep_get_ext1_wakeup_status();
       if (wakeup_source & (1 << P_LORA_DIO_1)) {  // received a LoRa packet (while in deep sleep)
--- a/variants/heltec_v4/HeltecV4Board.cpp
+++ b/variants/heltec_v4/HeltecV4Board.cpp
@@ -21,10 +21,12 @@ void HeltecV4Board::begin() {
     pinMode(P_LORA_PA_TX_EN, OUTPUT);
     digitalWrite(P_LORA_PA_TX_EN,LOW);
 
+    esp_reset_reason_t reason = esp_reset_reason();
+    if (reason != ESP_RST_DEEPSLEEP) {
+      delay(1);  // GC1109 startup time after cold power-on
+    }
 
     periph_power.begin();
-
-    esp_reset_reason_t reason = esp_reset_reason();
     if (reason == ESP_RST_DEEPSLEEP) {
       long wakeup_source = esp_sleep_get_ext1_wakeup_status();
       if (wakeup_source & (1 << P_LORA_DIO_1)) {  // received a LoRa packet (while in deep sleep)


beachmiles:
Nice, hope this fix will get merged soon for cases when powersave is enabled. Awesome.
Any chance I could get a build of this for repeaters and companions for the v4? Thanks.

variants/heltec_tracker_v2/HeltecTrackerV2Board.cpp
```diff
@@ -9,6 +9,7 @@ void HeltecTrackerV2Board::begin() {
     pinMode(P_LORA_PA_POWER, OUTPUT);
     digitalWrite(P_LORA_PA_POWER,HIGH);
 
+    rtc_gpio_hold_dis((gpio_num_t)P_LORA_PA_POWER);
```
Socalix, line 15: Should to move to line 8

variants/heltec_v4/HeltecV4Board.cpp
```diff
@@ -10,6 +10,7 @@ void HeltecV4Board::begin() {
     pinMode(P_LORA_PA_POWER, OUTPUT);
     digitalWrite(P_LORA_PA_POWER,HIGH);
 
+    rtc_gpio_hold_dis((gpio_num_t)P_LORA_PA_POWER);
```
Socalix, line 16: Should to move to line 9

Socalix:
@weebl2000 Thank you for replacing the original PR with this one. It is much simpler and straightforward code, and easy to see what exactly it's doing.
Just need to move the hold_dis() before write() calls.

beachmiles:
I finally got the meshcore build env working.
Built the heltec v4 repeater and v4 companion bluetooth version from the dev branch with pr1398 and added this pr1600 along with socalix's recommendation to move that rtc disconnect to line 9.
**Edit. Looks like socalixs recommendation was incorrect so I need to rebuild again. The zip attached were what I flashed and the companion only results below are with the mod that socalix recommended so hopefully someone can rebuild or I can do it on Monday**
I flashed these to my v4 repeater and companion and both appear to to be working. 

Dont know how to enable powersaving mode on the companion but I'm drawing ~0.74 watts (.143 amps at 5.15V) with the screen off where on the 1.11 ver I was previously running just under .6 watts in idle mode. Maybe this is because the LNA is now staying powered on while the esp32 is in idle or light sleep?

Edit. If I disconnect from bluetooth and reconnect the watttage drops back down under .6 watts for a like ~30 seconds but then jumps back up to .74 watts and never comes back down.

**Edit. This build doesn't have good powersaving for repeaters and higher power usage for companions. My build from the dev branch must have brought something bad in. But the RX fix seems to be working at least!**
[heltec_v4_repeater_and_companion_1.12_w_pr1398_and_pr1600_RX_fixes_rev2.zip](https://github.com/user-attachments/files/25118332/heltec_v4_repeater_and_companion_1.12_w_pr1398_and_pr1600_RX_fixes_rev2.zip)

The dev put a repeater .bin with his fixes here which has the good power savings.
https://github.com/meshcore-dev/MeshCore/pull/1600#issuecomment-3867505320

weebl2000:
@Socalix I think the current order is correct.

The key distinction here is between the GPIO peripheral registers and the pad hold logic.

digitalWrite() updates the GPIO output registers unconditionally — even while the pad is held. The hold only freezes the pad, not the register state. When rtc_gpio_hold_dis() is called, the pad immediately reflects whatever level is already programmed in the GPIO registers.

This is why the ESP-IDF docs for gpio_hold_dis() state:

“If you don't want the level changes, the gpio should be configured to a known state before this function is called.”

That instruction only makes sense if register writes are expected to occur while the pad is held.

With the current order:

Configure GPIO output HIGH (register updated, pad still held HIGH)

Release hold → pad transitions from held-HIGH to register-HIGH (no glitch)

Reversing the order would cause the pad to briefly revert to its default state on hold release, which would momentarily remove power from the GC1109 FEM.

So the existing order is intentional and avoids a power glitch during wake from deep sleep.

image
https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-reference/peripherals/gpio.html

alexsporn:
@weebl2000 adding this as a comment to the code might be valuable so it does not get changed in the future because "it looks wrong"

weebl2000:
> @weebl2000 adding this as a comment to the code might be valuable so it does not get changed in the future because "it looks wrong"

done

weebl2000:
> Edit. If I disconnect from bluetooth and reconnect the watttage drops back down under .6 watts for a like ~30 seconds but then jumps back up to .74 watts and never comes back down.
> [heltec_v4_repeater_and_companion_1.12_w_pr1398_and_pr1600_RX_fixes_rev2.zip](https://github.com/user-attachments/files/25118332/heltec_v4_repeater_and_companion_1.12_w_pr1398_and_pr1600_RX_fixes_rev2.zip)

I haven't tried it myself yet with the companion. It might be difficult to test if you're receiving messages all the time btw.

Unless you have a faraday cage. 😋

Socalix:
> @Socalix I think the current order is correct.
> The key distinction here is between the GPIO peripheral registers and the pad hold logic.
> digitalWrite() updates the GPIO output registers unconditionally — even while the pad is held. The hold only freezes the pad, not the register state. When rtc_gpio_hold_dis() is called, the pad immediately reflects whatever level is already programmed in the GPIO registers.

@weebl2000 This is interesting. Before commenting above, I checked the same documentation specifically for `rtc_gpio_hold_dis()` that's used in the code and there is no mention of switching to default state. It just says:
`Disabling hold function will allow the pad receive the values of input enable, output enable, output value, function, drive strength from RTC_IO peripheral.`

I don't know why there's difference between `gpio_hold_dis()` and  `rtc_gpio_hold_dis()` documentation and what is the correct answer. Maybe we need Heltec to provide info specifically for the v4 implementation, or do some more digging and hardware testing?

rtc_gpio_hold_dis

weebl2000:
I've updated the code comments to clarify a bit more.

towerviewcams:
@beachmiles I have tested your 1398 / 1600 rxfix.rev2 repeater firm file and V4 draws 37mA.  I loaded stock firm 1.12 and then yours. I also tried the no display firm that brings down to 11mA and then yours and it jumps to 37mA again.

ignisf:
@weebl2000 can we please get steps how to test this? Reading through the comments I think people are talking about different things and possibly testing incorrectly.

weebl2000:
> @weebl2000 can we please get steps how to test this? Reading through the comments I think people are talking about different things and possibly testing incorrectly.

Realistically, you can test by running a repeater with power saving enabled. Then, when it goes to sleep you should measure voltage across the GC1109 pins and see that it's still being powered.

The GC1109 is behind the screen so you would have to remove your heltec v4's screen.

But the real fix is in keeping it alive during sleep/wake cycles.

You can also "test" by checking RSSI during receive that happen when the v4 is sleeping.

towerviewcams:
@weebl2000 I'm measuring the GC1109 now. Pin 15 and 16 both have 3.7 volts. It looks like possibly that your file has the OLED screen still powered. This was found to be the case with 1.12 and then both there was a temporary "no display" firm file made by @Socalix to fix this. Then Kevin @IoTThinks made a more perm fix after that. was PR1570 and then looks like another opened with 1608 even though not much clarity there.

If I load the "no display" firmware, then I see 11mA....... Now, there is NO power on the GC1109 once sleep delay ends. pin 15 and 16 go to zero. When I do a telemetry check I see a SNR of -116 and then after that it goes up to -104 that shows in fact the LNA did NOT have power - so your correct. Power then stays on for the long delay and then once powersaving goes back into effect, GC1109 goes dark.

Lets test this and see if it truly draws 6 ish mA more on top of the 11mA. look forward to your next file.

weebl2000:
> @weebl2000 I'm measuring the GC1109 now. Pin 15 and 16 both have 3.7 volts. It looks like possibly that your file has the OLED screen still powered. This was found to be the case with 1.12 and then both there was a temporary "no display" firm file made by @Socalix to fix this. Then Kevin @IoTThinks made a more perm fix after that. was PR1570 and then looks like another opened with 1608 even though not much clarity there.
> 
> If I load the "no display" firmware, then I see 11mA....... Now, there is NO power on the GC1109 once sleep delay ends. pin 15 and 16 go to zero. When I do a telemetry check I see a SNR of -116 and then after that it goes up to -104 that shows in fact the LNA did NOT have power - so your correct. Power then stays on for the long delay and then once powersaving goes back into effect, GC1109 goes dark.
> 
> Lets test this and see if it truly draws 6 ish mA more on top of the 11mA. look forward to your next file.

Try building this branch, it contains all my open PRs + dev: https://github.com/weebl2000/MeshCore/tree/dev_plus

towerviewcams:
@weebl2000 there are a ton of files there. Not my strong suit. But if another test bin is created I will test.

weebl2000:
> @weebl2000 there are a ton of files there. Not my strong suit. But if another test bin is created I will test.

https://x230.weebl.me/public/v4_dev_plus.bin - v4 repeater build with all fixes. Flash without erase.

towerviewcams:
@weebl2000  Yes sir, we went from 11.1 mA to now 12.3mA and the GC1109 has power in sleep. I'm still testing but when I see the current drop from 47ma after the last signal is received, it holds, then after 20 seconds, sleeps and then drops back to 12.3 ish.......GC1109 power remains on

beachmiles:
> > @weebl2000 there are a ton of files there. Not my strong suit. But if another test bin is created I will test.
> 
> https://x230.weebl.me/public/v4_dev_plus.bin - v4 repeater build with all fixes. Flash without erase.

Could you please make a .bin for the companion Bluetooth as well for me to test?  Not sure if companions ever deep sleep since there is no way to run the "powersaving on" command that I know of, but am curious to see if there is less power usage on your build.

Sounds like something else in the dev branch I built from with your fix maybe broke powersaving on the repeater version that towercams tested?
Thanks!

weebl2000:
Here's a build for Heltec v4 companion (BLE): https://x230.weebl.me/public/v4_companion_ble_dev_plus_25c7ec3d.bin

This includes the deep sleep LNA fix from this PR, plus BLE companion power saving (#1347) — power saving kicks in when you disable Bluetooth with the button.

towerviewcams:
@weebl2000  Your companion build does work and current draw drops down from 110ma when BLE is connected to a very low 10.6mA when BLE is disabled via the long press button. The display is staying on and I cant get it to shut off.

weebl2000:
> @weebl2000 Your companion build does work and current draw drops down from 110ma when BLE is connected to a very low 10.6mA when BLE is disabled via the long press button. The display is staying on and I cant get it to shut off.

Is the dipslay staying on all the time, or do you mean it keeps waking up on message received?

That last one is to be expected. Every time it receives a message it wakes up from deep sleep. Since no companion is connected (BLE is off) - the screen turns on.

On my Heltec v4 the screen turns off correctly btw.

towerviewcams:
It was weird, I had to reboot, then long press to turn off BLE and then display went off. I've never tested current draw in companion mode on a V4 before this.....Lots of power for sure at 110ma......

weebl2000:
> It was weird, I had to reboot, then long press to turn off BLE and then display went off. I've never tested current draw in companion mode on a V4 before this.....Lots of power for sure at 110ma......

FYI, the build I shared has higher BLE power (P9), because I noticed bad range at the default bluetooth power for my Heltec v4. But duty cycle is very low so it shouldn't matter too much.

beachmiles:
> @weebl2000 Yes sir, we went from 11.1 mA to now 12.3mA and the GC1109 has power in sleep. I'm still testing but when I see the current drop from 47ma after the last signal is received, it holds, then after 20 seconds, sleeps and then drops back to 12.3 ish.......GC1109 power remains on

On my repeater with this latest "[v4_dev_plus.bin](https://x230.weebl.me/public/v4_dev_plus.bin)" build I am not seeing it enter powersaving mode much probably due to it being in a fairly high traffic location as I am seeing mainly drawing a steady ~43mA @ 5.09V  (.219 watts) with occasional short peaks at 729mA@ 4.82V (3.51 watts) along with a bright white LED flash from the heltec which does'nt appear to be due to transmissions as I see this bright LED flash sometimes without this large wattage peak.
But ya every once in a while I saw it go into powersave mode drawing 7.1mA at 5.09V ( .0349 watts) and when a RX came in it went back up to 43mA.

I was powering heltec with USB power brick. Will try to get my Nordic power profiler hooked up and power it from the 3.3v to get a better picture of what's happening 

Need to find out what is causing that 3.51 watt spike though, maybe it is a TX? Ive seen this seemingly random short spike on every build for this heltec v4.
20 seconds sure seems like too long of a delay before entering powersave mode. Would think we would want to enter powersave after a max of 1 seconds of inactivity, ideally maybe ~10ms max?

weebl2000:
1 second is very aggressive. There's usually things that need to happen after receiving packets, like transmitting, acking and the likes. Also, going to sleep and waking up every time is also a power hungry exercise. Good tradeoff might be 5 seconds.

towerviewcams:
@beachmiles There is allot more going on after the last received packet. IF we sleep to quick, then when the system is active there will be delays, missed messages etc. 20 seconds is to long yes. Normally Kevin's powersaving i've noticed is around 5-10 seconds.

beachmiles:
> 1 second is very aggressive. There's usually things that need to happen after receiving packets, like transmitting, acking and the likes. Also, going to sleep and waking up every time is also a power hungry exercise. Good tradeoff might be 5 seconds.

Maybe using the esp32 light sleep mode would make more sense during this long delay before finding a gap in traffic long enough to allow for deep sleep, not sure if this is already being done. The 1262 lora chip has a 256 byte RX buffer so it should give enough time for the esp32 to wake up from deep sleep to process the new packet. But if there is tons of other random meshcore housekeeping timers / time critical functions getting used maybe just using light sleep mode would make more sense?
Think there is a lot of optimization to be done still in powersaving mode like turning off the led flashes for TX and RX events, etc.

Can you verify in your compiled bins that the RX register fix https://github.com/meshcore-dev/MeshCore/pull/1398 is included with the addition of this flag SX126X_REGISTER_PATCH? Pretty sure it is.

weebl2000:
> Can you verify in your compiled bins that the RX register fix #1398 is included with the addition of this flag SX126X_REGISTER_PATCH? Pretty sure it is.

Yeah, it's all in there.

beachmiles:
1.13 released with v4 RX fix and powersaving fixes but it doesn't look like this fix made it in saving P_LORA_PA_POWER during deep sleep.

weebl2000:
> 1.13 released with v4 RX fix and powersaving fixes but it doesn't look like this fix made it in saving P_LORA_PA_POWER during deep sleep.

Correct, it's not merged yet. You can build it separately at https://mcimages.weebl.me - if you build "dev_plus" branch it's dev+ all my open PRs.

weebl2000:
@ripplebiz In support of this, someone else pointed out to me that Heltec does it the same way in their ESP32 library:

https://github.com/HelTecAutomation/Heltec_ESP32/blame/7663bf84c4c54afddb181d64706c7ced77aea0a5/src/radio/radio.c#L889-L901C4

Pretty sure it's the right way.

PR #1603, opened 2026-02-06
 Add RP2040 and STM32 support to build.sh

oltaco:
GithHub builds for RP2040 and STM32 environements are failing due to build.sh trying to run mergebin on them.

This PR adds a function get_platform_for_env() to get the build platform from the pio config for the environment that we are building and uses that for post-build actions so that we can copy the correct files to /out. It also adds -D ESP32_PLATFORM to esp32_base so that we can target those builds in the post-build actions.

~~It does rely on jq which is included as part of the ubuntu-latest runner that the GitHub builds use, but it might be worth adding jq to .github/actions/setup-build-environment/action.yml~~
Switched to using python so there's no requirement for jq

@liamcottle You're the master, can you take a look?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -64,6 +64,8 @@ case $1 in
     ;;
 esac
 
+# cache project config json for use in get_platform_for_env()
+PIO_CONFIG_JSON=$(pio project config --json-output)
 
 # $1 should be the string to find (case insensitive)
 get_pio_envs_containing_string() {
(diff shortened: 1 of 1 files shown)


oltaco:
On @liamcottle's suggestion, this is now using python for get_platform_for_env() so there's jq requirement.

liamcottle (approved):

PR #1605, opened 2026-02-06
Kiss modem spec compliance

ViezeVingertjes:
After quite a lengthy _brainstorm-session_ in Discord with @mepholic about the current `KISS-like` implementation vs actual `KISS`, gave it a shot and it seems to work out quite well, so why not go the extra mile?

Also included the fixes from @agessaman #1591 

## **kissattach / axlisten**
terminal-screenshot-kissattach

## **Basic decoding of incoming frames**
terminal-screenshot-monitor

## **Sending an advert** _(received on my companion)_
terminal-screenshot-advert

### Standard KISS

| Test | Result |
|------|--------|
| TXDELAY / Persistence / SlotTime / TXtail / FullDuplex produce no response | Pass |
| Return (0xFF) does not crash or produce response | Pass |
| Unknown cmd 0x07, 0x0A silently ignored | Pass |
| Non-zero port (1, 2) frames silently ignored | Pass |
| Back-to-back FEND handling | Pass |
| Byte escaping (FEND/FESC in payload) | Pass |
| Invalid escape (FESC+0x42) still processes | Pass |
| Empty data frame / empty param commands ignored | Pass |
| Ping → Pong | Pass |

### SetHardware – Identity & Crypto

| Test | Result |
|------|--------|
| GetVersion → version=1, reserved=0 | Pass |
| GetIdentity → 32-byte public key | Pass |
| GetRandom(16) → 16 bytes; two calls differ; 0/65 return InvalidParam | Pass |
| Hash matches SHA-256; empty returns InvalidLength | Pass |
| SignData → 64-byte signature; empty returns InvalidLength | Pass |
| VerifySignature valid→0x01, tampered→0x00; too short → InvalidLength | Pass |
| Encrypt/Decrypt round-trip; too short → InvalidLength; bad MAC → MacFailed | Pass |
| KeyExchange → 32-byte shared secret, deterministic; too short → InvalidLength | Pass |

### SetHardware – Radio & Diagnostics

| Test | Result |
|------|--------|
| SetRadio / GetRadio round-trip (freq=869618000, bw=62500, sf=8, cr=5) | Pass |
| SetRadio too short → InvalidLength | Pass |
| SetTxPower / GetTxPower round-trip (power=14) | Pass |
| SetTxPower empty → InvalidLength | Pass |
| GetBattery, GetStats, GetCurrentRssi, IsChannelBusy, GetAirtime, GetNoiseFloor | Pass |
| GetAirtime empty → InvalidLength | Pass |
| GetSensors(0x07); empty → InvalidLength | Pass |
| GetMCUTemp (24.0°C), GetDeviceName (Seeed Xiao-nrf52) | Pass |
| Unknown sub-cmd → Error 0x05 (UnknownCmd) | Pass |

### Data Frames & TX

| Test | Result |
|------|--------|
| Data frame TX → TxDone result=0x01 | Pass |
| 255-byte data frame TX succeeds | Pass |
| Data frame >255 bytes silently dropped | Pass |
| Duplicate TX → only one TxDone | Pass |
| Stats TX count incremented after TX | Pass |

### DIFF
--- a/examples/kiss_modem/main.cpp
+++ b/examples/kiss_modem/main.cpp
@@ -136,21 +136,11 @@ void loop() {
       int8_t rssi = (int8_t)radio_driver.getLastRSSI();
       modem->onPacketReceived(snr, rssi, rx_buf, rx_len);
     }
-    /* Sample noise floor right after drain: we're in STATE_RX so wrapper can collect. */
-    for (int i = 0; i < 16; i++) {
-      radio_driver.loop();
-    }
   }
 
-  /* Trigger starts a new 64-sample calibration window every 2s. */
   if ((uint32_t)(millis() - next_noise_floor_calib_ms) >= NOISE_FLOOR_CALIB_INTERVAL_MS) {
     radio_driver.triggerNoiseFloorCalibrate(0);
     next_noise_floor_calib_ms = millis();
   }
   radio_driver.loop();
-  if (!modem->isActuallyTransmitting()) {
-    for (int i = 0; i < 15; i++) {
-      radio_driver.loop();
-    }
-  }
 }


ViezeVingertjes:
If you could do a review of the changes when you have time? @mepholic
Reviewed and tested it, now first need some fresh air before looking at it once more, it will all look the same hehe. 👍🏼

liamcottle:
Would be interested to know if Winlink can send messages through the KISS interface after your changes. (I haven't read over it). I actually tried the other day, and nothing was being transmit when attempting a P2P packet session via serial/kiss. I even went as far as downloading a serial logger application. Commented out all other firmware replies, but CMD_DATA still wouldn't transmit. Worked fine in my simple nodejs test script... :D

ViezeVingertjes:
> Would be interested to know if Winlink can send messages through the KISS interface after your changes. (I haven't read over it). I actually tried the other day, and nothing was being transmit when attempting a P2P packet session via serial/kiss. I even went as far as downloading a serial logger application. Commented out all other firmware replies, but CMD_DATA still wouldn't transmit. Worked fine in my simple nodejs test script... :D

Not familiar with Winlink, but now would likely work if it uses the DATA frames. If you have time/chance to test it, would be a nice addition/confirmation in this PR though. o:)

ViezeVingertjes:
I also included your fixes @agessaman in case this would get merged. 💯

agessaman:
@ViezeVingertjes I sent over a PR on your fork. This one gets the noise floor working (the loop wasn't polling enough times to get a valid sample so it was getting stuck at -120) and making sure the radio polls in all states when not in TX mode. With these changes have pyMC_core working with this implementation, and pyMC_repeater things everying is working well.

ViezeVingertjes:
> @ViezeVingertjes I sent over a PR on your fork. This one gets the noise floor working (the loop wasn't polling enough times to get a valid sample so it was getting stuck at -120) and making sure the radio polls in all states when not in TX mode. With these changes have pyMC_core working with this implementation, and pyMC_repeater things everying is working well.

Great! it's in and resolved the conflicts that (the other PR was merged in the meantime hehe)

mepholic:
This looks pretty good to me!

# SetHardware response codes

My main comment is regarding the SetHardware commands and responses. To me it feels like it would me more in line with the rest of the spec to make the response code prefixes match the command codes themselves. This also reduces the amount of times you have to repeat yourself in your code's `#define`'s, and may be more intuitive for other developers interacting with the modem.

For instance, the SetHardware command for GetIdentity is `0x01`, but the Identity response seems to be indicated with `0x21`. I can't think of any good reasons to make the request and response codes different from one another. Is there something you have in mind that I'm missing?

The way I plan on implementing these is you send a SetHardware command, followed by a hardware specific command, such as GetIdentity. So for example, excluding the `FEND` and `FESC` semantics, you send `0xP6 0x01`, and then in response you get `0xP6 0x01 xx xx xx xx`, where `P` is the destination or source KISS port. The response uses the same format as the command, followed by whatever return value the command produces.

Generally, the KISS client can differentiate requests and responses based on what it sends, vs what it receives, with the sole exception of the scenario where there's another KISS client on the on the bus that is attempting to interact with the radios. In these cases, if a client really needs to differentiate requests and responses, it's feasible that it could be inferred based on what additional data is included with the code (could be input or output parameters, for instance)

If you _really_ need to use separate response codes from command codes, I suggest splitting the byte equally into "request space" (`00` - `7F`) and "response space" (`80` - `FF`), which should give you 128 commands and responses. Or... another solution could be to add another byte after SetHardware, but before the hardware specific command byte, to indicate if the following byte contains a request from a KISS client or a response from the modem.

# Get RSSI implementation

The final thing I was wondering was about your implementation of `GetCurrentRssi`. I imagine this gives you the RSSI of the last received packet, is that correct? If so, in the current state of things, someone interacting with the modem would have to send a `GetCurrentRssi` command after _every_ packet received in order to get signal strength reports on each. Not only might they not have sufficient time to do this between packets, but it also unnecessarily increases the amount of direct interactions that the client must perform on the KISS modem.

This command is probably fine to have for one-off operations, but I intended on implementing a `SetSignalReport` command, which would enable or disable of reporting RSSI and SNR for every received packet directly after sending the packet to the KISS client in CMD_DATA. 

It would look something like this (command codes sort of randomly chosen):
Command | Cmd Byte | Input Value | Immediate Reponse | Description
----------- | ---------- | ------------- | ---------- | -------------
`SetSignalReport` | `0x67` | `0x00` / `0x01` | N/A or echo back the user command to them to confirm | Enable or disable per-rx packet signal reports, 0x00 to disable, any other value to enable
`GetSignalReport` | `0x68` | None | `0x00` / `0x01` | Get current signal report status (enabled or disabled)

When signal reporting is enabled, for inline signal reports, we could probably use one of the undefined command values from the spec, such as `0xP7` and just embed the RSSI and SNR after it. If this seems uncouth, another option would be to just encode the response as an alternate form of `GetSignalReport` (when the command isn't invoked directly), which would look something like `0xP6 0x68 xx xx xx xx (rssi) xx xx xx xx (snr)`.

ViezeVingertjes:
> This looks pretty good to me!
> 
> # SetHardware response codes
> My main comment is regarding the SetHardware commands and responses. To me it feels like it would me more in line with the rest of the spec to make the response code prefixes match the command codes themselves. This also reduces the amount of times you have to repeat yourself in your code's `#define`'s, and may be more intuitive for other developers interacting with the modem.
> 
> For instance, the SetHardware command for GetIdentity is `0x01`, but the Identity response seems to be indicated with `0x21`. I can't think of any good reasons to make the request and response codes different from one another. Is there something you have in mind that I'm missing?
> 
> The way I plan on implementing these is you send a SetHardware command, followed by a hardware specific command, such as GetIdentity. So for example, excluding the `FEND` and `FESC` semantics, you send `0xP6 0x01`, and then in response you get `0xP6 0x01 xx xx xx xx`, where `P` is the destination or source KISS port. The response uses the same format as the command, followed by whatever return value the command produces.
> 
> Generally, the KISS client can differentiate requests and responses based on what it sends, vs what it receives, with the sole exception of the scenario where there's another KISS client on the on the bus that is attempting to interact with the radios. In these cases, if a client really needs to differentiate requests and responses, it's feasible that it could be inferred based on what additional data is included with the code (could be input or output parameters, for instance)
> 
> If you _really_ need to use separate response codes from command codes, I suggest splitting the byte equally into "request space" (`00` - `7F`) and "response space" (`80` - `FF`), which should give you 128 commands and responses. Or... another solution could be to add another byte after SetHardware, but before the hardware specific command byte, to indicate if the following byte contains a request from a KISS client or a response from the modem.
> 
> # Get RSSI implementation
> The final thing I was wondering was about your implementation of `GetCurrentRssi`. I imagine this gives you the RSSI of the last received packet, is that correct? If so, in the current state of things, someone interacting with the modem would have to send a `GetCurrentRssi` command after _every_ packet received in order to get signal strength reports on each. Not only might they not have sufficient time to do this between packets, but it also unnecessarily increases the amount of direct interactions that the client must perform on the KISS modem.
> 
> This command is probably fine to have for one-off operations, but I intended on implementing a `SetSignalReport` command, which would enable or disable of reporting RSSI and SNR for every received packet directly after sending the packet to the KISS client in CMD_DATA.
> 
> It would look something like this (command codes sort of randomly chosen):
> 
> Command	Cmd Byte	Input Value	Immediate Reponse	Description
> `SetSignalReport`	`0x67`	`0x00` / `0x01`	N/A or echo back the user command to them to confirm	Enable or disable per-rx packet signal reports, 0x00 to disable, any other value to enable
> `GetSignalReport`	`0x68`	None	`0x00` / `0x01`	Get current signal report status (enabled or disabled)
> When signal reporting is enabled, for inline signal reports, we could probably use one of the undefined command values from the spec, such as `0xP7` and just embed the RSSI and SNR after it. If this seems uncouth, another option would be to just encode the response as an alternate form of `GetSignalReport` (when the command isn't invoked directly), which would look something like `0xP6 0x68 xx xx xx xx (rssi) xx xx xx xx (snr)`.

Thanks for the review.

**Response codes**: Agreed on the high-bit split. Responses are now `cmd | 0x80` via a `HW_RESP(cmd)` macro, with generic OK/Error at `0xF0`/`0xF1` and unsolicited frames (TX_DONE, RX_META) at `0xF8`+. I kept separate response codes because unsolicited frames and shared OK/Error don't have a single command to echo back.

**Signal reporting**: There already was inline per-packet SNR+RSSI (`RX_META` after every `CMD_DATA`), but I've now made it toggleable via `SetSignalReport`/`GetSignalReport` (`0x19`/`0x1A`). Enabled by default.

`GetCurrentRssi` reads the radio's RSSI register, not tied to last-packet RSSI.

I think this will make it more solid, let me know if i missed anything.

mepholic:
Looks good! I see that you actually fully implemented the CSMA p-Persistence algorithm as well. Great work! I think this will really help with congestion control, and it can also be tuned on a case-by-case basis to make the radio more or less aggressive at transmitting.

The default TX delay could probably be turned down a bit? Not sure what a good value is there. Did you get those defaults from somewhere, or test it and decide on them on your own?

Regarding the code and formatting of the serial commands and responses, is that it seems like at a minimum it would be a  good idea to include lengths for variable length data in commands sent to the KISS modem and in any responses that contain variable length data as well. This provides a sanity check of "how much data comes next" to each side of the conversation, and can make interacting with it from certain programming languages easier.

Serial connections can be notoriously noisy, and noise in a serial line can corrupt data. Many serial protocols implement CRC's to add some level of data integrity checking, but I think that's probably overkill in this case. I imagine most users will be using the USB serial console, which is less likely to suffer from issues of data corruption.

ViezeVingertjes:
> The default TX delay could probably be turned down a bit? Not sure what a good value is there. Did you get those defaults from somewhere, or test it and decide on them on your own?

Yes; pretty sure it came straight from the protocol document as default value but it can also be set through `TXDELAY` if required.

> Regarding the code and formatting of the serial commands and responses, is that it seems like at a minimum it would be a good idea to include lengths for variable length data in commands sent to the KISS modem and in any responses that contain variable length data as well.

I am not sure i understand why this would be beneficial in addition to the framing (`FEND` etc)? Do you have a certain edge case in mind?


We probably could roll with this version if there arent any weird things popping up and go from there. Should be very usable as-is currently. So far seems to run like a charm here.

mepholic:
I guess with the framing it's probably fine. LGTM!

ViezeVingertjes:
And so it became a full KISS compliant firmware afterall. 👍🏼 

@ripplebiz @recrof

ViezeVingertjes:
Managed to get it to work on Heltec CubeCell's (HTCC-AB01 & HTCC-AB02A), which both only have 128kb flash and 16kb ram. 👍🏼 

```
RAM:   [==        ]  15.2% (used 2488 bytes from 16384 bytes)
Flash: [========= ]  92.0% (used 120572 bytes from 131072 bytes)
```

And even further by reducing the crypto & deduplicating the libs with https://github.com/meshcore-dev/MeshCore/pull/1632

```
RAM:   [===       ]  25.1% (used 4120 bytes from 16384 bytes)
Flash: [=======   ]  65.9% (used 86380 bytes from 131072 bytes)
```

PR #1608, opened 2026-02-07
Fix Heltec v4 power consumption

weebl2000:
This improves on PR #1570 but I couldn't push to their fork directly.

Please read discussion there.


towerviewcams:
@weebl2000 is there a test bin file to take a look at this?

weebl2000:
> @weebl2000 is there a test bin file to take a look at this?

Companion I presume?

PR #1609, opened 2026-02-07
Bridge always has work (prevents sleep)

### LINKED ISSUE #1606: Powersaving stops espnow bridge from working
#1266 Seems to stop the Espnow bridge form working on my Heltec V4 when set to powersaving on.
As soon as I disable powersaving it works again.
Please investigate, thank you.

weebl2000:
fixes #1606

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -1219,5 +1219,8 @@ void MyMesh::loop() {
 
 // To check if there is pending work
 bool MyMesh::hasPendingWork() const {
+#if defined(WITH_BRIDGE)
+  if (bridge.isRunning()) return true;  // bridge needs WiFi radio, can't sleep
+#endif
   return _mgr->getOutboundCount(0xFFFFFFFF) > 0;
 }


JDat:
What if my variant is interrupt driver and board sleep until got interrupt from Serial hardware or from Radio module?

I don't thing this should be in repeater logic. Move this logic to board/variant code.

Yes, I am using RP2040 instead of Heltec/ESP32 !

Never mess with high level code when it can intact with other platforms when you only need to work with one board/variant/platform level stuff!

weebl2000:
> What if my variant is interrupt driver and board sleep until got interrupt from Serial hardware or from Radio module?
> 
> I don't thing this should be in repeater logic. Move this logic to board/variant code.
> 
> Yes, I am using RP2040 instead of Heltec/ESP32 !
> 
> Never mess with high level code when it can intact with other platforms when you only need to work with one board/variant/platform level stuff!

If the bridge isn't running then it will still sleep. Are you saying RP2040 can keep a bridge functioning properly while also sleeping?

JDat:
Bridge - not yet, my simple repeater is already working by interrupt driving instead of main loop. There are still some glitches, but I am working on that. Need to take into account: this is platform dependent stuff.

IoTThinks:
Hi @weebl2000 
So this PR will skip sleep if bridge is ever started?
Could it sleeps then send data over to uart then sleep again?

If yes, then I don't need to put in wakeup on UART.

PR #1612, opened 2026-02-07
Default button polarity to active-LOW across all firmware types

weebl2000:
Nearly all LoRa boards use a boot button that pulls to ground when pressed. The T1000-E is the one exception — it uses a capacitive touch button that is active-HIGH (confirmed by Meshtastic's `BUTTON_ACTIVE_LOW false`).

Changes:
- Default `USER_BTN_PRESSED` to `LOW` where not already defined (ESP32Board.h, UITask.cpp, XiaoNrf52Board.h)
- Replace hardcoded `LOW` button checks with `USER_BTN_PRESSED` in UITask and board files
- Fix MinewSemi ME25LS01: was incorrectly set to `USER_BTN_PRESSED=HIGH` with `NRF_GPIO_PIN_SENSE_HIGH` — changed to active-LOW with `SENSE_LOW`
- Fix ThinkNode M3 / Wio-E5-Mini: remove redundant explicit `USER_BTN_PRESSED=LOW` (now the default)
- Fix T1000-E: keep `USER_BTN_PRESSED=HIGH` and `SENSE_HIGH` (correct for this hardware), but fix `buttonStateChanged()` to use `USER_BTN_PRESSED` instead of hardcoded `LOW`

I found out when I was working on #1347 — it was actually releasing the boot button that would trigger the wakeup, not the pressing.

---
**Build firmware:** [Build from this branch](https://mcimages.weebl.me/?commitId=default-button-low)

### DIFF
--- a/examples/simple_repeater/UITask.cpp
+++ b/examples/simple_repeater/UITask.cpp
@@ -2,6 +2,10 @@
 #include <Arduino.h>
 #include <helpers/CommonCLI.h>
 
+#ifndef USER_BTN_PRESSED
+#define USER_BTN_PRESSED LOW
+#endif
+
 #define AUTO_OFF_MILLIS      20000  // 20 seconds
 #define BOOT_SCREEN_MILLIS   4000   // 4 seconds
 
@@ -85,7 +89,7 @@ void UITask::loop() {
   if (millis() >= _next_read) {
     int btnState = digitalRead(PIN_USER_BTN);
     if (btnState != _prevBtnState) {
-      if (btnState == LOW) {  // pressed?
+      if (btnState == USER_BTN_PRESSED) {  // pressed?
         if (_display->isOn()) {
           // TODO: any action ?
         } else {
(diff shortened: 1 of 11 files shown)


LitBomb:
When you say "nearly all LoRa boards..." which boards do not pull to ground when the button pressed?  What would happen to those boards if this PR is merged?

weebl2000:
> When you say "nearly all LoRa boards..." which boards do not pull to ground when the button pressed? What would happen to those boards if this PR is merged?

All devices as far as I can tell, just went through the datasheets. Except T1000-E that is genuinely high. MinewSemi being high was a bug AFAIK.

LitBomb:
Thanks for checking.  T1000-E has been a problem child but with a lot of effort from the community it is finally working.  Sure don't want to break its button functionality.  thanks.

weebl2000:
> Thanks for checking. T1000-E has been a problem child but with a lot of effort from the community it is finally working. Sure don't want to break its button functionality. thanks.

Got my hand on a t1000e, and the button works fine with this PR.

PR #1632, opened 2026-02-08
Deduplicate crypto libraries, saving ~49KB flash

ViezeVingertjes:
Fixes a stack overflow in ed25519_verify() by making large local variables static, removing the need for rweather's Ed25519::verify() workaround.

Only vendors the AES128+SHA256 subset of rweather/Crypto instead of all 37 files.

No issues observed so far. Additional testing and confirmation are welcome. This change frees up stack space for future improvements and improves support for resource-limited devices.


lib/ed25519/ge.c
```diff
@@ -65,9 +65,9 @@ B is the Ed25519 base point (x,4/5) with x positive.
 */
 
 void ge_double_scalarmult_vartime(ge_p2 *r, const unsigned char *a, const ge_p3 *A, const unsigned char *b) {
-    signed char aslide[256];
-    signed char bslide[256];
-    ge_cached Ai[8]; /* A,3A,5A,7A,9A,11A,13A,15A */
+    static signed char aslide[256];
+    static signed char bslide[256];
+    static ge_cached Ai[8]; /* A,3A,5A,7A,9A,11A,13A,15A */
```
weebl2000, line 80: Making these `static` saves ~1.8KB of stack which fixes the overflow, but it also makes `ge_double_scalarmult_vartime()` non-reentrant. If `ed25519_verify()` is ever called concurrently from multiple FreeRTOS tasks, these shared buffers will be silently corrupted.

Currently this looks safe — `Identity::verify()` is only called from the packet receive path in `Mesh.cpp` — but it's worth documenting so future maintainers don't get bitten.

```suggestion
    static signed char aslide[256];   /* static: saves ~1.8KB stack; NOT reentrant */
    static signed char bslide[256];
    static ge_cached Ai[8]; /* A,3A,5A,7A,9A,11A,13A,15A */
```

weebl2000 (commented):
Nice improvement — vendoring the minimal subset is clean and the stack savings are real (~1.8KB moved to BSS). One concern about the `static` locals in `ge.c`.

weebl2000 (approved):

lib/ed25519/ge.c
ViezeVingertjes: What do you thnk about an Assert? just in case its not read... (guilty of it sometimes)

48019bfb512a0aaceb238a305dbf4534f3f5fcc1

weebl2000: Nice, the assert makes the constraint very visible. Two minor notes:

1. `assert()` is typically compiled out in release builds (`NDEBUG`), so this is really a development-time guard. That's fine for this purpose — the comment does the heavy lifting for future readers.
2. The `static volatile bool` check + set isn't atomic, so there's a theoretical TOCTOU race. In practice this doesn't matter — ESP32 FreeRTOS tasks on a single core won't preempt mid-statement, and even on dual-core the window is vanishingly small. Just mentioning it for completeness.

Overall this looks good to me.

src/helpers/RegionMap.cpp
```diff
@@ -1,6 +1,5 @@
 #include "RegionMap.h"
 #include <helpers/TxtDataHelpers.h>
-#include <SHA256.h>
```
JosiahWI, line 3: Good catch.

src/Identity.cpp
```diff
@@ -15,12 +15,13 @@ Identity::Identity(const char* pub_hex) {
...
-#if 0
-  // NOTE:  memory corruption bug was found in this function!!
-  return ed25519_verify(sig, message, msg_len, pub_key);
-#else
-  return Ed25519::verify(sig, this->pub_key, message, msg_len);
-#endif
+  // ed25519_verify uses static buffers internally (ge.c) and is NOT reentrant.
+  // Spinlock to serialize concurrent calls (e.g. from multiple FreeRTOS tasks).
+  static std::atomic<bool> in_verify{false};
+  while (in_verify.exchange(true, std::memory_order_acquire)) { /* spin */ }
```
JosiahWI: I like the spinlock idea.

JosiahWI (commented):
The code looks great. I'm giving my capstone team an opportunity to build this PR and test it and then I'll approve. :)

JosiahWI (approved):
The reduction in program size from this PR will be very useful for my school capstone. One of my teammates has built our branch with it and confirmed that the Flash saving makes it possible to fit the program on an nRF52840 Dongle alongside an InternalFS and DFU bootloader.

Edit: The above is only possible if you are not using a display.

nextgens:
I'd go even further than this: we don't keep key schedules around -> we should only keep AESTiny128 and AESTiny256.

You can get rid of AESSmall* and friends.

nextgens:
I've ended up benchmarking this (see PR above) on ESP32... the specific implementation does not make much difference for our usecase. We definitely don't need all of them though ;)

The most we ever encrypt is packet-length:
```
AES128 (184 bytes): 371 ms for 500 iterations (0.74 ms avg)
AESTiny128 (184 bytes): 361 ms for 500 iterations (0.72 ms avg)
AESSmall128 (184 bytes): 370 ms for 500 iterations (0.74 ms avg)
AES128_ESP (184 bytes): 370 ms for 500 iterations (0.74 ms avg)
```

AESTiny is bugged and stops encrypting after a bit, I haven't looked into why

ViezeVingertjes:
Rebased this branch onto the current dev branch and resolved merge conflicts while preserving the PR scope (vendored crypto subset plus ed25519 reentrancy fixes).

Perhaps we can get this one through if one could validate it on a different setup than mine? More space more better.

weebl2000:
> Rebased this branch onto the current dev branch and resolved merge conflicts while preserving the PR scope (vendored crypto subset plus ed25519 reentrancy fixes).
> 
> Perhaps we can get this one through if one could validate it on a different setup than mine? More space more better.

Which devices/platforms have you tested it on? I can have a go on devices I have.

src/Identity.cpp
nextgens: IMHO that's the wrong place for it though. Why not putting it in the library/verify() instead or the end function?

nextgens (changes requested):

src/Identity.cpp
ViezeVingertjes: > IMHO that's the wrong place for it though. Why not putting it in the library/verify() instead or the end function?

Februari has been a long time but i remember i tried some alternatives first and RP2040 gave me a hard time and in the discord we came to this solution but if you know a more elegant approach that builds/works for all platforms, then i would be totally down to merge it in.

Note that the lib is **C** instead of **C++** though if you meant just moving it over? Or maybe i completely misunderstood.

ViezeVingertjes:
> > Rebased this branch onto the current dev branch and resolved merge conflicts while preserving the PR scope (vendored crypto subset plus ed25519 reentrancy fixes).
> > Perhaps we can get this one through if one could validate it on a different setup than mine? More space more better.
> 
> Which devices/platforms have you tested it on? I can have a go on devices I have.

T-Echo and T-Beam back then, currently runs on my Wismesh Tag.
- Flashed it over my existing firmware which was all fine.
- Erase + flash to make sure nothing odd happens there.

Later imported my own key etc again and now using it normally.
Probably more than needed as it triggers the same code paths, but better safe than sorry.

src/Identity.cpp
nextgens: https://github.com/nextgens/MeshCore/commit/6e3749229cbf6fa5821bc291bb723ac4cbad52a6
is what I had in mind

ViezeVingertjes:
Added tests, rebased so its back in business, retested on ESP32 / nRF.

src/Identity.cpp
nextgens: @ViezeVingertjes what about a C pthread mutex, as above?

nextgens:
What about using a pthread C mutex instead? This is what I have in mind: https://github.com/nextgens/MeshCore/commit/6e3749229cbf6fa5821bc291bb723ac4cbad52a6

The idea being that it´s the library that's worth fixing and not just this caller

ViezeVingertjes:
> What about using a pthread C mutex instead? This is what I have in mind: [nextgens@6e37492](https://github.com/nextgens/MeshCore/commit/6e3749229cbf6fa5821bc291bb723ac4cbad52a6)
> 
> The idea being that it´s the library that's worth fixing and not just this caller

I remember looking into it; but that does not compile for all platforms MeshCore supports (RP2040 & STM32 do not support it). I do however agree that it might also be fixed in the library instead of the caller side, i just found out that i have the commit on a branch from back then, so will need to test it on hardware and could push it.

nextgens (approved):
LGTM

PR #1633, opened 2026-02-08
Fix GPS UART consuming +8mA when disabled (nRF52)

### LINKED ISSUE #1628: GPS causes +8mA power consumption even when disabled on Heltec T114
## Environment
- **Board:** Heltec T114 (nRF52840)
- **Build flag:** `-D ENV_INCLUDE_GPS=1`
- **GPS Module:** Connected to Serial1 (PIN_GPS_TX=37, PIN_GPS_RX=39)
- **Control pins:** PIN_GPS_EN=21, PIN_GPS_STANDBY=34, PIN_GPS_RESET=38

## Problem Description

When compiling with `-D ENV_INCLUDE_GPS=1`, the board consumes **~8mA additional current** even when GPS is disabled through the UI settings.

### Current Measurements:
- **With `-D ENV_INCLUDE_GPS=0`:** 14-16 mA (normal) ✅
- **With `-D ENV_INCLUDE_GPS=1` and GPS disabled:** 22-24 mA ❌
- **With `-D ENV_INCLUDE_GPS=1` and GPS enabled:** 50-56 mA

### Expected Behavior:
GPS disabled → 14-16 mA (same as `ENV_INCLUDE_GPS=0`)

### Actual Behavior:
GPS disabled → 22-24 mA (+8mA leak)

### What we checked:

1. **GPIO pins (PIN_GPS_EN, PIN_GPS_STANDBY, PIN_GPS_RESET)**
   - ✅ All properly initialized to OFF/SLEEP state in `UITask::begin()`
   - ✅ `setGPSHardwareState()` correctly manages all pins
   - ✅ Verified with oscilloscope - pins are in correct state

2. **Serial1 (UART)**
   - ❌ Adding `Serial1.end()` in `stop_gps()` causes issues:
     - Either no effect on power consumption, OR
     - Board hangs/doesn't boot, OR
     - GPS can't be toggled on/off again
   - Serial1 is initialized in `EnvironmentSensorManager::initBasicGPS()` but never stopped

3. **I2C Wire1** 
   - ✅ Tested by commenting out `-D ENV_PIN_SDA` and `-D ENV_PIN_SCL`
   - No change in power consumption

4. **MicroNMEALocationProvider constructor**
   - Creates global object before `setup()`
   - Initializes PIN_GPS_EN but NOT PIN_GPS_STANDBY (we fixed this, but no effect)

### Code Analysis:

The issue appears when this line executes in `target.cpp`:
```cpp
#if ENV_INCLUDE_GPS
#include <helpers/sensors/MicroNMEALocationProvider.h>
MicroNMEALocationProvider nmea = MicroNMEALocationProvider(Serial1);
EnvironmentSensorManager sensors = EnvironmentSensorManager(nmea);
```

## Comparison

**Important:** The same code works correctly on **Heltec V4** board with no extra power consumption when GPS is disabled.

## Questions

1. What peripheral/resource is consuming the extra 8mA when `ENV_INCLUDE_GPS=1`?
2. Why does `Serial1.end()` cause instability instead of saving power?
3. Is there something in the global object initialization that keeps a resource active?
4. How is this handled differently on Heltec V4 vs T114?

### Fix 1: Add PIN_GPS_STANDBY to MicroNMEALocationProvider
```cpp
// Added _pin_standby member and initialization in constructor
if (_pin_standby != -1) {
    pinMode(_pin_standby, OUTPUT);
    digitalWrite(_pin_standby, PIN_GPS_STANDBY_ACTIVE);  // Put GPS to sleep
}
```
**Result:** No change in power consumption

### Fix 2: Stop Serial1 in stop_gps()
```cpp
    _location->stop();
    Serial1.end();  // Stop UART
}
```
**Result:** Board becomes unstable or GPS won't restart

### Fix 3: Initialize all GPS pins to OFF in UITask::begin()
```cpp
digitalWrite(PIN_GPS_EN, !PIN_GPS_EN_ACTIVE);      // Disable power
digitalWrite(PIN_GPS_STANDBY, PIN_GPS_STANDBY_ACTIVE);  // Sleep mode
```
**Result:** No change in power consumption

## Request

Could someone help identify:
1. What resource/peripheral stays active when `ENV_INCLUDE_GPS=1`?
2. The correct way to fully power down GPS-related peripherals on nRF52840?
3. Why the same code works on V4 but not on T114?

## Additional Context

This is critical for battery-powered applications where +50% idle power consumption is unacceptable.

Any guidance would be greatly appreciated!

weebl2000:
## Summary

Fixes #1628 — GPS causes +8mA power consumption even when disabled on Heltec T114 (nRF52840).

### Root cause

On the T114, `GPS_RESET` (pin 38) is the same pin as `PIN_3V3_EN` — it controls a 3.3V power rail for the GPS module:

```cpp
// variants/heltec_t114/variant.h
#define PIN_3V3_EN              (38)
#define GPS_RESET               (38)  // same pin!
```

`MicroNMEALocationProvider::begin()` sets pin 38 HIGH (powering the 3V3 rail / releasing GPS reset), but `stop()` never touched it — it only set `GPS_EN` (pin 21) LOW. The 3V3 rail stayed on, the GPS module stayed powered, drawing +8mA.

Confirmed by comparing with the Meshtastic T114 variant, which has `GPS_RESET` commented out and uses `PIN_GPS_STANDBY` (pin 34) for GPS power control.

### Changes

- **Assert reset pin in `stop()`** — `MicroNMEALocationProvider::stop()` now calls `digitalWrite(_pin_reset, GPS_RESET_FORCE)`, mirroring what `begin()` does in reverse. On the T114 this pulls pin 38 LOW, cutting the 3V3 rail.
- **Guard `_location->loop()`** behind `gps_active` check to skip NMEA processing when GPS is disabled.

### Safety for other boards

- Boards without `GPS_RESET` defined: `_pin_reset` is -1, the new code is skipped entirely.
- Boards with `GPS_RESET`: The constructor already asserts reset at boot. `stop()` now returns to that same state. `begin()` de-asserts it as before. This is the symmetric counterpart that was missing.

### Verified power readings (Heltec T114, reported by @mileshkin)

| State | Current |
|---|---|
| Screen on, GPS on | ~53mA |
| Screen off, GPS on | ~44mA |
| Screen on, GPS off | ~28mA |
| Screen off, GPS off | ~16mA |

## Test plan

- [x] Verify idle current returns to ~16mA with GPS disabled (Heltec T114)
- [x] Verify GPS can be toggled on/off via settings and continues to acquire fix
- [x] Verify no regression on ESP32 GPS targets (Heltec Tracker v2, etc.)

---
**Build firmware:** [Build from this branch](https://mcimages.weebl.me/?commitId=fix/gps-uart-power-leak)

### DIFF
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -707,7 +707,9 @@ void EnvironmentSensorManager::loop() {
   static long next_gps_update = 0;
 
   #if ENV_INCLUDE_GPS
-  _location->loop();
+  if (gps_active) {
+    _location->loop();
+  }
   if (millis() > next_gps_update) {
 
     if(gps_active){
--- a/src/helpers/sensors/MicroNMEALocationProvider.h
+++ b/src/helpers/sensors/MicroNMEALocationProvider.h
@@ -79,7 +79,10 @@ public :
         if (_pin_en != -1) {
             digitalWrite(_pin_en, !PIN_GPS_EN_ACTIVE);
         }
-        if (_peripher_power) _peripher_power->release();  
+        if (_pin_reset != -1) {
+            digitalWrite(_pin_reset, GPS_RESET_FORCE);
+        }
+        if (_peripher_power) _peripher_power->release();
     }
 
     bool isEnabled() override {


mileshkin:
I tried fixing it this way. Like in your case, it resulted in a persistent "Loading..." message.

weebl2000:
The key difference from your Fix 2 is the `gps_active` guard on `_location->loop()`. Without it, the main loop keeps calling `_location->loop()` → `Serial1.available()` / `Serial1.read()` on the ended UART — which on nRF52 causes undefined behavior (hang/crash).

This PR gates that call behind `gps_active`, so after `Serial1.end()` nobody touches the ended UART. That should prevent the hang you saw.

Would you be able to test this branch on your T114?

mileshkin:
I cloned your branch and compiled a firmware from it - the first thing i did, flashed it to my T114, and got an infinite "Loading..."

weebl2000:
Found the cause of the hang. The Adafruit nRF52 `Uart::end()` [spins here](https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/cores/nRF5/Uart.cpp#L168):

```cpp
while ( !(nrfUart->EVENTS_TXSTOPPED && nrfUart->EVENTS_RXTO) ) yield();
```

`EVENTS_RXTO` only fires when the receiver has been **idle for one character period** after STOPRX. When `_location->stop()` cuts GPS power, the GPS module's TX line drops LOW — putting the nRF52 UART RX into a continuous break condition. The receiver is never idle, RXTO never fires, `end()` hangs forever.

Fix (just pushed): pull up the MCU's RX pin before calling `Serial1.end()`:
```cpp
_location->stop();
pinMode(PIN_GPS_TX, INPUT_PULLUP);  // keep RX line idle (HIGH) after GPS powers down
Serial1.end();
```

Could you test again?

mileshkin:
No... it still gets stuck on the "Loading..." stage. If I disconnect the GPS module and reboot, the loading proceeds normally.

weebl2000:
The internal pullup wasn't strong enough — GPS module's TX line has a low-impedance path to ground through ESD diodes when powered off, overpowering the ~13kΩ pull.

New approach (just pushed): instead of fighting the GPS pin state, **redirect PSEL.RXD to the UART's own TX pin** before calling `Serial1.end()`:

```cpp
_location->stop();
NRF_UARTE0->PSEL.RXD = NRF_UARTE0->PSEL.TXD;  // loopback RX to TX
Serial1.end();
```

The UARTE drives TX HIGH in idle state, so RX reads a guaranteed idle line regardless of what the GPS module is doing. RXTO fires immediately and `end()` completes.

Could you give this one a try?

mileshkin:
Still infinite "Loading..." 

I even replaced the GPS module with a similar one to rule out a hardware issue. But the module itself seems to be fine. The Meshtastic firmware works perfectly with this board, disabling the GPS perfectly. I tried to see how they solved this for the T114, but I'm not skilled enough to reverse engineer it properly.

weebl2000:
> Still infinite "Loading..."
> 
> I even replaced the GPS module with a similar one to rule out a hardware issue. But the module itself seems to be fine. The Meshtastic firmware works perfectly with this board, disabling the GPS perfectly. I tried to see how they solved this for the T114, but I'm not skilled enough to reverse engineer it properly.

Can you send some debug logging from restart?

weebl2000:
Took a different approach — compared with Meshtastic's T114 variant to understand how they handle GPS power.

### What we found

On the T114, `GPS_RESET` (pin 38) is the **same pin** as `PIN_3V3_EN` — it controls a 3.3V power rail:

```cpp
// variants/heltec_t114/variant.h
#define PIN_3V3_EN              (38)
#define GPS_RESET               (38)  // same pin!
```

`MicroNMEALocationProvider::begin()` sets pin 38 HIGH (powering the 3V3 rail / releasing GPS reset), but `stop()` never touched pin 38 — it only set `GPS_EN` (pin 21) LOW. We think this means the 3V3 rail stays on and the GPS module stays powered, which would explain the +8mA.

This could also explain why all the `Serial1.end()` attempts hung — if the GPS is still powered, it's actively sending NMEA data over its TX line.

Interestingly, Meshtastic's T114 variant has `GPS_RESET` commented out entirely and uses `PIN_GPS_STANDBY` (pin 34) instead.

### Changes in this push

1. **Assert reset pin in `stop()`** — `MicroNMEALocationProvider::stop()` now calls `digitalWrite(_pin_reset, GPS_RESET_FORCE)` to pull pin 38 LOW, which should cut the 3V3 rail.
2. **Reverted all `Serial1.end()`/PSEL workarounds** — shouldn't be needed if the GPS is actually powered off.
3. **Kept `gps_active` guard** on `_location->loop()` to skip NMEA processing when GPS is disabled.

@mileshkin could you give this latest push a try? Hoping this one actually addresses the root cause.

mileshkin:
Woohoo! It worked! Now everything is as it should be:

- screen on, GPS on: 53mA
- screen off, GPS on: 44mA
- screen on, GPS off: 28mA
- screen off, GPS off: 16mA

Just what we needed! Brilliant work colleague! I hesitated to link to Meshtastic because I thought it wouldn't be very ethical within the Meshcore framework but on the first try, this reverse engineering study you did paid off!
Thank you so much!

mileshkin:
I'm going to check the other boards that might have been affected - hopefully without any negative consequences.

weebl2000:
> I'm going to check the other boards that might have been affected - hopefully without any negative consequences.

Cool. Pretty sure it is safe for other boards. Only boards where GPS reset pin was shared with something else like 3.3V here are probably going to see a fix. The stop was simply missing the reset.

mileshkin:
> Pretty sure it is safe for other boards

I checked it on other boards - no unwanted consequences were found!

PR #1634, opened 2026-02-08
Add highly experimental support for Waveshare esp32 c6 and LR1121 Chipset

aslater3:
I have managed to get Waveshare esp32 c6 and Waveshare Core1121 / LR1121 support compiled and running. 

Tested basic functionality, screen needs work but is functional. Buttons support needs to be added aling with Lora TX LED status.

### DIFF
--- a/src/helpers/radiolib/CustomLR1121.h
+++ b/src/helpers/radiolib/CustomLR1121.h
@@ -1,0 +1,23 @@
+#pragma once
+
+#include <RadioLib.h>
+
+class CustomLR1121 : public LR1121 {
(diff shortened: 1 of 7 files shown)


recrof:
I will close this one, as it has list of issues:
* all PRs are merged to dev, you are merging to main
* custom-wired builds are not added to variants, as there are infinite configurations possible and every board gets built into the release
* lr1121 does have multiple radios and it needs special handling, as 2.4 radio needs limit of 20dBm and sub-ghz is limited to 22dBm

PR #1638, opened 2026-02-08
build.sh: add list and -l to list firmwares available to build

thanegill:
build.sh: add list and -l to list firmwares available to build


liamcottle:
Nice one! :)

PR #1641, opened 2026-02-09
Longer preamble test

OverkillFPV:
I have been testing a longer preamble (32) instead of (16) i have had major tx reliability improvements for strong and weak links. This wasn't a noticeable issue on wide (250khz, sf11) But narrow seems to have this issue. My theory is that narrow's higher spectral density and the lower noise floor is causing the AGC to have to make a larger jump and the preamble is not long enough for the AGC to settle and the lora chip to reliably sync the packet clock and freq.


OverkillFPV:
I think i may have done this PR wrong, it's trying to merge into main instead of dev i think

PR #1647, opened 2026-02-10
Proposal for refactoring variant definition architecture

jbrazio:
This PR outlines a proposed architectural refactoring of variant definitions within the MeshCore ecosystem. The t114 variant has been refactored as a proof-of-concept implementation. This proposal seeks to determine whether this architectural pattern should be adopted for all variant definitions. before investing more effort into it.

Current variant definitions suffer from significant maintainability challenges. The prevalent practice of copying and pasting configuration sections between variants has resulted in fragmented, inconsistent definitions.

This implementation introduces a modular environment composition system designed to maximize reusability and simplify the creation of new board targets. The refactored t114 variant demonstrates how board targets can be assembled from pre-built, reusable components rather than defined as monolithic configurations.

The core innovation is a library of standardized environment modules that can be mixed and matched to build new targets. Instead of starting from scratch or copying entire configurations, developers can now select base components, combine functionality and customize only what's necessary . This approach makes building new targets significantly easier and faster as common functionality is packaged into reusable modules that can be included with a single reference.

The modular architecture also improves consistency across targets. When a module is updated or improved, all targets using that module benefit automatically. This eliminates the need to manually propagate version upgrades or enhancements across dozens of individual variant files.

@ripplebiz @liamcottle @recrof @fdlamotte

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -87,9 +87,12 @@ build_flags = ${arduino_base.build_flags}
   -D NRF52_PLATFORM
   -D LFS_NO_ASSERT=1
   -D EXTRAFS=1
+  -I lib/nrf52/s140_nrf52_6.1.1_API/include
+  -I lib/nrf52/s140_nrf52_6.1.1_API/include/nrf52
 lib_deps =
   ${arduino_base.lib_deps}
   https://github.com/oltaco/CustomLFS @ 0.2.1
+
 ; ----------------- RP2040 ---------------------
 
 [rp2040_base]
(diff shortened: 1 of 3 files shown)


ripplebiz:
Looks good. Makes sense. 
but... someone's going to make a mistake. 100% chance of it.
Whoever commits to changing a variant will have to test every single target.

jbrazio:
I can finish the centralised modules on this PR and then migrate and validate the targets: T114, WSL3, v3, Lora32, Promicro and RAK4631 one at a time.

PR #1648, opened 2026-02-10
WifiManager to replace hardcoded ssid/password for ESP wifi companions

derekg612:
Instead of requiring hardcoded credentials, this library provides a web ui for listing nearby SSIDs and facilitates connecting to one.

https://github.com/tzapu/WiFiManager


mileshkin:
This would be sweet!

PR #1653, opened 2026-02-10
Implement remote LNA toggle CLI command

jbrazio:
Refactor of #1164.

This pull request adds support for configuring and persisting RX boosted gain settings for SX12xx series radios (SX1262, SX1268, SX1276) across the codebase. These settings are now stored in preferences, can be set and queried via CLI commands, and are applied during radio initialization.

You might hang a repeater [with a proper antenna] in a difficult to access place to find out that there is a lot of QRM and RX is deaf; most of the situations where I faced this, disabling the Boosted Gain helped a lot with the problem.

When we use small out-of-tune antennas (the ones shipping with the boards) having a boost on the RX helps, this is usually the client situation. For repeater where we usually have proper antennas it's a question of testing with it ON and OFF and check what gives you better performance.

### DIFF
--- a/src/helpers/radiolib/CustomLR1110.h
+++ b/src/helpers/radiolib/CustomLR1110.h
@@ -4,6 +4,8 @@
 #include "MeshCore.h"
 
 class CustomLR1110 : public LR1110 {
+  bool _rx_boosted = false;
+
   public:
     CustomLR1110(Module *mod) : LR1110(mod) { }
 
@@ -21,6 +23,13 @@ class CustomLR1110 : public LR1110 {
     }
     
     float getFreqMHz() const { return freqMHz; }
+
+    int16_t setRxBoostedGainMode(bool en) {
+      _rx_boosted = en;
+      return LR1110::setRxBoostedGainMode(en);
+    }
+
+    bool getRxBoostedGainMode() const { return _rx_boosted; }
 
     bool isReceiving() {
       uint16_t irq = getIrqStatus();


JDat:
Idea looks good, but... Can this be done without reconfiguring every platform? Let's say: every board/target have function to control LNA boost and... Depending on platform, LNA boost gain setting is transfered to radio chip. If board/platform/variant not support LNA Boss, it just do noting (empty function). This approach allow to write code without zillion of #define-s

jbrazio:
Yeah I know it's a lot of duplicated code but due to the current architecture the requirement of having the function prototypes on `target.cpp/h` will not go away, having them as empty pointers or adding a conditional macro.. depends on the coding style. And the current coding style is to use conditional macros.

towerviewcams:
The boost in the SX12xx board with large amounts of testing does not effect RX sensitivity or noise floor. We tested this allot with the V4 and the LNA bypassed with PCB jump wire. Then set the boost flag on and off on the SX12xx......There was no difference we could find in both quiet and noisy locations. This was for the Sx1262 in the V4.

If someone can really see a difference with this, what board are you using? Our group hasn't been able to duplicate that.

Using the term LNA might confuse people into thinking its the external 17db LNA that cannot be controlled by software. This is what happen in PR1398 at first.

jbrazio:
We had a couple of cases with WSL3 and T114 (433 band) in really good city center spots where the TX had a good coverage but RX was just deaf. Uploading a firmware with the LNA off solved the issue.

The ideia is exactly that, put the repeater in a high place and test with and without.. then choose what better works for you.

towerviewcams:
Yea, 433, different world for sure and not V4 as expected - different hardware. Cool

txkbaldlaw:
I'm looking for this in the RAK 3401/13302 1W booster.  

This would be nice, as I have a repeater in a noisy environment, and I think turning off the RX Boost is going to help.  I applied it via a compiled firmware to a more accessible repeater, and it improved connectivity.  When it put it back to regular FW with RXBoost on, connectivity became less stable.   I haven't applied it at the remote site yet, b/c it is difficult to access, but the ability to turn this on or off via a CLI switch remotely would make it easier.  

Several people are positing that on the RAK 3401/13302 boosters that deafening is occurring with the boosted RX gain set to on.

weebl2000:
Please see https://github.com/jbrazio/MeshCore/pull/1

jbrazio:
@ripplebiz Can you review this PR and provide a closure to it ?
Thanks for your time.

weebl2000:
Build this PR [here](https://mcimages.weebl.me/?commitId=remote-lna-on-dev)

weebl2000 (commented):
Thanks for the feature — being able to toggle RX boosted gain at runtime is useful. A few issues to address though.

src/helpers/CommonCLI.cpp
```diff
@@ -51,20 +51,21 @@ void CommonCLI::loadPrefsInt(FILESYSTEM* fs, const char* filename) {
...
     file.read((uint8_t *)&_prefs->sf, sizeof(_prefs->sf));                                         // 112
     file.read((uint8_t *)&_prefs->cr, sizeof(_prefs->cr));                                         // 113
     file.read((uint8_t *)&_prefs->allow_read_only, sizeof(_prefs->allow_read_only));               // 114
     file.read((uint8_t *)&_prefs->multi_acks, sizeof(_prefs->multi_acks));                         // 115
     file.read((uint8_t *)&_prefs->bw, sizeof(_prefs->bw));                                         // 116
     file.read((uint8_t *)&_prefs->agc_reset_interval, sizeof(_prefs->agc_reset_interval));         // 120
-    file.read((uint8_t *)&_prefs->path_hash_mode, sizeof(_prefs->path_hash_mode));                 // 121
-    file.read(pad, 2);                                                                             // 122
+    file.read((uint8_t *)&_prefs->sx126x_rx_boosted_gain, sizeof(_prefs->sx126x_rx_boosted_gain)); // 121
+    file.read((uint8_t *)&_prefs->path_hash_mode, sizeof(_prefs->path_hash_mode));                 // 122
```
weebl2000: This inserts `sx126x_rx_boosted_gain` at byte offset 121, shifting `path_hash_mode` from 121→122. Existing repeaters that upgrade will read their saved `path_hash_mode` as `sx126x_rx_boosted_gain`, and a pad byte (0) as `path_hash_mode` — silently corrupting it.

Fix: use the pad byte at offset 122 instead, keeping `path_hash_mode` at 121:

```suggestion
    file.read((uint8_t *)&_prefs->path_hash_mode, sizeof(_prefs->path_hash_mode));                 // 121
    file.read((uint8_t *)&_prefs->sx126x_rx_boosted_gain, sizeof(_prefs->sx126x_rx_boosted_gain)); // 122
    file.read(pad, 1); // 123 : 1 byte unused
```

(Same fix needed in `savePrefs` below.)

variants/heltec_v3/target.cpp
```diff
@@ -58,3 +58,13 @@ mesh::LocalIdentity radio_new_identity() {
   RadioNoiseListener rng(radio);
   return mesh::LocalIdentity(&rng);  // create new random identity
 }
+
+#if defined(USE_SX1262) || defined(USE_SX1268)
```
weebl2000: This identical block is copy-pasted into 30+ variant target files — that's the bulk of the 944-line diff.

`setRxBoostedGainMode()` / `getRxBoostedGainMode()` are methods on RadioLib's `SX126x` base class. Instead of free functions in every variant, add virtual methods to `RadioLibWrapper` (no-op default), override in `CustomSX1262Wrapper` / `CustomSX1268Wrapper` / `CustomLLCC68Wrapper`, and call through the existing `mesh::Radio&` reference. This eliminates all per-variant changes and the `-D USE_SX1262` additions to every platformio.ini.

src/helpers/CommonCLI.cpp
```diff
@@ -306,6 +311,10 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
         sprintf(reply, "> %s", StrHelper::ftoa(_prefs->node_lat));
       } else if (memcmp(config, "lon", 3) == 0) {
         sprintf(reply, "> %s", StrHelper::ftoa(_prefs->node_lon));
+#if defined(USE_SX1262) || defined(USE_SX1268)
```
weebl2000, line 324: The SX1262 doesn't have a discrete LNA — this controls the RX gain register (boosted vs power-saving mode). `radio.lna` is misleading and will be hard to rename once shipped. Consider `radio.rxboost` or `radio.rxgain`.

src/helpers/radiolib/CustomSX1262.h
```diff
@@ -92,4 +92,10 @@ class CustomSX1262 : public SX1262 {
       bool detected = (irq & SX126X_IRQ_HEADER_VALID) || (irq & SX126X_IRQ_PREAMBLE_DETECTED);
       return detected;
     }
+
```
weebl2000, line 94: `CustomLLCC68` is also SX126x-based and already calls `setRxBoostedGainMode()` in its `std_init()`, but this PR doesn't add `getRxBoostedGainMode()` to `CustomLLCC68.h`. Would be consistent to cover it (and moot if you centralize in the wrapper layer per the other comment).

jbrazio:
> Thanks for the feature — being able to toggle RX boosted gain at runtime is useful. A few issues to address though.

Thanks for the review.. all your suggestions are valid.
But I will not put any more effort into this PR unless it's clear if it will get merged. This PR + the original one has been dragging since Nov, despite the multiple request for comments from @ripplebiz

towerviewcams:
@weebl2000 @jbrazio @IoTThinks I have 5, of the 4.3.1 boards that have left the china warehouse Feb 27th and will be ready to perform testing once they arrive. Arrived at SFO customs this morning.

Happy to help if/when needed. Lets work together to make this happen for everyone.

terminalvelocity23:
I'm sorry, what's the CLI command for toggling it on and off? _radio.rxgain_ doesn't work, the repeater doesn't recognize it. I'm running a V4 with 1.14 with this PR merged.

upd: turns out, the USE_SX1262 wasn't defined anywhere for V4, so I've ended up manually adding it.

txkbaldlaw:
> I'm sorry, what's the CLI command for toggling it on and off? _radio.rxgain_ doesn't work, the repeater doesn't recognize it. I'm running a V4 with 1.14 with this PR merged.
> 
> upd: turns out, the USE_SX1262 wasn't defined anywhere for V4, so I've ended up manually adding it.

{get/set} radio.lna {off/on}

terminalvelocity23:
> {get/set} radio.lna {off/on}

This doesn't work. Look at the code, the command is {get/set} radio.rxgain {off/on}.

txkbaldlaw:
> > {get/set} radio.lna {off/on}
> 
> This doesn't work. Look at the code, the command is {get/set} radio.rxgain {off/on}.

Sorry, looks like it got changed somewhere along the way, and I missed it.  It was radio.lna in the earlier commits.   

Screenshot 2026-03-12 at 11 44 39 AM

ViezeVingertjes:
> upd: turns out, the USE_SX1262 wasn't defined anywhere for V4, so I've ended up manually adding it.

I think this issue isnt resolved yet, right?

liamcottle:
Is everyone happy with the state of the PR? I've reviewed the code and looks alright.

From what I can tell this will be usable on repeaters via CLI, but not on companions. The code is there for preferences in the companion firmware, but no way to actually get/set it.

That could probably be added in a follow up PR, otherwise if it's not really needed for companions or no intention to add a way to configure it for companions, we should probably not have the code using up the preference slot if it's not actually configurable.

weebl2000:
Code looks good, but I'm not sure this feature is actually needed. I think toggling of boosted gain off was mainly helping with deafness of SX1262 which was caused by the AGC not being able to properly reset.

In all my testing boosted gain ON is better as long as the AGC doesn't get stuck, which we have fixed now in 1.14.

If someone has wildly different results with the AGC properly resetting please share.

edit: actually we need to merge this https://github.com/jbrazio/MeshCore/pull/2 first - currently LR1110 won't build.

JDat:
> Code looks good, but I'm not sure this feature is actually needed. I think toggling of boosted gain off was mainly helping with deafness of SX1262 which was caused by the AGC not being able to properly reset.
> 
> In all my testing boosted gain ON is better as long as the AGC doesn't get stuck, which we have fixed now in 1.14.
> 
> If someone has wildly different results with the AGC properly resetting please share.
> 
> edit: actually we need to merge this [jbrazio#2](https://github.com/jbrazio/MeshCore/pull/2) first - currently LR1110 won't build.

Option to control settings is always better then no option. Don't want to use LNA toggle? Don't touch it! Need to set ON/OFF/Auto (thanks to PR mentioned), do it!

txkbaldlaw:
I noted significant performance enhancements in the RAK 3401 board in repeater firmware being able to turn it off in certain environments.  As mentioned by @JDat above, I'd rather have the option and not use it, rather than having to build a custom FW each time.

txkbaldlaw:
> I noted significant performance enhancements in the RAK 3401 board in repeater firmware being able to turn it off in certain environments. As mentioned by @JDat above, I'd rather have the option and not use it, rather than having to build a custom FW each time.

I rebooted to clear the neighbors list with radio.lna off, advert-ed from other repeaters, snapped a screenshot of the results.  Then, issued set radio.lna on, rebooted, and repeated the process.  Results below. 

It isn't as dramatic today, but at points, it's been a consistent and repeatable 3db shift.
 
 

Screenshot 2026-03-13 at 12 43 21 PM
Screenshot 2026-03-13 at 12 55 13 PM 1

jbrazio:
I don't understand the direction this discussion is heading. This boolean flag is just another tool in the toolbox, no matter the number of tests we do.. I prefer to give people tools that have a real impact on RF and let them decide what best fits their needs.

> From what I can tell this will be usable on repeaters via CLI, but not on companions. (..) That could probably be added in a follow up PR

I agree because this needs some coordination with the application side.

> I think this issue isnt resolved yet, right?

done.

weebl2000:
@jbrazio please have a look at merging this https://github.com/jbrazio/MeshCore/pull/2 - it fixes the build for LR1110 devices.

jbrazio:
@weebl2000 done

terminalvelocity23:
> Code looks good, but I'm not sure this feature is actually needed. I think toggling of boosted gain off was mainly helping with deafness of SX1262 which was caused by the AGC not being able to properly reset.
> 
> In all my testing boosted gain ON is better as long as the AGC doesn't get stuck, which we have fixed now in 1.14.
> 
> If someone has wildly different results with the AGC properly resetting please share.

I'm using this feature, it cuts down on the number of bad packets received somewhat. Got a lot of in-band noise from telemetry, and it's useful to have a toggle.

txkbaldlaw:
Thanks @liamcottle and thanks to all who worked on this.

towerviewcams:
> I'm using this feature, it cuts down on the number of bad packets received somewhat. Got a lot of in-band noise from telemetry, and it's useful to have a toggle.

@terminalvelocity23 
What boards do you use this on?

terminalvelocity23:
> What boards do you use this on?

V4s.

towerviewcams:
> > What boards do you use this on?
> 
> V4s.

your doing this on 4.2 boards right? not the new 4.3 where the LNA can use control logic to bypass the 17db LNA.

terminalvelocity23:
> your doing this on 4.2 boards right? not the new 4.3 where the LNA can use control logic to bypass the 17db LNA.

Yeah, 4.2 boards. Don't have any 4.3s yet.

jeroenvermeulen:
I have checked a couple of devices, and by default `get radio.rxgain` returns `off`.
Did see this both after a fresh install (erase flash) and an upgrade from 1.14.0 while keeping settings.
In the source code it looks like the goal was to have it on by default.

jbrazio:
I would argue this happens when doing an upgrade, if you use erase when flashing does it default to on ?

jeroenvermeulen:
**You are right about a wiped node + fresh install of v1.14.1**
- Took a Heltec V4
- Flashed repeater firmware v1.14.1 using the web flasher, with `Erase device` checked
- Connected to serial CLI
- All settings were wiped, default name 'Heltec Repeater' and new private key
- `get radio.rxgain` is indeed `on`

**Upgrade from v1.14.0 to v1.14.1 causes radio.rxgain to be OFF**
- Took a Heltec V4
- Flashed **previous** repeater firmware v1.14.0 using the web flasher, with `Erase device` checked
- Connected to serial CLI
- All settings were wiped, default name 'Heltec Repeater' and new private key
- `set radio 869.6179809,62.5,8,8`
- `reboot`
- Flashed repeater firmware v1.14.1 using the web flasher, with `Erase device` **not checked**
- Settings were kept.
- `get radio.rxgain` is `off` => UNEXPECTED

towerviewcams:
@jbrazio @jeroenvermeulen   Do I understand this correctly that in 1.14  that  radio.rxgain was off in that code until the .1 patch where now "IF" you do a fresh install it is now on?

jeroenvermeulen:
Did some more research.
Tested with 3 devices:
- Heltec HTIT-Tracker V1
- Heltec V4
- RAK19003+RAK4631

The results for all three devices are the same, but the problem only occurs in the repeater firmware, not the companion.

I have added the `getRxBoostedGainMode()` function to v1.14.0 to be able to see the status in the debug log without changing it. I have built the firmwares and flashed using the web flasher.

### Repeater firmware:

- Fresh 1.14.0 install: RX Boosted Gain Mode = ON (can't change)
- Fresh 1.14.1 install: RX Boosted Gain Mode = ON (default)
- Upgrade from 1.14.0 to 1.14.1 without erase: RX Boosted Gain Mode = **OFF** => UNEXPECTED

### Companion firmware:

- Fresh 1.14.0 install: RX Boosted Gain Mode = ON (can't change)
- Fresh 1.14.1 install: RX Boosted Gain Mode = ON (default, can't change in UI)
- Upgrade from 1.14.0 to 1.14.1 without erase: RX Boosted Gain Mode = ON (default, can't change in UI)

---
To repeat the problem on v1.14.0 the radio settings must have been set using `set radio 869.6179809,62.5,8,8` before the upgrade to 1.14.1.

PR #1666, opened 2026-02-11
Validate buffer length before reading fields in Packet::readFrom

weebl2000:
## Summary

`Packet::readFrom` reads the header byte, transport codes (4 bytes if present), and `path_len` from the source buffer before performing any length validation. With a short input (e.g. `len = 0`), these reads go past the end of the source buffer.

Unlike the main radio receive path (which parses packets inline in `Dispatcher::checkRecv` with proper bounds checking), `readFrom` is used by bridge interfaces (RS232, ESP-NOW) and `importContact` (flash blob storage). A corrupted blob or malformed bridge frame could trigger the over-read.

## Fix

Add upfront length checks:
- Minimum 2 bytes overall (header + path_len)
- Transport codes require 4 additional bytes
- Path bytes plus at least 1 byte of payload must fit before proceeding

## Test plan

- [ ] Bridge packet reception still works (RS232, ESP-NOW)
- [ ] Contact import/export still works
- [ ] Short/corrupt inputs return false without reading past the buffer
- [ ] Build tested on `Heltec_v3_companion_radio_ble`

---
**Build firmware:** [Build from this branch](https://mcimages.weebl.me/?commitId=fix/packet-readfrom-bounds)

### DIFF
--- a/src/Packet.cpp
+++ b/src/Packet.cpp
@@ -67,7 +67,7 @@ bool Packet::readFrom(const uint8_t src[], uint8_t len) {
   uint8_t i = 0;
   header = src[i++];
   if (hasTransportCodes()) {
-    if (i + 4 >= len) return false;  // need 4 bytes for transport codes + path_len after
+    if (i + 4 >= len) return false;  // need 4 transport bytes + the path_len byte
     memcpy(&transport_codes[0], &src[i], 2); i += 2;
     memcpy(&transport_codes[1], &src[i], 2); i += 2;
   } else {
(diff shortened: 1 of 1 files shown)


446564:
I'll start testing

446564:
at least nothing is broken in practical testing.

have you considered adding a few unit tests?

liamcottle:
> have you considered adding a few unit tests?

Once https://github.com/meshcore-dev/MeshCore/pull/925 is merged, it will make things like this easier to validate.

For example, create a new test with inputs that produce invalid/unexpected output, then fix the methods to ensure it behaves as expected.

Scott is reviewing #925 and which test framework is best to go with.

mtlynch:
I'm working on similar fixes in https://github.com/mtlynch/mesh-core/pull/3/changes

PR #1677, opened 2026-02-12
Add ChaChaPoly AEAD-4 encryption with nonce persistence

weebl2000:
**Build firmware:** [Build from this branch](https://mcimages.weebl.me/?commitId=feature/aead-4-encryption)

### Testing

- Current testling on Heltec v4 companion and Heltec v4 repeater. It's working so far. Would be great if others can try with different devices, especially more constrained devices with less ram/flash storage.

## Summary

Adds ChaCha20-Poly1305 (AEAD-4) encryption alongside the existing AES-128-ECB + HMAC-2 scheme, plus session key negotiation for Perfect Forward Secrecy. Updated nodes send AEAD-4 to peers that advertise support and fall back to ECB for legacy peers. All nodes can decode both formats. Old nodes continue to work unchanged.

Nonces are persisted to flash so they survive reboots without risk of reuse. Session keys are negotiated via ephemeral X25519 Diffie-Hellman and persisted immediately on establishment.

Relates to #259.

## What This Means in Practical Terms

The current encryption has a few weaknesses that this PR addresses:

- **Message tampering is too easy to attempt.** The existing 2-byte authentication code means an attacker only needs about 65,000 guesses to forge a valid-looking message. At LoRa speeds that's roughly 9 hours of continuous attempts. The new 4-byte tag raises this to over 4 billion guesses — at LoRa rates, that would take over a century.

- **Identical messages look identical on the air.** The current block cipher (ECB mode) produces the same ciphertext for the same plaintext, which can reveal patterns — for example, you could tell when someone sends the same message twice. The new scheme produces completely different ciphertext every time, even for identical messages.

- **Addressing fields are now protected.** Currently, only the message body is authenticated. With AEAD, the payload type and addressing hashes (which identify sender and recipient) are included in the authentication check, so an attacker cannot swap or modify them without detection. Outer routing fields like TTL and hop path are intentionally left unauthenticated so repeaters can still forward packets through the mesh.

- **Messages get slightly smaller.** ECB pads every message up to a 16-byte boundary, wasting airtime. The new scheme has no padding, so most messages shrink by a few bytes on the wire.

- **Compromise of a node doesn't reveal past messages.** Session key negotiation establishes fresh shared secrets via ephemeral key exchange. Even if a node's long-term private key is later compromised, previously recorded traffic cannot be decrypted (Perfect Forward Secrecy).

- **Nothing breaks.** Updated nodes send AEAD-4 to peers that advertise support, and fall back to ECB for legacy peers. Old nodes are completely unaffected — they never receive AEAD-4 messages because the sender checks their capability first.

- **Nodes advertise their capabilities.** Updated nodes include a flag in their advertisements saying "I understand the new encryption." When two updated nodes discover each other, they automatically start using AEAD-4 for their communication.

- **Nonces survive reboots.** Per-peer nonce counters are saved to flash periodically and before clean reboots. After a dirty reset (power loss, watchdog, brownout), nonces are bumped forward by a safety margin to guarantee no reuse.

## Wire Format

Current ECB:
```
[HMAC:2] [ECB_ciphertext:N×16]     (padded to block boundary)
```

New AEAD-4 (same position in payload):
```
[nonce:2] [ciphertext:M] [tag:4]    (exact plaintext length, no padding)
```

Average overhead: ~6 bytes (AEAD) vs ~9.5 bytes (ECB). Most messages get smaller.

## Cryptographic Design

**Per-message key derivation** (eliminates nonce-reuse catastrophe):
```
msg_key[32] = HMAC-SHA256(shared_secret, nonce || dest_hash || src_hash)
```

The `shared_secret` is either the static ECDH secret or a session key (see Session Key Negotiation below).

Including `dest_hash || src_hash` makes keys direction-dependent — Alice→Bob and Bob→Alice derive different keys even with the same nonce value (for 255/256 peer pairs; the 1/256 where dest_hash == src_hash is a residual limitation of 1-byte hashes).

**IV construction** (12 bytes, from on-wire fields):
```
iv[12] = { nonce_hi, nonce_lo, dest_hash, src_hash, 0, 0, 0, 0, 0, 0, 0, 0 }
```

**Associated data** (authenticated but not encrypted):
- Peer messages: `header || dest_hash || src_hash`
- Anonymous requests: `header || dest_hash`
- Group messages: `header || channel_hash`

Route type bits are masked out of the header in associated data (`header & ~PH_ROUTE_MASK`), since routing mode changes per hop as repeaters forward packets.

**Nonce management**: 16-bit counter per peer, persisted to flash. See "Nonce Persistence" section below.

## Session Key Negotiation (Perfect Forward Secrecy)

Session keys provide Perfect Forward Secrecy by establishing fresh shared secrets via ephemeral X25519 Diffie-Hellman. Compromise of either node's long-term private key cannot recover traffic encrypted with a session key.

### Protocol (2 messages + implicit confirmation)

```
Initiator                                   Responder
    |  1. REQ [REQ_TYPE_SESSION_KEY_INIT]        |
    |     [ephemeral_pub_A:32] (AEAD-4)          |
    | -----------------------------------------> |  derive session_key, persist, dual-decode
    |  2. RESPONSE [RESP_TYPE_SESSION_KEY_ACCEPT] |
    |     [ephemeral_pub_B:32] (static ECDH)     |
    | <----------------------------------------- |
    |  derive session_key, persist, nonce=1       |
    |  3. Any normal message (session key)       |
    | -----------------------------------------> |  confirm: drop old key
```

The INIT is encrypted with AEAD-4 (static ECDH or existing session key). The ACCEPT is always encrypted with the static ECDH secret, because the initiator hasn't derived the session key yet.

### Key Derivation

```
ephemeral_secret = X25519(their_ephemeral_pub, my_ephemeral_prv)
session_key[32]  = HMAC-SHA256(static_shared_secret, ephemeral_secret)
```

Uses existing `ed25519_key_exchange()` (X25519 Montgomery ladder) from `lib/ed25519`. No new dependencies.

### Who Initiates

- **Companion ↔ Repeater/Room/Sensor**: Companion initiates, server responds
- **Companion ↔ Companion**: Either side can initiate, both can respond

Repeaters, room servers, and sensors only implement the responder role — they never initiate session key negotiation.

### Automatic Triggers

Session key negotiation is triggered automatically based on message count. The trigger check runs inside `getEncryptionNonceFor()` — the single funnel all encrypted sends pass through — so no send path can silently skip it. Negotiation is deferred to the next `loop()` tick to avoid re-entrancy.

| Hop count | Current key | Trigger | Retry after failure |
|---|---|---|---|
| 0 (direct) | Static ECDH | Every 100 msgs | 100 msgs |
| 0 (direct) | Session key | nonce > 60000, then every 100 msgs | 100 msgs |
| 1–9 | Static ECDH | Every 500 msgs | 500 msgs |
| 1–9 | Session key | nonce > 60000, then every 300 msgs | 300 msgs |
| 10+ | Static ECDH | Every 1000 msgs | 1000 msgs |
| 10+ | Session key | nonce > 60000, then every 300 msgs | 300 msgs |

3 INIT attempts per negotiation (3-minute timeout each).

### Nonce Lifecycle

- **New contacts**: Static ECDH nonce seeded from RNG in range 1000–50000
- **Session key nonce**: Starts at 1 on establishment, full 65535 budget per session
- **Nonce exhaustion**: Fall back to static ECDH, keep retrying negotiation at tier intervals

### Encryption Key Selection

All node types use paired `getEncryptionKey()` / `getEncryptionNonce()` functions that return the correct key and nonce based on current session state:

```
has_session_key && sends_since_last_recv < 50  → AEAD with session key
has_session_key && sends_since_last_recv >= 50 → AEAD with static ECDH (stale probe)
CONTACT_FLAG_AEAD && nonce OK                  → AEAD with static ECDH
CONTACT_FLAG_AEAD && nonce exhausted           → ECB (pending renegotiation)
else                                           → ECB (legacy peer)
```

### Decode Order

```
has_session_key: session_key → prev_session_key (dual-decode) → static ECDH → ECB
CONTACT_FLAG_AEAD: static ECDH → ECB
else: ECB → static ECDH
```

### Dual-Decode Window

When the responder accepts a session key INIT, it enters DUAL_DECODE state: the new session key is active for sending, but both old and new keys are accepted for decoding. Once the initiator sends a message encrypted with the new session key (message 3), the responder confirms the transition and drops the old key.

This makes ACCEPT packet loss safe — the responder stays in dual-decode, the initiator times out and retries, and no messages are lost.

### Stale Session Detection

If a node sends 50 consecutive messages without receiving any session-key-encrypted reply, it falls back to static ECDH for sending (the peer may have lost the session key). At 100 unanswered sends, falls back to ECB. At 255, clears the AEAD capability flag and removes the session key entirely. The counter resets to 0 on any successful session-key-encrypted message from the peer.

### Session Key Persistence

Session keys use a two-tier storage model: a small RAM pool for active sessions and a larger flash-backed store for less recently used entries.

**RAM pool**: 8 slots (`MAX_SESSION_KEYS_RAM`), managed as an LRU cache. Each access touches a counter so the least-recently-used entry can be evicted when the pool is full. Entries in `INIT_SENT` state (ephemeral keys only) are never evicted — they must complete or time out.

**Flash store**: Up to 48 entries (`MAX_SESSION_KEYS_FLASH`), persisted to `/sess_keys` (companion) or `/s_sess_keys` (server firmware).

**Variable-length records**: Entries without a previous session key (no dual-decode) use 39 bytes (`SESSION_KEY_RECORD_MIN_SIZE`); entries with a previous key use 71 bytes (`SESSION_KEY_RECORD_SIZE`). The `SESSION_FLAG_PREV_VALID` flag bit distinguishes the two.

```
Without prev_key: [pub_key_prefix:4] [flags:1] [nonce:2] [session_key:32]         = 39 bytes
With prev_key:    [pub_key_prefix:4] [flags:1] [nonce:2] [session_key:32] [prev_session_key:32] = 71 bytes
```

**On-demand flash lookup**: When `findSessionKey()` misses the RAM pool, it reads the flash file to look for a matching entry. If found, the entry is loaded into RAM (evicting LRU if needed) and returned.

**Merge-save strategy**: When persisting, the code reads existing flash entries, filters out any that are already in the RAM pool or have been explicitly removed, then writes the merged result (RAM entries + surviving flash-only entries). This prevents flash from resurrecting deleted entries while preserving entries that were evicted from RAM.

**Removed-entry tracking**: When a session key is explicitly removed (e.g., invalidation after static ECDH fallback), its prefix is recorded in a small tracking array. The merge-save step skips these prefixes so the deleted entry doesn't reappear from stale flash data. The tracking array is cleared after each successful save.

## Nonce Persistence

Nonces are persisted to a dedicated file on flash (`/nonces` for companion radios, `/s_nonces` for server firmware).

**Periodic saves**: After every `NONCE_PERSIST_INTERVAL` (50) messages to a given peer, the nonce file is written. A dirty flag tracks whether any nonce has advanced since the last save.

**Clean reboot**: Software restarts and deep sleep wakes load the persisted nonces as-is. A `onBeforeReboot()` callback in CommonCLI flushes any dirty nonces before the restart.

**Dirty reboot**: Power-on, watchdog, and brownout resets are detected via `wasDirtyReset()` (platform-specific: `esp_reset_reason()` on ESP32, `RESETREAS` register on NRF52). After a dirty reset, all loaded nonces are bumped forward by `NONCE_BOOT_BUMP` (50), which is at least the persist interval, guaranteeing that even the worst-case unpersisted nonce is safely skipped. Session key nonces also receive the boot bump; if the bump causes a wrap, the nonce is forced to 65535 to trigger renegotiation.

**Format**: Simple array of `{pub_key_prefix[6], nonce[2]}` entries, matched to in-memory contacts/clients on load.

## Security Comparison

| Property | ECB + HMAC-2 (current) | AEAD-4 (new) | AEAD-4 + Session Key |
|---|---|---|---|
| Confidentiality | Identical blocks → identical ciphertext | Unique keystream per message | Same |
| Forgery resistance | 1/65K (~9 hours at LoRa rates) | 1/4.3B (~136 years) | Same |
| Key usage | 16 of 32 bytes (AES-128) | Full 32 bytes (ChaCha20-256) | Same |
| Addressing authentication | None | Payload type & address hashes via AAD | Same |
| MAC timing | `memcmp` (timing side-channel) | `secure_compare` (constant-time) | Same |
| Padding waste | 0-15 bytes per message | None | None |
| Perfect Forward Secrecy | No | No | Yes |
| Nonce reuse on reboot | N/A (no nonces) | Mitigated by persistence + boot bump | Same |

## Scope

| Payload type | AEAD-4 decode | AEAD-4 send | Session keys | Notes |
|---|---|---|---|---|
| TXT_MSG, REQ, RESPONSE, PATH | Yes | Yes (if peer advertises AEAD) | Yes | Per-peer secret, no collision risk |
| ANON_REQ | Yes | No (no prior capability exchange) | No | Ephemeral ECDH secret |
| GRP_TXT, GRP_DATA | Yes | No (see group considerations) | No | Shared channel key |

All node types (companion radio, repeater, room server, sensor) support AEAD-4 decode, AEAD-4 send, and session key negotiation (companion initiates or responds; server firmware responds only).

## Group Message Considerations

Group channels share a single key among all members. With a 2-byte nonce and multiple senders, cross-sender nonce collisions follow the birthday bound (~300 messages for 50% probability on an active channel). A collision leaks `P1 ⊕ P2` for that specific message pair via crib-dragging, but:

- **No key recovery** — per-message key derivation via HMAC-SHA256 is one-way
- **No cascade** — each collision is isolated, doesn't affect other messages
- **Bounded threat model** — the attacker must not have the channel PSK (if they do, they can already read everything)

This is mainly beneficial for public/hashtag channels where the PSK is already widely known and the ECB pattern leakage and weak MAC are a greater concern than the bounded nonce collision risk.

Potential future mitigations explored and deferred:
- **Per-sender derived keys** (`HMAC(channel_secret, sender_pub_key)`) — eliminates cross-sender collisions but requires receivers to know all senders' public keys, changing the group security model from "know the PSK = full access" to "know the PSK + sender discovery = access." Ruled out as a usability regression.
- **Expanded nonce** (4 bytes instead of 2) — pushes birthday bound to ~65,000 messages (~2 years at 100 msgs/day). Costs 2 extra bytes of airtime and creates a different wire format for groups vs peers.
- **Sender hash byte on wire** — differentiates senders for key derivation at 1 byte cost, but leaks sender identity metadata (traffic correlation, identification via adverts) that is currently hidden inside the encrypted payload.

## Decode Order

Adaptive per-peer: for peers with `CONTACT_FLAG_AEAD` set, try AEAD-4 first then ECB fallback. For unknown/legacy peers, try ECB first then AEAD-4 fallback. When a session key exists, decode order is: session key → prev session key (dual-decode window) → static ECDH → ECB. This avoids the 1/65536 ECB false-positive rate on AEAD packets (nonce bytes matching truncated HMAC) for known AEAD peers, while minimizing wasted CPU for legacy peers.

## Capability Advertisement

- `feat1` bit 0 (`FEAT1_AEAD_SUPPORT`) is set in adverts for all node types (chat, repeater, room, sensor)
- Receivers record peer capability in `ContactInfo.flags` bit 1 (`CONTACT_FLAG_AEAD`)
- Old nodes parse `feat1` but ignore the value (forward-compatible via existing `AdvertDataParser`)

### Core Library

- **`src/MeshCore.h`** — AEAD constants, session key constants (`SESSION_KEY_SIZE`, `REQ_TYPE_SESSION_KEY_INIT`, `RESP_TYPE_SESSION_KEY_ACCEPT`, `NONCE_REKEY_THRESHOLD`, `SESSION_KEY_*` thresholds and limits), two-tier pool sizing (`MAX_SESSION_KEYS_RAM=8`, `MAX_SESSION_KEYS_FLASH=48`), variable-length record sizes (`SESSION_KEY_RECORD_SIZE`, `SESSION_KEY_RECORD_MIN_SIZE`), `SESSION_FLAG_PREV_VALID`
- **`src/Utils.h` / `src/Utils.cpp`** — `aeadEncrypt()` and `aeadDecrypt()` using ChaChaPoly
- **`src/Mesh.h`** — `getPeerFlags()`, `getPeerNextAeadNonce()`, `getPeerSessionKey()`, `getPeerPrevSessionKey()`, `onSessionKeyDecryptSuccess()`, `getPeerEncryptionKey()`, `getPeerEncryptionNonce()` virtuals; `aead_nonce` param on `createDatagram`/`createPathReturn`
- **`src/Mesh.cpp`** — AEAD send path in `createDatagram`/`createPathReturn`; session key → prev session key → static ECDH → ECB adaptive decode order
- **`src/helpers/ContactInfo.h`** — `uint16_t aead_nonce` field, `nextAeadNonce()` helper
- **`src/helpers/SessionKeyPool.h`** — `SessionKeyEntry` struct and `SessionKeyPool` class (LRU-managed RAM pool with `last_used` tracking, eviction that skips `INIT_SENT` entries, removed-entry tracking for merge-save safety)

### Companion Radio (BaseChatMesh)

- **`src/helpers/BaseChatMesh.h` / `BaseChatMesh.cpp`** — Advertise AEAD, track peer capability, AEAD send for all peer message types, nonce persistence, session key negotiation (both initiator and responder roles), encryption key/nonce funnel (`getEncryptionKeyFor`/`getEncryptionNonceFor`), deferred rekey trigger via `_pending_rekey_idx`

### Server-Side (ClientACL + examples)

- **`src/helpers/ClientACL.h` / `ClientACL.cpp`** — Server-side AEAD nonce tracking and persistence, session key responder (`handleSessionKeyInit`), paired encryption key/nonce selection (`getEncryptionKey`/`getEncryptionNonce`), flash-backed session key wrappers with merge-save, peer-index forwarding helpers
- **`src/helpers/CommonCLI.h` / `CommonCLI.cpp`** — Advertise AEAD for repeaters/rooms/sensors; `onBeforeReboot()` callback for nonce/session key flush
- **`examples/simple_repeater/MyMesh.h` / `MyMesh.cpp`** — AEAD + session key support, nonce persistence, session key INIT handling in `onPeerDataRecv`
- **`examples/simple_room_server/MyMesh.h` / `MyMesh.cpp`** — Same
- **`examples/simple_sensor/SensorMesh.h` / `SensorMesh.cpp`** — Same

### Platform Support

- **`src/helpers/ArduinoHelpers.h`** — `wasDirtyReset()` helper (ESP32/NRF52 reset reason detection)
- **`examples/companion_radio/DataStore.h` / `DataStore.cpp`** — Nonce and session key file I/O, variable-length session key records, merge-save with flash-backed lookup (`loadSessionKeyByPrefix`)
- **`examples/companion_radio/MyMesh.h` / `MyMesh.cpp`** — Wire up nonce/session key persistence and reboot callback, flash-backed session key overrides (`loadSessionKeyRecordFromFlash`, `mergeAndSaveSessionKeys`)

## Build Verification

- ESP32 (`Heltec_v3_companion_radio_ble`): builds successfully
- ESP32 (`Heltec_v3_repeater`): builds successfully
- ESP32 (`Heltec_v3_room_server`): builds successfully
- NRF52 (`Xiao_nrf52_companion_radio_ble`): builds successfully

## Future Work

- Group messages: send AEAD-4 (all updated nodes can already decode it)
- ANON_REQ: remain ECB (no prior capability exchange possible)
- `rekey <peer>` CLI command for manual session key renegotiation

---
**Build firmware:** [Build from this branch](https://mcimages.weebl.me/?commitId=feature/aead-4-encryption)

### DIFF
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -156,15 +156,16 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
             uint8_t data[MAX_PACKET_PAYLOAD];
             int macAndDataLen = pkt->payload_len - i;
 
-            // Try ECB first (Phase 1: all senders use ECB), then AEAD-4 fallback.
-            // IMPORTANT: Phase 2 MUST swap to AEAD-first. ECB-first has a 1/65536
-            // false-positive rate on AEAD packets (nonce bytes matching truncated HMAC),
-            // producing garbage plaintext. AEAD-first has only 1/2^32 false-positive on
-            // ECB packets, which is negligible.
-            int len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
-            if (len <= 0) {
-              uint8_t assoc[3] = { pkt->header, dest_hash, src_hash };
+            // Try-both decode: AEAD-first for peers known to support it (avoids 1/65536
+            // ECB false-positive on AEAD packets), ECB-first for unknown/legacy peers.
+            uint8_t assoc[3] = { pkt->header, dest_hash, src_hash };
+            int len;
+            if (getPeerFlags(j) & CONTACT_FLAG_AEAD) {
               len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
+              if (len <= 0) len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
+            } else {
+              len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
+              if (len <= 0) len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
             }
             if (len > 0) {  // success!
               if (pkt->getPayloadType() == PAYLOAD_TYPE_PATH) {
@@ -183,7 +184,7 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
                 if (onPeerPathRecv(pkt, j, secret, path, path_len, extra_type, extra, extra_len)) {
                   if (pkt->isRouteFlood()) {
                     // send a reciprocal return path to sender, but send DIRECTLY!
-                    mesh::Packet* rpath = createPathReturn(&src_hash, secret, pkt->path, pkt->path_len, 0, NULL, 0);
+                    mesh::Packet* rpath = createPathReturn(&src_hash, secret, pkt->path, pkt->path_len, 0, NULL, 0, getPeerNextAeadNonce(j));
                     if (rpath) sendDirect(rpath, path, path_len, 500);
                   }
                 }
@@ -466,13 +467,13 @@ Packet* Mesh::createAdvert(const LocalIdentity& id, const uint8_t* app_data, siz
 
 #define MAX_COMBINED_PATH  (MAX_PACKET_PAYLOAD - 2 - CIPHER_BLOCK_SIZE)
 
-Packet* Mesh::createPathReturn(const Identity& dest, const uint8_t* secret, const uint8_t* path, uint8_t path_len, uint8_t extra_type, const uint8_t*extra, size_t extra_len) {
+Packet* Mesh::createPathReturn(const Identity& dest, const uint8_t* secret, const uint8_t* path, uint8_t path_len, uint8_t extra_type, const uint8_t*extra, size_t extra_len, uint16_t aead_nonce) {
   uint8_t dest_hash[PATH_HASH_SIZE];
   dest.copyHashTo(dest_hash);
-  return createPathReturn(dest_hash, secret, path, path_len, extra_type, extra, extra_len);
+  return createPathReturn(dest_hash, secret, path, path_len, extra_type, extra, extra_len, aead_nonce);
 }
 
-Packet* Mesh::createPathReturn(const uint8_t* dest_hash, const uint8_t* secret, const uint8_t* path, uint8_t path_len, uint8_t extra_type, const uint8_t*extra, size_t extra_len) {
+Packet* Mesh::createPathReturn(const uint8_t* dest_hash, const uint8_t* secret, const uint8_t* path, uint8_t path_len, uint8_t extra_type, const uint8_t*extra, size_t extra_len, uint16_t aead_nonce) {
   uint8_t path_hash_size = (path_len >> 6) + 1;
   uint8_t path_hash_count = path_len & 63;
 
@@ -504,7 +505,14 @@ Packet* Mesh::createPathReturn(const uint8_t* dest_hash, const uint8_t* secret, 
       getRNG()->random(&data[data_len], 4); data_len += 4;
     }
 
-    len += Utils::encryptThenMAC(secret, &packet->payload[len], data, data_len);
+    if (aead_nonce) {
+      uint8_t dh = packet->payload[0];
+      uint8_t sh = packet->payload[1];
+      uint8_t assoc[3] = { packet->header, dh, sh };
+      len += Utils::aeadEncrypt(secret, &packet->payload[len], data, data_len, assoc, 3, aead_nonce, dh, sh);
+    } else {
+      len += Utils::encryptThenMAC(secret, &packet->payload[len], data, data_len);
+    }
   }
 
   packet->payload_len = len;
@@ -512,9 +520,10 @@ Packet* Mesh::createPathReturn(const uint8_t* dest_hash, const uint8_t* secret, 
   return packet;
 }
 
-Packet* Mesh::createDatagram(uint8_t type, const Identity& dest, const uint8_t* secret, const uint8_t* data, size_t data_len) {
+Packet* Mesh::createDatagram(uint8_t type, const Identity& dest, const uint8_t* secret, const uint8_t* data, size_t data_len, uint16_t aead_nonce) {
   if (type == PAYLOAD_TYPE_TXT_MSG || type == PAYLOAD_TYPE_REQ || type == PAYLOAD_TYPE_RESPONSE) {
-    if (data_len + CIPHER_MAC_SIZE + CIPHER_BLOCK_SIZE-1 > MAX_PACKET_PAYLOAD) return NULL;
+    size_t max_overhead = aead_nonce ? (AEAD_NONCE_SIZE + AEAD_TAG_SIZE) : (CIPHER_MAC_SIZE + CIPHER_BLOCK_SIZE-1);
+    if (data_len + max_overhead > MAX_PACKET_PAYLOAD) return NULL;
   } else {
     return NULL;  // invalid type
   }
@@ -529,7 +538,15 @@ Packet* Mesh::createDatagram(uint8_t type, const Identity& dest, const uint8_t* 
   int len = 0;
   len += dest.copyHashTo(&packet->payload[len]);  // dest hash
   len += self_id.copyHashTo(&packet->payload[len]);  // src hash
-  len += Utils::encryptThenMAC(secret, &packet->payload[len], data, data_len);
+
+  if (aead_nonce) {
+    uint8_t dest_hash = packet->payload[0];
+    uint8_t src_hash = packet->payload[1];
+    uint8_t assoc[3] = { packet->header, dest_hash, src_hash };
+    len += Utils::aeadEncrypt(secret, &packet->payload[len], data, data_len, assoc, 3, aead_nonce, dest_hash, src_hash);
+  } else {
+    len += Utils::encryptThenMAC(secret, &packet->payload[len], data, data_len);
+  }
 
   packet->payload_len = len;
 
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -531,7 +531,8 @@ Packet* Mesh::createDatagram(uint8_t type, const Identity& dest, const uint8_t* 
   if (type == PAYLOAD_TYPE_TXT_MSG || type == PAYLOAD_TYPE_REQ || type == PAYLOAD_TYPE_RESPONSE) {
     size_t hash_prefix = PATH_HASH_SIZE * 2;  // dest_hash + src_hash
     size_t max_overhead = aead_nonce ? (AEAD_NONCE_SIZE + AEAD_TAG_SIZE) : (CIPHER_MAC_SIZE + CIPHER_BLOCK_SIZE-1);
-    if (data_len + hash_prefix + max_overhead > MAX_PACKET_PAYLOAD) return NULL;
+    size_t max_payload_data = MAX_PACKET_PAYLOAD - hash_prefix - max_overhead;
+    if (data_len > max_payload_data) return NULL;
   } else {
     return NULL;  // invalid type
   }
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -156,17 +156,46 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
             uint8_t data[MAX_PACKET_PAYLOAD];
             int macAndDataLen = pkt->payload_len - i;
 
-            // Try-both decode: AEAD-first for peers known to support it (avoids 1/65536
-            // ECB false-positive on AEAD packets), ECB-first for unknown/legacy peers.
             // Mask out route type bits — they are set after encryption and vary per hop.
             uint8_t assoc[3] = { (uint8_t)(pkt->header & ~PH_ROUTE_MASK), dest_hash, src_hash };
-            int len;
+            int len = 0;
             bool decoded_aead = false;
-            if (getPeerFlags(j) & CONTACT_FLAG_AEAD) {
+            bool decoded_session = false;
+
+            // Session key decode path: try session key(s) first if available
+            const uint8_t* sess_key = getPeerSessionKey(j);
+            if (sess_key) {
+              len = Utils::aeadDecrypt(sess_key, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
+              if (len > 0) {
+                decoded_session = true;
+                decoded_aead = true;
+              } else {
+                // Try prev_session_key (dual-decode window)
+                const uint8_t* prev_key = getPeerPrevSessionKey(j);
+                if (prev_key) {
+                  len = Utils::aeadDecrypt(prev_key, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
+                  if (len > 0) {
+                    decoded_session = true;
+                    decoded_aead = true;
+                  }
+                }
+              }
+              if (!decoded_session) {
+                // Session key failed — try static ECDH, then ECB
+                len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
+                if (len > 0) {
+                  decoded_aead = true;
+                } else {
+                  len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
+                }
+              }
+            } else if (getPeerFlags(j) & CONTACT_FLAG_AEAD) {
+              // No session key — standard AEAD-first decode for AEAD-capable peers
               len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
               if (len > 0) decoded_aead = true;
               else len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
             } else {
+              // Legacy ECB-first decode
               len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
               if (len <= 0) {
                 len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
@@ -174,7 +203,8 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
               }
             }
             if (len > 0) {  // success!
-              if (decoded_aead) onPeerAeadDetected(j);
+              if (decoded_session) onSessionKeyDecryptSuccess(j);
+              else if (decoded_aead) onPeerAeadDetected(j);
               if (pkt->getPayloadType() == PAYLOAD_TYPE_PATH) {
                 int k = 0;
                 uint8_t path_len = data[k++];
@@ -191,12 +221,12 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
                 if (onPeerPathRecv(pkt, j, secret, path, path_len, extra_type, extra, extra_len)) {
                   if (pkt->isRouteFlood()) {
                     // send a reciprocal return path to sender, but send DIRECTLY!
-                    mesh::Packet* rpath = createPathReturn(&src_hash, secret, pkt->path, pkt->path_len, 0, NULL, 0, getPeerNextAeadNonce(j));
+                    mesh::Packet* rpath = createPathReturn(&src_hash, getPeerEncryptionKey(j, secret), pkt->path, pkt->path_len, 0, NULL, 0, getPeerEncryptionNonce(j));
                     if (rpath) sendDirect(rpath, path, path_len, 500);
                   }
                 }
               } else {
-                onPeerDataRecv(pkt, pkt->getPayloadType(), j, secret, data, len);
+                onPeerDataRecv(pkt, pkt->getPayloadType(), j, getPeerEncryptionKey(j, secret), data, len);
               }
               found = true;
               break;
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -502,8 +502,6 @@ Packet* Mesh::createAdvert(const LocalIdentity& id, const uint8_t* app_data, siz
   return packet;
 }
 
-#define MAX_COMBINED_PATH  (MAX_PACKET_PAYLOAD - 2 - CIPHER_BLOCK_SIZE)
-
 Packet* Mesh::createPathReturn(const Identity& dest, const uint8_t* secret, const uint8_t* path, uint8_t path_len, uint8_t extra_type, const uint8_t*extra, size_t extra_len, uint16_t aead_nonce) {
   uint8_t dest_hash[PATH_HASH_SIZE];
   dest.copyHashTo(dest_hash);
@@ -514,7 +512,9 @@ Packet* Mesh::createPathReturn(const uint8_t* dest_hash, const uint8_t* secret, 
   uint8_t path_hash_size = (path_len >> 6) + 1;
   uint8_t path_hash_count = path_len & 63;
 
-  if (path_hash_count*path_hash_size + extra_len + 5 > MAX_COMBINED_PATH) return NULL;  // too long!!
+  size_t max_overhead = aead_nonce ? (AEAD_NONCE_SIZE + AEAD_TAG_SIZE) : (CIPHER_MAC_SIZE + CIPHER_BLOCK_SIZE-1);
+  size_t max_combined_path = MAX_PACKET_PAYLOAD - PATH_HASH_SIZE * 2 - max_overhead;
+  if (path_hash_count*path_hash_size + extra_len + 5 > max_combined_path) return NULL;  // too long!!
 
   Packet* packet = obtainNewPacket();
   if (packet == NULL) {
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -158,16 +158,23 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
 
             // Try-both decode: AEAD-first for peers known to support it (avoids 1/65536
             // ECB false-positive on AEAD packets), ECB-first for unknown/legacy peers.
-            uint8_t assoc[3] = { pkt->header, dest_hash, src_hash };
+            // Mask out route type bits — they are set after encryption and vary per hop.
+            uint8_t assoc[3] = { (uint8_t)(pkt->header & ~PH_ROUTE_MASK), dest_hash, src_hash };
             int len;
+            bool decoded_aead = false;
             if (getPeerFlags(j) & CONTACT_FLAG_AEAD) {
               len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
-              if (len <= 0) len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
+              if (len > 0) decoded_aead = true;
+              else len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
             } else {
               len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
-              if (len <= 0) len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
+              if (len <= 0) {
+                len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
+                if (len > 0) decoded_aead = true;
+              }
             }
             if (len > 0) {  // success!
+              if (decoded_aead) onPeerAeadDetected(j);
               if (pkt->getPayloadType() == PAYLOAD_TYPE_PATH) {
                 int k = 0;
                 uint8_t path_len = data[k++];
@@ -228,7 +235,7 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
           // Phase 2 MUST swap to AEAD-first (see peer message comment above).
           int len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
           if (len <= 0) {
-            uint8_t assoc[2] = { pkt->header, dest_hash };
+            uint8_t assoc[2] = { (uint8_t)(pkt->header & ~PH_ROUTE_MASK), dest_hash };
             len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 2, dest_hash, 0);
           }
           if (len > 0) {  // success!
@@ -265,7 +272,7 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
           // worthwhile for public/hashtag channels where the PSK is already widely known.
           int len = Utils::MACThenDecrypt(channels[j].secret, data, macAndData, macAndDataLen);
           if (len <= 0) {
-            uint8_t assoc[2] = { pkt->header, channel_hash };
+            uint8_t assoc[2] = { (uint8_t)(pkt->header & ~PH_ROUTE_MASK), channel_hash };
             len = Utils::aeadDecrypt(channels[j].secret, data, macAndData, macAndDataLen, assoc, 2, channel_hash, 0);
           }
           if (len > 0) {  // success!
@@ -508,7 +515,7 @@ Packet* Mesh::createPathReturn(const uint8_t* dest_hash, const uint8_t* secret, 
     if (aead_nonce) {
       uint8_t dh = packet->payload[0];
       uint8_t sh = packet->payload[1];
-      uint8_t assoc[3] = { packet->header, dh, sh };
+      uint8_t assoc[3] = { (uint8_t)(packet->header & ~PH_ROUTE_MASK), dh, sh };
       len += Utils::aeadEncrypt(secret, &packet->payload[len], data, data_len, assoc, 3, aead_nonce, dh, sh);
     } else {
       len += Utils::encryptThenMAC(secret, &packet->payload[len], data, data_len);
@@ -543,7 +550,7 @@ Packet* Mesh::createDatagram(uint8_t type, const Identity& dest, const uint8_t* 
   if (aead_nonce) {
     uint8_t dest_hash = packet->payload[0];
     uint8_t src_hash = packet->payload[1];
-    uint8_t assoc[3] = { packet->header, dest_hash, src_hash };
+    uint8_t assoc[3] = { (uint8_t)(packet->header & ~PH_ROUTE_MASK), dest_hash, src_hash };
     len += Utils::aeadEncrypt(secret, &packet->payload[len], data, data_len, assoc, 3, aead_nonce, dest_hash, src_hash);
   } else {
     len += Utils::encryptThenMAC(secret, &packet->payload[len], data, data_len);
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -153,9 +153,19 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
             uint8_t secret[PUB_KEY_SIZE];
             getPeerSharedSecret(secret, j);
 
-            // decrypt, checking MAC is valid
             uint8_t data[MAX_PACKET_PAYLOAD];
-            int len = Utils::MACThenDecrypt(secret, data, macAndData, pkt->payload_len - i);
+            int macAndDataLen = pkt->payload_len - i;
+
+            // Try ECB first (Phase 1: all senders use ECB), then AEAD-4 fallback.
+            // IMPORTANT: Phase 2 MUST swap to AEAD-first. ECB-first has a 1/65536
+            // false-positive rate on AEAD packets (nonce bytes matching truncated HMAC),
+            // producing garbage plaintext. AEAD-first has only 1/2^32 false-positive on
+            // ECB packets, which is negligible.
+            int len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
+            if (len <= 0) {
+              uint8_t assoc[3] = { pkt->header, dest_hash, src_hash };
+              len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
+            }
             if (len > 0) {  // success!
               if (pkt->getPayloadType() == PAYLOAD_TYPE_PATH) {
                 int k = 0;
@@ -210,9 +220,16 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
           uint8_t secret[PUB_KEY_SIZE];
           self_id.calcSharedSecret(secret, sender);
 
-          // decrypt, checking MAC is valid
           uint8_t data[MAX_PACKET_PAYLOAD];
-          int len = Utils::MACThenDecrypt(secret, data, macAndData, pkt->payload_len - i);
+          int macAndDataLen = pkt->payload_len - i;
+
+          // Try ECB first (Phase 1), then AEAD-4 fallback.
+          // Phase 2 MUST swap to AEAD-first (see peer message comment above).
+          int len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
+          if (len <= 0) {
+            uint8_t assoc[2] = { pkt->header, dest_hash };
+            len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 2, dest_hash, 0);
+          }
           if (len > 0) {  // success!
             onAnonDataRecv(pkt, secret, sender, data, len);
             pkt->markDoNotRetransmit();
@@ -237,9 +254,19 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
         int num = searchChannelsByHash(&channel_hash, channels, 4);
         // for each matching channel, try to decrypt data
         for (int j = 0; j < num; j++) {
-          // decrypt, checking MAC is valid
           uint8_t data[MAX_PACKET_PAYLOAD];
-          int len = Utils::MACThenDecrypt(channels[j].secret, data, macAndData, pkt->payload_len - i);
+          int macAndDataLen = pkt->payload_len - i;
+
+          // Try ECB first (Phase 1), then AEAD-4 fallback.
+          // Phase 2 MUST swap to AEAD-first (see peer message comment above).
+          // Note: group channels share a key, so nonce collisions across senders can leak
+          // P1 XOR P2 for colliding message pairs (no key recovery). Bounded risk, mainly
+          // worthwhile for public/hashtag channels where the PSK is already widely known.
+          int len = Utils::MACThenDecrypt(channels[j].secret, data, macAndData, macAndDataLen);
+          if (len <= 0) {
+            uint8_t assoc[2] = { pkt->header, channel_hash };
+            len = Utils::aeadDecrypt(channels[j].secret, data, macAndData, macAndDataLen, assoc, 2, channel_hash, 0);
+          }
           if (len > 0) {  // success!
             onGroupDataRecv(pkt, pkt->getPayloadType(), channels[j], data, len);
             break;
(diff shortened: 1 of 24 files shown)


jimdigriz:
> Per-message key derivation (eliminates nonce-reuse catastrophe):
>
> `msg_key[32] = HMAC-SHA256(shared_secret, nonce || dest_hash || src_hash)`

I do not understand how this prevents nonce re-use. After 65k messages from A->B the nonce looks like it will be reused.

I do not understand why concatenation with src/dst would change this.

The concatenation means you are partitioning the nonce value per (uni-directional) flow, in effect running different counters for A->B, B->A and C->A. Right?

> Nonce management: 16-bit counter per peer, seeded from hardware RNG on boot and on contact load. Not persisted to flash — always fresh on each boot cycle.

What happens for devices without access to a good early boot entropy source?

What if two different reboots generate the same nonce?

What happens for A->B if:
 * reboot initialises nonce=20
 * 3 messages are sent from A->B
 * reboot initialises nonce=15
 * 10 messages are sent from A->B

What does this method improve over a plain incremental counter?

Why not persist the nonce once every 100 messages, and on reboot increment by 200 (rounded down to nearest 100)? When the nonce wraps, regenerate the key.

weebl2000:
Yeah, it doesn't stop nonce re-use. I think in the end we might need more bytes for nonces.

jimdigriz:
> But in the end maybe we need more bytes for nonces.

You do not, you can also change the key.

Just negotiate a dedicated key for this. It is a *lot* easier to understand and make safe.

It would require a round trip but then only need to be done every 65k messages; you could then also share that key for both directions (ie. A->B and B->A).

Then when `nonce=0` negotiate a new key, which allows you to pick if you want to persist the nonce or reset to zero on boot.

weebl2000:
> It would require a round trip but then only need to be done every 65k messages; you could then also share that key for both directions (ie. A->B and B->A).

Might be a good option. But the protocol will become a bit more complex and brittle. Then again, we can always fallback to ECB if nothing was negotiated.

src/Mesh.cpp
```diff
@@ -467,17 +502,26 @@ Packet* Mesh::createPathReturn(const uint8_t* dest_hash, const uint8_t* secret,
...
   return packet;
 }
 
-Packet* Mesh::createDatagram(uint8_t type, const Identity& dest, const uint8_t* secret, const uint8_t* data, size_t data_len) {
+Packet* Mesh::createDatagram(uint8_t type, const Identity& dest, const uint8_t* secret, const uint8_t* data, size_t data_len, uint16_t aead_nonce) {
   if (type == PAYLOAD_TYPE_TXT_MSG || type == PAYLOAD_TYPE_REQ || type == PAYLOAD_TYPE_RESPONSE) {
-    if (data_len + CIPHER_MAC_SIZE + CIPHER_BLOCK_SIZE-1 > MAX_PACKET_PAYLOAD) return NULL;
+    size_t hash_prefix = PATH_HASH_SIZE * 2;  // dest_hash + src_hash
+    size_t max_overhead = aead_nonce ? (AEAD_NONCE_SIZE + AEAD_TAG_SIZE) : (CIPHER_MAC_SIZE + CIPHER_BLOCK_SIZE-1);
+    if (data_len + hash_prefix + max_overhead > MAX_PACKET_PAYLOAD) return NULL;
```
jcjones: Since we're working in unsigned and `data_len` is controlled by the caller, lengths close to max will cause this check's additions to wrap and potentially bypass the check.

Consider something like calculating:
```cpp
size_t max_payload_data = MAX_PACKET_PAYLOAD - hash_prefix - max_overhead;                                                                                                                                      
if (data_len > max_payload_data) return NULL; 
```

src/Utils.cpp
```diff
@@ -87,6 +88,120 @@ int Utils::MACThenDecrypt(const uint8_t* shared_secret, uint8_t* dest, const uin
...
+ * collisions are possible (~300 msgs for 50% chance with random nonces).
+ * Damage is bounded (message pair leak, no key recovery).
+ */
+int Utils::aeadEncrypt(const uint8_t* shared_secret,
+                       uint8_t* dest,
+                       const uint8_t* src, int src_len,
+                       const uint8_t* assoc_data, int assoc_len,
+                       uint16_t nonce_counter,
+                       uint8_t dest_hash, uint8_t src_hash) {
+  if (src_len <= 0) return 0;
```
jcjones: Consider a bit more defense here?
```suggestion
  if (src_len <= 0 || src_len > MAX_PACKET_PAYLOAD) return 0;
  if (assoc_len < 0 || assoc_len > MAX_PACKET_PAYLOAD) return 0;
```

src/Utils.cpp
```diff
@@ -87,6 +88,120 @@ int Utils::MACThenDecrypt(const uint8_t* shared_secret, uint8_t* dest, const uin
...
+  ChaChaPoly cipher;
+  cipher.setKey(msg_key, 32);
+  cipher.setIV(iv, 12);
+  cipher.addAuthData(assoc_data, assoc_len);
+  cipher.decrypt(dest, src + AEAD_NONCE_SIZE, ct_len);
+
+  bool valid = cipher.checkTag(src + AEAD_NONCE_SIZE + ct_len, AEAD_TAG_SIZE);
+  cipher.clear();
+  memset(msg_key, 0, 32);
+  if (!valid) memset(dest, 0, ct_len);
```
jcjones, line 288: The ECB path doesn't do this. Maybe it should?

src/helpers/ContactInfo.h
```diff
@@ -15,6 +15,17 @@ struct ContactInfo {
   uint32_t lastmod;  // by OUR clock
   int32_t gps_lat, gps_lon;    // 6 dec places
   uint32_t sync_since;
+  mutable uint16_t aead_nonce;  // per-peer AEAD nonce counter for DMs (not used for group messages), seeded from HW RNG
+
+  // Returns next AEAD nonce (post-increment) if peer supports AEAD, 0 otherwise.
+  // When 0, callers use ECB encryption.
+  uint16_t nextAeadNonce() const {
+    if (flags & CONTACT_FLAG_AEAD) {
+      if (++aead_nonce == 0) ++aead_nonce;  // skip 0 (sentinel for ECB)
```
jcjones: At least for development, is it possible to log when we hit a nonce wraparound? We wouldn't know when we hit reuse, I suppose, but ... hmm. Thoughts.

jcjones (commented):
Not a casual review, but I like the design, and the directionality of the KDF. Good doc comments, too.

weebl2000:
Thanks for all the comments so far. I will look into them. Just tested this branch with a Heltec v4 repeater and Heltec v4 companion client, and I can confirm communicating between them works using AEAD-4.

It's a request for status from the repeater and the repeater response is understood correctly by the client.

# AEAD-4 Packet Decode Verification

## Wire Format

```
[header:1] [path_len:1] [path:N] [dest_hash:1] [src_hash:1] [nonce:2] [ciphertext:M] [tag:4]
```

## Sent Packet — REQ (23 bytes)

Raw: `0200DD130E659F0B0C02D86AC2508DF6B7B3B671F6638A`

| Field | Hex | Value |
|---|---|---|
| Header | `02` | Route=DIRECT(2), Type=REQ(0), Ver=0 |
| Path length | `00` | 0 (no path) |
| dest_hash | `DD` | Destination peer |
| src_hash | `13` | Source |
| AEAD nonce | `0E 65` | 3685 |
| Ciphertext | `9F 0B 0C 02 D8 6A C2 50 8D F6 B7 B3 B6` | 13 bytes plaintext |
| Tag | `71 F6 63 8A` | Poly1305 (truncated to 4 bytes) |

Format confirmed AEAD-4: 17 bytes after hashes is not a multiple of 16, ruling out legacy ECB.

## Received Packet — RESPONSE (70 bytes)

Raw: `060013DD830B84757DB841545969BA39A62BDD0D6AD9E2CD70B25208219F964F51E8AFB0E800130BBAFC23C9C0712B7E28CE72DE17508E30A3359222A2A7DD4B2375E5AE33AC`

| Field | Hex | Value |
|---|---|---|
| Header | `06` | Route=DIRECT(2), Type=RESPONSE(1), Ver=0 |
| Path length | `00` | 0 (no path) |
| dest_hash | `13` | Receiver |
| src_hash | `DD` | Responding peer |
| AEAD nonce | `83 0B` | 33547 |
| Ciphertext | `84 75 ... 4B 23 75` | 60 bytes plaintext |
| Tag | `E5 AE 33 AC` | Poly1305 (truncated to 4 bytes) |

Note: legacy ECB is structurally possible here (64 bytes is a multiple of 16), but context confirms AEAD-4.

## Associated Data

Per the route-mask fix, assoc data masks out route type bits:

| Packet | assoc bytes |
|---|---|
| REQ | `{0x00, 0xDD, 0x13}` — `(0x02 & ~0x03)=0x00`, dest, src |
| RESPONSE | `{0x04, 0x13, 0xDD}` — `(0x06 & ~0x03)=0x04`, dest, src |

## Observations

- Both packets use AEAD-4 wire format: `[nonce:2] [ciphertext:N] [tag:4]`
- dest/src hashes (`0xDD`, `0x13`) correctly swapped between REQ and RESPONSE
- Both routed DIRECT with empty path (single hop, no relaying)
- Nonce values (3685, 33547) are non-zero, consistent with independent per-peer counters seeded from HW RNG

src/Mesh.cpp
weebl2000: Added!

src/helpers/ContactInfo.h
weebl2000: Yes!

src/Utils.cpp
weebl2000: 🛡️ added.

src/Utils.cpp
weebl2000: I don't think we need it. ECB path currently:

1. computes HMAC
2. compares it.
3. if it's a match -> decrypt is called which writes to `dest`. On mismatch `dest` isn't touched, return 0 immediately.

AEAD path  is decrypt-then-verify, we write plaintext to `dest`, then check the tag. On failure we zero `dest` because it contains unauthenticated plaintext.

(I'm not a die-hard crypto person please correct me if this doesn't make sense)

weebl2000: Added small comment to the code to clarify.
image

ignisf:
@weebl2000 thank you for your contributions. Have you considered jumping straight to something proven like the double ratchet instead? Used in Signal.

weebl2000:
> @weebl2000 thank you for your contributions. Have you considered jumping straight to something proven like the double ratchet instead? Used in Signal.

I don't think double ratchet is practical, we would need to send a new key every message and rely on strict ordering of messages. With LoRa packet limits and out-of-order delivery it would be a disaster.

I'm working on session key negotiation though. That will fix the nonce problem, but requires an exchange first.

src/Mesh.cpp
```diff
@@ -151,10 +151,58 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
...
+                  decoded_aead = true;
+                } else {
+                  len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
+                }
+              }
+            } else if (getPeerFlags(j) & CONTACT_FLAG_AEAD) {
+              // No session key — standard AEAD-first decode for AEAD-capable peers
+              len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
+              if (len > 0) decoded_aead = true;
+              else len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
```
3dpgg, line 196: If a peer indicates support for AEAD, and we fail to decrypt using AEAD, do we really need to fall back to trying ECB? I'm struggling to imagine why such a peer would support AEAD but intentionally want to use ECB, given its problems. In such a case, surely the peer would instead decline to set the AEAD flag in the first place.

src/Mesh.cpp
```diff
@@ -151,10 +151,58 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
...
+                  }
+                }
+              }
+              if (!decoded_session) {
+                // Session key failed — try static ECDH, then ECB
+                len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
+                if (len > 0) {
+                  decoded_aead = true;
+                } else {
+                  len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
```
3dpgg, line 189: In this scenario, we have a session key but it failed to decrypt, and the previous session key also failed to decode. And then we also failed to decode using AEAD on the long-term secret. And now here, we fall back to ECB.

But if we had a session key, presumably that means we were already talking to a peer that supported AEAD. Why would such an AEAD-capable peer be falling back to ECB? If we don't fall-back to ECB, do we lose anything?

I ask this from two perspectives: removing unnecessary computation for corrupted packets, and tightening up what inputs are permitted from AEAD_capable peers.

src/Mesh.cpp
3dpgg: Can we remove the comment about "4 matches", since I think the max is actually 8. It confused me when trying to understand the context.

src/helpers/BaseChatMesh.cpp
```diff
@@ -875,4 +1005,375 @@ void BaseChatMesh::loop() {
...
+  if (i >= 0 && i < num_contacts) {
+    auto entry = findSessionKey(contacts[i].id.pub_key);
+    if (entry) {
+      bool changed = (entry->state == SESSION_STATE_DUAL_DECODE);
+      if (changed) {
+        memset(entry->prev_session_key, 0, SESSION_KEY_SIZE);
+        entry->state = SESSION_STATE_ACTIVE;
+      }
+      entry->sends_since_last_recv = 0;
+      if (changed) onSessionKeysUpdated();
```
3dpgg: Not a logic issue, but could this call to `onSessionKeysUpdated()` been placed within the `if (changed)` above? Was there something else that used to be here?

src/Mesh.cpp
```diff
@@ -151,10 +151,58 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
...
+                }
+              }
+            } else if (getPeerFlags(j) & CONTACT_FLAG_AEAD) {
+              // No session key — standard AEAD-first decode for AEAD-capable peers
+              len = Utils::aeadDecrypt(secret, data, macAndData, macAndDataLen, assoc, 3, dest_hash, src_hash);
+              if (len > 0) decoded_aead = true;
+              else len = Utils::MACThenDecrypt(secret, data, macAndData, macAndDataLen);
+            } else {
+              // Legacy ECB-first decode
```
3dpgg, line 198: It would be ideal if there was a flag or overrideable method that controls whether this peer will ever use the legacy ECB-first decoding method. That way, peers can decide to refuse ECB outright.

src/Mesh.cpp
```diff
@@ -201,9 +249,16 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
           uint8_t secret[PUB_KEY_SIZE];
           self_id.calcSharedSecret(secret, sender);
 
-          // decrypt, checking MAC is valid
           uint8_t data[MAX_PACKET_PAYLOAD];
-          int len = Utils::MACThenDecrypt(secret, data, macAndData, pkt->payload_len - i);
+          int macAndDataLen = pkt->payload_len - i;
+
+          // Try ECB first (Phase 1), then AEAD-4 fallback.
```
3dpgg, line 264: For this phase 1, could it have been AEAD first and then ECB fallback? If there's a reason for ECB first, I don't immediately understand it yet. May be worth filling in a comment that the order does/doesn't matter.

3dpgg (commented):
Thanks for putting together a PR to address AES-ECB! I had a cursory look, so not all files yet.

src/Mesh.cpp
weebl2000: I put it in for cases where peers flash old firmware and no longer support newer encryption, but haven't advertised yet. In future this fallback might be removed.

src/Mesh.cpp
weebl2000: Same as above, mainly useful during transition phase where clients flash older firmware.

src/Mesh.cpp
weebl2000: I was assuming in the beginning most clients won't support AEAD yet, so we try old encryption first. Can probably be reverse order when most clients support it.

src/Mesh.cpp
weebl2000: I think it'll be safer to add this later. Definitely want to have this tested in the field widely before allowing it to be disabled.

src/Mesh.cpp
weebl2000: It was already there in dev, but I can change it.

src/helpers/ContactInfo.h
```diff
@@ -17,6 +17,24 @@ struct ContactInfo {
...
+  uint16_t nextAeadNonce() const {
+    if (flags & CONTACT_FLAG_AEAD) {
+      if (++aead_nonce == 0) {
+        ++aead_nonce;  // skip 0 (sentinel for ECB)
+        MESH_DEBUG_PRINTLN("AEAD nonce wrapped for peer: %s", name);
+      }
+      if (aead_nonce < NONCE_INITIAL_MIN) {
+        aead_nonce = 1;  // stay stuck in exhaustion zone, always return ECB
+        return 0;
+      }
```
jcjones, line 33: When the nonce wraps from 65535 => 0 => 1, this check happens on every subsequent call. Since 1 < `NONCE_INITIAL_MIN`, it repeatedly resets `aead_nonce = 1` and returns `0`, permanently locking that peer to ECB encryption.

src/MeshCore.h
```diff
@@ -9,13 +9,42 @@
...
 
+// AEAD-4 (ChaChaPoly) encryption
+#define AEAD_TAG_SIZE        4
+#define AEAD_NONCE_SIZE      2
+#define CONTACT_FLAG_AEAD    0x02   // bit 1 of ContactInfo.flags (bit 0 = favourite)
+#define FEAT1_AEAD_SUPPORT   0x0001 // bit 0 of feat1 uint16_t
+
+// AEAD nonce persistence
+#define NONCE_PERSIST_INTERVAL  50   // persist every N messages per peer
+#define NONCE_BOOT_BUMP         50   // add this on load after dirty boot (must be >= PERSIST_INTERVAL)
```
jcjones, line 29: Note this doesn't match your PR description in 'Nonce Persistence':
> After a dirty reset, all loaded nonces are bumped forward by NONCE_BOOT_BUMP (100), which is at least 2× the persist interval

but I agree that 50 seems OK!

src/helpers/BaseChatMesh.cpp
```diff
@@ -875,4 +1005,381 @@ void BaseChatMesh::loop() {
...
+
+const uint8_t* BaseChatMesh::getEncryptionKeyFor(const ContactInfo& contact) {
+  auto entry = findSessionKey(contact.id.pub_key);
+  if (canUseSessionKey(entry)) {
+    return entry->session_key;
+  }
+  return contact.getSharedSecret(self_id);
+}
+
+uint16_t BaseChatMesh::getEncryptionNonceFor(const ContactInfo& contact) {
```
jcjones, line 1202: I don't think having two separate methods here take `ContactInfo` as an arg  is a good idea, because this is a race waiting to happen. Both of these call `findSessionKey` but there's nothing ensuring that they get the same result, it certainly looks like the key could change in between the calls.

In reality how often do keys change? Well, it's going to be a real fun bug during a rollover event. And since this hardware isn't fast, I think this is too risky to do this way. 

IMO these two methods should become one that returns both the key and the nonce.

src/helpers/BaseChatMesh.cpp
```diff
@@ -875,4 +1005,381 @@ void BaseChatMesh::loop() {
...
+    SHA256 sha;
+    sha.resetHMAC(static_secret, PUB_KEY_SIZE);
+    sha.update(ephemeral_secret, PUB_KEY_SIZE);
+    sha.finalizeHMAC(static_secret, PUB_KEY_SIZE, new_session_key, SESSION_KEY_SIZE);
+  }
+  memset(ephemeral_secret, 0, PUB_KEY_SIZE);
+
+  // 4. Store in pool (dual-decode: new key active, old key still valid)
+  auto entry = allocateSessionKey(from.id.pub_key);
+  if (!entry) return 0;
```
jcjones: This early failout skips the zeroization on `:1266`.

jcjones:
```suggestion
  if (!entry) {
    memset(new_session_key, 0, SESSION_KEY_SIZE);
    return 0;
  }
```

src/helpers/ClientACL.cpp
```diff
@@ -141,3 +235,313 @@ bool ClientACL::applyPermissions(const mesh::LocalIdentity& self_id, const uint8
...
+    SHA256 sha;
+    sha.resetHMAC(client->shared_secret, PUB_KEY_SIZE);
+    sha.update(ephemeral_secret, PUB_KEY_SIZE);
+    sha.finalizeHMAC(client->shared_secret, PUB_KEY_SIZE, new_session_key, SESSION_KEY_SIZE);
+  }
+  memset(ephemeral_secret, 0, PUB_KEY_SIZE);
+
+  // 4. Store in pool (dual-decode: new key active, old key still valid)
+  auto entry = allocateSessionKey(client->id.pub_key);
+  if (!entry) return 0;
```
jcjones: Like above in BaseChatMesh.cpp
```suggestion
  if (!entry) {
    memset(new_session_key, 0, SESSION_KEY_SIZE);
    return 0;
  }
```

src/helpers/BaseChatMesh.cpp
```diff
@@ -875,4 +1005,381 @@ void BaseChatMesh::loop() {
...
+    return entry->session_key;
+  }
+  return contact.getSharedSecret(self_id);
+}
+
+uint16_t BaseChatMesh::getEncryptionNonceFor(const ContactInfo& contact) {
+  uint16_t nonce = 0;
+  auto entry = findSessionKey(contact.id.pub_key);
+  if (canUseSessionKey(entry)) {
+    ++entry->nonce;
```
jcjones: This can push the nonce right up against the rollover/fallback logic... it seems to work correctly, but it might be a good idea to write a quite comment that it's, uh, sensitive to get it right, for posterity that touches this in the future.

src/Mesh.cpp
```diff
@@ -425,13 +490,13 @@ Packet* Mesh::createAdvert(const LocalIdentity& id, const uint8_t* app_data, siz
 
 #define MAX_COMBINED_PATH  (MAX_PACKET_PAYLOAD - 2 - CIPHER_BLOCK_SIZE)
```
jcjones: 184 - 2 - 16 = 166 (for ECB)
but using ChaChaPoly here, the overhead is only 6 bytes, not 18... This is inconsistent with the new `createDatagram` method below where you're mode-aware and calculate the exact overhead.

Now, at the path lengths we're working with, this is kinda immaterial, but ISTM a path packet could carry ~10 more bytes than this check allows, when in AEAD mode.

But I think the real argument is to match this calculation up with the `createDatagram` somehow, perhaps by abstracting out the mode-aware calculation to its own method?

src/MeshCore.h
```diff
@@ -9,13 +9,42 @@
...
+
+// Session key negotiation (Phase 2)
+#define REQ_TYPE_SESSION_KEY_INIT   0x08
+#define RESP_TYPE_SESSION_KEY_ACCEPT 0x08  // response type byte in PAYLOAD_TYPE_RESPONSE
```
jcjones, line 33: Nit: While they're in different payload types, sharing the literal `0x08` is IMO confusing and can cause false a non-session-key response to be misinterpreted as a session key acceptance... if we have the bits to spare, maybe make them be different?

src/helpers/BaseChatMesh.cpp
```diff
@@ -774,12 +900,16 @@ bool BaseChatMesh::removeContact(ContactInfo& contact) {
   }
   if (idx >= num_contacts) return false;   // not found
 
-  // remove from contacts array
+  removeSessionKey(contact.id.pub_key);  // also remove session key if any
```
jcjones, line 1024: I think we have to touch `_pending_rekey_idx` here, too, otherwise the `contacts` array changes size but the rekey doesn't.

jcjones (commented):
Probably worth noting that it looks like if a peer goes from supporting AEADs to _not_, then the sender needs to forget the peer and re-discover it.

Looking real good here!

src/helpers/ContactInfo.h
weebl2000: Yeah, I did this intentionally. I reserved nonces 1-999, because ideally it shouldn't happen that a peer rolls over 65535 without renegotiation of a session key. The only thing missing is trying to renegotiate a session key, but I think it's a theoretical case currently.

Most realistic scenario for this happening is that our contact no longer supports AEAD-4, so they only support ECB anyway. If they support AEAD-4 in future, I would expect them to trigger a session key initiation.

If someone comes up with a good solution in the future nonces 1-999 are still safe.

src/MeshCore.h
weebl2000: Updated PR description. Yeah, 50 is right at the edge but I think it's safe enough. The only risk is sending 49 messages then your 50th message somehow gets sent but isn't persisted.

src/helpers/BaseChatMesh.cpp
weebl2000: I think your concern is legitimate but the actual risk is close to zero on this single-threaded system. It touches a lot of call sites and virtual wrappers. Addressing this isn't trivial. I was thinking about just using session keys only and doing away with static ECDH - but the code overhead to keep ECDH isn't that large either.

Deciscions, decisions. Thoughts?

src/helpers/BaseChatMesh.cpp
weebl2000: Added

src/Mesh.cpp
weebl2000: Going to inline it for now.

src/MeshCore.h
weebl2000: Let's make the accept become 0x09 if maintainers decide this PR is worth merging. I don't want to break dev builds in the meantime. We can break them if this PR is selected for merging.

I haven't heard much from maintainers about this PR so far.

src/helpers/BaseChatMesh.cpp
weebl2000: Good catch, adding it.

weebl2000:
> Probably worth noting that it looks like if a peer goes from supporting AEADs to _not_, then the sender needs to forget the peer and re-discover it.
> 
> Looking real good here!

Thanks for your review once more. Addressed your remarks. And yes, if a peer goes froom AEAD to not supporting it, an advert will fix other contacts trying AEAD-4 instantly. If an advert never happens it takes a max of 50 messages before we fallback to ECB.

darconeous:
Pardon me if I missed it, but why use ChaCha20-256 when AES encryption is supported on pretty much all hardware?

weebl2000:
> Pardon me if I missed it, but why use ChaCha20-256 when AES encryption is supported on pretty much all hardware?

Main reasons for this are:

- easy to use full 256-bit key since it natively uses all 32 bytes from the ECDH shared secret, current AES-128 only uses half
- fast in software on all platforms (ESP32, nRF52, raspberry), not all have AES hw acceleration and when they do they do not always support AEAD algorithms at hw level, ESP32-S3 for example doesnt support AES GCM or AES CCM at hw level
- better nonce-reuse resilience — per-message key derivation via HMAC means a nonce collision only leaks P1 XOR P2 for that one message pair, no key recovery. AES-GCM catastrophically leaks the authentication key on nonce reuse
- already supported by the existing crypto library rweather/Crypto

jcjones:
Regarding @nextgens' comment on https://github.com/meshcore-dev/MeshCore/issues/259#issuecomment-4038808873 about authenticated DH - This is a legitimate gap that concerns me, too, and I think we should have a concrete plan around it before merging.

This PR uses bare X25519 ECDH for session key negotiation -- the AEAD protects integrity and confidentiality, but an active attacker could MITM the ephemeral exchange and have valid session keys for both sides. Poly1305 can't protect us if the key is negotiated with the wrong peer in the first place. 

I'd like to suggest that we don't need to solve this in this PR but we should have a top-level plan, in the description, containing a forward-compatible path before merging. I think such a plan should hit these three points:

1: Opportunistic upgrade to authenticated DH. 

If both sides advertise support (e.g., via a capability flag in the session key request/response), they perform a (`Noise_IK`-like?) handshake using their existing long-term identity keys to authenticate the ephemeral exchange. If one side doesn't support it, fall back to the current bare ECDH -- so we avoid another wire break, and can show degraded auth in the UI or whatever.

2: Out-of-Band safety number verification.

Once a session key is established, supporting clients can derive a (short) verification code (like Signal's safety numbers) from the session transcript and, if users care about MITM resistance, they can compare these OOB. This way we get trust-on-first-use with verification. Again, like Signal.

3: Wire compatibility.

The session key request/response already has a type byte and there's space available in the AD. Whatever's chosen for the authenticated handshake should be able to fit into the existing message flow by extending those payloads without changing the AEAD wire format itself.

Of course `rweather/Crypto` doesn't have any Noise implementations (but hey, it has Blake2b!) so there's a whole bunch of work to do to approach authenticated DH. I don't want the perfect to be the enemy of the good, so I am quite supportive of proceeding with this PR without waiting on having everything we could want ready to go. However, we should spend the little bit of research time to confirm, update this PR with a plan, and maybe toss some code comments in to make clear that the grounds here are prepared for the follow-on work.

nextgens:
I don't particularly care about the choice of primitives but I want to point out that there's also ``rweather/CryptoLW`` if we wanted to: https://rweather.github.io/lightweight-crypto/algorithms.html

The key schedule of ascon is significantly faster... and it has a nice XOF.

@jcjones if you want to sleep at night I recommend you refrain from looking at https://github.com/meshcore-dev/MeshCore/blob/main/lib/ed25519/key_exchange.c ;)

Avamander:
Looking at this PR I think that it would actually be quite viable and **significantly safer** to adopt something like Noise and to implement a ratcheting mechanism. Like briefly mentioned in earlier comments (and reordering or missing messages can be handled when a ratchet is being used). This way we could utilize battle-tested implementations for at least some parts of the entire setup, with potentially significantly fewer footguns.

Right now for example I'm pretty sure 2-byte nonce is way way too small for securing shared-key groups beyond a tiny amount of messages.

For the price that the ratchet introduces, it would actually provide perfect forward security (compared to just coarse forward security through periodic rekeying). It also means fewer messages could ever be compromised at a single time. I'm sure there are more benefits as well. And these points apply to both one-to-one communications and room servers. The current approach in this PR fails (or can be broken) in hard to recover ways.

This current PR is indeed **much better** than what we have now but it leaves gaps (or vulnerabilities) that would still require an another iteration, I fear.

----

1) For starters I'd consider switching to using something like Noise that has been battle-tested for doing the initial interactive handshake, that way the handshake itself is also authenticated. Hand-rolling DH is very error-prone.

2) Reuse or find a parallel approach for room servers, so that we wouldn't need a brand new approach for those later. This very likely also opens up opportunities to offload some parts to willing room servers for better ease-of-use.

If these two are addressed and if we have a willing room server, we could use the room server as a (better-connected) trust-free temporary storage or "key bag" of sorts. It could also theoretically make it possible to start one-to-one AEAD communications with a node that was temporarily out of reach (offline). And it could be used to make the handshake more reliable and potentially require fewer hops, especially if there's a shared room as well. Plenty of clever ways to improve usability, really.

However with an actual ratchet, the biggest difference would be that it would also allow FS in room servers. Which this PR doesn't currently provide, but would be a huge benefit.

PR #1680, opened 2026-02-12
Wio Tracker L1: Improve battery reading and add power management

reidt1:
wio-tracker-l1: improve battery voltage reading accuracy
    
    - Add 100us settling delay after enabling voltage divider
    - Average 8 ADC samples to reduce noise
    - Disable voltage divider after reading to save power
    - Move ADC configuration to begin() for consistency
    - Move getBattMilliVolts implementation to .cpp file

wio-tracker-l1: add power management support
    
    - Add boot voltage protection (won't boot below 3.3V)
    - Add LPCOMP wake configuration for voltage recovery
    - Implement initiateShutdown() with LPCOMP wake on low voltage
    - Enable NRF52_POWER_MANAGEMENT build flag
    
    Device will enter protective shutdown on low battery and wake
    automatically when voltage recovers to ~3.7V or USB is connected.

### DIFF
--- a/variants/wio-tracker-l1/WioTrackerL1Board.cpp
+++ b/variants/wio-tracker-l1/WioTrackerL1Board.cpp
@@ -3,16 +3,46 @@
 
 #include "WioTrackerL1Board.h"
 
+#ifdef NRF52_POWER_MANAGEMENT
+// Static configuration for power management
+// Values set in variant.h defines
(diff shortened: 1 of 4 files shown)


reidt1:
@recrof @liamcottle Not sure who exactly to ping, but was wondering if there were any questions or concerns about this PR

PR #1682, opened 2026-02-13
nrf52: allow companion to sleep when idle

oltaco:
Following on from the [NRF52 repeater power savings](https://github.com/meshcore-dev/MeshCore/pull/1562), this adds the same for NRF52 companions.

I've been running it on WioTrackerL1/ThinkNode M1/Heltec T114/Meshtiny for the last week with no issues.

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -225,4 +225,9 @@ void loop() {
   ui_task.loop();
 #endif
   rtc_clock.tick();
+
+#if defined(NRF52_PLATFORM)
+  board.sleep(1800);
+#endif
+
 }


oltaco:
Forgot to remove some commented testing code, fixed now. 🤦‍♂️

examples/companion_radio/main.cpp
```diff
@@ -225,4 +225,9 @@ void loop() {
   ui_task.loop();
 #endif
   rtc_clock.tick();
+
+#if defined(NRF52_PLATFORM)
+  board.sleep(1800);
```
weebl2000, line 230:
```suggestion
  if (!the_mesh.hasPendingWork()) {
    board.sleep(1800);
  }
```
Maybe?

Probably need to add hasPendingWork(), _mgr->getOutboundCount(0xFFFFFFFF) > 0 - not sure if it's actually useful.

towerviewcams:
I have a WioTracker L1, put up a test bin and I can check this in my lab and report back.

weebl2000:
> I have a WioTracker L1, put up a test bin and I can check this in my lab and report back.

Build it here:

http://mcimages.weebl.me/?commitId=nrf52-lowpower

lbibass:
Here are the results of testing with an NRF PPK and a Heltec T114. Hope this helps! 

Before:
image

After:
image

oltaco:
> Here are the results of testing with an NRF PPK and a Heltec T114. Hope this helps!
> 
> Before: image
> 
> After: image

I hope the before and after are around the wrong way! 😂

lbibass:
> > Here are the results of testing with an NRF PPK and a Heltec T114. Hope this helps!
> > Before: image
> > After: image
> 
> I hope the before and after are around the wrong way! 😂

Check the average power consumption! it dropped from almost 12 mA to almost 8 mA.

towerviewcams:
> Here are the results of testing with an NRF PPK and a Heltec T114. Hope this helps!

@lbibass  can you point me on how/where to get this power meter/software your using here? I do allot of testing and would like to add this to my lab..  Thank you.

Found it!!!! yea I'm getting one!

lbibass:
It would be great to get this merged!

terminalvelocity23:
I've tested it with T114, and it works great. Dropped idle current from 14-15 mA to 8.2 mA.

examples/companion_radio/main.cpp
robekl: Seems like the right direction.

`board.sleep()` is unconditional here, so this does not actually implement “sleep when idle”. On nRF52 `NRF52Board::sleep()` is a blocking sd_app_evt_wait() / __WFE() wait (src/helpers/NRF52Board.cpp:254), which means every companion build now becomes interrupt-paced even while `MyMesh::loop()` still has active serial or radio work to drain. That is a behavior change from the existing repeater logic, which only sleeps after `!the_mesh.hasPendingWork()` (examples/simple_repeater/main.cpp:137). If the intent is only idle power saving, this needs an explicit idle predicate; otherwise active USB/BLE sessions and contact syncs now pay an avoidable wait on every loop iteration.

weebl2000, line 230:
```suggestion
  if (_mgr->getOutboundTotal() == 0)
    board.sleep(1800);
```
This is probably sufficient, but even without it the code should still work, but it sleeps an additional millisecond? :)

weebl2000: I think just checking outbound is an easy extra, it will wakeup regardless though.

chiefMarlin:
Also would like to see this merged! Thanks for hard work!

examples/companion_radio/main.cpp
terminalvelocity23: I concur, added a similar check to my version.

mweinelt, line 230:
```suggestion
  board.sleep(0);
```

Sleep on NRF52 is event-based, not time-based. See `NRF52Board::sleep` in `src/helpers/NRF52Board.cpp`, where `secs` is completely unused.

mweinelt:
This is what I'm running with now. Delegating the decision to the MyMesh class was simpler, since it has access to all the relevant data.

```patch
diff --git a/examples/companion_radio/MyMesh.cpp b/examples/companion_radio/MyMesh.cpp
index 32e36260..e1267623 100644
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -391,6 +391,10 @@ int MyMesh::getRecentlyHeard(AdvertPath dest[], int max_num) {
   return max_num;
 }
 
+bool MyMesh::isIdleForSleep() {
+  return _mgr->getOutboundTotal() == 0;
+}
+
 void MyMesh::onContactPathUpdated(const ContactInfo &contact) {
   out_frame[0] = PUSH_CODE_PATH_UPDATED;
   memcpy(&out_frame[1], contact.id.pub_key, PUB_KEY_SIZE);
diff --git a/examples/companion_radio/MyMesh.h b/examples/companion_radio/MyMesh.h
index 4271729b..bebc462c 100644
--- a/examples/companion_radio/MyMesh.h
+++ b/examples/companion_radio/MyMesh.h
@@ -101,6 +101,7 @@ public:
   void enterCLIRescue();
 
   int  getRecentlyHeard(AdvertPath dest[], int max_num);
+  bool isIdleForSleep();
 
 protected:
   float getAirtimeBudgetFactor() const override;
diff --git a/examples/companion_radio/main.cpp b/examples/companion_radio/main.cpp
index eff9efca..034c7bb6 100644
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -225,4 +225,9 @@ void loop() {
   ui_task.loop();
 #endif
   rtc_clock.tick();
+#ifdef NRF52_PLATFORM
+  if (the_mesh.isIdleForSleep()) {
+    board.sleep(0);
+  }
+#endif
 }
 ```

mweinelt:
#2286 includes the relevant changes and should superseed this PR.

Blake-Latchford:
@oltaco Can this PR be closed given mweinelt's comment above?

PR #1690, opened 2026-02-13
DS18B20 temperature sensor

thebino:
This PR will add support for a DS18B20 3-wire temperature sensor.

image

Screenshot_20260214-142459


awhite2000:
In the diagram above - technically speaking, the 1-Wire bus doesn't require that the sensor is connected to the VCC line, it's enough to connect it to GND and connect the data line to a pull-up resistor.

thebino:
@awhite2000 you're not wrong about this. Those D18B20 support two power modes
- normal mode (3-wire)
- parasite mode (2-wire)

For temperature sensors a stable and continuous power supply is recommended. Also the parasite mode requires additional configuration which might not be supported by every board. So I kept it simple with 3-wire

PR #1691, opened 2026-02-13
Be more patient: increase companion timeouts

weebl2000:
Increase base timeout constants and scale flood timeouts by attempt number.

## Problem

The flood timeout (`500ms + 16× airtime`) only covers ~3-4 hop round trips. Each forwarding node adds ~airtime + random(0-5)×airtime/2 delay, so a 10-20 hop destination needs 15-31 seconds round trip — but the timeout fires after ~6s, causing premature retries that add congestion and make subsequent attempts even less likely to succeed.

## Changes

**Increased base constants**:
- `SEND_TIMEOUT_BASE_MILLIS`: 500 → 1000
- `FLOOD_SEND_TIMEOUT_FACTOR`: 16 → 32
- `DIRECT_SEND_PERHOP_FACTOR`: 6 → 10
- `DIRECT_SEND_PERHOP_EXTRA_MILLIS`: 250 → 500

**Attempt-based scaling** for flood timeouts: `base_timeout × (attempt + 1)`

For a typical 345ms airtime packet (29B, SF8, BW62.5kHz, CR4/8):

| Attempt | Timeout |
|---|---|
| 0 (first send) | 12.0s (~7 hop RT) |
| 1 | 24.1s (~15 hop RT) |
| 2 | 36.1s (~23 hop RT) |
| 3 | 48.2s (~31 hop RT) |

Only `sendMessage` and `sendCommandData` pass `attempt` — other send functions (login, anon req, contact request) default to `attempt=0`, no behavior change.

---
**Build firmware:** [Build from this branch](https://mcimages.weebl.me/?commitId=be-more-patient)

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -100,10 +100,10 @@
 
 #define MAX_CHANNEL_DATA_LENGTH       (MAX_FRAME_SIZE - 9)
 
-#define SEND_TIMEOUT_BASE_MILLIS        500
-#define FLOOD_SEND_TIMEOUT_FACTOR       16.0f
-#define DIRECT_SEND_PERHOP_FACTOR       6.0f
-#define DIRECT_SEND_PERHOP_EXTRA_MILLIS 250
+#define SEND_TIMEOUT_BASE_MILLIS        1000
+#define FLOOD_SEND_TIMEOUT_FACTOR       32.0f
+#define DIRECT_SEND_PERHOP_FACTOR       10.0f
+#define DIRECT_SEND_PERHOP_EXTRA_MILLIS 500
 #define LAZY_CONTACTS_WRITE_DELAY       5000
 
 #define PUBLIC_GROUP_PSK                "izOH6cXN6mrJ5e26oRXNcg=="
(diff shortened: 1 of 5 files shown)


KiserDesigns:
Would love to see this tested in a dense mesh. I think this has a good chance of reducing flood packets from retries.

ripplebiz:
These timeouts are only suggestions, and the mobile app does it's own timeout calcs, and does do linear back-off according to Liam. So, these changes would add back-offs to back-offs (I believe).

weebl2000:
> These timeouts are only suggestions, and the mobile app does it's own timeout calcs, and does do linear back-off according to Liam. So, these changes would add back-offs to back-offs (I believe).

@liamcottle Can you confirm? I know for sure this PR changes the behavior of my companion on the regular client app.

liamcottle:
> @liamcottle Can you confirm? I know for sure this PR changes the behavior of my companion on the regular client app.

The app does do some random delays between retries, but it's not exponential at this time.

I plan to increase the maximum attempts in a future app update, as we have recently added support for making ack packets unique. Previously they were limited to ~4 attempts as intermediate repeaters would drop the packet if they'd previously seen it. Dev branch firmware will allow for ack payloads to be unique every time, so using the same ack code or attempt will not be dropped.

I would be against having the firmware increase the estimated timeout exponentially based on the attempt count.

I'd prefer the clients to implement their own backoff strategies, as it's easier/faster for clients to be updated than firmware.

Otherwise the clients will end up ignoring what the firmware suggests as a timeout, as sending a message with attempt=20 would result in an insane timeout duration.

I'd probably be okay with a slight increase to estimated timeout defaults, but not basing it exponentially off the retry count.

PR #1705, opened 2026-02-15
Experimental lr1121 generic support

aslater3:
## Summary
This PR adds LR1121-specific TX power handling for multi-band operation by enforcing different maximum TX power limits based on active RF band.

## What Changed
- Updated `CustomLR1121` to track the currently configured band from `setFrequency(...)`.
- Added LR1121-specific TX power clamping in both `setOutputPower(...)` overloads:
  - 2.4 GHz band: max `20 dBm`
  - sub-GHz band: max `22 dBm`
- Kept changes scoped to `CustomLR1121` so non-LR1121 radios and boards are unaffected.

## Why
This addresses maintainer feedback that LR1121 needs special handling because it supports multiple radios with different TX power limits by band.

## Scope / Compatibility
- No behavior change for non-LR1121 targets.
- No board-level logic changes required for existing non-LR1121 variants.

## Testing
- Code-level validation only in this PR.
- Full hardware/runtime validation should be done on LR1121 targets in both sub-GHz and 2.4 GHz configurations. Tested locally on subGhz in repeater and companion mode.

### DIFF
--- a/src/helpers/radiolib/CustomLR1121.h
+++ b/src/helpers/radiolib/CustomLR1121.h
@@ -6,6 +6,24 @@ class CustomLR1121 : public LR1121 {
 public:
   CustomLR1121(Module* mod) : LR1121(mod) { }
 
+  int16_t setFrequency(float freq) override {
+    _is_24ghz = (freq >= 2400.0f);
+    return LR1121::setFrequency(freq);
+  }
+
+  int16_t setFrequency(float freq, bool skipCalibration, float band = 4) {
+    _is_24ghz = (freq >= 2400.0f);
+    return LR1121::setFrequency(freq, skipCalibration, band);
+  }
+
+  int16_t setOutputPower(int8_t power) override {
+    return LR1121::setOutputPower(clampTxPower(power));
+  }
+
+  int16_t setOutputPower(int8_t power, bool forceHighPower, uint32_t rampTimeUs = 48) {
+    return LR1121::setOutputPower(clampTxPower(power), forceHighPower, rampTimeUs);
+  }
+
   size_t getPacketLength(bool update) override {
     size_t len = LR1121::getPacketLength(update);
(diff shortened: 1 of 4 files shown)


src/helpers/radiolib/CustomLR1121.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#include <RadioLib.h>
+
+class CustomLR1121 : public LR1121 {
```
weebl2000, line 5: How is the radio actually initialized?

src/helpers/radiolib/CustomLR1121.h
```diff
@@ -0,0 +1,49 @@
...
+public:
+  CustomLR1121(Module* mod) : LR1121(mod) { }
+
+  int16_t setFrequency(float freq) override {
+    _is_24ghz = (freq >= 2400.0f);
+    return LR1121::setFrequency(freq);
+  }
+
+  int16_t setFrequency(float freq, bool skipCalibration, float band = 4) {
```
weebl2000, line 14: This isn´t actually overriding.

src/helpers/radiolib/CustomLR1121.h
```diff
@@ -0,0 +1,49 @@
...
+    _is_24ghz = (freq >= 2400.0f);
+    return LR1121::setFrequency(freq, skipCalibration, band);
+  }
+
+  int16_t setOutputPower(int8_t power) override {
+    return LR1121::setOutputPower(clampTxPower(power));
+  }
+
+  int16_t setOutputPower(int8_t power, bool forceHighPower, uint32_t rampTimeUs = 48) {
```
weebl2000, line 23: Same

weebl2000 (changes requested):
needs a bit of love.

liamcottle:
Hi there, thanks for the contribution. I'll close this one out since it's experimental, was opened against `main` instead of `dev` and the review comments haven't been addressed.

PR #1714, opened 2026-02-16
Default to zerohop advert with 'advert' CLI-Command 

mesher-de:
This is an alternative approach to #1490 

The goal is to reduce mesh-load cause by too many repeater adverts.

This PR modifies the logic of the `advert` command to send a zerohop advert by default and introduces a new `advert.flood` command for sending a flood advert.

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -200,10 +200,13 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
       // Reset clock
       getRTCClock()->setCurrentTime(1715770351);  // 15 May 2024, 8:50pm
       _board->reboot();  // doesn't return
-    } else if (memcmp(command, "advert", 6) == 0) {
+    } else if (memcmp(command, "advert.flood", 12) == 0 && (command[12] == 0 || command[12] == ' ')) {
       // send flood advert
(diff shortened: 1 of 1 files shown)


liamcottle:
Hey, thanks for the time/effort putting in the PR :)

To be honest, would prefer to leave `advert` cli command as flood, otherwise now you have a command that does different things across different firmware versions.

This also means the app would need to determine which firmware version you're running so it knows to use `advert.flood` instead of `advert`. Otherwise this changes what that buttons does in the app, which may be unexpected to the user.

Would prefer to have a new command `advert.zerohop` that sends a zero hop advert. I haven't had a chance to review that linked PR but looks like it does something along those lines.

mesher-de:
> Hey, thanks for the time/effort putting in the PR :)
> 
> To be honest, would prefer to leave `advert` cli command as flood, otherwise now you have a command that does different things across different firmware versions.
> 
> This also means the app would need to determine which firmware version you're running so it knows to use `advert.flood` instead of `advert`. Otherwise this changes what that buttons does in the app, which may be unexpected to the user.
> 
> Would prefer to have a new command `advert.zerohop` that sends a zero hop advert. I haven't had a chance to review that linked PR but looks like it does something along those lines.

Hi Liam,

Thanks for taking the time to review this!

Will update the linked #1490 to use `advert.zerohop` as you suggested and close this one.

PR #1718, opened 2026-02-16
discover sends a single repeater discovery request and populates the neighbor list; self is excluded

realtag-github:
discover sends a single repeater discovery request and populates the neighbor list; self repeater is excluded.

Allows to quickly build a list of neighbors without waiting for repeaters to advert. Useful for right after restart when the list is empty.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -767,8 +767,6 @@ void MyMesh::onControlDataRecv(mesh::Packet* packet) {
 }
 
 void MyMesh::sendNodeDiscoverReq() {
-  if (_prefs.disable_fwd) return;
-
   uint8_t data[10];
   data[0] = CTL_TYPE_NODE_DISCOVER_REQ; // prefix_only=0
   data[1] = (1 << ADV_TYPE_REPEATER);


examples/simple_repeater/MyMesh.cpp
```diff
@@ -738,6 +738,37 @@ void MyMesh::onControlDataRecv(mesh::Packet* packet) {
...
+  }
+}
+
+void MyMesh::sendNodeDiscoverReq() {
+  if (_prefs.disable_fwd) return;
+
+  uint8_t data[10];
+  data[0] = CTL_TYPE_NODE_DISCOVER_REQ; // prefix_only=0
+  data[1] = (1 << ADV_TYPE_REPEATER);
+  getRNG()->random(&data[2], 4); // tag
```
ripplebiz, line 773: Should save the 4-byte tag to a member variable, then make sure the response packets match. Otherwise, a bad actor could pollute everyone's neighbors lists with random garbage.

ripplebiz:
Please change base branch to 'dev'.

examples/simple_repeater/MyMesh.cpp
```diff
@@ -1168,6 +1199,15 @@ void MyMesh::handleCommand(uint32_t sender_timestamp, char *command, char *reply
     } else {
       strcpy(reply, "Err - ??");
     }
+  } else if (memcmp(command, "discover", 8) == 0) {
```
liamcottle: Could we make it `discover.neighbours` so it's more explicit that the intention is to discover nearby neighbours? otherwise `discover.repeaters` so in the future if we wanted to discover other types the `discover` prefix doesn't gobble up the request on old firmware.

examples/simple_repeater/MyMesh.cpp
liamcottle: Possibly with a timeout too. So if the request tag is reused by some other nearby repeater in the future, it isn't using those responses as well. Probably unlikely since there's 4.2 billion combinations for a `uint32_t` tag, but if the member variable is never cleared, it's still another way to pollute, as anyone can listen to those tags as they go over the air.

liamcottle:
I like this idea :)

- `discover.neighbours` or `discover.repeaters` cli command sends a 0 hop control packet request with a random tag
- nearby neighbouring repeaters should reply to control packet
- repeater should filter out control packets that are not from repeater node type
- repeater should filter out control packets that are not for the random tag
- repeater should timeout (clear the random tag) after a while so much later replies with same tag aren't used

This will be a nice addition!

realtag-github:
> Please change base branch to 'dev'.

Hopefully I did it correctly. it seems to have picked up 2 unrelated commits though.

liamcottle:
If you checkout latest dev branch locally, then rebase your discover branch onto the local dev branch, then force push, it should fix it.

The PR needs to be opened against out dev branch though. It's currently set to main branch.

Regarding the timeout, wonder if we should bump to 60 sec, since there's some chOnky meshes out there now? The repeaters have quite a high delay before sending a reply to the control packets, since lots of nodes can reply to it.

realtag-github:
- Renamed discover to discover.neighbors  I used American spelling of the word to stay consistent with "neighbor" command. 
- Added 30 seconds timeout after which DISOVER_RESP packets are ignored.
- Not accepting all responses anymore but checking it now.

realtag-github:
> If you checkout latest dev branch locally, then rebase your discover branch onto the local dev branch, then force push, it should fix it.

right I forgot to switch it again here. Only showing 2 commits now as it should. 

> Regarding the timeout, wonder if we should bump to 60 sec, since there's some chOnky meshes out there now? The repeaters have quite a high delay before sending a reply to the control packets, since lots of nodes can reply to it.

Yeah 30 seconds might be too short. It works in my tests but I am not very far from other repeaters and our mesh is fairly small.

examples/simple_repeater/MyMesh.cpp
```diff
@@ -738,6 +738,49 @@ void MyMesh::onControlDataRecv(mesh::Packet* packet) {
...
+    mesh::Identity id(&packet->payload[6]);
+    if (id.matches(self_id)) {
+      return;
+    }
+    putNeighbour(id, rtc_clock.getCurrentTime(), packet->getSNR());
+  }
+}
+
+void MyMesh::sendNodeDiscoverReq() {
+  if (_prefs.disable_fwd) return;
```
liamcottle: Prob remove this, so repeater can still perform discovery, even if it's not actively allowing packets to be repeated.

liamcottle:
Sweet!

One last comment, you have a check for `if (_prefs.disable_fwd) return;` in the cli command logic.

I would say if you're explicitly sending the command, it should still send the request, even if the repeater you're sending the request from has repeat disabled. It should still be able to discover nearby neighbours.

I'm currently away for the day, but will be keen to test this on my repeaters through the app later tonight.

examples/simple_repeater/MyMesh.cpp
realtag-github: Removed it now and pushed updated PR.  Confirmed that discovery still works regardless of the forwarding mode now.

liamcottle:
Looks good, thanks! Will flash and test on my repeaters tonight :)

ripplebiz:
Looks good.

liamcottle:
Tested and working on my repeaters. The next app update will also have a new `Discover Neighbours` button in the repeater neighbours dropdown menu. Thanks for contributing :)

PR #1724, opened 2026-02-17
Simply send SSD1306 off/on commands to keep I2C bus on Wire functioning correctly

weebl2000:
Following comment by @IoTThinks  that the code killed the I2C bus. Even though pins 17/18 are mentioned to be dedicated to the OLED, you could still use the I2C bus on there for other things. I started out by stopping/restarting the Wire, but maybe it's easier to just put SSD11306 to sleep and avoid VEXT claim/release altogether? Power consumption should be very very minimal (< 10µA).

Confirmed that my BME280 works again when I comment out dedicated env pins:

image

Before this change with these lines commented out the BME280 would not be read because the bus would be broken.

I don't have a Heltec v4 with the TFT display, would be great if someone could test. I'd suspect the behavior is unchanged.

@Quency-D thoughts?

Tested this on a USB companion and v4 repeater, both working fine. No increased power consumption. Theoretically a 3500 mAh 18650 could power the SSD11306 display for 40 years if it's in the off state.

The only concern might be keeping VEXT on, but currently only the heltec boards have VEXT defined for the display and there's nothing else on there.

For anyone that can't build themselves easily but wants to try this: https://mcimages.weebl.me?commitId=fix-ssd1306display

### DIFF
--- a/src/helpers/ui/SSD1306Display.cpp
+++ b/src/helpers/ui/SSD1306Display.cpp
@@ -19,30 +19,14 @@ bool SSD1306Display::begin() {
 
 void SSD1306Display::turnOn() {
   if (!_isOn) {
-    if (_peripher_power) {
-      _peripher_power->claim();
-#if defined(PIN_BOARD_SDA) && defined(PIN_BOARD_SCL)
-      Wire.begin(PIN_BOARD_SDA, PIN_BOARD_SCL);  // re-init after Wire.end() in turnOff()
-#else
-      Wire.begin();
-#endif
-    }
-    _isOn = true;  // set before begin() to prevent double claim
-    if (_peripher_power) begin();  // re-init display after power was cut
+    _isOn = true;
+    display.ssd1306_command(SSD1306_DISPLAYON);
   }
-  display.ssd1306_command(SSD1306_DISPLAYON);
 }
 
 void SSD1306Display::turnOff() {
-  display.ssd1306_command(SSD1306_DISPLAYOFF);
   if (_isOn) {
-    if (_peripher_power) {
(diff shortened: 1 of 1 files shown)


IoTThinks:
You may check this PR.
We can pass NULL to perpher_power to skip all release/claim.
OLED in Heltec v4 does not use vext.

And set back the value in MC 1.11 for vext.

https://github.com/meshcore-dev/MeshCore/pull/1569

Please keep the GPIO 17,18 for all Meshcore roles on Heltec boards.

IoTThinks:
@weebl2000 I believe when vext is on, it would comsume extra 26mA as in MeshCore 1.12.

The minimum current change you see, I suspect not due to vext.
OLED is likely not using vext at all.

The extra small current is OLED still consumes tiny current when off.

weebl2000:
> @weebl2000 I believe when vext is on, it would comsume extra 26mA as in MeshCore 1.12.
> 
> The minimum current change you see, I suspect not due to vext. OLED is likely not using vext at all.
> 
> The extra small current is OLED still consumes tiny current when off.

The current increase when the screen turns off isn't from OLED sleep current (~10µA, unmeasurable on USB). It's from the inverted VEXT polarity — release() with ACTIVE=LOW drives GPIO36 HIGH, which turns VEXT ON (NPN Q6 + PMOS Q5 high-side switch). So the firmware turns on the Vext rail at the exact moment it intends to turn it off. Your fix (display(NULL) + ACTIVE=HIGH) correctly avoids this.

IoTThinks:
I would love to have the root cause fixed.

My PR may be a hint. I dont mind if any friends here can push a fix to dev.

Thanks a lot for your finding about vext polarity.

PR #1726, opened 2026-02-17
Fix packet pool leak when rx queue is full

weebl2000:
PacketQueue::add() silently dropped packets when the queue was at capacity. The packet pointer was lost — never enqueued, never returned to the unused pool. Each occurrence permanently shrank the 32-packet pool until allocNew() returned NULL and the node went deaf. Return bool from add() and free the packet back to the pool on failure.

I'm not sure how realistic it is to hit 32 packet limit.


KiserDesigns:
This looks like a very straightforward fix for a not-so-obvious issue. Would love to see this merged.

PR #1728, opened 2026-02-18
NRF52: Add get bootloader.ver command for NRF52

oltaco:
This PR adds a function for finding out the NRF52 bootloader version. It relies on searching the flash memory from offset 0xF9000 to 0xFC000 for the "UF2 Bootloader " string which is found in the INFO.TXT file in bootloaders based on Adafruit nRF52 Bootloader.

Tested on RAK 4631 and Heltec V3.

RAK4631 stock bootloader:
```
get bootloader.ver
  -> > 0.4.2
```

RAK4631 latest OTAFIX bootloader:
```
get bootloader.ver
  -> > 0.9.2-OTAFIX1.2-BP1.1
```
Heltec V3:
```
get bootloader.ver
  -> ERROR: unsupported
```

### DIFF
--- a/src/helpers/NRF52Board.h
+++ b/src/helpers/NRF52Board.h
@@ -50,6 +50,7 @@ public:
   virtual uint8_t getStartupReason() const override { return startup_reason; }
   virtual float getMCUTemperature() override;
   virtual void reboot() override { NVIC_SystemReset(); }
+  virtual bool getBootloaderVersion(char* version, size_t max_len) override;
   virtual bool startOTAUpdate(const char *id, char reply[]) override;
   virtual void sleep(uint32_t secs) override;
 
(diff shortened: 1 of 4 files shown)


src/helpers/NRF52Board.h
```diff
@@ -50,6 +50,7 @@ class NRF52Board : public mesh::MainBoard {
   virtual uint8_t getStartupReason() const override { return startup_reason; }
   virtual float getMCUTemperature() override;
   virtual void reboot() override { NVIC_SystemReset(); }
+  virtual bool getBootloaderVersion(char* version, size_t max_len) override;
```
ripplebiz, line 53: Just wondering whether this should be non-virtual method, as it's only an nRF concept? (and you're guarding it with a #if )

oltaco: Yeah good point, I think to make it non-virtual it will require including NRF52Board.h into CommonCLI and casting the _board call to NRF52Board, is that OK?

PR #1730, opened 2026-02-18
Add option to change between SPI0 and SPI1 on RP2040

JDat:
This is useful when somebody want to make/solder own custom Meshcore hardware

Open variants/rpi_picow/platformio.ini file
Comment out unneeded SPI line
Uncomment needed SPI line

Changes to be committed:
	modified:   variants/rpi_picow/platformio.ini
	modified:   variants/rpi_picow/target.cpp


JDat:
There is much more better way to fix this by adding another variant instead of tinkering with picow.

PR #1733, opened 2026-02-18
Merge main into dev

weebl2000:
There's a few things that were done on main but didn't get merged to dev.


weebl2000:
@ripplebiz out of curiosity: will this be done separately or?

liamcottle:
> out of curiosity: will this be done separately or?

It's mostly just the documentation. PRs for docs get opened against `main` so they show up on the documentation sites and default git branch instantly. Whereas code change PRs are opened against `dev` so they can be tested.

The main branch should be treated as always working, and is the code used in the most recent release tag.

I don't think merging the docs into dev branch is really needed, and just causes extra noise on the timeline.

weebl2000:
> I don't think merging the docs into dev branch is really needed, and just causes extra noise on the timeline.

It's usually a good approach to have dev up to date with main. Imo it adds more noise seeing commits that are in main but not dev.👀

PR #1745, opened 2026-02-19
Add counter-based relay suppression for flood packets

ViezeVingertjes:
Implements a MANET-style counter-based relay suppression mechanism for repeaters. During the random backoff before relaying a flood packet, if the node hears the same packet relayed by enough neighbors, it cancels its own relay since the area is already covered.

- Adds onDuplicateFloodRecv hook in Mesh, called when a duplicate flood packet arrives
- Repeater override scans the outbound queue for a matching pending relay and increments a per-packet counter
- When the counter reaches the configurable threshold, the queued relay is removed immediately
- New relay.redundancy preference (default 3, 0 to disable), remotely configurable via get/set relay.redundancy
- Persisted to prefs file, backward-compatible with existing saved configs

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -484,8 +484,13 @@ int MyMesh::calcRxDelay(float score, uint32_t air_time) const {
 }
 
 uint32_t MyMesh::getRetransmitDelay(const mesh::Packet *packet) {
-  uint32_t t = (_radio->getEstAirtimeFor(packet->path_len + packet->payload_len + 2) * _prefs.tx_delay_factor);
-  return getRNG()->nextInt(0, 5*t + 1);
+  uint32_t airtime = _radio->getEstAirtimeFor(packet->path_len + packet->payload_len + 2);
+  uint32_t t = airtime * _prefs.tx_delay_factor;
+  uint32_t delay = getRNG()->nextInt(0, 5*t + 1);
+  if (_prefs.relay_redundancy > 0 && delay < airtime) {
+    delay = airtime;
+  }
+  return delay;
 }
 uint32_t MyMesh::getDirectRetransmitDelay(const mesh::Packet *packet) {
   uint32_t t = (_radio->getEstAirtimeFor(packet->path_len + packet->payload_len + 2) * _prefs.direct_tx_delay_factor);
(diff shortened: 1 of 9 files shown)


ViezeVingertjes:
Any thoughts on such approach to reduce congestion?

Although i still think https://github.com/meshcore-dev/MeshCore/issues/1223 is a more proper solution, perhaps something like this would be more in favor?

awhite2000:
I'm not sure it's a good idea.

Let's consider an edge (or a bottleneck) repeater R, where there are multiple well-connected neighbors on one side (group A) and another group of repeaters on the other side (group B). However, group A and B are not connected directly between themselves except for this repeater.

With the proposed workaround, since the repeater R receives multiple copies from neighbors in group A, it decides to drop the packet. As a consequence the group B will never receive the packet. Net result is that depending on timing a large portion of the traffic between group A and B will be dropped. This will surely reduce the congestion ;) but not in a good way.

ViezeVingertjes:
> I'm not sure it's a good idea.
> 
> Let's consider an edge (or a bottleneck) repeater R, where there are multiple well-connected neighbors on one side (group A) and another group of repeaters on the other side (group B). However, group A and B are not connected directly between themselves except for this repeater.
> 
> With the proposed workaround, since the repeater R receives multiple copies from neighbors in group A, it decides to drop the packet. As a consequence the group B will never receive the packet. Net result is that depending on timing a large portion of the traffic between group A and B will be dropped. This will surely reduce the congestion ;) but not in a good way.

We did think of such scenario, but in reality it would not just be hearing one direction unless its a Yagi (and then it doesnt make any difference). In case it would actually be happening in reality in some weird case, it could be disabled for that specific repeater.

examples/simple_repeater/MyMesh.cpp
```diff
@@ -509,6 +514,37 @@ bool MyMesh::filterRecvFloodPacket(mesh::Packet* pkt) {
   return false;
 }
 
+void MyMesh::onDuplicateFloodRecv(mesh::Packet* pkt) {
+  if (_prefs.relay_redundancy == 0) return;
+
+  uint8_t dup_hash[MAX_HASH_SIZE];
+  pkt->calculatePacketHash(dup_hash);
+
+  int n = _mgr->getOutboundCount(0xFFFFFFFF);
```
weebl2000, line 523: Maybe UINT32_MAX instead?

examples/simple_repeater/MyMesh.cpp
```diff
@@ -509,6 +514,37 @@ bool MyMesh::filterRecvFloodPacket(mesh::Packet* pkt) {
   return false;
 }
 
+void MyMesh::onDuplicateFloodRecv(mesh::Packet* pkt) {
+  if (_prefs.relay_redundancy == 0) return;
+
+  uint8_t dup_hash[MAX_HASH_SIZE];
```
weebl2000, line 520:
```suggestion
  // Don't suppress if we have few neighbors — likely a bridge/edge node
  int active = countActiveNeighbours();
  if (active <= _prefs.relay_redundancy) return;
  
  uint8_t dup_hash[MAX_HASH_SIZE];
```

Maybe this would be a good defensive addition? Do not inhibit forwarding if we have less neighbors than relay redundancy. It will also save some computation time.

awhite2000:
> We did think of such scenario, but in reality it would not just be hearing one direction unless its a Yagi

Umm.. not really? even with an omni antenna this scenario is possible if neighbors are clumped unevenly in a way that there are often no other paths between the groups. This is an actual scenario for my node with an omni antenna.

ViezeVingertjes:
> > We did think of such scenario, but in reality it would not just be hearing one direction unless its a Yagi
> 
> Umm.. not really? even with an omni antenna this scenario is possible if neighbors are clumped unevenly in a way that there are often no other paths between the groups. This is an actual scenario for my node with an omni antenna.

Then you're probably not the use-case and won't need to use it. Another idea was this https://github.com/meshcore-dev/MeshCore/issues/1223

But we're kinda at the point that postponing isnt an option anymore, no matter what the solution will be, we need to get things moving for sure. So any better idea's are welcome ofcourse! ^^

dreirund:
> *Let's consider an edge (or a bottleneck) repeater R, where there are multiple well-connected neighbors on one side (group A) and another group of repeaters on the other side (group B). However, group A and B are not connected directly between themselves except for this repeater.*
> 
> *With the proposed workaround, since the repeater R receives multiple copies from neighbors in group A, it decides to drop the packet. As a consequence the group B will never receive the packet.*

This should be an option repeater admins can toggle depending on the situation.

ViezeVingertjes:
We're going to test run this; will re-open if that provides some more data to work with here. 👍🏼

PR #1743, opened 2026-02-19
fix agc reset on SX126x, SX1276 & LR11x0 chips

weebl2000:
**UPDATE March 2nd 2026:** this one is good to merge AFAIK, bunch of people have tested it and there's no regression found - only big improvements to AGC actually being reset.

Note: also made a PR to RadioLib upstream: https://github.com/jgromes/RadioLib/pull/1718 - so we can hopefully use those abstractions when that gets merged.

- tried to simply explicitly call sleep briefly to reset AGC -> not sufficient it seems
- do more thorough reset using Calibrate (0x7F) -> success
- reset noise_floor so firmware shows accurate values
- so far it seems to work great on a Heltec v4
- should work on all SX126x based chips by using Calibrate(0x7F)
- shoudl work on all LR11x0 based chips by using Calibrate(0x3F)
- should work on SX1276 chips too, calling sleep there should be sufficient - SX1276's AGC is simpler — just LNA gain stepping. Sleep mode powers down the entire analog frontend, and re-entering RX restarts the AGC from G1 (max gain).
- default calibration is in 902-928 MHz range, made sure to follow up with actual configured frequency
- build firmware for yourself here: https://mcimages.weebl.me/?commitId=fixagcreset

fixes #1716- see discussion there for initial test results
fixes #1209

I've referenced [SX1262 datasheet](https://www.mouser.com/datasheet/2/761/DS_SX1261-2_V1.1-1307803.pdf) to implement the calibration reset.

- p25 4.3 ,64-65 table 12-1 listing ΣΔ ADC architecture
- p55 table 9-1 sleep enable vs stdby_rc enable
- p55 9.2 states which blocks can be calibrated
- p61 9.2.1 calibrate specific freq
- p66 warm sleep restores register state confirms we need more aggressive Calibrate(0x7F) reset
- p73 total calibration time is 3.5 ms and must be launched in STDBY_RC mode

For LR1110 similar implementation, the 'AGC' is completely black boxed behind firmware so you can only reset it by resetting registers.

[LR1110 datasheet ](https://semtech.my.salesforce.com/sfc/p/E0000000JelG/a/RQ00000AXyaj/EQ4jOcJX3lpB41OWGz0VBBLb_avBZzvqrAZfl2P8ID0) page 9 section 1.2.1 states:
•Air interface fully compatible with the SX1261/2/8 family

Also see [LR1110 user docs](https://www.mouser.com/pdfDocs/UM_LR1110_W_APP_V10.pdf)

So reset is very similar.

PR #1746, opened 2026-02-20
Add Muzi Base Uno board support

johngag:
New variant for the Muzi Works Base Uno (nRFLR1262, SX1262) and Base Duo (nRFLR1121, LR1121) boards. Both share the same PCB with QWIIC I2C, QSPI flash, USB-C, and battery management. The Duo reuses the Uno board class and pin definitions, differing only in the radio driver (CustomLR1110 vs CustomSX1262).

Build environments per board:
- repeater (headless), repeater_oled (with SSD1306 display)
- companion_radio_usb, companion_radio_ble
- room_server


thetooth:
Thanks for this, testing on DUO hardware it appears as though radio is only able to transmit, it's completely deaf on 915Mhz band. Any ideas?

johngag:
> Thanks for this, testing on DUO hardware it appears as though radio is only able to transmit, it's completely deaf on 915Mhz band. Any ideas?

hmm, to be honest I only have an UNO to test it out. I think I have an idea why the DUO is not working. Let me trying to put something together.

johngag:
@thetooth Ok, I posted a fix for you to try. Let me know

thetooth:
Tested and no change. There are no entries in the RX log or any errors/messages in the device log when sending from another device. I tested a completely fresh one since the DUO I was using had been set to 2.4Ghz before and maybe it was stuck in that mode but that's not the case.

Looking at the noise floor value, it's -119dB and doesn't change so the RX is completely off, I'll play around with some things and report back if I can get it working.

johngag:
@thetooth Can you give your DUO a try now?

thetooth:
@johngag no luck still. I have been looking into the meshtastic implementation to see if there is a bug fix for LR1121 that we haven't picked up, [#9705](https://github.com/meshtastic/firmware/pull/9705) and #1743 looked promising but no change.

However, I FINALLY got somewhere purely by accident. I have been using the BLE companion for all of the testing and when I forgot to connect the BLE client to the device while looking at the serial output all of a sudden I get:

```
: RX, len=22 (type=2, route=D, payload_len=20) SNR=14 RSSI=-22 score=1000 time=312 hash=5C83E51875E5317B [97 -> CC]
DEBUG: RadioLibWrapper: noise_floor = -106
: TX, len=6 (type=3, route=D, payload_len=4)
DEBUG:  Dispatcher::loop(): WARNING: outbound packed send timed out!
```

So looks like RX is working with the hardware configuration fine, but whenever we have serial companion in the connected state it breaks. Tested the different build modes, BLE, USB, repeater, etc are all affected in the same way as soon as the comms class is active with the client RX stops.

```
DEBUG: App meshcore-flutter connected
: TX, len=22 (type=2, route=D, payload_len=20) [CC -> 97]
DEBUG:  Dispatcher::loop(): WARNING: outbound packed send timed out!
DEBUG:  Dispatcher::checkSend(): CAD busy max duration reached!
: TX, len=22 (type=2, route=D, payload_len=20) [CC -> 97]
DEBUG:  Dispatcher::loop(): WARNING: outbound packed send timed out!
DEBUG:  Dispatcher::checkSend(): CAD busy max duration reached!
--- BLE disconnected here ---
: TX, len=22 (type=2, route=D, payload_len=20) [CC -> 97]
DEBUG:  Dispatcher::loop(): WARNING: outbound packed send timed out!
: RX, len=6 (type=3, route=D, payload_len=4) SNR=13 RSSI=-21 score=1000 time=181 hash=B449B287F11F66B1
```

lbibass:
One quick note: The variant.h file only specifies a 2 MB QSPI flash chip. That's fine for the Uno (according to the specs on the shop page) but the Duo uses a 16 MB chip according to the schematic on their site. But the spec sheet says the Duo uses 8MB... and the Meshtastic firmware only calls out a W25Q32JVSS. That is a 4MB chip, and doesn't match any of the rest of the information. I need to take a closer look at the chip that's actually mounted to my Duo this evening, because this just doesn't make any sense.

lbibass:
IMG_7360
my Duo uses the 25Q128JVPQ. 16MB.

lbibass:
Okay. There are a LOT of issues with this PR. The battery pins are incorrect, the user button pins are incorrect, the lr1121 pins appear to be incorrect. I'm working on fixing these issues right now. 

EDIT: the QSPI pins are also entirely incorrect. 

EDIT 2: there's no power enable pin for the SX1262.

lbibass:
[I've opened a draft PR with proper support for the muzi duo](https://github.com/meshcore-dev/MeshCore/pull/2054) if you want to give it a shot @thetooth. The issue was that basically none of the LR1121 radio pins were defined correctly in this PR. I've based my PR off of the Seeed T1000-e code, and I've been running it the last few days. It actually works!

johngag:
> [I've opened a draft PR with proper support for the muzi duo](https://github.com/meshcore-dev/MeshCore/pull/2054) if you want to give it a shot @thetooth. The issue was that basically none of the LR1121 radio pins were defined correctly in this PR. I've based my PR off of the Seeed T1000-e code, and I've been running it the last few days. It actually works!

I will remove duo support from this PR since I didnt have a duo to test with.

PR #1748, opened 2026-02-20
Camp Mode

tmacinc:


tmacinc:
Merged, kept both options for most and will need to confirm operation. Changed custom codes and will need to update app accordingly.

PR #1753, opened 2026-02-20
Fix the bug where no information can be received the first time after…

MGJ520:
Fix issue where no messages can be received after reconfiguring the radio, specifically on the first receive attempt.

For example: 1. First, after the two devices successfully communicate through direct connection, they simultaneously modify the radio configuration. 

2. After modification, only one side is allowed to send information, while the other side will be unable to receive it.


examples/companion_radio/MyMesh.cpp
```diff
@@ -1259,6 +1259,9 @@ void MyMesh::handleCmdFrame(size_t len) {
       savePrefs();
 
       radio_set_params(_prefs.freq, _prefs.bw, _prefs.sf, _prefs.cr);
+      //You need to clear the old registers, otherwise the LoRa module won't receive any information the first time.
+      _radio->onSendFinished();
```
robekl, line 1268: Calling `_radio->onSendFinished()` unconditionally from app-layer reconfiguration code is unsafe when a TX is still in progress. In the dispatcher flow, send completion drives packet accounting/release (`src/Dispatcher.cpp:62-79`). For radios like ESP-NOW, `onSendFinished()` just flips completion state (`src/helpers/esp32/ESPNOWRadio.cpp:90-92`), so this can mark an unsent/unfinished packet as sent and release it prematurely on the next dispatcher cycle.  
Why this matters:
This is a correctness issue (possible packet loss with false success accounting) and can be timing-dependent, making field failures hard to reproduce.  
Suggested direction:
Use a dedicated “re-enter RX after reconfigure” API that does not mutate TX-completion state, or gate reconfiguration on “no outbound TX in progress” before forcing RX restart.

MGJ520: ok,I understand, thank you for reviewing.

MGJ520: But it seems that the CMD_SET_RADIO_PARAMS command doesn't take these issues into account either.

MGJ520: And ESPNOW will not run here to set up the radio configuration.

PR #1757, opened 2026-02-20
Fix bus arrival time display and sorting

wokhouse:
## Summary
- Fixed ISO8601 parser to handle UTC 'Z' suffix format returned by 511.org API (was only handling `-08:00` offset format, causing 8-hour display error)
- Calculate Unix timestamps manually to avoid `mktime()` timezone issues on ESP32
- Sort arrivals by time (soonest first) before display
- Reduce display rows from 4 to 3 to match typical API response

## Problem
The bus arrival display was showing buses arriving in 8 hours when they were actually arriving in a few minutes. This was caused by two issues:
1. The 511.org API returns timestamps with `Z` suffix (UTC) but our parser only handled `-08:00` offset format
2. `mktime()` on ESP32 was incorrectly converting UTC times due to local timezone assumptions

## Solution
- Rewrote `parseISO8601()` to manually calculate Unix timestamps, treating input as UTC
- Added proper handling for both `Z` suffix and timezone offset formats
- Added arrival sorting by minutes (ascending) to show soonest buses first

## Test plan
- [x] Built and flashed firmware to Heltec E290 device
- [x] Verified arrival times match SF MTA website
- [x] Confirmed sorting shows soonest buses first

### DIFF
--- a/examples/bap_firmware/BAPAPI.cpp
+++ b/examples/bap_firmware/BAPAPI.cpp
@@ -87,11 +87,12 @@ int BAPAPIClient::fetchArrivals(uint32_t stop_id, BusArrival* arrivals, int max_
 
   MESH_DEBUG_PRINTLN("Fetching path: %s", path.c_str());
 
-  // Use WiFiClient directly for more control
-  WiFiClient client;
+  // Use WiFiClientSecure for HTTPS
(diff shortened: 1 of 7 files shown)


wokhouse:
sorry--will fix cc!

PR #1758, opened 2026-02-20
Update admin password command description

jeroenvermeulen:
Tested in v1.13.
These commands don't work anymore:
```
get password
  -> ??: password
set password secret
  -> unknown config: password secret
```
This does work:
```
password secret
  -> password now: secret
```

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -293,10 +293,9 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 
 ---
 
-#### View or change this node's admin password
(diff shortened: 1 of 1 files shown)


liamcottle:
I'm not sure that `get password` and `set password` were ever commands that we supported.

From memory, only `password <new password>` existed.

Whoever wrote those docs might have been confused with the guest password and assumed it worked the same, since you can in fact do:

`get guest.password` and `set guest.password <new password>`, but the admin password is restricted to being changed only. It can't be fetched via cli.

Will merge, thanks :)

PR #1770, opened 2026-02-21
Add memory protection and additional thread safety

robekl:
AI based analysis found these potential issues. Fixes included. Each item is in a separate commit to allow ease of cherry-picking. 

## Summary

This PR addresses the confirmed, active issues from the AI-assisted review and removes low-risk/theoretical items from the actionable set.  
The fixes focus on memory safety, bounds validation, and thread/concurrency safety in callback-driven paths.

### 1. Route path payload length was not validated before parsing/copying

**Why this is an issue:**  
Incoming route/path payloads could declare lengths beyond the supported maximum. Without strict validation, this can lead to out-of-bounds reads/writes, memory corruption, and unstable behavior under malformed or hostile input.

**Mitigation in this PR:**  
Added strict length checks before route-path parsing and copy operations, rejecting oversize payloads early and keeping all operations within defined bounds.

---

### 2. Fixed-size temporary CLI buffers used unbounded string copies

**Why this is an issue:**  
Unbounded copies into stack buffers can overflow with long input tokens, causing memory corruption and potential crashes.

**Mitigation in this PR:**  
Replaced unsafe copies with bounded copy operations sized to destination buffers, ensuring null termination and preventing overflow from oversized input.

---

### 3. Thermistor interpolation had edge-condition failures (OOB/div-by-zero)

**Why this is an issue:**  
Interpolation logic could hit out-of-range table access or divide-by-zero on edge values, resulting in invalid readings, undefined behavior, or crashes.

**Mitigation in this PR:**  
Reworked interpolation to clamp edges, validate intermediate values, and guard zero-denominator paths. This preserves expected behavior while eliminating unsafe math/index cases.

---

### 4. ESP-NOW bridge secret handling allowed unbounded length scanning and zero-key path hazards

**Why this is an issue:**  
Length discovery that can walk beyond expected bounds risks invalid memory reads. Separately, XOR-style operations with an empty key can produce invalid behavior (e.g., modulo-by-zero patterns or ineffective protection).

**Mitigation in this PR:**  
Introduced bounded secret-length handling, enforced safe key assumptions, and guaranteed explicit null termination on loaded secret values so cryptographic helper logic always operates on valid, bounded data.

---

### 5. ESP-NOW RX shared buffer had callback/main-loop data races

**Why this is an issue:**  
Concurrent access to a shared RX buffer from callback context and loop context can corrupt frames, produce partial reads, and cause nondeterministic behavior under load.

**Mitigation in this PR:**  
Added synchronization around shared RX state and copy operations, including bounds-aware transfer semantics to ensure consistent frame visibility across producer/consumer contexts.

---

### 6. Bridge RX callback performed non-trivial processing in callback context

**Why this is an issue:**  
Doing heavy packet parsing/queueing directly in callback context increases reentrancy risk, contention risk, and susceptibility to timing-related bugs.

**Mitigation in this PR:**  
Refactored RX flow to a lightweight callback mailbox/hand-off model, with full packet processing moved to the main loop context. This serializes processing and reduces callback-side concurrency hazards.

---

### 7. BLE serial queue access was unprotected across callbacks and loop

**Why this is an issue:**  
Send/receive queue mutation from multiple contexts without locking can cause queue corruption, dropped frames, and intermittent hard-to-reproduce faults.

**Mitigation in this PR:**  
Added mutex protection for queue operations and adjusted pop/copy patterns so data is moved safely between callback and loop code paths.

---

#### Issue
`Mesh::onRecvPacket` had multiple branches where fixed payload fields were read before validating that the payload was large enough for those reads.  
This affected direct and flood parsing paths (TRACE, CONTROL, ACK, ADVERT, PATH/REQ/RESPONSE/TXT, ANON_REQ, and GRP payloads).

#### Why this matters
In real deployments, short/truncated packets can occur due to RF corruption, partial bridge delivery, or malformed input from peers.  
Reading before validating length can lead to out-of-bounds access and undefined behavior, including:
- incorrect ACK/path/message processing
- non-deterministic forwarding decisions
- potential crashes on constrained embedded targets

#### What changed
Length checks were moved/added so every fixed-field parse is guarded before access:

- TRACE direct path now requires minimum header bytes before reading `tag/auth/flags`.
- CONTROL direct fast-path now checks payload has at least 1 byte before inspecting flags.
- Direct ACK fast-path now requires 4 bytes before CRC extraction.
- Main ACK parser now validates length before CRC read.
- PATH/REQ/RESPONSE/TXT parser now validates minimum 2-byte hash header before reads.
- ANON_REQ parser now validates minimum destination-hash + pubkey + MAC + ciphertext-presence before reads.
- Group payload parser now validates minimum channel-hash + MAC + ciphertext-presence before reads.
- ADVERT parser now validates minimum pubkey + timestamp + signature before parsing.

#### Behavioral impact
- Valid packets: behavior is unchanged.
- Invalid short packets: now deterministically rejected/logged instead of being partially parsed from unsafe memory.

#### Why this is correct per protocol docs
The guards align with documented payload structure requirements:
- ACK requires 4-byte checksum.
- ADVERT requires fixed `pubkey(32) + timestamp(4) + signature(64)` header.
- PATH/REQ/RESPONSE/TXT require destination/source hash bytes before MAC/ciphertext.
- ANON_REQ requires destination hash + sender pubkey + MAC before ciphertext.
- Group payloads require channel hash + MAC before ciphertext.
- CONTROL requires at least one flags byte.

---

## Validation performed

Compilation was verified with PlatformIO across affected targets/paths:

- `Heltec_v2_repeater_bridge_espnow`
- `Heltec_v2_companion_radio_ble`
- `t1000e_repeater`
- `t1000e_companion_radio_ble`

### DIFF
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -46,6 +46,10 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
 
   if (pkt->isRouteDirect() && pkt->getPayloadType() == PAYLOAD_TYPE_TRACE) {
     if (pkt->path_len < MAX_PATH_SIZE) {
+      if (pkt->payload_len < 9) {
+        MESH_DEBUG_PRINTLN("%s Mesh::onRecvPacket(): incomplete TRACE packet", getLogDateTime());
+        return ACTION_RELEASE;
+      }
       uint8_t i = 0;
       uint32_t trace_tag;
       memcpy(&trace_tag, &pkt->payload[i], 4); i += 4;
@@ -69,7 +73,8 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
     return ACTION_RELEASE;
   }
 
-  if (pkt->isRouteDirect() && pkt->getPayloadType() == PAYLOAD_TYPE_CONTROL && (pkt->payload[0] & 0x80) != 0) {
+  if (pkt->isRouteDirect() && pkt->getPayloadType() == PAYLOAD_TYPE_CONTROL
+      && pkt->payload_len > 0 && (pkt->payload[0] & 0x80) != 0) {
     if (pkt->path_len == 0) {
       onControlDataRecv(pkt);
     }
@@ -80,11 +85,12 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
   if (pkt->isRouteDirect() && pkt->path_len >= PATH_HASH_SIZE) {
     // check for 'early received' ACK
     if (pkt->getPayloadType() == PAYLOAD_TYPE_ACK) {
-      int i = 0;
-      uint32_t ack_crc;
-      memcpy(&ack_crc, &pkt->payload[i], 4); i += 4;
-      if (i <= pkt->payload_len) {
+      if (pkt->payload_len >= 4) {
+        uint32_t ack_crc;
+        memcpy(&ack_crc, pkt->payload, 4);
         onAckRecv(pkt, ack_crc);
+      } else {
+        MESH_DEBUG_PRINTLN("%s Mesh::onRecvPacket(): incomplete direct ACK packet", getLogDateTime());
       }
     }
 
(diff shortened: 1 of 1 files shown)


recrof:
1. MeshCore firmware doesn't use FreeRTOS and doesn't plan to.
2. we don't have time to read and test LLM generated code
3. we never merge monster PRs that cover multiple issues. 
4. we never cherry-pick any changes from large PRs.

fschrempf:
> MeshCore firmware doesn't use FreeRTOS and doesn't plan to.

The Arduino core frameworks for ESP32 and NRF52 (and probably other MCUs) are just wrappers around FreeRTOS. Just because you usually don't see it doesn't mean you are not using it.

robekl:
> we never merge monster PRs that cover multiple issues.
>  we never cherry-pick any changes from large PRs.

would you rather that I create 8 individual PRs?
@recrof

PR #1774, opened 2026-02-21
Improve duplicate suppression hot paths and remove redundant outbound queue scans

robekl:
This PR applies two focused performance improvements in packet processing and scheduling paths.

### 1) Packet-hash caching with faster duplicate-table prefiltering
- Packet hash computation is now cached on the packet object and reused when packet content is unchanged.
- The cache is explicitly invalidated at mutation/reuse boundaries so duplicate checks always reflect current packet state.
- Duplicate detection now uses small constant-size prefilters (for packet-hash prefix and ACK low-byte) to quickly reject obvious misses before any full table scan.
- Existing exact-match duplicate checks remain in place when a prefilter indicates a possible hit.

### 2) Outbound scheduler no-double-scan path
- The send loop no longer performs a separate “ready count” scan before selecting the next outbound packet.
- Scheduling now relies on a single retrieval pass, preserving existing timing, CAD, and transmit behavior.

### Why this is beneficial

- Reduces repeated CPU work in high-frequency duplicate-suppression paths by:
  - avoiding repeat hash recomputation for unchanged packets
  - avoiding many unnecessary full duplicate-table scans
- Reduces scheduler overhead by removing redundant linear queue traversal.
- Keeps memory usage static and embedded-friendly with simple fixed-size structures.

### Reviewer-focused implementation notes

- Duplicate prefilters are intentionally conservative:
  - They can produce false positives (which fall through to exact checks),
  - but do not introduce false negatives for duplicate detection.
- Hash cache correctness is maintained by explicit invalidation where packet fields are updated or packet instances are recycled.
- Behavior is unchanged for packet matching semantics and transmit gating; this PR targets execution cost, not routing policy.

### DIFF
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -224,7 +224,6 @@ void Dispatcher::processRecvPacket(Packet* pkt) {
 }
 
 void Dispatcher::checkSend() {
-  if (_mgr->getOutboundCount(_ms->getMillis()) == 0) return;  // nothing waiting to send
   if (!millisHasNowPassed(next_tx_time)) return;   // still in 'radio silence' phase (from airtime budget setting)
   if (_radio->isReceiving()) {   // LBT - check if radio is currently mid-receive, or if channel activity
     if (cad_busy_start == 0) {


robekl:
low ROI; closing and reconsidering more simple changes to provide benefit

PR #1779, opened 2026-02-21
Add SerialDualInterface that allows ble companions to be used over usb while ble is disconnected

alexsporn:
Right now when the BLE firmware is used there is no way to connect the companion over USB.

This change introduces a `SerialDualInterface` that supports connecting to both BLE and USB. As soon as BLE is connected, the USB serial connection is disabled. When BLE is disconnected USB becomes available again.

This was tested on the nRF52 platform with a T1000-E.

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -43,7 +43,8 @@ static uint32_t _atoi(const char* sp) {
     #endif
   #elif defined(BLE_PIN_CODE)
     #include <helpers/esp32/SerialBLEInterface.h>
-    SerialBLEInterface serial_interface;
+    #include <helpers/SerialDualInterface.h>
+    SerialDualInterface<SerialBLEInterface> serial_interface;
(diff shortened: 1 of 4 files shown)


JDat:
Interesting idea! Sometimes I want that my RP2040 can work both on USB Serial and on "real" serial using GPIO pins...

alexsporn:
Added proper detection of USB serial connection on nrf52 and esp32 platforms so that devices with display properly show `Connected` or the BLE pin.

Tested successfully on:
- NRF52:
  * SenseCap T1000-E
  * Wio Tracker L1 Pro
  * Heltec T114v2 with display
- ESP32:
  * Heltec V4 with display

alexsporn:
Toggling BLE in the UI only disables BLE connection. I also changed ui-new to not show the BLE pin if the connection is disabled.

liamcottle:
Superseded by https://github.com/meshcore-dev/MeshCore/pull/3049

PR #1797, opened 2026-02-22
Add stats-extpower serial CLI command for INA3221 power monitoring

MrAlders0n:
## Summary

Add a new serial-only `stats-extpower` CLI command that reads voltage and current from an INA3221 external power monitoring board. Returns JSON containing only enabled channels (e.g. `{"ch1_voltage_mv":12450,"ch1_current_ma":150}`), or a human-readable error message if no supported hardware is detected.

This enables tools like LetsMesh and other serial monitoring tools to source battery voltage from the INA3221 rather than the built-in ADC, providing accurate power readings for nodes using external BMS, MPPT charge controllers, or other power management boards. By collecting telemetry over the serial connection, repeater health can be monitored without adding traffic to the mesh itself.

## Changes

- **`src/helpers/CommonCLI.cpp/h`** — Register `stats-extpower` command and add virtual callback
- **`src/helpers/SensorManager.h`** — Add `formatExtPowerStats()` virtual method (default returns false)
- **`src/helpers/sensors/EnvironmentSensorManager.cpp/h`** — Implement INA3221 channel reading and JSON formatting
- **`examples/simple_repeater/MyMesh.cpp/h`** — Override `formatExtPowerStatsReply` to delegate to sensor manager
- **`docs/cli_commands.md`** — Document the new command


src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -489,6 +489,31 @@ bool EnvironmentSensorManager::querySensors(uint8_t requester_permissions, Cayen
...
+#if ENV_INCLUDE_INA3221
+  if (!INA3221_initialized) return false;
+  char* dp = reply;
+  *dp++ = '{';
+  bool first = true;
+  for (int i = 0; i < TELEM_INA3221_NUM_CHANNELS; i++) {
+    if (INA3221.isChannelEnabled(i)) {
+      int ch = i + 1;
+      int voltage_mv = (int)(INA3221.getBusVoltage(i) * 1000);
+      int current_ma = (int)(INA3221.getCurrentAmps(i) * 1000);
```
ripplebiz, line 555: Some sort of guard for max reply buffer length would be good here

examples/simple_repeater/MyMesh.cpp
```diff
@@ -1015,10 +1015,16 @@ void MyMesh::formatRadioStatsReply(char *reply) {
...
 
 void MyMesh::formatPacketStatsReply(char *reply) {
-  StatsFormatHelper::formatPacketStats(reply, radio_driver, getNumSentFlood(), getNumSentDirect(), 
+  StatsFormatHelper::formatPacketStats(reply, radio_driver, getNumSentFlood(), getNumSentDirect(),
                                        getNumRecvFlood(), getNumRecvDirect());
 }
 
+void MyMesh::formatExtPowerStatsReply(char *reply) {
+  if (!sensors.formatExtPowerStats(reply)) {
+    strcpy(reply, "No external power monitoring board detected");
```
ripplebiz: To help the JSON parsers consuming this, the error prob should be wrapped in Json: eg
  { "err": "No external ... " }

src/helpers/sensors/EnvironmentSensorManager.cpp
MrAlders0n: Just pushed a change to add a guard. @ripplebiz

examples/simple_repeater/MyMesh.cpp
MrAlders0n: Fixed this, thanks! @ripplebiz

MrAlders0n:
Just wanted to follow up @ripplebiz to see if you had any more issues with this change.

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -540,6 +540,35 @@ bool EnvironmentSensorManager::querySensors(uint8_t requester_permissions, Cayen
 }
 
 
+bool EnvironmentSensorManager::formatExtPowerStats(char* reply) {
+#if ENV_INCLUDE_INA3221
+  if (!INA3221_initialized) return false;
+  char* dp = reply;
+  int remaining = 160;  // max reply buffer length
```
446564, line 547: should reserve two bytes for the null, like a secondary overflow gaurd

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -540,6 +540,35 @@ bool EnvironmentSensorManager::querySensors(uint8_t requester_permissions, Cayen
...
+  remaining--;
+  bool first = true;
+  for (int i = 0; i < TELEM_INA3221_NUM_CHANNELS; i++) {
+    if (INA3221.isChannelEnabled(i)) {
+      int ch = i + 1;
+      int voltage_mv = (int)(INA3221.getBusVoltage(i) * 1000);
+      int current_ma = (int)(INA3221.getCurrentAmps(i) * 1000);
+      if (!first) { *dp++ = ','; remaining--; }
+      int n = snprintf(dp, remaining, "\"ch%d_voltage_mv\":%d,\"ch%d_current_ma\":%d", ch, voltage_mv, ch, current_ma);
+      if (n >= remaining) break;
```
446564, line 558: we should also terminate the string before the break i think

446564 (changes requested):

PR #1808, opened 2026-02-23
Added build flags for M5Stack Unit C6L

callum5892:
Enabled USB-CDC on boot for M5Stack_Unit_C6L_companion_radio_usb to fix serial connection issues


recrof:
@liamcottle @ripplebiz this is safe to be merged

PR #1820, opened 2026-02-24
Update default preset to EU/UK (Narrow)

recrof:


LitBomb:
I have no issue with this PR as this is how it always been done since day one when there wasn't a proper client.

However, a year later, the firmware should be updated to support worldwide by setting no default preset settings, and no transmission until preset is set via USB serial for repeaters and room servers and BLE via an app client.  This should be part of the overall first use user experience.

recrof:
> However, a year later, the firmware should be updated to support worldwide by setting no default preset settings, and no transmission until preset is set via USB serial for repeaters and room servers and BLE via an app client. This should be part of the overall first use user experience.

We hear you and this is planned.

Pytlicek:
+1 for this. On current default preset, I see here and there new devices (for a longer time, not only during setup)
And people can be confused why they are there alone.

formtapez:
You missed to set LORA_CR to 8, which is not included in platformio.ini for some reason.
It defaults now to 5, which does not correspond to the preset "EU/UK (Narrow)".
It is compatible, but much less reliable.

PR #1825, opened 2026-02-24
Feature: Added firmware support for the new Orobas hardware

Cyber8RU7U5:
OROBAS has released a new SEIR (v5).

This device is going into production and these files allow MeshCore to be built for that device.

More information:
[Website](https://orobas.nl/)
[Mail](contact@orobas.nl)

### DIFF
--- a/variants/orobas_seir/target.h
+++ b/variants/orobas_seir/target.h
@@ -105,5 +105,5 @@ extern ESP32RTCClock rtc_clock;
 bool radio_init();
 uint32_t radio_get_rng_seed();
 void radio_set_params(float freq, float bw, uint8_t sf, uint8_t cr);
-void radio_set_tx_power(uint8_t dbm);
+void radio_set_tx_power(int8_t dbm);
 mesh::LocalIdentity radio_new_identity();
--- a/variants/orobas_seir/target.h
+++ b/variants/orobas_seir/target.h
@@ -1,0 +1,109 @@
+#pragma once
+
+#include "esp_mac.h"
+
+#include <helpers/ESP32Board.h>
+#include <helpers/SensorManager.h>
+#include <helpers/radiolib/CustomSX1262Wrapper.h>
+#include <helpers/sensors/LocationProvider.h>
+
+// GPS
+#define HAS_GPS         1
+
+// BATTERY
+#define BATTERY_PIN     0
+#define BATTERY_DIVIDER 2.43f
+
+//--------------------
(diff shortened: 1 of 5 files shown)


variants/orobas_seir/target.h
```diff
@@ -0,0 +1,109 @@
...
+    uint32_t now = millis();
+  }
+
+  bool querySensors(uint8_t requester_permissions, CayenneLPP &telemetry) override {
+    int next_available_channel = TELEM_CHANNEL_SELF + 1;
+
+    // Location
+    if ((requester_permissions & TELEM_PERM_LOCATION)) {
+      if (_loc && _loc->isValid()) {
+        telemetry.addGPS(TELEM_CHANNEL_SELF, node_lat, node_lon, node_altitude);
```
robekl, line 73: do values from `_loc` need to be copied into `node_lat`, `node_lon`, `node_altitude`?

variants/orobas_seir/target.h
```diff
@@ -0,0 +1,109 @@
...
+
+extern SEIRv5Board board;
+extern CustomSX1262Wrapper radio_driver;
+extern ESP32RTCClock rtc_clock;
+
+/* ---------- Radio ---------- */
+bool radio_init();
+uint32_t radio_get_rng_seed();
+void radio_set_params(float freq, float bw, uint8_t sf, uint8_t cr);
+void radio_set_tx_power(uint8_t dbm);
```
robekl: should `dbm` be `int8_t`?

Cyber8RU7U5: Correct, will modify thanks!

Cyber8RU7U5: ~It does not build with `int8` only with `uint8`, looking into it now. Setting this as unresolved.~

Fixed in 488b404

variants/orobas_seir/target.h
Cyber8RU7U5: They are not copied into those variables. The data is calculated in `src/helpers/SensorManager.h` and used for the telemetry, hope this clarifies it.

liamcottle:
Hi there, thanks for the PR.

Do you have any product listing page with photos of the hardware, or somewhere this hardware can be purchased?

Looking at the links provided, it just says "coming soon".

Unfortunately we aren't looking to merge in board support if users can't actually acquire the units.

It would be best to keep this as an experimental board in your custom repo until users/developers can access the hardware for testing.

PR #1836, opened 2026-02-24
Fix RAK3401 SKY66122-11 FEM control: enable CSD/CPS for proper PA/LNA operation

weebl2000:
preamble: I was talking to someone about their RAK3401 board on Discord and dived into datasheets and current implementation. I asked them to test and come back with findings, but I strongly believe the FEM isn't being controlled properly and this can cause RX sensitivity to be bad and perhaps even cause power/overheating issues.

The RAK13302 1W module uses a Skyworks SKY66122-11 front-end module with three digital control pins (CSD, CTX, CPS) that must be actively driven by the host MCU. The previous code only managed CTX (GPIO 31) — toggling it for TX/RX — but never initialized CSD (GPIO 24) or CPS (GPIO 21), leaving them floating with no pull-up/pull-down resistors on the PCB. CTX (GPIO 31) shouldn't be managed by us manually, since DIO2 handles TX path. 

With floating CSD and CPS, the SKY66122 was in an undefined operating mode:
- The 30 dB TX PA may not have been reliably engaging
- The 16 dB RX LNA was never reliably active, degrading receive sensitivity

I have 4 variants for you to try, but I highly suggest to just try variant #4, it's this PR + the AGC reset functionality from PR #1743 - I think enough testing was done to conclude this is the best variant.

~~1. [This PR, fix SKY66122-11 FEM being driven correctly](https://mcimages.weebl.me?commitId=fix-rak3401-sky66122-11-fem-nopatch)~~
~~2. [This PR, fix SKY66122-11 FEM being driven correctly, and SX126X register patch](https://mcimages.weebl.me?commitId=fix-rak3401-sky66122-11-fem)~~
~~3. [This PR, fix SKY66122-11 FEM being driven correctly + AGC reset SX1262](https://mcimages.weebl.me?commitId=rak3401-agc-and-lna-remote)~~
4. [This PR, fix SKY66122-11 FEM being driven correctly + AGC reset SX1262 + SX126X register patch](https://mcimages.weebl.me/?commitId=fix-sky-fem-and-agc)

**I expect variant number 4 to have the best improvement. When replying with your findings please mention the builds you have tested!**

Following [the comment](https://github.com/meshtastic/firmware/pull/9571#issuecomment-3958036679) from Heltec engineer I also applied the register patch here. It should help with keeping AGC in check.

I don't have a RAK13302 device so I cannot test this myself.

References:

- [RAK13302 docs](https://docs.rakwireless.com/product-categories/wisblock/RAK13302/overview/)
- [SKY66122-11 FEM datasheet](https://www.skyworksinc.com/-/media/SkyWorks/Documents/Products/2901-3000/SKY66122-11_204782G.pdf)
- [SX1262 datasheet](https://semtech.my.salesforce.com/sfc/p/E0000000JelG/a/RQ000008nKCH/hp2iKwMDKWl34g1D3LBf_zC7TGBRIo2ff5LMnS8r19s)
rak13302-schematic
rak13302-power-explanation

### DIFF
--- a/variants/rak3401/RAK3401Board.cpp
+++ b/variants/rak3401/RAK3401Board.cpp
@@ -20,37 +20,27 @@ void RAK3401Board::begin() {
 
   Wire.begin();
 
+  // PIN_3V3_EN (WB_IO2, P0.34) controls the 3V3_S switched peripheral rail
+  // AND the 5V boost regulator (U5) on the RAK13302 that powers the SKY66122 PA.
+  // Must stay HIGH during radio operation — do not toggle for power saving.
   pinMode(PIN_3V3_EN, OUTPUT);
   digitalWrite(PIN_3V3_EN, HIGH);
 
-  // Initialize SKY66122-11 FEM on the RAK13302 module.
-  // CSD (P0.24) and CPS (P0.21) must be HIGH for both TX and RX modes.
-  // CTX (P0.31) selects TX(HIGH) vs RX(LOW) and also enables the 5V boost
-  // converter that powers the PA section (VCC1/VCC2).
-  // The LNA section (VSUP1/VCC0) runs on 3.3V and works with boost off.
-  //
-  // Drive CTX LOW first to prevent transient TX mode (Mode 2) while CSD/CPS
-  // are being enabled — the RAK13302 has no pull-downs on these pins.
-  pinMode(P_LORA_PA_EN, OUTPUT);
-  digitalWrite(P_LORA_PA_EN, LOW);         // CTX=0: RX mode, boost off
-
-  pinMode(P_LORA_PA_CSD, OUTPUT);
-  digitalWrite(P_LORA_PA_CSD, HIGH);       // CSD=1: enable FEM
-
+  // Enable SKY66122-11 FEM on the RAK13302 module.
+  // CSD and CPS are tied together on the RAK13302 PCB, routed to IO3 (P0.21).
+  // HIGH = FEM active (LNA for RX, PA path available for TX).
+  // TX/RX switching (CTX) is handled by SX1262 DIO2 via SetDIO2AsRfSwitchCtrl.
   pinMode(SX126X_POWER_EN, OUTPUT);
-  digitalWrite(SX126X_POWER_EN, HIGH);     // CPS=1: enable TX/RX paths
-
-  delay(1);  // SKY66122 turn-on settling time
+  digitalWrite(SX126X_POWER_EN, HIGH);
+  delay(1);  // SKY66122 turn-on settling time (tON = 3us typ)
 }
 
 #ifdef NRF52_POWER_MANAGEMENT
 void RAK3401Board::initiateShutdown(uint8_t reason) {
-  // Put SKY66122 in guaranteed <1 uA shutdown (Mode 4: CSD=0, CTX=0, CPS=0)
-  digitalWrite(P_LORA_PA_EN, LOW);         // CTX=0, boost off
-  digitalWrite(SX126X_POWER_EN, LOW);      // CPS=0
-  digitalWrite(P_LORA_PA_CSD, LOW);        // CSD=0
+  // Disable SKY66122 FEM (CSD+CPS LOW = shutdown, <1 uA)
+  digitalWrite(SX126X_POWER_EN, LOW);
 
-  // Disable 3V3 switched peripherals
+  // Disable 3V3 switched peripherals and 5V boost
   digitalWrite(PIN_3V3_EN, LOW);
 
   enterSystemOff(reason);
--- a/variants/rak3401/RAK3401Board.h
+++ b/variants/rak3401/RAK3401Board.h
@@ -38,14 +38,6 @@ public:
     return "RAK 3401";
   }
 
-  // SKY66122 FEM TX/RX switching via CTX pin.
-  // CTX=HIGH: TX mode + 5V boost ON (PA powered from VCC1/VCC2)
-  // CTX=LOW:  RX mode + 5V boost OFF (LNA powered from VSUP1 at 3.3V)
-  void onBeforeTransmit() override {
-    digitalWrite(P_LORA_PA_EN, HIGH);  // CTX=1: TX mode, boost on
-  }
-
-  void onAfterTransmit() override {
-    digitalWrite(P_LORA_PA_EN, LOW);   // CTX=0: RX mode, boost off
-  }
+  // TX/RX switching is handled by SX1262 DIO2 -> SKY66122 CTX (hardware-timed).
+  // No onBeforeTransmit/onAfterTransmit overrides needed.
 };
--- a/variants/rak3401/variant.h
+++ b/variants/rak3401/variant.h
@@ -147,14 +147,15 @@ static const uint8_t AREF = PIN_AREF;
 #define SX126X_BUSY (9)
 #define SX126X_RESET (4)
 
-// SKY66122-11 FEM control pins (active HIGH, active LOW = shutdown <1uA)
-// CSD+CPS must be HIGH for TX and RX; CTX selects TX(HIGH) vs RX(LOW)
-// CTX also enables the 5V boost converter for the PA during TX
-#define P_LORA_PA_CSD (24)          // P0.24 -> SKY66122 CSD (pin 11) - FEM enable
-#define SX126X_POWER_EN (21)        // P0.21 -> SKY66122 CPS (pin 1) - path select
-#define P_LORA_PA_EN  (31)          // P0.31 -> SKY66122 CTX (pin 2) - TX/RX + boost EN
+// SKY66122-11 FEM control on the RAK13302 module:
+//   CSD + CPS are tied together on the PCB, routed to WisBlock IO3 (P0.21).
+//   Setting IO3 HIGH enables the FEM (LNA for RX, PA path for TX).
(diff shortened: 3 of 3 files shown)


txkbaldlaw:
I had good results from using variant 4 above.  Stock 3401/13302 1W booster from Rak.  No other add-on boards. 

Firmware appears stable.  Turning off the RXBoost via 'set radio.lna off' results in immediately and reproducible 3-3.5 db gain in advert packet reception from distant repeaters.

variants/rak3401/RAK3401Board.cpp
```diff
@@ -23,10 +23,33 @@ void RAK3401Board::begin() {
...
+
   pinMode(P_LORA_PA_EN, OUTPUT);
-  digitalWrite(P_LORA_PA_EN, LOW);  // Start with PA disabled
-  delay(10);  // Allow PA module to initialize
+  digitalWrite(P_LORA_PA_EN, LOW);         // CTX=0: RX mode, boost off
+
+  delay(1);  // SKY66122 turn-on settling time
+}
+
+#ifdef NRF52_POWER_MANAGEMENT
```
robekl, line 38: is `NRF52_POWER_MANAGEMENT` enabled in the build for rak3401?

variants/rak3401/platformio.ini
```diff
@@ -11,6 +11,7 @@ build_flags = ${nrf52_base.build_flags}
   -D LORA_TX_POWER=22
   -D SX126X_CURRENT_LIMIT=140
   -D SX126X_RX_BOOSTED_GAIN=1
+  -D SX126X_REGISTER_PATCH=1       ; Patch register 0x8B5 for improved RX with SKY66122 FEM
```
robekl, line 14: with this enabled, do the return values of readRegister/writeRegister need to be checked?

robekl:
could we see this problem?

FEM control sequencing in `begin()` can briefly enter TX mode before CTX is forced low.
- In RAK3401Board.cpp:31-35, `CSD` and `CPS` are driven HIGH first.
- `CTX` is only configured/forced LOW at RAK3401Board.cpp:37-38.
- SKY66122 mode table says `CSD=1, CPS=1, CTX=1` is TX mode. If CTX is floating/high during boot, you can get a transient TX/boost-on condition before RX-safe state is asserted.
- maybe set CTX LOW first (or configure pulldown very early), then enable CSD/CPS.

variants/rak3401/platformio.ini
weebl2000: If so, that would be a separate PR. This just enables the functionality.

weebl2000:
> could we see this problem?
> 
> FEM control sequencing in `begin()` can briefly enter TX mode before CTX is forced low.
> 
> * In RAK3401Board.cpp:31-35, `CSD` and `CPS` are driven HIGH first.
> * `CTX` is only configured/forced LOW at RAK3401Board.cpp:37-38.
> * SKY66122 mode table says `CSD=1, CPS=1, CTX=1` is TX mode. If CTX is floating/high during boot, you can get a transient TX/boost-on condition before RX-safe state is asserted.
> * maybe set CTX LOW first (or configure pulldown very early), then enable CSD/CPS.

Agreed, it's probably cleaner that way. I doubt it would have caused any issues currently. At most it would have amplified a bit of noise for a few microseconds. :) Changing it.

variants/rak3401/RAK3401Board.cpp
weebl2000: It's not currently it seems.

weebl2000:
Update from a Discord user who has tested this:

image

Received positive feedback from multiple users so far.

improper-bleep:
I observed a significant improvement with variation 4. When I got my RAK 13302 I noticed right away that Rx performance was worse than my RAK 4631 even though it should be same or better. I never noticed any heat issues.

Comparing Rx performance between the 13302 and the 4631 on the exact same packet transmissions with the 13302 mounted in my attic and the 4631 about 8 feet lower inside my house, the 4631 has better Rx with the both devices running the current 1.13.0 on the web flasher site. So the 4631 outperforms the 13302 even though it doesn't have the filter and is mounted 8 feet lower. Running firmware variant 4 above the 13302 now noticeably outperforms the 4631, as it should given its design and height advantage.

I can post more details of my test results if that would be helpful. I only had time to test the current 1.13.0, variant 1, and variant 4 but I can also test the variants 2 and 3 soon if that would help.

improper-bleep:
I had some more time today to test variants 2 and 3 as well. I get the best results with variants 2 and 4.

I tested by comparing the relative Rx performance on the same packets between a RAK 4631 running stock 1.13.0 and a RAK 13302 running stock 1.13.0 and all 4 variants above. The 13302 is mounted about 8 ft higher but both are in the same house. This is because I already mounted it in my attic but when I had both down in my house running stock 1.13.0 the 4361 received many more packets from distant repeaters than the 13302. Swapping places and antennas made no difference. Tx was/is much better on the 13302 so it seems to be an Rx issue only. To test, I sent 3 messages to a private channel from 9 different locations that I know based on past experience to be near the limits of where the 4631 can receive from. Sometimes it works and sometimes not based on weather, rf conditions, etc. A second companion and phone were left at home to collect results. 4631 had a txdelay of 2.0 set to avoid the repeaters stepping on each other.

In the table below, Scenario 0 is with stock 1.13.0 and scenarios 1-4 are with variants 1-4, respectively. 

🟩Packet heard by RAK 13302 but not RAK 4631
🟢Packet heard by both repeaters
⚪Packet not heard by either repeater
🔴Packet heard by RAK 4631 but not RAK 13302 

|                   |   Scenario 0   |   Scenario 1   |   Scenario 2   |   Scenario 3   |   Scenario 4   |
| ------------ | ------------ | ------------- | ------------- |------------- |------------- |
| Location A | 🔴🔴🟢 | ⚪⚪⚪ | 🟩 🟩⚪ | ⚪⚪🔴 | 🟩 🟩 🟩 | 
| Location B | ⚪⚪⚪ | ⚪⚪⚪ | 🟩 🟩 🟩 | ⚪⚪⚪ | 🟩 🟩 🟩 |  
| Location C | ⚪⚪⚪ | 🟩🟢🟢 | 🟩 🟩 🟩 | ⚪⚪⚪ | 🟩 🟩 🟩 | 
| Location D | 🟢🟢🟢 | 🔴🔴⚪ | 🟢🟢🟢 | ⚪⚪⚪ | 🟢🟢🟢 | 
| Location E | ⚪⚪⚪ | ⚪⚪🟩 | ⚪🟩🟢 | 🟩🟢🟢 | 🟩 🟩 🟩 | 
| Location F | 🟩🟢🟢 | 🔴🔴🟩 | 🟢🟢🟢 | 🔴🔴🟢 | 🟢🟢 🟩 | 
| Location G | 🟢🟢🟢 | ⚪⚪⚪ | 🟢🟢🟢 | 🔴🔴🔴 | 🟢🟢🟢 |  
| Location H | ⚪🔴🟢 | 🔴🔴🔴 | 🟢🟢🟢 | 🔴🔴🟢 | 🟢🟢🟢 |  
| Location I | 🟢🔴🟢 | 🔴🟢🟢 | missed  | 🟢🟢🟢 | 🟢🟢🟢 |  


It's best not to compare between scenarios because they were done on different days, I didn't always get the exact same parking spot, sometimes I was surrounded by large vehicles, and other variables. What I'm really looking at is whether the 4631 heard a packet which the 13302 did not (denoted by 🔴) as that really should not be happening much, if at all, because the 13302 has an 8ft height advantage and an rf filter.

In addition to the tests above, I left the 13302 running on variant 4 overnight and received around 80 packets from distant repeaters which is more than I ever have before on the stock firmware. Usually, with stock firmware, it would be no more than 10 on a good night but there were also many nights with 0.

da808fokai:
image

This is on Variant 4 linked above that was updated with agc-and-lna-remote-registerpatch
AGC reset set at 16

Updated Astor Repeater and tested for ~12hr now.   It's hearing very well vs a comparable observer in capturing all packets, indicating that it is not going deaf for extended seconds at a time.  I've got a hill node LOS that I put up on 1s delay in case I miss a packet while tx at this observer.   

Noise floor has been stable at about -102 with noise spikes correlating with high rx utilization in the graphs.  I'll keep this up for a few days at least and update if any issues arise.

Thanks!

chrisperelstein:
I don't have as rigorous data as other users, but variant 4 made my 3401 test repeater go from being slightly worse than a 4631 right next to it to substantially better. Huge improvement.

variants/rak3401/RAK3401Board.cpp
```diff
@@ -20,13 +20,29 @@ void RAK3401Board::begin() {
...
+void RAK3401Board::initiateShutdown(uint8_t reason) {
+  // Disable SKY66122 FEM (CSD+CPS LOW = shutdown, <1 uA)
+  digitalWrite(SX126X_POWER_EN, LOW);
+
+  // Disable 3V3 switched peripherals and 5V boost
+  digitalWrite(PIN_3V3_EN, LOW);
+
+  enterSystemOff(reason);
```
robekl, line 46: `initiateShutdown()` now overrides the NRF52 shutdown path but does not implement the low-voltage recovery pattern used by other NRF52 variants (`configureVoltageWake(...)` on SHUTDOWN_REASON_LOW_VOLTAGE / SHUTDOWN_REASON_BOOT_PROTECT). If low-voltage shutdown is triggered, this variant can remain in SYSTEMOFF until manual intervention instead of auto-waking on battery recovery.

This might be worth implementing in a subsequent PR.

PR #1844, opened 2026-02-25
Fix for stale gps data being returned when device has no fix

tmacinc:
Adds a gps_fix_valid flag used during telemetry building to ignore gps data if the device doesn't currently have an active fix.


robekl:
should `gps_fix_valid` be set to false in `start_gps()` and/or `stop_gps()`?

liamcottle:
I would probably opt for the "has fix" flag to be returned as a separate telemetry item, rather than hiding the gps data entirely. Otherwise, you might have a tracker enabled, and the gps goes into some trees and loses the gps fix. With this change, you wouldn't be able to retrieve the last known position.

tmacinc:
> I would probably opt for the "has fix" flag to be returned as a separate telemetry item, rather than hiding the gps data entirely. Otherwise, you might have a tracker enabled, and the gps goes into some trees and loses the gps fix. With this change, you wouldn't be able to retrieve the last known position.

That's fair, I did it this way because it's how it responds before a fix is found, or if the companion never gets a fix. When gps is enabled currently, if its zeros it doesn't return gps data, but once a fix is made it returns the last known location forever.

I'll update this tomorrow as I agree that it would be better to add a the qualifier flag.

tmacinc:
> I would probably opt for the "has fix" flag to be returned as a separate telemetry item, rather than hiding the gps data entirely. Otherwise, you might have a tracker enabled, and the gps goes into some trees and loses the gps fix. With this change, you wouldn't be able to retrieve the last known position.

I've updated this to use the CayenneLPP LPP_SWITCH type as it seems the most appropriate for this. Now the gps data is sent anytime the gps is active and the switch confirms if the device has_valid_fix. Let me know if this is what you had in mind. It requires the client to know what the switch represents, but there isn't a more intuitive type that I can see.

liamcottle:
I'm not sure `LPP_SWITCH` makes sense, since the apps/clients wouldn't display it with a correct label. I think this one should probably wait until we've looked into a key/value telemetry type.

That way, this PR could do something like:

```
telemetry.addKeyValue("gps_fix", gps_has_fix ? "1" : "0");
```

And the UI can display the custom label, and the user knows what it means without needing prior knowledge of internal usage of `LPP_SWITCH`.

tmacinc:
I agree this would be best. I was working with what's available in the current library that's being used and since the clients require a lot knowledge anyways its not a stretch to use the switch as long as its documented. If you've got plans for overhauling the telemetry to key/value it would be ideal. But since that would require a major refactor for both the firmware and the clients, does it make sense to add this in the meantime?

The current bug is pretty real. If the companion hasn't had a fix, it doesn't return gps data at all, after it acquires a fix it starts adding data to the telemetry, and if the fix is lost it keeps sending the last known position without any qualifier. If Im using it for tracking, I have no way of knowing if the data is valid after the initial fix has been obtained. That's why I initially made it not return data if the fix isnt valid. It was consistent behavior with the device before obtaining a fix. So you easily know if the telemetry doesn't contain gps data, the device has no GPS or its disabled. This didnt require knowledge of additional info, but you require the client to maintain the last known position.

tmacinc:
It sounds like there is a possible plan for this in a future telemetry update. I'll keep using my fork for now.

PR #1847, opened 2026-02-25
Initial RAK One-Wire implementation for WisMesh Solar Repeater

hb9fxq:
WisMesh Solar Repeater boundle  (RAK9154 Solar Battery + RAK2560_HZB + RAK4630 Module) to report battery pack telemetry in channel 2

- Battery Pack: https://docs.rakwireless.com/product-categories/accessories/rak9154/datasheet/
- https://store.rakwireless.com/products/wismesh-meshtastic-solar-repeater?variant=44485612765382
- Board: https://docs.rakwireless.com/product-categories/wisnode/rak2560/overview/

### DIFF
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -331,6 +331,15 @@ bool EnvironmentSensorManager::begin() {
   }
   #endif
 
+  #ifdef ENV_INCLUDE_ONEWIRE
+  OneWire_initialized = _oneWireHub.begin();
+  if (OneWire_initialized) {
+    MESH_DEBUG_PRINTLN("OneWire SensorHub initialized with %d probe(s)", _oneWireHub.getNumPids());
+  } else {
+    MESH_DEBUG_PRINTLN("OneWire SensorHub: No probes found");
+  }
+  #endif
+
   return true;
 }
 
@@ -480,6 +489,27 @@ bool EnvironmentSensorManager::querySensors(uint8_t requester_permissions, Cayen
         telemetry.addTemperature(TELEM_CHANNEL_SELF, BMP085.readTemperature());
         telemetry.addBarometricPressure(TELEM_CHANNEL_SELF, BMP085.readPressure() / 100);
         telemetry.addAltitude(TELEM_CHANNEL_SELF, BMP085.readAltitude(TELEM_BMP085_SEALEVELPRESSURE_HPA * 100));
+    }
+    #endif
+
+    #ifdef ENV_INCLUDE_ONEWIRE
+    if (OneWire_initialized) {
+      bool has_onewire_data = _oneWireHub.hasVoltage() || _oneWireHub.hasCurrent() || _oneWireHub.hasBatteryPercent() || _oneWireHub.hasTemperature();
+      if (has_onewire_data) {
+        MESH_DEBUG_PRINTLN("OneWire Telemetry ch=%d: V=%d C=%d SOC=%d T=%d",
+          next_available_channel,
+          _oneWireHub.hasVoltage(), _oneWireHub.hasCurrent(),
+          _oneWireHub.hasBatteryPercent(), _oneWireHub.hasTemperature());
+        if (_oneWireHub.hasVoltage())
+          telemetry.addVoltage(next_available_channel, _oneWireHub.getVoltage());
+        if (_oneWireHub.hasCurrent())
+          telemetry.addCurrent(next_available_channel, _oneWireHub.getCurrent());
+        if (_oneWireHub.hasBatteryPercent())
+          telemetry.addPercentage(next_available_channel, _oneWireHub.getBatteryPercent());
+        if (_oneWireHub.hasTemperature())
+          telemetry.addTemperature(next_available_channel, _oneWireHub.getTemperature());
+        next_available_channel++;
+      }
     }
     #endif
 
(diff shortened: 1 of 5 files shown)


hb9fxq:
Post (German) https://community.meshcore.ch/t/rak-wismesh-repeater-battery-pack-telemetrie/230

src/helpers/sensors/OneWireSensorHub.cpp
```diff
@@ -0,0 +1,267 @@
...
+      registerPid(msg[0]);
+      break;
+
+    case SNHUBAPI_EVT_ADD_SID:
+      MESH_DEBUG_PRINTLN("OneWire: Added SID 0x%02X for PID %d", msg[0], pid);
+      break;
+
+    case SNHUBAPI_EVT_SDATA_REQ: {
+      uint8_t ipso_type = msg[0];
+      uint16_t val_len = len - 1;
```
robekl, line 123: can `len` be passed in as 0? would the resulting value for val_len (65536) for the subsequent loop be a problem?

src/helpers/sensors/OneWireSensorHub.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+      uint8_t ordered[256];
+      for (uint16_t i = 0; i < val_len; i += 2) {
```
robekl, line 127: if `val_len` is >= 256, the writes to `ordered` would be beyond the end of its allocated size. can `val_len` be that value?

src/helpers/sensors/OneWireSensorHub.cpp
```diff
@@ -0,0 +1,267 @@
...
+    default:
+      MESH_DEBUG_PRINTLN("OneWire: Unhandled IPSO %d (len=%d)", ipso_type, data_len);
+      break;
+  }
+}
+
+bool OneWireSensorHub::hasVoltage() const { return _has_voltage; }
+float OneWireSensorHub::getVoltage() const { return _cached_voltage; }
+bool OneWireSensorHub::hasCurrent() const { return _has_current; }
+float OneWireSensorHub::getCurrent() const { return (float)_cached_current_ma / 100.0f; }
```
robekl: should this be dividing by 1000 to go from mA to A?

src/helpers/sensors/OneWireSensorHub.cpp
```diff
@@ -0,0 +1,267 @@
...
+      _next_poll_ms = millis() + 500;
+    }
+  }
+}
+
+void OneWireSensorHub::handleEvent(uint8_t pid, uint8_t sid, SNHUBAPI_EVT_E eid,
+                                    uint8_t* msg, uint16_t len) {
+  switch (eid) {
+    case SNHUBAPI_EVT_QSEND:
+      oneWireSerial.write(msg, len);
```
robekl: should the return value of `write()` be checked to handle failed or partial writes?

src/helpers/sensors/OneWireSensorHub.cpp
hb9fxq: To my surprise it is indeed 100

hb9fxq:
Thx @robekl for the inputs, will update probably tonight or on the weekend! Do you also have the HW to verify?

src/helpers/sensors/OneWireSensorHub.cpp
robekl: if the value is 1/100 of an amp, then the variable should not be named "ma" (milli means 1/1000). I would suggest keeping the variable representing ma and divide by 10 when fetching from the hardware, and fix this line to divide by 1000. I prefer consistency, and most other variants use ma, ma is a more common unit, etc

robekl:
> Thx @robekl for the inputs, will update probably tonight or on the weekend! Do you also have the HW to verify?

No, unfortunately I don't have this hardware. Sorry I can't help with testing there.

src/helpers/sensors/OneWireSensorHub.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+  oneWireSerial.begin(9600);
+  RakSNHub_Protocl_API.init(_onewire_callback);
+
+  MESH_DEBUG_PRINTLN("OneWire: Scanning for sensor probes (%d ms)...", ONEWIRE_DISCOVERY_TIMEOUT_MS);
+
+  unsigned long start = millis();
+  unsigned long last_rx = 0;
+
+  while ((millis() - start) < ONEWIRE_DISCOVERY_TIMEOUT_MS) {
```
robekl, line 39: can a `delay()` or `yield()` be added to this busy-wait loop to avoid starving other threads, etc?

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -483,6 +492,27 @@ bool EnvironmentSensorManager::querySensors(uint8_t requester_permissions, Cayen
     }
     #endif
 
+    #ifdef ENV_INCLUDE_ONEWIRE
+    if (OneWire_initialized) {
+      bool has_onewire_data = _oneWireHub.hasVoltage() || _oneWireHub.hasCurrent() || _oneWireHub.hasBatteryPercent() || _oneWireHub.hasTemperature();
```
robekl, line 497: Stale telemetry can be reported indefinitely because `_has_*` flags stay true after first successful parse and there is no freshness/TTL check before publishing. If the probe disconnects or polling fails, old cached values continue to be emitted as current battery data.

The fix for this may be a bigger issue and outside of the scope of this PR.

hb9fxq:
> > Thx @robekl for the inputs, will update probably tonight or on the weekend! Do you also have the HW to verify?
> 
> No, unfortunately I don't have this hardware. Sorry I can't help with testing there.

Thank you for the good code review inputs! I'm about to call this "working fine on my RAKmachine" and will mount the unit permanently soon.

src/helpers/sensors/EnvironmentSensorManager.cpp
hb9fxq: I will look into it, to report a "Trap" only if the age of the data is beond the expected max age.

src/helpers/sensors/OneWireSensorHub.cpp
hb9fxq: looking into it

PR #1850, opened 2026-02-25
Seeed Studio T1000-E: move to new UI

sybrenstuvel:
The compiler is switched from `examples/companion_radio/ui-orig` to
`examples/companion_radio/ui-new`. The handling of the momentary button
was switched from `T1000eBoard::buttonStateChanged()` to a
`MomentaryButton` in `variants/t1000-e/target.cpp`.

Everything still seems to work properly. I briefly tested the room
server and repeater modes, and did a bit more testing of the companion
(USB & BLE) modes.

The desired GPS state is now saved, and survives reboots. The
momentary button also still works properly.

Ref: #989

---------

I've included one little extra change in a separate commit, to remove trailing whitespace and ensure each file (related to the T1000e, I didn't go over the entire source tree) has consistent newlines. Let me know if I can keep that commit in this PR, or if such cleanups should be done in a separate PR.


variants/t1000-e/target.cpp
```diff
@@ -14,7 +14,15 @@ MicroNMEALocationProvider nmea = MicroNMEALocationProvider(Serial1, &rtc_clock);
 T1000SensorManager sensors = T1000SensorManager(nmea);
 
 #ifdef DISPLAY_CLASS
-  NullDisplayDriver display;
+  DISPLAY_CLASS display;
```
robekl: should this match the type of `display` in target.cpp? (both DISPLAY_CLASS)

sybrenstuvel: It should -- I started to clean things up (because what's the point of the `DISPLAY_CLASS` define if it's not used), but then decided against this change to keep this PR simple. But, evidently, I didn't roll it back entirely.

There's more of these issues in the T1000 code, with mismatched `#ifdef`s and other non-uses of the define (like using `NullDisplayDriver` here directly, instead of the defined one).

How about this: I'll remove this change from this PR, and after this has landed, go over `variants/t1000-e` for another cleanup PR.

sybrenstuvel:
I've removed the one unintended change from the PR, so that it's purely about moving to the new UI code now.

weebl2000 (approved):
LGTM.

Tested on my T1000e, can confirm GPS toggle is now persisted. Triple click toggles buzzer off/on correctly.

I'd say this is good to go.

PR #1853, opened 2026-02-25
limit tx power to LORA_ MAX_TX_POWER

lincomatic:
Certain radios, such as eByte E229xxM33S will BLOW UP their PAs if tx power is set too high. Limit max tx power to a configurable level.  if LORA_MAX_TX_POWER isn't set, it uses LORA_TX_POWER

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -4,6 +4,10 @@
 #include "AdvertDataHelpers.h"
 #include <RTClib.h>
 
+#ifndef LORA_MAX_TX_POWER
+#define LORA_MAX_TX_POWER LORA_TX_POWER
+#endif
+
 // Believe it or not, this std C function is busted on some platforms!
 static uint32_t _atoi(const char* sp) {
   uint32_t n = 0;
@@ -92,7 +96,7 @@ void CommonCLI::loadPrefsInt(FILESYSTEM* fs, const char* filename) {
     _prefs->bw = constrain(_prefs->bw, 7.8f, 500.0f);
     _prefs->sf = constrain(_prefs->sf, 5, 12);
     _prefs->cr = constrain(_prefs->cr, 5, 8);
-    _prefs->tx_power_dbm = constrain(_prefs->tx_power_dbm, -9, 30);
+    _prefs->tx_power_dbm = constrain(_prefs->tx_power_dbm, -9, LORA_MAX_TX_POWER);
     _prefs->multi_acks = constrain(_prefs->multi_acks, 0, 1);
     _prefs->adc_multiplier = constrain(_prefs->adc_multiplier, 0.0f, 10.0f);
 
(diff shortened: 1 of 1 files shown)


ViezeVingertjes:
Pretty sure they are designed in such way that the max TX clamps and doesnt do any damage. Although it's good to constain it to avoid confusion, as now people thought they were able to set it to 30 and it actually doing something.

One note; now it cannot be set above `LORA_TX_POWER` if there is no `LORA_MAX_TX_POWER` defined, all devices have to be checked & configured properly, as some might have a lower value than the maximum. Those cannot be set to a normal value after this change if thats the case.

lincomatic:
> Pretty sure they are designed in such way that the max TX clamps and doesnt do any damage. Although it's good to constain it to avoid confusion, as now people thought they were able to set it to 30 and it actually doing something.
> 
> One note; now it cannot be set above `LORA_TX_POWER` if there is no `LORA_MAX_TX_POWER` defined, all devices have to be checked & configured properly, as some might have a lower value than the maximum. Those cannot be set to a normal value after this change if thats the case.

You're pretty sure, but you're wrong. The E22xxxM33S documentation *specifically* says that the PA will blow up if you set it above 9. If you put accidentally put in a firmware that allows higher settings or your tx setting gets corrupted, your M33 will definitely blow up. It's happened to many people.

LORA_MAX_TX_POWER is just an override that sets the max w/o affecting the default value, which is set by LORA_TX_POWER. If you don't set LORA_MAX_TX_POWER, the code makes it equal to LORA_TX_POWER. Just gives more flexibility

ViezeVingertjes:
> You're pretty sure, but you're wrong.

Any PA can be damaged if you overdrive it, there is more to a device than just the PA though. But indeed; i see that of the default variants, Heltec V4 for example allows max power and is safe with it due to using the GC1109 and for the G2 we've defined the max manually. Would have thought the design of more devices would be... more foolproof, but apparantly not.

Back to the change itself:
Instead of always defaulting LORA_MAX_TX_POWER to LORA_TX_POWER, we should probably define `LORA_MAX_TX_POWER` explicitly? E22 30/33 dBm builds already ship with a default that’s not compliant (e.g. >27 dBm) and also the T-Beam 1W it seems, which was added recently. If we lower the default for compliance or any other reason and still rely on “max = LORA_TX_POWER when MAX is missing,” we could accidentally cap devices that are allowed to go higher in some regions.

The other way around would also be possible, that someone thinks its not the maximum and turns it up instead, frying the PA because it's unclear that it's also the max safe value as input for the PA. That would be solved by actually using `LORA_MAX_TX_POWER` which clearly states what it is and what it does.

I think it was discussed in the past, cant remember the outcome but most likely one of the others will. 👍🏼

lincomatic:
> > You're pretty sure, but you're wrong.
> 
> Any PA can be damaged if you overdrive it, there is more to a device than just the PA though. But indeed; i see that of the default variants, Heltec V4 for example allows max power and is safe with it due to using the GC1109 and for the G2 we've defined the max manually. Would have thought the design of more devices would be... more foolproof, but apparantly not.
> 
> Back to the change itself: Instead of always defaulting LORA_MAX_TX_POWER to LORA_TX_POWER, we should probably define `LORA_MAX_TX_POWER` explicitly? E22 30/33 dBm builds already ship with a default that’s not compliant (e.g. >27 dBm) and also the T-Beam 1W it seems, which was added recently. If we lower the default for compliance or any other reason and still rely on “max = LORA_TX_POWER when MAX is missing,” we could accidentally cap devices that are allowed to go higher in some regions.
> 
> The other way around would also be possible, that someone thinks its not the maximum and turns it up instead, frying the PA because it's unclear that it's also the max safe value as input for the PA. That would be solved by actually using `LORA_MAX_TX_POWER` which clearly states what it is and what it does.
> 
> I think it was discussed in the past, cant remember the outcome but most likely one of the others will. 👍🏼

I only put the LORA_MAX_TX_POWER default to LORA_TX_POWER as a safeguard. Yes, LORA_MAX_TX_POWER should be specified in the platformio.ini for the builds where a power limit should be enforced. I kept them separate so to allow flexibility.. there can be a separate default via LORA_TX_POWER from the max LORA_TX_MAX_POWER. The line where it defaults LORA_TX_MAX_POWER to LORA_TX_POWER if LORA_TX_MAX_POWER is undefined is just a safeguard for existing builds that haven't been updated. If it's a big deal to you, I can take it out and just always force LORA_TX_POWER to be the maximum, but do we really want to make the highest allowable power to be the default? IMHO, we should separate the two parameters

PR #1578, opened 2026-02-02
initial port of M5Stack Unit C6L

recrof:
+update pioarduino to newer bugfix release

PR #1884, opened 2026-03-01
Fix M5Stack Unit C6L: correct pin config and add IO expander support

TheRealHaoLiu:
## Summary

Fixes several incorrect hardware configurations in the M5Stack Unit C6L variant introduced in #1578. All corrections verified against the [Meshtastic official variant](https://github.com/meshtastic/firmware/blob/master/variants/esp32c6/m5stack_unitc6l/variant.h) and tested on actual hardware.

- **I2C pins**: SDA=10, SCL=8 (was SDA=16, SCL=17)
- **TCXO voltage**: 3.0V (was 1.8V)
- **TX indicator**: NeoPixel on GPIO 2 (was P_LORA_TX_LED=15)
- **Remove SX126X_RXEN=5**: LNA enable is on PI4IO I2C expander P5, not a direct GPIO
- **Add USB CDC**: Required for serial over USB on ESP32-C6
- **Add PI4IO I2C GPIO expander** (addr 0x43): Controls LoRa reset (P7), LNA enable (P5), RF switch (P6), and button (P0)
- **Fix ESP32-C6 BLE build**: Add missing `#include <esp_mac.h>` for `esp_efuse_mac_get_default()`

> **Note:** This PR contains only the hardware fixes. #1885 is a superset that includes these fixes plus display support and OTA switching. Merge one or the other, not both.

## Test plan

- [x] Repeater build compiles
- [x] Companion BLE build compiles
- [x] Tested on actual M5Stack Unit C6L hardware
- [x] Verified pin configuration matches Meshtastic reference variant

### DIFF
--- a/src/helpers/esp32/SerialBLEInterface.cpp
+++ b/src/helpers/esp32/SerialBLEInterface.cpp
@@ -1,4 +1,5 @@
 #include "SerialBLEInterface.h"
+#include <esp_mac.h>
 
 // See the following for generating UUIDs:
 // https://www.uuidgenerator.net/
--- a/variants/m5stack_unit_c6l/UnitC6LBoard.h
+++ b/variants/m5stack_unit_c6l/UnitC6LBoard.h
@@ -1,15 +1,130 @@
 #pragma once
 
 #include <Arduino.h>
+#include <Wire.h>
 #include <helpers/ESP32Board.h>
 
+// PI4IO I/O Expander (I2C address 0x43)
+// Pin mapping:
+//   P0 = Button (active low)
+//   P1 = (unused input)
+//   P5 = LNA_EN (LNA Enable)
+//   P6 = ANT_SW (RF Switch)
+//   P7 = NRST (LoRa Reset)
+#define PI4IO_ADDR 0x43
+
+// PI4IO registers
+#define PI4IO_REG_CHIP_RESET 0x01
+#define PI4IO_REG_IO_DIR 0x03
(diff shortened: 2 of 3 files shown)


robekl:
I dug through the history on this one and the pin-fix side of the PR looks justified.

The current Unit C6L mapping came in with the original support commit `84e68cf4` ("initial port of M5Stack Unit C6L, update pioarduino to newer bugfix release") and was then reinforced by `9f4eeeec` ("Added build flags for M5Stack Unit C6L"). Those commits established values like `PIN_BOARD_SDA=16`, `PIN_BOARD_SCL=17`, `P_LORA_TX_LED=15`, `P_LORA_RESET=-1`, `SX126X_RXEN=5`, `GPS_RX=4`, and `GPS_TX=5`. I looked through the git history and the original PR context and did not find a hardware-based rationale for those assignments; the initial support PR was effectively just an import of the board support.

That makes the corrections here plausible as a real board-definition fix rather than a speculative rewrite. The authoritative M5Stack Unit C6L docs show a very different hardware layout, including the board pin map and the SX1262/expander wiring, and those docs line up with the direction of this PR: the existing tree appears to treat several board nets as direct GPIOs even though the actual hardware routes them differently.

Authoritative sources:
- https://docs.m5stack.com/en/unit/Unit%20LoRaWAN%20C6L
- https://docs.m5stack.com/en/unit/Unit%20LoRaWAN%20C6L%20PinMap

PR #1885, opened 2026-03-01
M5Stack Unit C6L: display support, dual-firmware OTA switching

TheRealHaoLiu:
## Summary

Complete M5Stack Unit C6L support with correct hardware configuration, display, and dual-firmware OTA switching. This is a superset of #1884 — includes all the pin/IO expander fixes plus additional features.

> **Note:** This PR includes all fixes from #1884. Merge one or the other, not both.

### Hardware fixes (same as #1884)
- Correct I2C pins, TCXO voltage, NeoPixel, USB CDC
- Add PI4IO I2C GPIO expander support
- Remove incorrect SX126X_RXEN

### New features
- **SSD1306 64x48 SPI display driver**: New `SSD1306SPIDisplay` class with fix for Adafruit library missing 64x48 COM pins configuration
- **Dual-firmware OTA switching**: Hold BOOT button during power-on to switch between repeater and companion firmware stored in separate OTA partitions, with buzzer feedback
- **Repeater display UI**: Boot screen, node name with marquee scrolling, radio parameters, I2C expander button wake
- **Companion display UI**: Boot screen, BLE PIN display, node name with marquee scrolling, frequency, unread message count
- **USB companion radio**: Added USB serial companion mode alongside BLE
- **File organization**: Variant-specific UITasks in `companion/` and `repeater/` subdirectories

### Code review fixes
- Removed hardcoded US-only LoRa frequency overrides — inherit defaults from `arduino_base` like all other variants
- Extracted `MarqueeScroller` struct into `DisplayDriver.h` to deduplicate scroll logic between companion and repeater UITasks
- Fixed USB companion display incorrectly showing "Connected" — now shows "USB Ready"

## Test plan

- [x] All four environments build (companion BLE, companion USB, repeater, room server)
- [x] Tested on actual M5Stack Unit C6L hardware
- [x] OTA partition switching works via BOOT button hold
- [x] Display renders correctly on 64x48 SSD1306 OLED
- [x] Marquee scrolling works for long node names
- [x] Button wakes display after auto-off
- [x] Both firmware identities preserved independently in SPIFFS
- [x] USB companion mode connects and works with companion app
- [x] BLE companion mode shows PIN correctly


variants/m5stack_unit_c6l/platformio.ini
```diff
@@ -1,104 +1,132 @@
...
   +<../examples/companion_radio/*.cpp>
+  +<../variants/m5stack_unit_c6l/companion/UITask.cpp>
 lib_deps =
-  ${M5Stack_Unit_C6L.lib_deps}
+  ${m5stack_unit_c6l.lib_deps}
   densaugeo/base64 @ ~1.4.0
   end2endzone/NonBlockingRTTTL@^1.3.0
+  adafruit/Adafruit SSD1306 @ ^2.5.13
 
-[env:M5Stack_Unit_C6L_companion_radio_usb]
```
wbijen, line 93: Thanks for the PR, I was working on the same device!
Was this intentional? The C6L does have USB-C, and the USB companion mode is useful for desktop-tethered setups.

variants/m5stack_unit_c6l/platformio.ini
```diff
@@ -1,104 +1,132 @@
...
+  ${m5stack_unit_c6l.lib_deps}
+  bakercp/CRC32 @ ^2.0.0
+  adafruit/Adafruit SSD1306 @ ^2.5.13
   end2endzone/NonBlockingRTTTL@^1.3.0
+
+[env:m5stack_unit_c6l_room_server]
+extends = m5stack_unit_c6l
+build_flags = ${m5stack_unit_c6l.build_flags}
+  ; LoRa settings - must match other devices on your mesh
+  -D LORA_FREQ=910.525
```
wbijen: No other variant overrides LORA_FREQ/SF/BW/CR — they all inherit from [arduino_base] in the root platformio.ini (869.618 / 62.5 / SF8). These are only first-boot defaults anyway; saved NodePrefs from flash take priority on subsequent boots, and the companion app can reconfigure at runtime. Hardcoding US-only values here is inconsistent with the rest of the repo and could mislead users into thinking a recompile is needed to change regions. Suggest removing these overrides and letting the standard defaults apply.

variants/m5stack_unit_c6l/repeater/UITask.cpp
```diff
@@ -0,0 +1,128 @@
...
+      if (!_display->isOn()) {
+        _display->turnOn();
+      }
+      _auto_off = millis() + AUTO_OFF_MILLIS;
+    }
+    _next_read = millis() + 200;
+  }
+
+  if (_display->isOn()) {
+    // Update marquee scroll
```
wbijen, line 95: The marquee scroll logic is duplicated between companion/UITask.cpp and repeater/UITask.cpp (with inconsistent state storage — class members vs file-scope statics). Since the firmware already has drawTextEllipsized() for static text overflow, consider adding a small MarqueeScroller struct alongside DisplayDriver to handle the animated case. Callers would just hold one as a member and call update(textW, displayW, millis()) to get the pixel offset — no copy-paste needed per variant.

variants/m5stack_unit_c6l/M5StackUnitC6LBoard.h
```diff
@@ -0,0 +1,182 @@
...
+    Wire.beginTransmission(addr);
+    Wire.write(reg);
+    if (Wire.endTransmission() != 0) return false;
+    if (Wire.requestFrom(addr, (uint8_t)1) != 1) return false;
+    if (!Wire.available()) return false;
+    *value = Wire.read();
+    return true;
+  }
+
+  void checkOTASwitch() {
```
wbijen, line 48: Nice feature — OTA partition switching makes good use of the 16MB flash and gives users a way to swap between repeater and companion without reflashing. That said, this feels like it could be its own PR since it's not C6L-specific — any dual-partition ESP32 board could benefit. Splitting it out would make it easier to review the hardware/display changes here independently, and let other variants adopt OTA switching too.

wbijen (commented):
Great work getting the C6L fully functional with display, IO expander, and buzzer support. Left some review comments.

**One thing: I think this should target dev rather than main.**

variants/m5stack_unit_c6l/platformio.ini
TheRealHaoLiu: didn't realize i can use the web client to try usb tethered with, adding it now

variants/m5stack_unit_c6l/platformio.ini
TheRealHaoLiu: fixed

variants/m5stack_unit_c6l/repeater/UITask.cpp
TheRealHaoLiu: fixed

variants/m5stack_unit_c6l/M5StackUnitC6LBoard.h
TheRealHaoLiu: if its ok i will submit a PR to refactor this as a follow up

TheRealHaoLiu:
> Great work getting the C6L fully functional with display, IO expander, and buzzer support. Left some review comments.
> 
> **One thing: I think this should target dev rather than main.**

rebased against dev

wbijen:
Looks goot to me but I don't have merge permissions.

TheRealHaoLiu:
> Looks goot to me but I don't have merge permissions.

any idea who I would need to get some attention from?

skyepn:
Yes please...

skyepn:
Is this build available anywhere for testing? I'm having a hard time building your `main-m5stack-unit-c6l` branch, there seem to be a bunch of missing FreeRTOS dependancies

s4ntz:
any chance that we can have this build for testing?

TheRealHaoLiu:
@skyepn @s4ntz https://github.com/TheRealHaoLiu/MeshCore-M5Burner-UnitC6L this is where i been putting the compiled firmwares

GSzabados:
Any update on this PR to support the M5Stack C6L?

@TheRealHaoLiu thanks for your work, I have seen your repo as well, I will give it a try maybe during the weekend.

TheRealHaoLiu:
Doesn't seem like any maintainer is looking at this 😢

GSzabados:
I know the feeling. I was just recently contributing to Home Assistant, just a simple fix to support some devices, which I even do not own. A simple few line update in a library and then in the core component's code took 4 months. Maybe the new standard in OSS, but it is quite discouraging.

PR #1886, opened 2026-03-01
Adding code to make wifi work better on companions:

NickDunklee:
* Retry handling if disconnected
      * Properly uses WIFI_STA
      * Adding first pass CLI commands for provisioning
      * Gated only for ESP32


NickDunklee:
Dirty push

Issue #1880, opened 2026-02-28
region name memory space max out


jirogit:
Thank you for your contribution to MeshCore!

The repeaters are supposed to hold 32 regions, but it looks like the current limit is way smaller. I use pretty long names, so 12 or so for me.

When the region name memory space maxes out, new region names will be shortened unintentionally.

It looks like you are able to add a new region, but when you come back to the "Manage Regions" menu again, you don't find the new region name you just added, because the region name memory has been maxed out.

recrof:
I don't think it's advisable to use region names this long. MCUs have limited RAM and we might need it for other features.

jirogit:
> I don't think it's advisable to use region names this long. MCUs have limited RAM and we might need it for other features. 

Thank you very much for your advice, recrof!

Does that mean the region name length limit of 30 characters needs to be changed to smaller numbers?

jirogit:
My guess is like below. I will check reality if I have energy.

---

Title: RegionMap::save() always returns true even when file write fails
src/helpers/RegionMap.cpp

Description
RegionMap::save() tracks write success using a local success variable, but ignores its final value and unconditionally returns true at the end of the function. As a result, partial or failed writes are silently reported as successful to the caller.
Affected code (src/helpers/RegionMap.cpp):
cppbool RegionMap::save(FILESYSTEM* _fs, const char* path) {
  File file = openWrite(_fs, path ? path : "/regions2");
  if (file) {
    // ...
    bool success = file.write(...) == ...;
    // success may become false if any write fails
    // ...
    file.close();
    return true;   // ← BUG: always returns true, ignoring 'success'
  }
  return false;
}
Fix:
cpp    file.close();
    return success;   // return the actual result

Impact
The caller in examples/simple_repeater/MyMesh.cpp relies on the return value to report success or failure to the user:
cppbool success = region_map.save(_fs);
strcpy(reply, success ? "OK" : "Err - save failed");
Because save() always returns true, the CLI always replies "OK" even when the region data was not actually written to flash. The user has no way to detect the failure.

Steps to Reproduce

Flash Repeater firmware on a device with limited flash (e.g. Seeed XIAO nRF52840 without QSPI, env:Xiao_nrf52_repeater)
Add regions until flash write begins to fail
Run region save
CLI replies "OK" even though the save failed


Expected Behavior
CLI replies "Err - save failed" when the file write does not complete successfully.
Actual Behavior
[...]

PR #1892, opened 2026-03-01
fix: RegionMap::save() always returned true on write failure

jirogit:
Return 'success' instead of hardcoded 'true' so that failed flash writes are correctly reported to the caller.

I hope this will be helpful for the issue "region name memory space max out #1880".


jirogit:
I tested. This did not work.

PR #1894, opened 2026-03-01
Increase the delay when powering up the T1000-E to avoid hangs on boot

### LINKED ISSUE #1780: T1000-E reboot breaks companion
It seems the filesystem or the prefs get corrupted and only a full erase and re-flash brings the device back.

**Steps to reproduce:**
- Erase Flash
- Flash 1.13.0 BLE Companion
- Connect MeshCore App to companion
- Trigger reboot (device will never be reachable again)

The only way to get the device out of this state is to do a full erase flash.

alexsporn:
It seems that 10ms is too short to allow the LR1110 to power up and stabilize before we call `radio_init()`.
This increases the delay to 150ms which seems to work fine on my devices.

Fixes #1780

### DIFF
--- a/variants/t1000-e/T1000eBoard.cpp
+++ b/variants/t1000-e/T1000eBoard.cpp
@@ -19,5 +19,5 @@ void T1000eBoard::begin() {
 
   Wire.begin();
 
-  delay(10);   // give sx1262 some time to power up
+  delay(150);  // give LR1110 some time to power up
 }


weebl2000:
RadioLib already handles the startup timings of LR1110. This seems like a workaround for a different problem. Waiting 150ms should not really be necessary. I think it's coincidence/placebo 👀 .

alexsporn:
> RadioLib already handles the startup timings of LR1110. This seems like a workaround for a different problem. Waiting 150ms should not really be necessary. I think it's coincidence/placebo 👀 .

I tried it on 2 devices and waiting on boot at this point is the only way to get it to boot consistently.
The device won't boot and will hang. It is pretty easily identified because you won't hear the booting buzzer sound.
See my issue here https://github.com/meshcore-dev/MeshCore/issues/1780
And a new issue by a different user with the same observation: https://github.com/meshcore-dev/MeshCore/issues/1921

I actually found this issue trying to do the same, changing the BLE pin, but found out that the simple reboot already breaks it. Even deploying a new version won't fix it, only a full wipe and restart.

I tried to debug this by waiting on the serial to be connected before continuing the boot (so that I could get all the debug logs while booting), but I could not get it to hang due to the implicit delay I added by waiting on a serial connection.

The waiting a couple of milliseconds longer (150ms is to be safe) on board init consistently fixed this issue.

weebl2000:
Hopefully #1939 fixes this.

alexsporn:
Rebased on latest `1.15.0`. It is still the only way that my 6x T1000e boot successfully after an update/reboot.

Issue #1546, opened 2026-01-31
Contact Settings


drnnyzcvwm-ops:
I’d love to see an auto-add feature in MeshCore that adds peers automatically if they’re within a configurable hop limit (e.g., 2–3 hops). Manually adding 350+ contacts per day is becoming overwhelming.

stevenlafl:
Clarifying this, you don't want hops beyond the limit so you have Auto Add turned off currently?

ghost:
Yes, I have Auto Add turned off. I’m manually adding repeaters and people that are within 3-4 hops of me. If I had Auto Add on, I was getting loads of people up north and contacts would fill, which isn’t ideal. That’s why I was hoping for an auto-add feature that respects a configurable hop limit.

themelisx:
Possible [duplicate](https://github.com/meshcore-dev/MeshCore/issues/1533)

liamcottle:
Implemented in #1900

PR #1900, opened 2026-03-02
Add configurable max hops filter for auto-add contacts

### LINKED ISSUE #1533: [Feature Request] Auto add hop limit
It looks like all the data is available on companions to be able to implement a configurable hop limit for the auto add contact feature.

It would be great if we could set a hop limit for auto add so that only nodes learned under the configured number of hops are auto added to our companion nodes.

wbijen:
## Add configurable max hops filter for auto-add contacts

Closes #1533, relates to #1546

### Problem

Devices auto-add every contact they hear, regardless of distance. In dense or wide meshes this fills the contact list with far-away nodes that are unreliable to reach, pushing out closer, more useful contacts.

### Solution

Add a `autoadd_max_hops` setting that filters auto-add by hop count. Only new contacts within the configured hop limit are auto-added. Value `0` means no limit (current behavior, the default).

- **Only new contacts are filtered** — existing contacts continue to get updated regardless of hop count
- **Filtered contacts still appear in "recently heard"** — `onDiscoveredContact()` is called so the companion app UI can show them, they just aren't persisted to the contact list
- **Configurable from companion app** — extends the existing `CMD_SET/GET_AUTOADD_CONFIG` protocol with an optional 3rd byte

### Protocol change

`CMD_SET_AUTOADD_CONFIG` (0x3A):
- Old: `[cmd] [config_bitmask]` (2 bytes)
- New: `[cmd] [config_bitmask] [max_hops]` (3 bytes, 3rd byte optional)

`CMD_GET_AUTOADD_CONFIG` response:
- Old: `[resp_code] [config_bitmask]` (2 bytes)
- New: `[resp_code] [config_bitmask] [max_hops]` (3 bytes)

Backwards compatible — old companion apps send 2 bytes (max_hops stays 0), and ignore the extra response byte per the length-prefixed frame protocol.

### Files changed

| File | Change |
|------|--------|
| `src/helpers/BaseChatMesh.h` | Add `virtual getAutoAddMaxHops()` (default 0) |
| `src/helpers/BaseChatMesh.cpp` | Hop count check in `onAdvertRecv()` for new contacts |
| `examples/companion_radio/NodePrefs.h` | Add `autoadd_max_hops` field |
| `examples/companion_radio/DataStore.cpp` | Persist new field (offset 88) |
| `examples/companion_radio/MyMesh.h` | Override declaration |
| `examples/companion_radio/MyMesh.cpp` | Override impl + extended SET/GET handlers |

### Testing

- Builds clean on `heltec_v4_companion_radio_usb` (ESP32-S3)
- Default value 0 (no limit) preserves current behavior
- Old prefs files without the new byte: `file.read()` fails silently, field stays 0 via `memset` in constructor

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -1790,7 +1790,7 @@ void MyMesh::handleCmdFrame(size_t len) {
   } else if (cmd_frame[0] == CMD_SET_AUTOADD_CONFIG) {
     _prefs.autoadd_config = cmd_frame[1];
     if (len >= 3) {
-      _prefs.autoadd_max_hops = min(cmd_frame[2], (uint8_t)63);
+      _prefs.autoadd_max_hops = min(cmd_frame[2], (uint8_t)64);
     }
     savePrefs();
     writeOKFrame();
--- a/examples/companion_radio/NodePrefs.h
+++ b/examples/companion_radio/NodePrefs.h
@@ -30,5 +30,5 @@ struct NodePrefs {  // persisted to file
   uint8_t autoadd_config;    // bitmask for auto-add contacts config
   uint8_t client_repeat;
   uint8_t path_hash_mode;    // which path mode to use when sending
-  uint8_t autoadd_max_hops;  // 0 = no limit, 1 = direct (0 hops), N = up to N-1 hops (max 63)
+  uint8_t autoadd_max_hops;  // 0 = no limit, 1 = direct (0 hops), N = up to N-1 hops (max 64)
 };


liamcottle:
Hey,

Thanks for the PR!

This looks pretty good to me, nice and simple.
Just wondering if we should allow for the use case of only allowing zero hop contacts to be added as well.

Currently, the PR uses zero to indicate no limit, since the preferences are always initialized to zero by default.

We could possibly fetch the value from prefs, and if it's 1 or higher, we know we have a limit set, but, subtract 1 from that value to indicate what the limit is.

This would allow for enforcing auto add of direct/zero hop contacts too, as well as any other hop limit is desired.

Options would be:
- 0 (no limit)
- 1 (0 hop / direct only)
- 2 (1-hop limit)

Since it's a protocol/companion level thing, it's up to the clients/apps to know that it works this way.

Ideally, we would use something like -1 or 255 as no limit, and 0 as 0-hop and 1 as 1-hop, but this is just a limitation in the firmware where all values are zero by default.

examples/companion_radio/MyMesh.cpp
```diff
@@ -1785,12 +1789,16 @@ void MyMesh::handleCmdFrame(size_t len) {
     }
   } else if (cmd_frame[0] == CMD_SET_AUTOADD_CONFIG) {
     _prefs.autoadd_config = cmd_frame[1];
+    if (len >= 3) {
+      _prefs.autoadd_max_hops = cmd_frame[2];
```
robekl: `autoadd_max_hops` is accepted as an unconstrained byte, but the protocol/inline comments define valid values as 0 or 1..63. Values 64..255 currently become “effectively no limit” because packet->getPathHashCount() is capped at 63. This makes invalid client input persist as a semantically different config than documented; clamp on write/read (e.g. `min(value, 63)`).

examples/companion_radio/MyMesh.cpp
```diff
@@ -1785,12 +1789,16 @@ void MyMesh::handleCmdFrame(size_t len) {
...
+    }
     savePrefs();
-    writeOKFrame();  
+    writeOKFrame();
   } else if (cmd_frame[0] == CMD_GET_AUTOADD_CONFIG) {
     int i = 0;
     out_frame[i++] = RESP_CODE_AUTOADD_CONFIG;
     out_frame[i++] = _prefs.autoadd_config;
+    out_frame[i++] = _prefs.autoadd_max_hops;
```
robekl, line 1801: does increasing the length of this frame for `CMD_GET_AUTOADD_CONFIG` impact consumers of the frame? eg consider `app_target_ver` or ? or does it not matter?

examples/companion_radio/MyMesh.cpp
wbijen: Good catch — fixed it. Values are now clamped on write: min(cmd_frame[2], (uint8_t)63).

examples/companion_radio/MyMesh.cpp
wbijen: No impact - the protocol is length-prefixed. writeFrame sends a 3-byte header (> + 16-bit length) before the payload ([ArduinoSerialInterface.cpp:30-33](https://github.com/meshcore-dev/MeshCore/tree/main/src/helpers/ArduinoSerialInterface.cpp#L30-L33)). Consumers read the length first, then parse only the bytes they understand. An old app receiving a 3-byte response when it expects 2 will simply ignore the extra byte.

wbijen:
> Hey,
> 
> Thanks for the PR!
> 
> This looks pretty good to me, nice and simple. Just wondering if we should allow for the use case of only allowing zero hop contacts to be added as well.
> 
> Currently, the PR uses zero to indicate no limit, since the preferences are always initialized to zero by default.
> 
> We could possibly fetch the value from prefs, and if it's 1 or higher, we know we have a limit set, but, subtract 1 from that value to indicate what the limit is.
> 
> This would allow for enforcing auto add of direct/zero hop contacts too, as well as any other hop limit is desired.
> 
> Options would be:
> 
> * 0 (no limit)
> * 1 (0 hop / direct only)
> * 2 (1-hop limit)
> 
> Since it's a protocol/companion level thing, it's up to the clients/apps to know that it works this way.
> 
> Ideally, we would use something like -1 or 255 as no limit, and 0 as 0-hop and 1 as 1-hop, but this is just a limitation in the firmware where all values are zero by default.

Thanks! Good call - implemented. The encoding is now:

0 = no limit
1 = direct only (0 hops)
N = up to N-1 hops
Also added a clamp to 64 on write per robekl's suggestion.

PR #1901, opened 2026-03-02
Fix ikoka handheld build

weebl2000:
I noticed it wasn't working: 
image


weebl2000:
@liamcottle Low-hanging fruit, currently the ikoka handheld build does not work at all. This fixes that.

image

PR #1902, opened 2026-03-02
Peek & pop separately when pushing messages to BLE client

weebl2000:
I've noticed very often that the first few messages can get lost after reconnecting with BLE. This should fix that.

Try for yourself by [building here](https://mcimages.weebl.me?commitId=no-fire-and-forget-ble-message-popping)

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -254,6 +254,24 @@ int MyMesh::getFromOfflineQueue(uint8_t frame[]) {
   return 0; // queue is empty
 }
 
+int MyMesh::peekOfflineQueue(uint8_t frame[]) {
+  if (offline_queue_len > 0) {
+    size_t len = offline_queue[0].len;
+    memcpy(frame, offline_queue[0].buf, len);
+    return len;
+  }
+  return 0;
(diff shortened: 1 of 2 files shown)


examples/companion_radio/MyMesh.cpp
```diff
@@ -1224,11 +1242,13 @@ void MyMesh::handleCmdFrame(size_t len) {
     }
   } else if (cmd_frame[0] == CMD_SYNC_NEXT_MESSAGE) {
     int out_len;
-    if ((out_len = getFromOfflineQueue(out_frame)) > 0) {
-      _serial->writeFrame(out_frame, out_len);
+    if ((out_len = peekOfflineQueue(out_frame)) > 0) {
+      if (_serial->writeFrame(out_frame, out_len) > 0) {
```
robekl: should a partial write (`writeFrame(...) > 0` and less than `out_len`) be considered success, eg enough to pop the frame from the queue?

weebl2000: As far as I can tell, it's an all or nothing operation. We either receive 0 or len. No partial writes. But hey it's more future-proof and it should work too.

Issue #1906, opened 2026-03-03
Proposal: Advertise repeat capability via FEAT1 flags (room servers + companions with repeat)


wbijen:
## Problem

Nodes that repeat packets but aren't dedicated repeaters (room servers with `set repeat on`, companions with `client_repeat`) advertise as their base type (`ADV_TYPE_ROOM`, `ADV_TYPE_CHAT`). This causes two issues:

1. **Path display** — hops through these nodes show as "unknown" in companion apps because the app only recognizes `ADV_TYPE_REPEATER` as a valid relay node
2. **Path hash collisions** — when a room server collides with a repeater, the app picks the repeater (wrong node) because it doesn't consider the room server as a valid path hop

Reported in #1866.

## Proposed solution

Use the existing `FEAT1` field in the advert data as a capabilities bitmask. The node keeps its true type and additionally advertises what it can do.

### Advert flags byte (current)

```
bits 0-3: type (CHAT=1, REPEATER=2, ROOM=3, SENSOR=4)
bit 4:    has lat/lon
bit 5:    has feat1 (uint16_t) — currently unused
bit 6:    has feat2 (uint16_t) — currently unused
bit 7:    has name
```

`FEAT1` and `FEAT2` are already wired up in `AdvertDataBuilder`/`AdvertDataParser` but never set by any firmware. Old clients that see the `ADV_FEAT1_MASK` bit just skip 2 bytes — fully backwards compatible.

### Proposed FEAT1 capabilities bitmask

```cpp
#define ADVERT_CAP_REPEAT  0x0001  // this node forwards packets
// bits 1-15: reserved for future capabilities
```

### Firmware changes

**Room server** — `examples/simple_room_server/MyMesh.cpp`:
```cpp
mesh::Packet *MyMesh::createSelfAdvert() {
  uint8_t app_data[MAX_ADVERT_DATA_SIZE];
  AdvertDataBuilder builder(ADV_TYPE_ROOM, ...);
  if (!_prefs.disable_fwd) {
    builder.setFeat1(ADVERT_CAP_REPEAT);
  }
  ...
}
```

**Companion radio** — `examples/companion_radio/MyMesh.cpp`:
```cpp
// In createSelfAdvert(), if client_repeat is enabled:
if (_prefs.client_repeat) {
  builder.setFeat1(ADVERT_CAP_REPEAT);
}
```

[...]

Issue #1866, opened 2026-02-27
Room Servers with repeat not treated as repeaters for pathing and collisions.


MrPatzy:
I have a couple room servers, one at work, a stationary hardwired Heltec V3, and a mobile room.server in my vehicle, RAK 19007. My work is downtown so I wanted to let people near by have a place for messages, and same with my car,  but also to use the store and forward feature for if I miss messages during my commute. Additionally, I need(ed) my mobile room server to repeat just to get a signal out of my office with a hope at getting to the wider mesh. Its relatively crucial for my MC operations.

Anyways ive found that when a message is passed through a room server, when looking at the message path, the hop is labeled as unknown. Seems like if they're sending adverts, they should be able to be labeled. Testing today showed hops going through my stationary node were labeled as unknown even though its in my contacts, and I Remote admin that server from my companion. 

Further complication, my mobile room server has a ID collision with a solar repeater in the area. So because room servers are unknown, MeshCore is labeling the hop with the solar repeater instead. That repeater is a few cities away from where I was testing, and doesnt have LOS into the downtown area. Particularly odd, because in the same message path, I came across a different collision, and they were both repeaters, and the app identified that and showed a message, but didnt do so for the room server/repeater collision. 

![Screenshot_20260226_104312_MeshCore.jpg](https://github.com/user-attachments/assets/2f5ac1cc-2fdd-4892-9ffa-717f4a6624ac)

![Screenshot_20260226_104316_MeshCore.jpg](https://github.com/user-attachments/assets/3710c544-1f1b-4ca9-a963-15add1152d00)

So here you can see the conflict for the fd hop, and the second screenshot is the app telling me which of my contacts have the collision. The e7 hop however doesnt show that message. The e7 repeater that is listed is the one way out of range that my mobile room server has a collision with. There is nothing listed as to a possible collision, and makes it seem like the message is popping way out of the city to come right back. 

I dont think this is how room servers are meant to be treated although I do understand that they occupy a Grey area between repeater and companion. I don't know what the right solution is, but this behavior is hard for understanding mesh pathing especially as MC has taken off in my area in 2026.

n2aws:
This is a known issue when enabling "repeat" on a room server.

The room server firmware doesn't have all of the workings and logic of the repeater firmware,  and the missing route info is a known issue as a result.   This is why "roompeater" configurations are generally not recommended, as it can make troubleshooting the mesh much more difficult when the pathing info it missing.

Issue #1707, opened 2026-02-15
Client-repeat not recognized in message paths


1nv:
When client-repeat mode is enabled on node A with public key 'aa...' and messages are routed via that node, other nodes don't recognize node A as the repeater in message paths, and show it as "unknown repeater aa", even though A's advert has been received before.
The problem is that A is adverted as a companion, which ignores it in the code that searches for a known repeater via its short id. A special flag for client-repeat mode might be introduced to solve that issue.

liamcottle:
Once the firmware supports multi-byte path segments, as discussed in #1613, the app could be updated to include more contact types for name resolution in the repeat path lists. However, with 1-byte IDs, conflicts are too high to allow companions to be included in this at this time.

1nv:
@liamcottle Yeah, I understand that, that's why a change to protocol might be required to explicitly mark nodes capable of repeating (0x0 in Appdata Flags?) for 1-byte (and most likely for 2-byte too) IDs.

PR #1909, opened 2026-03-03
Advertise repeat capability via FEAT1 flags

wbijen:
## Summary

Implements [#1906](https://github.com/meshcore-dev/MeshCore/issues/1906) — nodes that forward packets but aren't dedicated repeaters (room servers with repeat enabled, companions with `client_repeat`) can now advertise this via a bitmask in the existing FEAT1 advert field.

This helps companion apps distinguish relay-capable nodes in path displays and resolves hash collisions where a room server was being confused with a repeater (as reported in #1866 and #1707).

### What changed

- **Added `ADVERT_CAP_REPEAT` (0x0001)** — first bit in the FEAT1 bitmask, meaning "this node forwards packets".
- **`buildAdvertData()`** and **`createSelfAdvert()`** gain an optional `feat1` parameter (default 0, no extra bytes on wire).
- **Room server** sets `ADVERT_CAP_REPEAT` when repeat is enabled (`!disable_fwd`).
- **Companion radio** sets `ADVERT_CAP_REPEAT` when `client_repeat` is on (3 call sites: manual advert, self-export, periodic advert).
- **Dedicated repeater** sets `ADVERT_CAP_REPEAT` for consistency, even though `ADV_TYPE_REPEATER` already implies it.

The generic FEAT1 naming is kept intentionally — the 16-bit field carries different kinds of information (capabilities, flags, properties) depending on context, so a neutral name fits better than something specific.

### Backwards compatibility

Fully backwards compatible in both directions:
- **Old clients** that see the `ADV_FEAT1_MASK` bit simply skip the 2-byte field (already wired up in the parser).
- **Old firmware** sends `feat1 = 0`, so the field is omitted from the wire — no extra bytes.

### Future FEAT1 bits

The uint16_t bitmask has 15 bits remaining. Some candidates:

| Bit | Idea | Rationale | Related |
|-----|------|-----------|---------|
| TBD | `ADVERT_FLAG_OPEN_ROOM` | Room server has no password (or uses default) — apps could show a lock/unlock icon without needing a round-trip and skip the login screen | PR #901 |
| TBD | `ADVERT_FLAG_READONLY` | Room server has `allow_read_only` enabled — apps could indicate guest access is available | |

These don't have concrete firmware/app consumers yet, so they're not included — just noted for future reference.

## Test plan

- [x] Builds cleanly against `heltec_v4_companion_radio_usb`
- [x] Builds cleanly against `heltec_v4_room_server`
- [x] Builds cleanly against `heltec_v4_repeater`
- [x] Builds cleanly against `heltec_v4_sensor` (unchanged, still passes `feat1 = 0` via default)
- [ ] Verify on-air: companion with `client_repeat` on → advert includes FEAT1 field with 0x0001
- [ ] Verify on-air: room server with repeat → advert includes FEAT1 field with 0x0001
- [ ] Verify old companion app still parses adverts correctly (skips unknown 2 bytes)

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -1111,11 +1111,12 @@ void MyMesh::handleCmdFrame(size_t len) {
       writeErrFrame(ERR_CODE_ILLEGAL_ARG);
     }
   } else if (cmd_frame[0] == CMD_SEND_SELF_ADVERT) {
+    uint16_t caps = _prefs.client_repeat ? ADVERT_CAP_REPEAT : 0;
     mesh::Packet* pkt;
     if (_prefs.advert_loc_policy == ADVERT_LOC_NONE) {
-      pkt = createSelfAdvert(_prefs.node_name);
+      pkt = createSelfAdvert(_prefs.node_name, caps);
     } else {
-      pkt = createSelfAdvert(_prefs.node_name, sensors.node_lat, sensors.node_lon);
+      pkt = createSelfAdvert(_prefs.node_name, sensors.node_lat, sensors.node_lon, caps);
     }
     if (pkt) {
       if (len >= 2 && cmd_frame[1] == 1) { // optional param (1 = flood, 0 = zero hop)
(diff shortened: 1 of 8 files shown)


jbrazio:
Please see https://github.com/meshcore-dev/meshcore.js/issues/17

PR #1912, opened 2026-03-03
feat: add basic GPIO support for repeaters and use pin mappings for the Heltec MeshSolar node

MoppelMat:
I cloned the code from the (already merged) simple_sensor folder and put it into the repeater code.
The specific pins are from the Heltec MeshSolar node on its pin header H3 (pin GPIO 30 and 5).

The pins are not used in a normal installation.
Use this picture for reference on this node https://docs.heltec.org/en/node/meshsolar/_images/pinmap.jpg
image

I changed it in a way that you can not only enable or disable a pin, but you can pulse it also.
I needed this, to "press the power button" on my RaspBerry Pi in my repeater case.

Usage in admin CLI:
ior 1 (disable the first pin, GPIO 30 in this case)
ios 1 (enable the first pin, GPIO 30 in this case)
iot 1(toggle the first GPIO pin 30 in this case)
iop 1 (enable the the first GPIO pin 30 in this case and disable it 50ms later)

ior 2 (disable the second pin, GPIO 5 in this case)
...

Specifics to this use case:
I put the second GPIO to a Matek BEC enable pin at a later stage, where I not only could "press the power button" but enable or disable the power supply from the RaspBerry Pi with a GPIO pin.

IMG_1199
IMG_1200
IMG_1209


examples/simple_repeater/MyMesh.cpp
```diff
@@ -1236,7 +1236,35 @@ void MyMesh::handleCommand(uint32_t sender_timestamp, char *command, char *reply
       sendNodeDiscoverReq();
       strcpy(reply, "OK - Discover sent");
     }
-  } else{
+  } else if (memcmp(command, "io", 2) == 0) { 
+      uint32_t val = 0;
+      uint32_t current = board.getGpio();
+      char mode = command[2]; // This is ' ', 'r', 's', 't', or 'p'
+
```
ripplebiz: Would be best to keep the syntax consistent with sensor firmware, ie. "io {r|s|t}mask-value"
Looks like you moved the space ' '.

awhite2000:
How does this work with internal pull-up / pull-down? maybe there should be a persisted config section (plus commands) to specify the pin direction (in / out) and pull-up / pull-down if supported.

Edit: see here: https://docs.espressif.com/projects/arduino-esp32/en/latest/api/gpio.html

MoppelMat:
> How does this work with internal pull-up / pull-down? maybe there should be a persisted config section (plus commands) to specify the pin direction (in / out) and pull-up / pull-down if supported.
> 
> Edit: see here: https://docs.espressif.com/projects/arduino-esp32/en/latest/api/gpio.html

interesting. sounds good, but I am not qualified to do such a change.
is it working for the nRF MCU, too?

awhite2000:
> interesting. sounds good, but I am not qualified to do such a change. is it working for the nRF MCU, too?

Yes, see here: https://docs.nordicsemi.com/bundle/ps_nrf52840/page/gpio.html

variants/heltec_mesh_solar/MeshSolarBoard.h
```diff
@@ -32,4 +32,15 @@ class MeshSolarBoard : public NRF52BoardDCDC {
...
+
+  void setGpio(uint32_t values) override {
+      // set led values
+      digitalWrite(30, values & 1);
+      digitalWrite(5, (values & 2) >> 1);
+  }
+
+  uint32_t getGpio() override {
+      // get led value
+      return (digitalRead(30) << 1) | digitalRead(5);
```
robekl, line 44: setGpio() treats bit 0 of the value as pin 30 and bit 1 as pin 5: values & 1 is written to pin 30, and (values & 2) >> 1 is written to pin 5. But getGpio() reads them back as (digitalRead(30) << 1) | digitalRead(5), which makes pin 30 become bit 1 and pin 5 become bit 0

Issue #1878, opened 2026-02-28
RAK WisBlock 4631 crashes or won’t start when running firmware v1.13.0


Emmpunkt:
Description:
After running my RAK 4631 as a repeater for 9 days without issues, the device suddenly stopped working. It no longer boots reliably. I attempted several restarts and reflashed the firmware multiple times. I also erased the flash completely and tested older firmware versions.

Findings so far:
- Firmware 1.13.0 → device does not start or crashes immediately
- Firmware 1.12.x → same issue
- Firmware 1.11.0 → device works normally again
- Battery voltage is 4.11 V, so low power should not be the cause
- Device behaves as if it is stuck in sleep mode
Question:
Has anyone experienced similar issues with firmware 1.13.0 on the RAK 4631?
Any hints on what might cause the device to fail after running for several days?

I went down to 1.10. because of an other Issue. (BME260)

IoTThinks:
Do you use the CLI "powersaving on"?

Emmpunkt:
Yes, i had "powersavings on" , but switched it off before downgrading

IoTThinks:
So "powersaving on" is not the issue.
Firmware 1.13.0 → device does not start or crashes immediately

Do you wire your battery to battery pads or via usb-c?
You may tell us how you do the wiring from panel to board to battery.

Emmpunkt:
I´m using the battery-plug, the solar-plug is not used. 
If i unplug the battery and use usb-c, it starts   (as i remember ;) ) If it helps you, i will check it again to be shure...

Wiring:
Solar panel -> CN3795 (MPPT-Charger) -> Battery
Battery -> Battery Plug

The thing that i don´t understand:
It was running for 9 days. I remember that this happend before with 1.12 or 1.11, but was something like 14 days. But there was a low battery and after charging and resetting it was working again.
But now, even with deleting the flash, i was not able to start it again. Is there something staying in the memory even with emty flash?

IoTThinks:
> Wiring: Solar panel -> CN3795 (MPPT-Charger) -> Battery 
> Battery -> Battery Plug

This setup should be fine for NRF52 Power Saving management in 1.13 as you wire Battery -> Battery Plug
Have you check the Battery plug connection? It may loose.

Emmpunkt:
The (silikon) wires where soldered directly to the pins because i had not space in my box.  But i put a connector inline yesterday after this happened.
The measured voltage was OK. Measured directly on the pins. 
[...]

PR #1914, opened 2026-03-04
Fix rak4631 boot with DC/DC

### LINKED ISSUE #1879: RAK WisBlock 4631 reports incorrect temperature in telemetry starting with firmware v1.11.
Description:
My RAK 4631 is sending incorrect temperature values in the telemetry data. The reported temperature is consistently about 8°C higher than the temperature measured by the attached BME sensor.
This issue appears to start with firmware v1.11.0.
When I downgrade to v1.10.0, the temperature readings are correct again.
Observed behavior:
- Firmware 1.11.0 and newer → temperature is ~8°C too high
- Firmware 1.10.0 → correct BME temperature
- It looks as if the device is reporting the CPU temperature instead of the BME sensor temperature
Question:
Is anyone else seeing incorrect temperature readings with firmware 1.11.0 or later?
Could there be a regression where the wrong sensor source is used?

weebl2000:
Only enable DC/DC converter when SoftDevice is managing power.

fixes #1879 

Build here https://mcimages.weebl.me/?commitId=fix-rak4631-boot&variant=RAK_4631_repeater

### DIFF
--- a/src/helpers/NRF52Board.cpp
+++ b/src/helpers/NRF52Board.cpp
@@ -245,13 +245,15 @@ bool NRF52Board::configureVoltageWake(uint8_t ain_channel, uint8_t refsel) {
 void NRF52BoardDCDC::begin() {
   NRF52Board::begin();
 
-  // Enable DC/DC converter for improved power efficiency
+  // Enable DC/DC converter only when SoftDevice is managing the POWER peripheral.
+  // Without SoftDevice, NRF_POWER->DCDCEN = 1 leaves the DC/DC permanently on,
+  // which causes boot failures on some boards when running on battery alone.
+  // The SoftDevice manages DC/DC dynamically (on during radio, off during idle),
+  // which is the safe way to use it.
   uint8_t sd_enabled = 0;
   sd_softdevice_is_enabled(&sd_enabled);
   if (sd_enabled) {
     sd_power_dcdc_mode_set(NRF_POWER_DCDC_ENABLE);
-  } else {
-    NRF_POWER->DCDCEN = 1;
   }
 }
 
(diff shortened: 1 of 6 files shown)


weebl2000:
This fixes the problem @Emmpunkt was having - not being able to boot with DC/DC on.

src/helpers/NRF52Board.cpp
```diff
@@ -235,19 +250,23 @@ void NRF52Board::configureVoltageWake(uint8_t ain_channel, uint8_t refsel) {
...
+  // Without SoftDevice, NRF_POWER->DCDCEN = 1 leaves the DC/DC permanently on,
+  // which causes boot failures on some boards when running on battery alone.
+  // The SoftDevice manages DC/DC dynamically (on during radio, off during idle),
+  // which is the safe way to use it.
   uint8_t sd_enabled = 0;
   sd_softdevice_is_enabled(&sd_enabled);
   if (sd_enabled) {
     sd_power_dcdc_mode_set(NRF_POWER_DCDC_ENABLE);
-  } else {
-    NRF_POWER->DCDCEN = 1;
```
fschrempf, line 246: I guess this will increase the power consumption of all NRF52 repeaters (running without softdevice) by something like 20 to 30%.

I have a board with RAK4630 (uart.cz Solar Mesh Baseboard) running with DCDC enabled for several months now without any issues. And a lot of other people are running similar RAK4630 and RAK4631 hardware without such issues.

So I really wonder what is different for @Emmpunkt in https://github.com/meshcore-dev/MeshCore/issues/1878 and if this change is justified.

IoTThinks: If "powersaving on" is used, repeaters dont need this DC2DC to achieve 6mA.

This DC2DC is more meaningful to companions and repeaters without powersaving on.

I guess DC2DC is less reliable at low voltage than the default LDO way.

IoTThinks:
So besides disabling DC2DC, this PR is to fix the NRF52 power management too?

Will disabling of DC2DC alone can fix this problem?

PR #1923, opened 2026-03-04
Add SNR-driven dynamic coding rate selection for repeater retransmits

weebl2000:
When a repeater forwards a direct-routed pkt, it looks up the next-hop neighbor's SNR, computes the link margin against the SF-specific demodulation floor, then determines coding rate (CR) based on SNR. Weaker links get higher CR. We apply CR on a per-packet basis and restore to default configuration after TX.

```
Margin thresholds:
<3 dB   - CR 4/8
3-6 dB  - CR 4/7
6-10 dB - CR 4/6
>=10 dB - CR 4/5
```

Flood repeats adhere to configured CR, as do forwards to unknown neighbors.

If you want to try this you can [BUILD HERE](https://mcimages.weebl.me/?commitId=snr-drives-cr).

I saw this was on the roadmap so I had a go at it. For now we leave flood repeats as-is. In future we might do something smart with average SNR of all our neighbors, but let's take small steps.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -100,7 +100,7 @@ int8_t MyMesh::findNeighbourSNR(const uint8_t* hash, uint8_t hash_size) {
 }
 
 // Approximate SNR demod floor per SF (same as RadioLibWrappers.cpp)
-static float cr_snr_thresholds[] = {
+static const float cr_snr_thresholds[] = {
   -7.5f,  // SF7
   -10.0f, // SF8
   -12.5f, // SF9
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -1,5 +1,6 @@
 #include "MyMesh.h"
 #include <algorithm>
+#include <climits>
 
 /* ------------------------------ Config -------------------------------- */
 
@@ -85,6 +86,42 @@ void MyMesh::putNeighbour(const mesh::Identity &id, uint32_t timestamp, float sn
   neighbour->heard_timestamp = getRTCClock()->getCurrentTime();
   neighbour->snr = (int8_t)(snr * 4);
 #endif
+}
+
+int8_t MyMesh::findNeighbourSNR(const uint8_t* hash, uint8_t hash_size) {
+#if MAX_NEIGHBOURS
+  for (int i = 0; i < MAX_NEIGHBOURS; i++) {
+    if (neighbours[i].heard_timestamp != 0 && neighbours[i].id.isHashMatch(hash, hash_size)) {
+      return neighbours[i].snr;
+    }
+  }
+#endif
+  return INT8_MAX;
+}
+
+// Approximate SNR demod floor per SF (same as RadioLibWrappers.cpp)
+static float cr_snr_thresholds[] = {
+  -7.5f,  // SF7
+  -10.0f, // SF8
+  -12.5f, // SF9
+  -15.0f, // SF10
+  -17.5f, // SF11
+  -20.0f  // SF12
+};
+
+uint8_t MyMesh::selectCodingRateForPeer(const uint8_t* hash, uint8_t hash_size) {
+  int8_t snr4 = findNeighbourSNR(hash, hash_size);
+  if (snr4 == INT8_MAX) return 0;  // unknown neighbor, use default
+
+  float snr = snr4 / 4.0f;
+  float threshold = (_prefs.sf >= 7 && _prefs.sf <= 12)
+    ? cr_snr_thresholds[_prefs.sf - 7] : -15.0f;
+  float margin = snr - threshold;
+
+  if (margin < 3.0f)  return 8;
+  if (margin < 6.0f)  return 7;
+  if (margin < 10.0f) return 6;
+  return 5;  // good margin, use lightest CR
 }
 
 uint8_t MyMesh::handleLoginReq(const mesh::Identity& sender, const uint8_t* secret, uint32_t sender_timestamp, const uint8_t* data, bool is_flood) {
@@ -957,6 +994,7 @@ void MyMesh::begin(FILESYSTEM *fs) {
 
   radio_driver.setParams(_prefs.freq, _prefs.bw, _prefs.sf, _prefs.cr);
   radio_driver.setTxPower(_prefs.tx_power_dbm);
+  setDefaultCR(_prefs.cr);
 
   radio_driver.setRxBoostedGainMode(_prefs.rx_boosted_gain);
   MESH_DEBUG_PRINTLN("RX Boosted Gain Mode: %s",
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -575,12 +575,12 @@ int MyMesh::calcRxDelay(float score, uint32_t air_time) const {
 }
 
 uint32_t MyMesh::getRetransmitDelay(const mesh::Packet *packet) {
-  uint32_t t = (_radio->getEstAirtimeFor(packet->getPathByteLen() + packet->payload_len + 2) * _prefs.tx_delay_factor);
-  return getRNG()->nextInt(0, 5*t + 1);
+  uint32_t tx_airtime_ms = (estimateTxAirtimeFor(packet->getPathByteLen() + packet->payload_len + 2, packet->_tx_cr) * _prefs.tx_delay_factor);
+  return getRNG()->nextInt(0, 5 * tx_airtime_ms + 1);
 }
 uint32_t MyMesh::getDirectRetransmitDelay(const mesh::Packet *packet) {
-  uint32_t t = (_radio->getEstAirtimeFor(packet->getPathByteLen() + packet->payload_len + 2) * _prefs.direct_tx_delay_factor);
-  return getRNG()->nextInt(0, 5*t + 1);
+  uint32_t tx_airtime_ms = (estimateTxAirtimeFor(packet->getPathByteLen() + packet->payload_len + 2, packet->_tx_cr) * _prefs.direct_tx_delay_factor);
+  return getRNG()->nextInt(0, 5 * tx_airtime_ms + 1);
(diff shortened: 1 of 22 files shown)


robekl:
> determines coding rate (CR) based on SNR

Is that the local node's SNR when receiving packets from the remote node? So the calculation assumes symmetric link SNR's at each end? I'm always leering about making decisions/calculations based on this assumption.

weebl2000:
> > determines coding rate (CR) based on SNR
> 
> Is that the local node's SNR when receiving packets from the remote node? So the calculation assumes symmetric link SNR's at each end? I'm always leering about making decisions/calculations based on this assumption.

Correct, we assume the link is fairly symmetrical. In practice this might not be the case. This can go both ways, either we're spot on with our CR, but it can just as easily be too low or too high. The system will be somewhat self-correcting though, if our CR isn't spot on our SNR will drop for neighbors and thus yield a bump in CR.

I think it's a reasonable starting point for SNR-driven CR. Knowing our neighbor's SNR receive values from our signals is a heavier protocol change. The 3dB margin before dropping to a lower CR seems like a conservative enough choice too.

robekl:
Is there data that shows that each CR has higher success of being received successfully at those SNRs? eg how was it determined that CR 4/5 will work adequately at 10 dB margin vs 20 dB (or some other value), etc?

examples/simple_repeater/MyMesh.cpp
```diff
@@ -87,6 +88,42 @@ void MyMesh::putNeighbour(const mesh::Identity &id, uint32_t timestamp, float sn
...
+    if (neighbours[i].heard_timestamp != 0 && neighbours[i].id.isHashMatch(hash, hash_size)) {
+      return neighbours[i].snr;
+    }
+  }
+#endif
+  return INT8_MAX;
+}
+
+// Approximate SNR demod floor per SF (same as RadioLibWrappers.cpp)
+static float cr_snr_thresholds[] = {
```
robekl: can this be const/constexpr so it stays out of ram? (I think that's how that works)

src/Dispatcher.cpp
```diff
@@ -68,6 +68,9 @@ void Dispatcher::loop() {
       next_tx_time = futureMillis(t * getAirtimeBudgetFactor());
 
       _radio->onSendFinished();
+      if (outbound->_tx_cr != 0 && outbound->_tx_cr != _default_cr) {
+        _radio->setCodingRate(_default_cr);
```
robekl, line 132: what happens if the underlying call fails? the return value is not propagated to here

src/Dispatcher.cpp
```diff
@@ -275,14 +281,21 @@ void Dispatcher::checkSend() {
     } else {
       memcpy(&raw[len], outbound->payload, outbound->payload_len); len += outbound->payload_len;
 
+      if (outbound->_tx_cr != 0 && outbound->_tx_cr != _default_cr) {
```
robekl, line 356: `_tx_cr` can accidentally carry over from an old packet to a new packet because packets are reused from a pool and not always fully reset when reallocated.
At line 284, send logic trusts _tx_cr; if that field is stale, a packet that never asked for a CR override may still transmit with one.

weebl2000:
> Is there data that shows that each CR has higher success of being received successfully at those SNRs? eg how was it determined that CR 4/5 will work adequately at 10 dB margin vs 20 dB (or some other value), etc?

The demodulation floors are from the SX1262 datasheet. The CR-specific thresholds are heuristic starting points based on ~2 dB per CR step approximation (it's probably less). It should be refined with real-world testing.

Without this PR there will be no real-world testing though 😉

robekl:
> self-correcting though, if our CR isn't spot on our SNR will drop for neighbors and thus yield a bump in CR

What's the fallback path? If the remote doesn't decode the packet successfully at the higher CR, will we retry at a lower CR? or just flood? 

I don't understand the part of your statement, "our SNR will drop for neighbors". How will that happen?

src/Dispatcher.cpp
weebl2000: RadioLib's setCodingRate() only fails for out-of-range values (< 5 or > 8). We only ever pass 5–8 from selectCodingRateForPeer(), so it can't fail in practice. Not worth adding error handling for an impossible case IMO.

src/Dispatcher.cpp
weebl2000: Good thought, but this is already covered. obtainNewPacket() resets _tx_cr = 0 on every allocation from the pool (Dispatcher.cpp:327). For received packets that go through raw allocNew(), they enter via onRecvPacket() which calls selectCodingRateForPeer() to set _tx_cr before retransmit — so no stale carryover.

examples/simple_repeater/MyMesh.cpp
weebl2000: yes!

robekl:
Re: testing, what metrics can be analyzed to determine how often this is executed and the success rate?

weebl2000:
> What's the fallback path? If the remote doesn't decode the packet successfully at the higher CR, will we retry at a lower CR? or just flood?
> 
> I don't understand the part of your statement, "our SNR will drop for neighbors". How will that happen?

There is no fallback path really. If a direct-routed packet is sent with too little error correction, a neighbor simply cannot decode it. In a healthy mesh this shouldn't be an issue. Stronger links will pick up the slack and stronger routes will establish more naturally. Mind you, this "deafness" would only occur with highly asymmetrical links. Asymmetrical links are troublesome in their own right, but aren't the most common. Most links imo are fairly symmetrical.

If there is no other route? Your default CR will still apply to any flood routed packages, so if direct messages fail with a lower CR, if you have CR8 configured your flood message might still go through.

Might it break weaker direct links on the edge of working? Yes, maybe. But these links would still work on flood if your default CR is higher than the determined CR. 

Will it encourage stronger links to have preference over weaker links? Also yes definitely. Stronger links that work with lower CR will take less airtime and be more succesful, leading to a healthier mesh.

In the end, the main benefit of this change is not about fixing one specific link, it's about not forcing every link to pay the cost of our **WORST** link.

weebl2000:
> Re: testing, what metrics can be analyzed to determine how often this is executed and the success rate?

In my opinion? Test if the dynamic rate selection works for your repeaters. I.e. does it adjust based on observed SNRs from neighbors? If so: deploy this to the mesh as a whole and see what it does. Are direct links more succesful, less airtime spent? There is no way of analysing the net result of this by looking at just one node.

We definitely also need improvements like #1691 - our companions are way too eager to spam the network currently.

robekl:
The recommended CR varies... in UK it's 8 but in US it's 5. I can see how going "faster" (lower CR value, eg fewer bits of correction) for strong links can reduce airtime, etc, and be beneficial. I can see how going "slower" (higher CR value, more bits) can help marginal links get messages through, to reduce the number of retries or dropped messages, at the cost of one slower direct message (hop). But if the slower message still fails, and the default (in the case of US/default=CR5) is less resilient, would this still be helpful? Is the dynamic adjustment beneficial in theory to both default settings? I need to think about this more...

robekl:
A separate topic, slightly related... how about automatically increasing CR on very-low airtime nodes?

robekl:
Are calculated airtime estimates something we need to consider with this change (or any change to make CR dynamic per packet)?

AI analysis:

**Impact on retransmit delays**
- `getDirectRetransmitDelay()` and `getRetransmitDelay()` still use `getEstAirtimeFor(...)` *before* Dispatcher applies per-packet CR override.
- In PR flow, override is decided in Mesh, delay is computed next, but radio CR is not switched until send stage.
- Result: delay jitter windows are estimated using default CR, not per-packet CR.
  - If packet gets bumped to `CR7/CR8`, true airtime is longer than delay model expected -> tighter-than-intended retransmit spacing, more collision risk.
  - If forced to `CR5` from higher default, opposite bias (more conservative delay than needed).

**Impact on send-timeout scaling**
- Dispatcher outbound timeout (`outbound_expiry = now + 1.5 * est_airtime`) is now computed after setting per-packet CR, so this timeout tracks the actual selected CR better than before.
- Airtime budget backoff (`next_tx_time`) is based on measured TX duration, so duty-cycle spacing also naturally reflects chosen CR.
- App-level message timeout formulas (chat/client `calcFloodTimeoutMillisFor` / `calcDirectTimeoutMillisFor`) are unchanged by PR and still based on estimated airtime in existing code paths; PR mainly changes repeater forwarding behavior.

**Net effect**
- PR improves correctness for TX-stage timeout/budget handling under dynamic CR.
- PR introduces a mismatch in retransmit-delay estimation vs actual per-packet airtime (most visible for direct forwarded packets with CR override), which can affect contention behavior.
- Secondary edge case: repeater sets Dispatcher default CR once at startup; temporary runtime radio CR changes may not update that stored default, so post-send CR restore may snap back to stale default.

terminalvelocity23:
I'll give it a try, thanks.

robekl:
@weebl2000 I had a try at implementing the timing calculations based on the dynamic CR. please have a look:

https://github.com/weebl2000/meshcore/pull/38

terminalvelocity23:
FWIW this PR is pretty stable, I've been running it over a few weeks and it didn't cause any issues. Can't say anything about the impact on the performance tho.

robekl:
`getDirectRetransmitDelay()` is still calculated using `_radio->getEstAirtimeFor(...)` before the per-packet CR override is applied. The actual radio CR is only switched later in `Dispatcher::checkSend()`. That means the jitter window is based on default CR, not the real packet airtime. If a packet is bumped to CR7/8, its true airtime is longer than the spacing model expects, which increases overlap/collision risk for repeater forwards.
Straightforward fix: compute retransmit delay using the selected CR, either by passing the effective CR into the airtime estimator or by adding a helper that estimates airtime for a packet length plus explicit CR.

robekl:
`findNeighbourSNR()` accepts any neighbor entry with `heard_timestamp != 0`; it does not check recency. That means CR selection can be driven by an old advertisement-derived SNR long after link conditions changed, which makes the per-packet override stale and potentially wrong in exactly the cases where links are fluctuating.

PR #1924, opened 2026-03-05
SEN0658 weather station for RAK4631

johnterickson:
Adds support for [SEN0658](https://wiki.dfrobot.com/sen0658) Ultrasonic 9-in-1 RS485 Weather Sensor to the RAK4631.

Test configuration:

| [RS485 adapter](https://www.dfrobot.com/product-2392.html) | RAK4630 on [RAK19003 Mini Base Board](https://store.rakwireless.com/products/wisblock-base-board-rak19003) |
| - | - |
| Blue | RX0 |
| Green | TX0 |
| Black | GND |
| Red | + pin of battery connector |

| [SEN0658](https://wiki.dfrobot.com/sen0658) | [RS485 adapter](https://www.dfrobot.com/product-2392.html)  |
| - | - |
| Brown | 12V |
| Black | GND |
| Yellow | RS485 A |
| Blue | RS485 B |

There's not perfect mappings for the sensors so what I have is:

| Match | Telemetry Sensor | Actual Sensor |
| - | - | - |
| YES | Temperature | " |
| YES | Humidity | " |
| YES | Air Pressure | " |
| YES | Luminosity | " |
| NO | Concentration | PPM 10 |
| NO | Distance (m) | Wind Speed (m/s) |
| NO | Direction (degrees) | Wind Direction (degrees) |
| NO | Analog Input | PPM 2.5 |
| NO | Generic | Noise dB |

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -1800,15 +1800,11 @@ void MyMesh::handleCmdFrame(size_t len) {
   } else if (cmd_frame[0] == CMD_GET_CUSTOM_VARS) {
     out_frame[0] = RESP_CODE_CUSTOM_VARS;
     char *dp = (char *)&out_frame[1];
-    for (int i = 0; i < sensors.getNumSettings() && dp - (char *)&out_frame[1] < 140; i++) {
-      if (i > 0) {
-        *dp++ = ',';
-      }
-      strcpy(dp, sensors.getSettingName(i));
-      dp = strchr(dp, 0);
-      *dp++ = ':';
-      strcpy(dp, sensors.getSettingValue(i));
-      dp = strchr(dp, 0);
+    char *end = (char *)out_frame + sizeof(out_frame);
+    for (int i = 0; i < sensors.getNumSettings(); i++) {
+      if (i > 0) *dp++ = ',';
+      dp += snprintf(dp, end - dp, "%s:", sensors.getSettingName(i));
+      dp += sensors.getSettingValue(i, dp, end - dp);
     }
     _serial->writeFrame(out_frame, dp - (char *)out_frame);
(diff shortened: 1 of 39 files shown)


KPrivitt:
I found that all the existing sensors use Channel 1 to display their values.

If multiple sensors are installed that utilize the same "add" function for example:

telemetry.addTemperature()

The CayenneLLP software will only display one.  Any sensor that adds another Temperature will not be displayed and it's silently dropped.  Only one Temperature can be displayed in a channel.

Since they all use Channel one, I found that hard coding my values to Channel 2 would work fine.  I presume that your next_available_channel will always be greater than one, so there will be no conflict and your values are displayed together and all will be displayed.

johnterickson:
@KPrivitt yeah I am unsure about how multiple sensors are supposed to work on the same device. Some sensors look like they are written to be hardcoded to a certain channel (so it's predictable) and others (like mine) err on "playing nice" with other sensors (but adding one sensor might change the channel of another). I'm not sure what's better. 🤷‍♂️

My RAK4631 already has a built in temp + voltage on channel 1 and my thought was to keep different "things" (repeater itself, weather station) on different channels, but to try to keep any one "thing" as just one channel - that way multiple instances of a sensor (some of the RAK demos have 4 soil sensors) each have exactly 1 channel. I feel it keeps it tidy but I admit it's an arbitrary opinion. 😄 The downside here is that I have two concentration values and I had to force one in to a type other than "concentration". Actually, now that i am talking through it, maybe it would be better to just to say that "each instance of a sensor type has a hardcoded number of values" and that "each instance of a sensor should start a new channel for a different 'thing'". What I mean by that is that the RAK barometer sensor board would use the SELF channel because it is part of the mesh radio system itself, but the barometer in my external weather station should be on a different channel from SELF.

That said, these are all just thoughts I had while reading through the code by myself so I don't know how much they align with the project.

PR #1928, opened 2026-03-05
feat: Add support for PAYLOAD_TYPE_GRP_DATA

dz0ny:
- Docs changes are to reflect how it is currently in fw
- This adds ability to send datagram data to everyone in channel

### DIFF
--- a/docs/companion_protocol.md
+++ b/docs/companion_protocol.md
@@ -295,7 +295,7 @@ Bytes 4+: Binary payload bytes (variable length)
 
 **Data Type / Transport Mapping**:
 - `0x0000` is invalid for this command.
-- `0xFFFF` (`DATA_TYPE_CUSTOM`) is the generic custom-app namespace.
+- `0xFFFF` (`DATA_TYPE_DEV`) is the developer namespace for experimenting and developing apps.
 - Other non-zero values can be used as assigned application/community namespaces.
 
 **Note**: Applications that need a timestamp should encode it inside the binary payload.
--- a/src/helpers/TxtDataHelpers.h
+++ b/src/helpers/TxtDataHelpers.h
@@ -6,7 +6,7 @@
 #define TXT_TYPE_PLAIN          0      // a plain text message
 #define TXT_TYPE_CLI_DATA       1      // a CLI command
 #define TXT_TYPE_SIGNED_PLAIN   2      // plain text, signed by sender
-#define DATA_TYPE_CUSTOM        0xFFFF // generic custom app namespace for group/channel datagrams
+#define DATA_TYPE_DEV           0xFFFF // developer namespace for experimenting with group/channel datagrams and building apps
 
 class StrHelper {
 public:


liamcottle:
Howdy, thanks for the PR!

We have a previously opened PR for similar functionality here https://github.com/meshcore-dev/MeshCore/pull/1530.

We left some review comments on it, but it was never followed up so didn't get merged.
Please have a read through for more info.

I noticed in this PR you have merged both group text and group data into the single `CMD_SEND_GROUP_TEXT` command, personally I'd prefer to have separate commands.

- `CMD_SEND_CHANNEL_TXT_MSG`
- `CMD_SEND_CHANNEL_DATA`

We also need to have a distinct payload type at the start of the raw group data as well, so we can reserve certain types for specific tasks.

I have some more info about this in the previous PR comments.

dz0ny:
Ah, the SOS, reaction thing yep. Pushed changes, so one can specify type and then binary 0xff is for custom apps.

examples/companion_radio/MyMesh.cpp
```diff
@@ -564,6 +565,41 @@ void MyMesh::onChannelMessageRecv(const mesh::GroupChannel &channel, mesh::Packe
 #endif
 }
 
+void MyMesh::onChannelDataRecv(const mesh::GroupChannel &channel, mesh::Packet *pkt, uint32_t timestamp, uint8_t txt_type,
```
liamcottle: please refactor `txt_type` to `data_type`, since this isn't intended for text based data at this time.

examples/companion_radio/MyMesh.cpp
```diff
@@ -564,6 +565,41 @@ void MyMesh::onChannelMessageRecv(const mesh::GroupChannel &channel, mesh::Packe
...
+                               const uint8_t *data, size_t data_len) {
+  int i = 0;
+  if (app_target_ver >= 3) {
+    out_frame[i++] = RESP_CODE_CHANNEL_MSG_RECV_V3;
```
liamcottle: we should probably use a new `RESP_CODE_CHANNEL_DATA_RECV`

examples/companion_radio/MyMesh.cpp
```diff
@@ -564,6 +565,41 @@ void MyMesh::onChannelMessageRecv(const mesh::GroupChannel &channel, mesh::Packe
...
+
+  uint8_t channel_idx = findChannelIdx(channel);
+  out_frame[i++] = channel_idx;
+  out_frame[i++] = pkt->isRouteFlood() ? pkt->path_len : 0xFF;
+  out_frame[i++] = txt_type;
+  memcpy(&out_frame[i], &timestamp, 4);
+  i += 4;
+
+  size_t available = MAX_FRAME_SIZE - i;
+  if (data_len > available) data_len = available;
```
liamcottle: I would probably prefer to have an error returned if provided data is too long, rather than truncating and sending anyway

examples/companion_radio/MyMesh.cpp
```diff
@@ -1031,26 +1067,48 @@ void MyMesh::handleCmdFrame(size_t len) {
...
       ChannelDetails channel;
-      bool success = getChannel(channel_idx, channel);
-      if (success && sendGroupMessage(msg_timestamp, channel.channel, _prefs.node_name, text, len - i)) {
+      if (!getChannel(channel_idx, channel)) {
+        writeErrFrame(ERR_CODE_NOT_FOUND); // bad channel_idx
+      } else if (sendGroupMessage(msg_timestamp, channel.channel, _prefs.node_name, text, text_len)) {
         writeOKFrame();
       } else {
-        writeErrFrame(ERR_CODE_NOT_FOUND); // bad channel_idx
+        writeErrFrame(ERR_CODE_TABLE_FULL);
```
liamcottle: we should probably revert all the changes made to `CMD_SEND_CHANNEL_TXT_MSG` as they're not relevant for the new `CMD_SEND_CHANNEL_DATA`. Changes to the error codes and lookup should be a separate PR :)

examples/companion_radio/MyMesh.cpp
```diff
@@ -1031,26 +1067,48 @@ void MyMesh::handleCmdFrame(size_t len) {
...
       }
     }
+  } else if (cmd_frame[0] == CMD_SEND_CHANNEL_DATA) { // send GroupChannel datagram
+    int i = 1;
+    uint8_t txt_type = cmd_frame[i++];
```
liamcottle: Please refactor `txt_type` to `data_type`

examples/companion_radio/MyMesh.h
```diff
@@ -137,6 +137,8 @@ class MyMesh : public BaseChatMesh, public DataStoreHost {
                            const uint8_t *sender_prefix, const char *text) override;
   void onChannelMessageRecv(const mesh::GroupChannel &channel, mesh::Packet *pkt, uint32_t timestamp,
                             const char *text) override;
+  void onChannelDataRecv(const mesh::GroupChannel &channel, mesh::Packet *pkt, uint32_t timestamp, uint8_t txt_type,
```
liamcottle: Rename `txt_type` to `data_type`.

src/helpers/BaseChatMesh.cpp
```diff
@@ -353,8 +353,10 @@ int BaseChatMesh::searchChannelsByHash(const uint8_t* hash, mesh::GroupChannel d
 #endif
 
 void BaseChatMesh::onGroupDataRecv(mesh::Packet* packet, uint8_t type, const mesh::GroupChannel& channel, uint8_t* data, size_t len) {
+  if (len < 5) return;
```
liamcottle: Probably need some debug log here so we know data was dropped.

src/helpers/BaseChatMesh.cpp
```diff
@@ -454,6 +460,26 @@ bool BaseChatMesh::sendGroupMessage(uint32_t timestamp, mesh::GroupChannel& chan
   return false;
 }
 
+bool BaseChatMesh::sendGroupData(uint32_t timestamp, mesh::GroupChannel& channel, uint8_t txt_type, const uint8_t* data, int data_len) {
```
liamcottle: Rename `txt_type` to `data_type`.

src/helpers/BaseChatMesh.cpp
```diff
@@ -454,6 +460,26 @@ bool BaseChatMesh::sendGroupMessage(uint32_t timestamp, mesh::GroupChannel& chan
...
+  if (data_len < 0) return false;
```
liamcottle: Probably need debug log here too

src/helpers/BaseChatMesh.cpp
```diff
@@ -454,6 +460,26 @@ bool BaseChatMesh::sendGroupMessage(uint32_t timestamp, mesh::GroupChannel& chan
...
+  // createGroupDatagram() accepts at most (MAX_PACKET_PAYLOAD - CIPHER_BLOCK_SIZE)
+  // plaintext bytes; subtract our 5-byte {timestamp, txt_type} header.
+  const int max_group_data_len = (MAX_PACKET_PAYLOAD - CIPHER_BLOCK_SIZE) - 5;
+  if (data_len > max_group_data_len) data_len = max_group_data_len;
```
liamcottle: Same as previous comment, we should probably not truncate the data, and instead return an error code and log the error.

src/helpers/BaseChatMesh.cpp
```diff
@@ -454,6 +460,26 @@ bool BaseChatMesh::sendGroupMessage(uint32_t timestamp, mesh::GroupChannel& chan
...
+
+  uint8_t temp[MAX_PACKET_PAYLOAD];
+  memcpy(temp, &timestamp, 4);
+  temp[4] = txt_type;
+  if (data_len > 0) memcpy(&temp[5], data, data_len);
+
+  auto pkt = createGroupDatagram(PAYLOAD_TYPE_GRP_DATA, channel, temp, 5 + data_len);
+  if (pkt) {
+    sendFloodScoped(channel, pkt);
```
liamcottle: currently group text messages are always flood routed to the whole network, I wonder if we should allow for the ability for group data to have an optional path supplied, to allow for targeted group data to a specific area via specific repeaters... maybe something for @ripplebiz to comment on?

src/helpers/BaseChatMesh.h
```diff
@@ -111,6 +111,8 @@ class BaseChatMesh : public mesh::Mesh {
   virtual uint32_t calcDirectTimeoutMillisFor(uint32_t pkt_airtime_millis, uint8_t path_len) const = 0;
   virtual void onSendTimeout() = 0;
   virtual void onChannelMessageRecv(const mesh::GroupChannel& channel, mesh::Packet* pkt, uint32_t timestamp, const char *text) = 0;
+  virtual void onChannelDataRecv(const mesh::GroupChannel& channel, mesh::Packet* pkt, uint32_t timestamp, uint8_t txt_type,
```
liamcottle: please change `txt_type` to `data_type`

src/helpers/BaseChatMesh.h
```diff
@@ -148,6 +150,7 @@ class BaseChatMesh : public mesh::Mesh {
   int  sendMessage(const ContactInfo& recipient, uint32_t timestamp, uint8_t attempt, const char* text, uint32_t& expected_ack, uint32_t& est_timeout);
   int  sendCommandData(const ContactInfo& recipient, uint32_t timestamp, uint8_t attempt, const char* text, uint32_t& est_timeout);
   bool sendGroupMessage(uint32_t timestamp, mesh::GroupChannel& channel, const char* sender_name, const char* text, int text_len);
+  bool sendGroupData(uint32_t timestamp, mesh::GroupChannel& channel, uint8_t txt_type, const uint8_t* data, int data_len);
```
liamcottle: please change `txt_type` to `data_type`

src/helpers/TxtDataHelpers.h
```diff
@@ -6,6 +6,7 @@
 #define TXT_TYPE_PLAIN          0    // a plain text message
 #define TXT_TYPE_CLI_DATA       1    // a CLI command
 #define TXT_TYPE_SIGNED_PLAIN   2    // plain text, signed by sender
+#define TXT_TYPE_CUSTOM_BINARY  0xFF // custom app binary payload (group/channel datagrams)
```
liamcottle: since this will be a new type dedicated to group data, instead of group text, we should prob rename to `DATA_TYPE_CUSTOM`

src/helpers/BaseChatMesh.cpp
```diff
@@ -454,6 +460,26 @@ bool BaseChatMesh::sendGroupMessage(uint32_t timestamp, mesh::GroupChannel& chan
   return false;
 }
 
+bool BaseChatMesh::sendGroupData(uint32_t timestamp, mesh::GroupChannel& channel, uint8_t txt_type, const uint8_t* data, int data_len) {
+  if (data_len < 0) return false;
+  // createGroupDatagram() accepts at most (MAX_PACKET_PAYLOAD - CIPHER_BLOCK_SIZE)
+  // plaintext bytes; subtract our 5-byte {timestamp, txt_type} header.
+  const int max_group_data_len = (MAX_PACKET_PAYLOAD - CIPHER_BLOCK_SIZE) - 5;
```
liamcottle: we should probably move `max_group_data_len` to a constant defined in `Meshcore.h` near `MAX_PACKET_PAYLOAD`. probably named `MAX_GROUP_DATA_LENGTH`

dz0ny:
@liamcottle Should users be allowed to send packets on idx=0 (the public channel)? I think maybe not the 0xff type, but the others should be allowed. What do you think?

I know users want SOS, REACTION types, so those should be allowed. Just custom protocols probably not.

src/helpers/BaseChatMesh.cpp
dz0ny: I think as an optional param, not required. Or maybe hop based, so only allow one hop.But yea heavily depends on use case and how to prevent spamming.

docs/companion_protocol.md
```diff
@@ -391,11 +396,15 @@ Messages are received via the RX characteristic (notifications). The device send
    - `PACKET_CHANNEL_MSG_RECV` (0x08) - Standard format
    - `PACKET_CHANNEL_MSG_RECV_V3` (0x11) - Version 3 with SNR
 
-2. **Contact Messages**:
+2. **Channel Data**:
+   - `PACKET_CHANNEL_DATA_RECV` (0x1B) - Standard format
```
liamcottle: We can probably just have a single type `PACKET_CHANNEL_DATA_RECV` which also includes the SNR and reserved bytes. The reason we have two types for text message packets, is legacy reasons. We didn't have a way to add SNR data to existing response, because there was no data length byte. So we couldn't add it to the end of the payload.

docs/companion_protocol.md
```diff
@@ -556,6 +590,8 @@ Use `CMD_SEND_CHANNEL_TXT_MSG` for plain text, and `CMD_SEND_CHANNEL_DATA` for b
 | 0x10  | PACKET_CONTACT_MSG_RECV_V3 | Contact message (V3 with SNR) |
 | 0x11  | PACKET_CHANNEL_MSG_RECV_V3 | Channel message (V3 with SNR) |
 | 0x12  | PACKET_CHANNEL_INFO        | Channel information           |
+| 0x1B  | PACKET_CHANNEL_DATA_RECV   | Channel data (standard)       |
+| 0x1C  | PACKET_CHANNEL_DATA_RECV_V3| Channel data (V3 with SNR)    |
```
liamcottle: Suggested to remove `PACKET_CHANNEL_DATA_RECV_V3` and only have `PACKET_CHANNEL_DATA_RECV`. SNR would be included in `PACKET_CHANNEL_DATA_RECV`.

examples/companion_radio/MyMesh.cpp
```diff
@@ -92,6 +92,8 @@
 #define RESP_CODE_STATS               24   // v8+, second byte is stats type
 #define RESP_CODE_AUTOADD_CONFIG      25
 #define RESP_ALLOWED_REPEAT_FREQ      26
+#define RESP_CODE_CHANNEL_DATA_RECV   27
+#define RESP_CODE_CHANNEL_DATA_RECV_V3 28
```
liamcottle: No need for `_V3` based on previous review comments

examples/companion_radio/MyMesh.cpp
```diff
@@ -565,27 +568,30 @@ void MyMesh::onChannelMessageRecv(const mesh::GroupChannel &channel, mesh::Packe
 #endif
 }
 
-void MyMesh::onChannelDataRecv(const mesh::GroupChannel &channel, mesh::Packet *pkt, uint32_t timestamp, uint8_t txt_type,
+void MyMesh::onChannelDataRecv(const mesh::GroupChannel &channel, mesh::Packet *pkt, uint32_t timestamp, uint8_t data_type,
                                const uint8_t *data, size_t data_len) {
   int i = 0;
   if (app_target_ver >= 3) {
-    out_frame[i++] = RESP_CODE_CHANNEL_MSG_RECV_V3;
+    out_frame[i++] = RESP_CODE_CHANNEL_DATA_RECV_V3;
```
liamcottle: Rename `RESP_CODE_CHANNEL_DATA_RECV_V3` to `RESP_CODE_CHANNEL_DATA_RECV`.

examples/companion_radio/MyMesh.cpp
```diff
@@ -565,27 +568,30 @@ void MyMesh::onChannelMessageRecv(const mesh::GroupChannel &channel, mesh::Packe
 #endif
 }
 
-void MyMesh::onChannelDataRecv(const mesh::GroupChannel &channel, mesh::Packet *pkt, uint32_t timestamp, uint8_t txt_type,
+void MyMesh::onChannelDataRecv(const mesh::GroupChannel &channel, mesh::Packet *pkt, uint32_t timestamp, uint8_t data_type,
                                const uint8_t *data, size_t data_len) {
   int i = 0;
   if (app_target_ver >= 3) {
```
liamcottle: No need for the target version check, since these are brand new request/response codes, which can be checked against a protocol version bump later on.

examples/companion_radio/MyMesh.cpp
```diff
@@ -565,27 +568,30 @@ void MyMesh::onChannelMessageRecv(const mesh::GroupChannel &channel, mesh::Packe
...
 
   size_t available = MAX_FRAME_SIZE - i;
-  if (data_len > available) data_len = available;
+  if (data_len > available) {
+    MESH_DEBUG_PRINTLN("onChannelDataRecv(): payload_len=%d exceeds frame space=%d, truncating", (uint32_t)data_len, (uint32_t)available);
+    data_len = available;
+  }
   int copy_len = (int)data_len;
   if (copy_len > 0) {
     memcpy(&out_frame[i], data, copy_len);
```
liamcottle, line 595: I feel like we should also have a `data_length` byte that's in the payload so we can in a future update, add extra data at the end if there is room for it. This will help prevent the issue where we needed new packet format versions to add the SNR to txt messages.

src/helpers/BaseChatMesh.cpp
```diff
@@ -460,24 +468,28 @@ bool BaseChatMesh::sendGroupMessage(uint32_t timestamp, mesh::GroupChannel& chan
...
+    MESH_DEBUG_PRINTLN("sendGroupData: data_len=%d exceeds max=%d", data_len, MAX_GROUP_DATA_LENGTH);
+    return false;
+  }
 
-  uint8_t temp[MAX_PACKET_PAYLOAD];
+  uint8_t temp[5 + MAX_GROUP_DATA_LENGTH];
   memcpy(temp, &timestamp, 4);
-  temp[4] = txt_type;
+  temp[4] = data_type;
   if (data_len > 0) memcpy(&temp[5], data, data_len);
```
liamcottle: As mentioned in previous review comment, wondering if we should have a `data_length` byte as well.

This would allow adding extra data at the end later on, as clients shouldn't be reading more than `data_length`. This should also fix the issue where packets have extra zero padding passed back to the client, since the firmware has no idea what the actual data length is. This is a problem we currently have with cayenne lpp. Where there's extra zeros at the end, and clients have to do funky things to know to stop reading.

I would prefer to do this at the `sendGroupData` level, e.g `temp[5] = data_len;` otherwise we would have to add this to all other payload types. Probably best to just make it standard that we know the length up front.

liamcottle:
> Should users be allowed to send packets on idx=0 (the public channel)? I think maybe not the 0xff type, but the others should be allowed. What do you think?

I'll discuss this one with Scott.

src/helpers/BaseChatMesh.cpp
liamcottle: Yeah definitely optional, just thinking it could be useful to be able to send some sort of flood advert locally, but without propagating to the entire network, since you could pick the direct path.

liamcottle:
We should probably go with `uint16_t` for `data_type`, instead of `uint8_t` so we have a larger pool of data types to hand out to the community for custom applications etc. ~256 might be too limiting, and we'd have to be careful how many IDs we give out. An ID could be given to an application/organisation/person, rather than a specific purpose, and the app author can embed their own internal type inside their data payload.

src/helpers/BaseChatMesh.cpp
dz0ny: I was thinking the same. When we deploy, we usually set up a single relay (on a drone or static), and that alone is used for relaying across roughly 10 km², which is also the typical limit of a search area for one group or command.

If you need a larger area, you simply duplicate the setup, since you already have to do so for logistical reasons.

dz0ny:
Addressed the current review items in 2fe3c36

dz0ny:
> We should probably go with `uint16_t` for `data_type`, instead of `uint8_t` so we have a larger pool of data types to hand out to the community for custom applications etc. ~256 might be too limiting, and we'd have to be careful how many IDs we give out. An ID could be given to an application/organisation/person, rather than a specific purpose, and the app author can embed their own internal type inside their data payload.

Sounds a bit premature optimization. App author can wire their own type in paylaod and this would be the only such type while the protocol uses uint8_t.

src/helpers/BaseChatMesh.cpp
```diff
@@ -353,8 +353,18 @@ int BaseChatMesh::searchChannelsByHash(const uint8_t* hash, mesh::GroupChannel d
 #endif
 
 void BaseChatMesh::onGroupDataRecv(mesh::Packet* packet, uint8_t type, const mesh::GroupChannel& channel, uint8_t* data, size_t len) {
-  uint8_t txt_type = data[4];
-  if (type == PAYLOAD_TYPE_GRP_TXT && len > 5 && (txt_type >> 2) == 0) {  // 0 = plain text msg
+  if (type == PAYLOAD_TYPE_GRP_TXT) {
+    if (len < 5) {
```
liamcottle, line 357: old check was `if len > 5` (disallowing exactly 5) and new logic is `len < 5`, which means exactly 5 is now allowed? Will this be problematic?

liamcottle:
> Addressed the current review items in [2fe3c36](https://github.com/meshcore-dev/MeshCore/commit/2fe3c36b8fc249db6cf78f4ed3f852e425f68029)

Looking good! Thank you for the updates :)

liamcottle:
> Sounds a bit premature optimization. App author can wire their own type in paylaod and this would be the only such type while the protocol uses uint8_t.

After discussions with Scott, I think we should go with `uint16_t` so we can maintain a registry of unique app/developer/purpose IDs in the project files.

Meshtastic does similar with their portnums, which goes up to 500+ IDs. `uint8_t` would limit us to ~256.
https://github.com/meshtastic/protobufs/blob/master/meshtastic/portnums.proto

Ideally, in your MeshCore SAR app, the MeshCore project would allocate you a "unique ID" from the `uint16_t` ~65,535 possible types pool. And then you have free reign to allocate your own internal IDs for your own custom payload types in your raw payload data, without other custom applications causing you trouble.

It was noted that if everyone is using `0xFF` for custom data, then you will end up with other applications sending unexpected data to your application. For example if you expect `0xFF` (custom data) and the next byte to be `0x01` for "image data", someone using `0xFF` for custom data in a different application would trigger your image handler if their first custom byte happens to be `0x01`. So, we need to be able to separate other applications from each other. 

I also think it would be beneficial to provide a larger pool of development/prototype ids, such as `0xFF00` - `0xFFFF`, which would help to avoid these issues during testing. Once the project allocates a unique ID, you wouldn't have garbage data from other apps using the same development IDs.

I would suggest to go with the larger pool from day one. I'd expect we would eat up ~256 unique application/type ids pretty quickly at the rate we are seeing networks grow, and new custom applications come about.

One could argue that we could just use a custom channel key to segregate applications, such as "#meshcore-sar", but I can see usefulness in being able to send "well known" payload types to any channel. For example, we could have a new "share contact" payload, and allow that to be sent to any channel.

src/helpers/BaseChatMesh.cpp
liamcottle: I can see usefulness in being able to share a group packet within a small area, either zero hop, or via 1 repeater in the area without needing to flood the entire network.

We could probably add a new command to send with custom path in the future. It would have to limit the max length of custom data though, due to the limitation of `MAX_FRAME_SIZE` between the firmware and app/client.

Just wondering if it should be allowed from day one. Would remove the need for clients to check firmware versions when it's added. If path is supplied, it uses that with `sendDirect` instead of `sendFlood`.

One use case could be to send an emergency alert group message to `#alerts` in a specific area of the mesh, by passing it through a chain of repeaters until it gets to the destination area.

ripplebiz: Possibly better to use a custom region/scope, but there's also nothing stopping using sendDirect() with GRP_DATA.

ripplebiz:
Could you please change the base branch to 'dev'?

ripplebiz:
Looks good, but I'd rather drop the enforced 'timestamp' field in all the methods (and the new companion protocol parts), and just have a data byte array (with explicit data_len). If various data_types need a timestamp they can just include one.
And, also widen data_type to uint16_t  (and remove the concept of CUSTOM)

dz0ny:
@ripplebiz changed to dev, you can merge now.

dz0ny:
Added support in https://github.com/dz0ny/meshcore-sar/pull/34

liamcottle:
Hey @dz0ny, thanks for your work on this :)
Sorry been a bit slow to move onward, have been away for a bit...

All is looking good, and we have approved the changes, but we are looking to add the ability to send these channel data packets as zero hop, and via direct paths (along with the existing flood mode) before it gets merged in.

We've been receiving a lot of feedback about the existing flood traffic on larger meshes, and that it's near impossible to manage/block. Regions serve to help here, but at the moment the onboarding of this is a bit difficult.

I believe by adding this new feature, that will be well used by other applications to send large/arbitrary data, we will run into the issue where flood traffic is amplified even larger, with no way to contain it to nearby users if that's all that's needed.

So, ideally, from day one we will need to have the ability to allow these group data packets to be sent zero hop, or via a specific set of repeaters to reach an area without flooding the entire network. And, if needed users can flood as a fallback if zero or direct doesn't meet their use case.

My proposal is to add a couple of extra parameters to the send channel data command. It would accept a path length, and path data, if the path length is `0xFF` as used in other places, it would send as flood. If the path length is `0x00` it would send zero hop, and any other variation of path length would use the provided path bytes and send via `sendDirect` instead of `sendFlood`.

Is this something you'd be willing to add into the PR? If not, we will look to add this in before merging :)

Again, thank you very much for your contribution here. I'm sure this feature will be well used. Just need to make sure we aren't flooding the entire mesh if we can avoid it.

liamcottle:
Merging. I'll send a follow up PR with the changes I requested above :)

dz0ny:
@liamcottle I was also thinking of hardcoding 0xffff as a region. That way, any packet on the companion would be assigned to the mc-dev region, and unless explicitly allowed, it could be denied. For other extensions, we could follow a similar approach.

This would give relay admins a tool to block traffic if something goes wrong.

Essentially, on the repeater, we would disallow:
- Alert messages
- Reactions to messages
- All the data stuff ...

If the user sets a region for the channel, even better—we could append it to the final region, e.g., mc-alert-de, mc-react-at, etc.

Issue #1214, opened 2025-12-14
[Feature Request] Add an option to disable LED in repeater


denis1482:
Please add an option to disable LED in repeater to save power.
Heltec T114 being placed at a good high spot in a waterproof case solar powered has no-one to observe it's LEDs, wasting energy.
It would be good for this option not only to disable LED flashing, but also to disable MCU wake-ups to process LED heartbeats.

2E0KYB:
Yes, please, this would be excellent 👍

scottfuhrman:
+1 this would be a nice feature

toshionline3000:
Yes please!

Koreanman01:
I just pushed an update for the Heltec v3 to have the option to disable the LED. I've tested it and it's working with persistance.
Also with the repeaters, commands "screen off", "led off", "screen on", "led on", and "status" are in the push for the v3. 

Going to work on this same thing for the T114 and the v4 now and test it out.
Will try to have an update for the T114 after testing Wednesday.

toshionline3000:
Will the option work for nrf52 as well?
Like for the SenseCap Solar Node. The LED are quite bright.

Koreanman01:
I see the SenseCap is slightly different.
I will check that board out and see what I can do with it.

rdmarsh:
This would be appreciated for the Heltec v3. I’ve set powersaving on and it’s still doing it, Google got me here.

Use case is the repeater is currently in our bedroom until I can roof mount it. It’s the best place in the house to reach more repeaters, but the xyl isn't impressed with the blinkenlights.

omik:
> This would be appreciated for the Heltec v3. I’ve set powersaving on and it’s still doing it, Google got me here.
> 
> Use case is the repeater is currently in our bedroom until I can roof mount it.

Similar issue here :) 

Would be great being able to do a simple "LED off" command 👍

Koreanman01:
I have a fix submitted to make this possible on the v3. Just waiting to see if it gets merged. 
The submitted feature addition is listed here. https://github.com/meshcore-dev/MeshCore/pull/1599

Koreanman01:
> Will the option work for nrf52 as well? Like for the SenseCap Solar Node. The LED are quite bright.

I should have a SenseCap in soon so I can do some testing on it. I'm hoping it comes in before I leave for vacation so I can try to push an update.

PR #1934, opened 2026-03-06
Preferences for BLE, heartbeat status, and LoRa TX activity LEDs

andyshinn:
This adds some settings to control the BLE and status LEDs:

* `led.ble`
  * `0`: Enables the normal behavior of blinking when disconnected from BLE and solid when connected to BLE.
  * `1`: Blinks BLE LED only when disconnected from BLE.
  * `2`: Turns on the BLE LED solid when connected but doesn't blink when disconnected.
  * `3`: Disables the BLE LED.
  
* `led.status`:
  * `0`: Existing behavior to enable the status LED blinking.
  * `1`: Disables the status LED.

* `led.activity`:
  * `0`: Existing behavior to enable blinking `P_LORA_TX_LED` pin during TX.
  * `1`: Disable blinking the `P_LORA_TX_LED` on TX.
  
For the BLE LED to work it required a patch to the upstream Bluefruit SDK so I added that here as well.

~~Will need to be added in downstream clients.~~ Now done as custom vars  so we already get them in clients. Can also be tested in CLI Rescue with `set led.status #` and `set led.ble #`

Possibly related to #1214.

### DIFF
--- a/src/helpers/esp32/SerialBLEInterface.cpp
+++ b/src/helpers/esp32/SerialBLEInterface.cpp
@@ -10,7 +10,8 @@
 
 #define ADVERT_RESTART_DELAY  1000   // millis
 
-void SerialBLEInterface::begin(const char* prefix, char* name, uint32_t pin_code) {
+void SerialBLEInterface::begin(const char* prefix, char* name, uint32_t pin_code, uint8_t led_ble_mode) {
+  (void)led_ble_mode;  // LED mode control not implemented on ESP32
   _pin_code = pin_code;
 
   if (strcmp(name, "@@MAC") == 0) {
--- a/arch/nrf52/extra_scripts/patch_bluefruit.py
+++ b/arch/nrf52/extra_scripts/patch_bluefruit.py
@@ -7,6 +7,8 @@ when BLE central disconnects unexpectedly (e.g., going out of range, supervision
 Patches applied:
 1. BLEConnection.h: Add _hvn_qsize member to track semaphore queue size
 2. BLEConnection.cpp: Store hvn_qsize and restore semaphore on disconnect
+3. bluefruit.h: Add bluefruit_blinky_cb as friend so it can access _led_conn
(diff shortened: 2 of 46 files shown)


fdlamotte:
Hi, interesting commit

But I find this is very specific to a device to make a new command in the protocol

Why didn't you use a custom var instead (as for the gps toggle) ? It's a generic approach and it removes the burden of writing the client part (that does not have to worry about if the feature exists on the companion)

andyshinn:
Sure that makes sense. I reworked it to be a custom var instead.

src/helpers/esp32/SerialBLEInterface.cpp
```diff
@@ -10,7 +10,8 @@
 
 #define ADVERT_RESTART_DELAY  1000   // millis
 
-void SerialBLEInterface::begin(const char* prefix, char* name, uint32_t pin_code) {
+void SerialBLEInterface::begin(const char* prefix, char* name, uint32_t pin_code, uint8_t led_ble_mode) {
+  (void)led_ble_mode;  // LED mode control not implemented on ESP32
```
robekl, line 14: The setting is exposed, saved, and appears configurable, but has no effect on ESP32. Can it be implemented or hidden for this target?

a3vd:
CLI command  **set led.status**  is not working on **Heltec V4**  with latest meshcore installed

andyshinn:
> CLI command **set led.status** is not working on **Heltec V4** with latest meshcore installed

Do you mean you added this patch to latest 1.14.0 sources, compiled and flashed, and it didn’t work? What color is that LED?

src/helpers/esp32/SerialBLEInterface.cpp
andyshinn: All the nrf52 boards seem to have a concept of a `LED_BLUE`. But the ESP32 boards only have a `LED_BUILTIN`. Do you think it makes sense to implement for `LED_BLUE` and `BLE_LED_PIN` or should we use another name?

Issue #1588, opened 2026-02-03
[Feature Request] Repeater limit channel flooding


jeremeguenther:
I am concerned about the long term ability of the mesh to sustain flood messages for all Channels.

I would like a config feature on repeaters where the repeater can choose to not repeat channel messages over X hops.  This would still allow for flood discovery, and memorized long distance routes.  But would allow us to exclude the vast majority of channel traffic from far away places we do not care about.  In addition to this limitation if we were able to add an exclude list of channels we did not want this restriction to apply to, then we could continue to pass all traffic for the public channel, or known channels like #emergency etc.

446564:
You can do this better with region scopes, instead of a limit on hops you can say only repeat our-region-scope.

Repeaters can be configured to drop other flood traffic or not, and users can choose to scope their channel messages or not.

You can also create nested scopes unique to each repeater to further control what can go where.

jeremeguenther:
My understanding of the current implementation of region scopes is that they rely on every single companion node to self regulate away from the default settings to prevent their traffic from flooding the network.

While you can prevent All flood traffic and only allow the local regions data, this also prevents path finding flooding which many people would find desirable to allow.

I do love the idea of using regions and scopes to help control data, and I can see where you could force companions to self tag their channels to participate.  But I do not see how they can work to both control most channels, while continuing to allow all flood traffic for path finding and the public channel.

If I am understanding this wrong, or there is documentation on how to configure something like this, I would love to learn how to do it.

melbournefan:
There are trade offs between each feature

- LoRA is not a control channel based protocol, that is unlike WiFi where signal measurements can be taken with minimal difficulty, Meshcore repeaters typically advert from upwards of 12 hours per day

- A few people on Meshcore are using the protocol to conduct long distance transmissions and experimentation, you practically need long hops to achieve that

- Overcoming rate limiting requires the use of more frequent beaconing / adverts, which in itself floods the mesh.

-  Meshcore is a infrastructure dependant application, unlike Tastic which is a hybrid (in itself has issues with airtime factors)

- letting users set rate limits leads to fragmentation, where a repeater could essentially block messages in the mesh because they're "opting out" of repeating a long hop packet

The fact is the mesh topology is a mesh, yes I agree you can't have a packet getting daisy chained across thousands of kilometres however there a plan does need to be figured out about dealing with resource heavy uses

jeremeguenther:
[...]

Issue #1502, opened 2026-01-27
Proposal: token bucket rate limiting on repeater firmware


Beanow:
I've [worked on this for a bit](https://forum.letsmesh.net/t/rate-limiting-algorithms/236) and think it's promising.
I tend to call it the "fairness limiter", because it's actually a map of multiple rate limiters and could use a name 😄

Using a [token bucket](https://smudge.ai/blog/ratelimit-algorithms#token-buckets) rate limiting algorithm for repeaters to decide whether they should repeat or not.

I'm looking for feedback from maintainers on whether you would welcome this change to be upstreamed or not.
Happy to work on the PRs, but checking on the idea first 😄

_Watching my test firmware with a 2D grid_
<img width="600" height="488" alt="Image" src="https://github.com/user-attachments/assets/78f856de-950b-4686-97c0-846f46ecbdb2" />

## Goals & motivation

The goal is for a repeater to **avoid amplifying** problematic levels of traffic, with the limited knowledge it has and minimal coordination.

Because larger meshes, like the NL one I'm part of, already regularly experience congestion problems in spite of attempts to politely ask for slower repeater adverts, removing #test bots and using flooding less for testing.

Particularly rate limiting is for when this is **ignorant, accidental or maliciously high traffic**. Think repeater adverts on too short intervals, excessive #test messages, or a DOS script forging adverts to draw a face on the ocean.

It aims to be objective and neutral in what traffic is "useful". For example Public and #test are treated equal to any other group. But adverts have much lower limits than group texts.

It's intended as complementary to other means of reducing traffic and keeping the mesh healthy. Especially scopes, because scopes is how a conscious sender can limit their floods voluntarily. And gives meshes something to coordinate and improve their experience. Which is a different angle than rate limiting is going for, providing some relief when faced with uncoordinated high traffic.

My approach is also intended to be used without protocol changes. And therefore has limited options to identify the cause of traffic.

And adding more repeaters to the mesh with these limits should improve overall connectivity, rather than make an already congested mesh worse.

**Non-goal** is for this to solve scaling meshes on it's own. It's one piece of the puzzle.
Nor does it intend to provide configuration to police the mesh (like block #ai because I can't get any RAM).

## Implementation details

Several `uint8_t` arrays are allocated where each array represents a traffic category. And each index into the array represents one bucket.
Every category and bucket has a max capacity and a refill rate.

[...]

PR #1938, opened 2026-03-06
Add adaptive congestion control for backbone repeaters

wbijen:
### Summary

This PR adds a lightweight, adaptive congestion control system to repeater nodes. When the network gets busy, repeaters automatically and gracefully reduce flood hop limits for group messages and advertisements, keeping the mesh healthy without manual intervention.

This addresses #1588 (repeater limit channel flooding) with a busy-adaptive approach rather than a static cap.

### How it works

**BusyTracker** (`src/helpers/BusyTracker.h`) computes a composite busy score (0.0–1.0) from three weighted signals over a 60-second tumbling window:

| Signal | Weight | Rationale |
|--------|--------|-----------|
| Airtime ratio | 50% | "Am I personally overloaded?" — measures actual radio utilization (TX + RX time as a fraction of the window). This is the most direct indicator of channel saturation. |
| Queue pressure | 30% | "Is my TX pipeline backed up?" — instantaneous queue fill level. High queue depth means packets are waiting longer to transmit, a leading indicator that airtime will spike next. |
| Duplicate ratio | 20% | "Is the mesh around me oversaturated?" — proportion of received floods that are duplicates. High dups mean many neighbors are reforwarding the same packets. This doesn't show up in airtime (dups are short RX, not TX) or queue (dups are dropped, never queued), but it tells us that reducing our own flood reach won't hurt delivery — other paths already cover those destinations. |

All three weights are compile-time configurable (`BUSY_WEIGHT_AIRTIME`, `BUSY_WEIGHT_QUEUE`, `BUSY_WEIGHT_DUP`).

The busy score drives a **piecewise-linear ramp** that smoothly reduces effective hop limits as congestion builds:

```
effective_hops
     ^
  8  |███
  7  |     \        ← upper segment: max → mid
  6  |        \
  5  |...........█      ← knee (mid)
  4  |              \
  3  |                 \    ← lower segment: mid → floor
  2  |                     ████████
     +--+--+--+--+--+--+--+--> busy
     0  .15       .58     1.0
        ^onset    ^knee
```

| Busy range | Label | Behavior |
|-----------|-------|----------|
| 0.00 – 0.15 | **Low** | Dead zone — full reach, no penalty |
| 0.15 – 0.58 | **Medium** | Progressive shedding: max → mid |
| 0.58 – 1.00 | **High** | Aggressive shedding: mid → floor |

**Example:** A repeater running at 40% busy will forward group messages up to ~6 hops (down from 8), while advertisements drop to ~5 hops. At 80% busy, groups are limited to ~3 hops and adverts to ~2. Direct/unicast traffic is never throttled.

### Traffic-aware retransmit delay

`getRetransmitDelay()` scales the random jitter window by `(1 + busy × 2)`:
- At `busy = 0` (idle): delay unchanged — same as current behavior
- At `busy = 0.5` (moderate): 2× wider jitter window, spreading retransmissions
- At `busy = 1.0` (saturated): 3× wider jitter window

This naturally reduces collision probability under load by spreading retransmissions over a wider time window. `getDirectRetransmitDelay()` is intentionally **not** scaled — DM and ACK traffic should not be penalized during congestion, as these are high-value, low-volume packets that need reliable delivery.

### Defaults

All thresholds are compile-time `#define`s, overridable via `build_flags` in `platformio.ini`:

| Parameter | Default | Purpose |
|-----------|---------|---------|
| `BUSY_ONSET` | 0.15 | Busy score below which no shedding occurs |
| `BUSY_WINDOW_MS` | 60000 | Tumbling window for score computation (ms) |
| `BUSY_WEIGHT_AIRTIME` | 0.5 | Weight for airtime ratio component |
| `BUSY_WEIGHT_QUEUE` | 0.3 | Weight for queue pressure component |
| `BUSY_WEIGHT_DUP` | 0.2 | Weight for duplicate ratio component |
| `GROUP_FLOOD_MAX` | 8 | Group message hop limit at idle |
| `GROUP_FLOOD_MID` | 5 | Group hops at the knee of the ramp |
| `GROUP_FLOOD_FLOOR` | 2 | Group hops at full congestion |
| `ADVERT_FLOOD_MAX` | 8 | Advertisement hop limit at idle |
| `ADVERT_FLOOD_MID` | 4 | Advert hops at the knee |
| `ADVERT_FLOOD_FLOOR` | 1 | Advert hops at full congestion |
| `PACKET_POOL_SIZE` | 32 | Packet pool capacity (shared with StaticPoolPacketManager) |
| `BUSY_TX_DELAY_SCALE` | 3 | TX delay scaling factor |

Operators can tune per-deployment, for example a high-traffic urban repeater:

```ini
build_flags =
  -DGROUP_FLOOD_FLOOR=1
  -DADVERT_FLOOD_FLOOR=0
  -DBUSY_ONSET=0.10
  -DBUSY_WEIGHT_AIRTIME=0.6f
  -DBUSY_WEIGHT_QUEUE=0.2f
  -DBUSY_WEIGHT_DUP=0.2f
```

### What's included

- **Adaptive hop limits** in `allowPacketForward()` — per-type flood limiting for group, advert, and fallback to `_prefs.flood_max`
- **Traffic-aware retransmit delay** — `getRetransmitDelay()` scales jitter by `(1 + busy × 2)`, spreading retransmissions when congested. Direct message delays are intentionally untouched to protect DM/ACK delivery
- **Counter wrap protection** — airtime counters wrap at ~49 days; BusyTracker detects implausible deltas (exceeding elapsed wall time), skips the window, re-baselines, and preserves the previous valid score
- **Misconfiguration guards** — `getEffectiveFloodMax()` clamps `mid` and `floor` if build flag overrides violate `ceiling >= mid >= floor`
- **`stats-busy` CLI command** — prints busy score, component breakdown, and effective hop limits over serial
- **OLED display line** — live status: `GRP:5/8 ADV:4/8 B:32%`
- **Build flag docs** in `variants/heltec_v4/platformio.ini`

### Design choices

- Zero heap allocation at runtime — everything is stack/static
- Header-only BusyTracker avoids touching core library build
- Piecewise-linear ramp gives smooth, predictable behavior without cliff edges
- Dead zone prevents flapping during normal low-traffic operation
- `PACKET_POOL_SIZE` constant shared between pool allocator and BusyTracker to prevent silent drift

### Is this overengineered?

Honest question — the core mechanism is simple (composite score → hop limit ramp), but there are a lot of configurable knobs and a linear ramp. The rationale is that different deployments face very different traffic patterns (urban dense mesh vs rural sparse backbone), and we won't know the right defaults until we have field data. The `#ifndef` guards make these zero-cost to ignore if you don't need them.

That said, if the consensus is that fewer knobs and hardcoded defaults would be better to start with, happy to strip the configurability down to just `GROUP_FLOOD_MAX`/`FLOOR` and `ADVERT_FLOOD_MAX`/`FLOOR` and bake the rest in. We can always add knobs back later if field testing shows they're needed. Would love to hear thoughts on this.

### Files changed

| File | Change |
|------|--------|
| `src/helpers/BusyTracker.h` | **New** — busy score computation, `getEffectiveFloodMax()`, all compile-time defines |
| `examples/simple_repeater/MyMesh.h` | BusyTracker member, getter, `PACKET_POOL_SIZE` define |
| `examples/simple_repeater/MyMesh.cpp` | Wired into loop/forward/delay/stats/CLI |
| `examples/simple_repeater/UITask.h` | Accept BusyTracker pointer |
| `examples/simple_repeater/UITask.cpp` | Render effective hop limits on display |
| `examples/simple_repeater/main.cpp` | Pass BusyTracker to UITask |
| `variants/heltec_v4/platformio.ini` | Documented congestion build flags |

### Related work

| Issue | Title | Relationship |
|-------|-------|-------------|
| #1502 | Token bucket rate limiting | **Complementary.** Per-sender fairness vs our global congestion response. Composes well. |
| #1588 | Repeater limit channel flooding | **Directly addressed.** Our `GROUP_FLOOD_MAX` is this feature, but busy-adaptive rather than static. |

Looking forward to feedback! Happy to tune defaults or simplify if it feels like too much for a first pass.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -415,10 +415,26 @@ bool MyMesh::isLooped(const mesh::Packet* packet, const uint8_t max_counters[]) 
 
 bool MyMesh::allowPacketForward(const mesh::Packet *packet) {
   if (_prefs.disable_fwd) return false;
-  if (packet->isRouteFlood() && packet->getPathHashCount() >= _prefs.flood_max) return false;
-  if (packet->isRouteFlood() && recv_pkt_region == NULL) {
-    MESH_DEBUG_PRINTLN("allowPacketForward: unknown transport code, or wildcard not allowed for FLOOD packet");
-    return false;
+  if (packet->isRouteFlood()) {
+    if (recv_pkt_region == NULL) {
+      MESH_DEBUG_PRINTLN("allowPacketForward: unknown transport code, or wildcard not allowed for FLOOD packet");
+      return false;
+    }
+    // per-type adaptive hop limits (groups and adverts use busy-driven ramp,
+    // but never exceed the operator-configured flood_max)
+    uint8_t hops = packet->getPathHashCount();
+    uint8_t limit = _prefs.flood_max;
+    uint8_t type = packet->getPayloadType();
+    if (type == PAYLOAD_TYPE_GRP_TXT || type == PAYLOAD_TYPE_GRP_DATA) {
+      uint8_t adaptive = getEffectiveFloodMax(busy_tracker.busy,
+          GROUP_FLOOD_MAX, GROUP_FLOOD_MID, GROUP_FLOOD_FLOOR);
+      limit = (adaptive < limit) ? adaptive : limit;
+    } else if (type == PAYLOAD_TYPE_ADVERT) {
+      uint8_t adaptive = getEffectiveFloodMax(busy_tracker.busy,
+          ADVERT_FLOOD_MAX, ADVERT_FLOOD_MID, ADVERT_FLOOD_FLOOR);
+      limit = (adaptive < limit) ? adaptive : limit;
+    }
+    if (hops >= limit) return false;
   }
   if (packet->isRouteFlood() && _prefs.loop_detect != LOOP_DETECT_OFF) {
     const uint8_t* maximums;
@@ -525,7 +541,8 @@ int MyMesh::calcRxDelay(float score, uint32_t air_time) const {
 
 uint32_t MyMesh::getRetransmitDelay(const mesh::Packet *packet) {
   uint32_t t = (_radio->getEstAirtimeFor(packet->getPathByteLen() + packet->payload_len + 2) * _prefs.tx_delay_factor);
-  return getRNG()->nextInt(0, 5*t + 1);
+  float scale = 1.0f + busy_tracker.busy * BUSY_TX_DELAY_SCALE;
+  return getRNG()->nextInt(0, (uint32_t)(5 * t * scale) + 1);
 }
 uint32_t MyMesh::getDirectRetransmitDelay(const mesh::Packet *packet) {
   uint32_t t = (_radio->getEstAirtimeFor(packet->getPathByteLen() + packet->payload_len + 2) * _prefs.direct_tx_delay_factor);
@@ -830,7 +847,7 @@ void MyMesh::sendNodeDiscoverReq() {
 
 MyMesh::MyMesh(mesh::MainBoard &board, mesh::Radio &radio, mesh::MillisecondClock &ms, mesh::RNG &rng,
                mesh::RTCClock &rtc, mesh::MeshTables &tables)
-    : mesh::Mesh(radio, ms, rng, rtc, *new StaticPoolPacketManager(32), tables),
+    : mesh::Mesh(radio, ms, rng, rtc, *new StaticPoolPacketManager(PACKET_POOL_SIZE), tables),
       _cli(board, rtc, sensors, acl, &_prefs, this), telemetry(MAX_PACKET_PAYLOAD - 4), region_map(key_store), temp_map(key_store),
       discover_limiter(4, 120),  // max 4 every 2 minutes
       anon_limiter(4, 180)   // max 4 every 3 minutes
@@ -907,6 +924,10 @@ void MyMesh::begin(FILESYSTEM *fs) {
     bridge.begin();
   }
 #endif
+
+  busy_tracker.reset(millis(), getTotalAirTime(), getReceiveAirTime(),
+                     getNumRecvFlood(),
+                     ((SimpleMeshTables *)getTables())->getNumFloodDups());
 
   radio_set_params(_prefs.freq, _prefs.bw, _prefs.sf, _prefs.cr);
   radio_set_tx_power(_prefs.tx_power_dbm);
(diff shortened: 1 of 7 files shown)


examples/simple_repeater/MyMesh.cpp
```diff
@@ -415,10 +415,24 @@ bool MyMesh::isLooped(const mesh::Packet* packet, const uint8_t max_counters[])
 
 bool MyMesh::allowPacketForward(const mesh::Packet *packet) {
```
robekl, line 416: could a max based on `min(_prefs->flood_max, getEffectiveFloodMax(...))` be used, to adhere to a configured lower value?

examples/simple_repeater/MyMesh.cpp
```diff
@@ -1314,6 +1355,12 @@ void MyMesh::loop() {
   uint32_t now = millis();
   uptime_millis += now - last_millis;
   last_millis = now;
+
+  // update busy score
+  auto t = (SimpleMeshTables *)getTables();
+  busy_tracker.update(now, getTotalAirTime(), getReceiveAirTime(),
+                      _mgr->getOutboundCount(0xFFFFFFFF), PACKET_POOL_SIZE,
```
robekl: Busy queue pressure is computed from `_mgr->getOutboundCount(0xFFFFFFFF)`, but `StaticPoolPacketManager::getOutboundCount()` delegates to `countBefore(now)` and treats now as a real timestamp, not an “all entries” sentinel. With 0xFFFFFFFF, many valid scheduled entries are excluded, so `queue_ratio` is systematically inaccurate (often near zero), which skews congestion control decisions. This should use current `now` (or a dedicated “count all queued” API) instead.

wbijen: Good catch @robekl on the `getOutboundCount(0xFFFFFFFF)` sentinel, fixed in the latest push, now using `getOutboundCount(now)`.

Worth noting: the same `0xFFFFFFFF` pattern exists in `StatsFormatHelper.h:17` (pre-existing, not from this PR). It has the same signed-comparison issue in `countBefore()` — after ~25 days uptime, `millis() + delay` lands in the `0x80000000` range and the `(int32_t)(schedule_time - 0xFFFFFFFF) > 0` check incorrectly skips those entries, causing `queue_len` in stats responses to report 0 even with packets queued. Lower impact since it's display-only, but might be worth a separate fix.

examples/simple_repeater/MyMesh.cpp
wbijen: Lowest now wins, fixed and pushed! thanks.

It now has GROUP_FLOOD_MAX and ADVERT_FLOOD_MAX configured at 8. Do you think that value makes sense or you want me to set it to 64 so by default there is no change in behavior at quiet times?

terminalvelocity23:
This sounds sensible enough, and it's useful to see the load indicator on your repeater. The group chat limits are a bit too restrictive for our huge mesh, but it's not hard to fine-tune.

samuk:
@wbijen does your branch update cleanly to latest Meshcore main branch?

I wonder if @liamcottle or @ripplebiz might give some indication here if they are interested in this feature/ approach

PR #1939, opened 2026-03-06
nRF52 BLE: wait for LFXO crystal before enabling SoftDevice (fixes t1000e hanging at reboot)

weebl2000:
The Adafruit framework starts the LFXO in init() but never waits for EVENTS_LFCLKSTARTED. On first boot the filesystem format (~600ms of page erases) gives the crystal time to stabilize. On subsequent boots only quick mounts happen, so sd_softdevice_enable() inside Bluefruit.begin() can be called while the crystal is still starting, which hangs indefinitely.

Poll EVENTS_LFCLKSTARTED before Bluefruit.begin() so the wait is exact rather than a blind delay.

fixes: #1780

Please try the firmware for yourself by [building here](https://mcimages.weebl.me?commitId=wait-for-crystal-t1000e).


MarijnDoeve:
~Hi, I have a T1000-E that never boots while running MeshCore firmware. Do you think that problem is related to this issue?~
Does not seem to be the case. Created issue #2211

PR #1952, opened 2026-03-06
Update cli_commands.md to include path.hash.mode and loop.detect

AI7NC:
Adding the new repeater cli commands introduced in 1.14

Ref: https://buymeacoffee.com/ripplebiz/path-diagnostics-improvements

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -397,7 +397,7 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 
 **Default:** `0`
 
-**Note:** the 'path.hash.mode' sets the low-level ID/hash encoding size used the for when the repeater adverts. This setting has no impact on what packet ID/hash size this repeater forwards, all sizes should be forwarded on firmware >= 1.14. This feature was added in firmware 1.14
+**Note:** the 'path.hash.mode' sets the low-level ID/hash encoding size used when the repeater adverts. This setting has no impact on what packet ID/hash size this repeater forwards, all sizes should be forwarded on firmware >= 1.14. This feature was added in firmware 1.14
 
 **Temporary Note:** adverts with ID/hash sizes of 2 or 3 bytes may have limited flood propogation in your network while this feature is new as v1.13.0 firmware and older will drop packets with multibyte path ID/hashes as only 1-byte hashes are suppored. Consider your install base of firmware >=1.14 has reached a criticality for effective network flooding before implementing higher ID/hash sizes. 
 


docs/cli_commands.md
```diff
@@ -383,6 +383,45 @@ This document provides an overview of CLI commands that can be sent to MeshCore
...
 ---
 
+### View or change this node's avert path hash size
+**Usage:**
+- `get path.hash.mode`
+- `set path.hash.mode <value>`
+
+**Parameters:**
+- `value`: Path hash size (0-2)
+  - `0`: 1 Byte hash size (256 count)[64 max flood]
```
liamcottle: prob reword `256 count` to `~256 unique ids`. no idea what count means without context. Same for the others below.

docs/cli_commands.md
```diff
@@ -383,6 +383,45 @@ This document provides an overview of CLI commands that can be sent to MeshCore
...
+  - `1`: 2 Byte hash size (65,536 count)[32 max flood]
+  - `2`: 3 Byte hash size (16,777,216 count)[21 max flood]
+
+**Default:** `0`
+
+**Note:** the 'mode' is the low-level encoding (0..3) where 0 -> 1-byte (legacy), 1 -> 2-byte, 2 -> 3-byte for the ID/hash of the repeaters adverts. This feature was added in firmware 1.14
```
liamcottle: Might be good to make it a list.

Path Hash Modes:
- `0` - 1-byte path hashes
- `1` - 2-byte path hashes
- `2` - 3-byte path hashes
- `3` - (reserved) do not use

docs/cli_commands.md
```diff
@@ -383,6 +383,45 @@ This document provides an overview of CLI commands that can be sent to MeshCore
...
+
+**Temporary Note:** adverts with hash sizes of 2 or 3 bytes may have limited flood propogation in your network while this feature is new. Consider your install base of firmware >=1.14 before implementing.
```
liamcottle: Might be good to mention specifically, v1.13.0 firmware and older will drop packets with multibyte path hashes. Only 1-byte is supported. v1.14.0+ firmware supports 1,2 and 3-byte path hashes.

docs/cli_commands.md
```diff
@@ -383,6 +383,45 @@ This document provides an overview of CLI commands that can be sent to MeshCore
 
 ---
 
+### View or change this node's avert path hash size
```
weebl2000: Two issues: heading should be `####` (h4) to match other entries under the `### Routing` section, and "avert" should be "advert".

```suggestion
#### View or change this node's advert path hash size
```

docs/cli_commands.md
weebl2000: Typo: "propogation" → "propagation". Also trailing whitespace.

```suggestion
**Temporary Note:** adverts with hash sizes of 2 or 3 bytes may have limited flood propagation in your network while this feature is new. Consider your install base of firmware >=1.14 before implementing.
```

docs/cli_commands.md
```diff
@@ -383,6 +383,45 @@ This document provides an overview of CLI commands that can be sent to MeshCore
...
+
+**Default:** `0`
+
+**Note:** the 'mode' is the low-level encoding (0..3) where 0 -> 1-byte (legacy), 1 -> 2-byte, 2 -> 3-byte for the ID/hash of the repeaters adverts. This feature was added in firmware 1.14
+
+**Temporary Note:** adverts with hash sizes of 2 or 3 bytes may have limited flood propogation in your network while this feature is new. Consider your install base of firmware >=1.14 before implementing. 
+
+---
+
+### View or change this node's loop detection
```
weebl2000: Same heading level issue — should be `####` to match other Routing subsections. Also trailing whitespace.

```suggestion
#### View or change this node's loop detection
```

docs/cli_commands.md
```diff
@@ -383,6 +383,45 @@ This document provides an overview of CLI commands that can be sent to MeshCore
...
+**Usage:**
+- `get loop.detect`
+- `set loop.detect <state>`
+
+**Parameters:**
+- `state`: 
+  - `off`: no loop detection is performed
+  - `minimal`: packets are dropped if repeaters ID/hash shows more times than 4(1-byte), 2(2-byte), 1(3-byte)
+  - `moderate`: packets are dropped if repeaters ID/hash shows more times than 2(1-byte), 1(2-byte), 1(3-byte)
+  - `strict`: packets are dropped if repeaters ID/hash shows more times than 1(1-byte), 1(2-byte), 1(3-byte)
```
weebl2000: "More times than N" implies strictly greater than (>N), but the firmware uses `n >= max_counters[hash_size]` (`MyMesh.cpp:456`), so packets are dropped when the count reaches N — not exceeds it. For example, `strict` with 1-byte (max=1) drops when the hash appears even once, but "more times than 1" implies 2+.

This is also inconsistent with your example below which correctly says "already encoded 4 times, it will reject".

```suggestion
  - `minimal`: packets are dropped if repeater's ID/hash appears 4 or more times (1-byte), 2 or more (2-byte), 1 or more (3-byte)
  - `moderate`: packets are dropped if repeater's ID/hash appears 2 or more times (1-byte), 1 or more (2-byte), 1 or more (3-byte)
  - `strict`: packets are dropped if repeater's ID/hash appears 1 or more times (1-byte), 1 or more (2-byte), 1 or more (3-byte)
```

docs/cli_commands.md
AI7NC: Take a look at commit [fe32f16](https://github.com/meshcore-dev/MeshCore/pull/1952/commits/fe32f16aa40a00dfed9d7315261680d84b9bf174). I decided to drop the list in the note as we have an informative list under parameters. I've added the selection of `3` to the parameters since it's technically an option for the future but made it clear it's not to be used with all caps. Removing the in-line list I reworded the note to make it simpler to understand so take a look.

docs/cli_commands.md
AI7NC: See commit [bb45486](https://github.com/meshcore-dev/MeshCore/pull/1952/commits/bb454861c7b893c76813c3e29f0f19a871f67c71)

docs/cli_commands.md
AI7NC: See [fe32f16](https://github.com/meshcore-dev/MeshCore/pull/1952/commits/fe32f16aa40a00dfed9d7315261680d84b9bf174)

AI7NC:
Please take a fresh look, I think I've resolved all the comments with proper commits but I'm new to working through a pull request like this on github so I'm still learning the process flow here.

PR #1954, opened 2026-03-07
Lora longer preamble

OverkillFPV:
After testing i have found that for the narrow settings i am using (62.5khz, sf7) that there is a reliability issue of transmissions that did not exist on wide (250khz,sf11). I have come to the conclusion that the preamble is too short for the low spreading factor as it is 100ms shorter going from 131ms for wide, down to 32ms for narrow. the increase to a preamble length of 32 brings the preamble time up to 65ms (32ms extra). In my opinion this extra on air time is offset by the lower need to resend an entire packet. I have been running this longer preamble for a while with good results, and everyone who has tested the modification so far has either noticed a slight, or decent improvement in reliability.

This change changes the preamble length to 32 <sf9 and the standard 16 >sf8. It allows the sf to be changed through the cli and will adjust the preamble on the next tx


JDat:
Is this compatible with existing MeshCore versions? By changing preamble will this code work with existing repeaters on network? Just asking, not dive deep into SX12xx datasheets yet.

OverkillFPV:
Yes, this affects the tx of the node it is on, and is completely interchangeable between nodes. Currently all my repeaters and companions are running 32 and are working perfectly with the rest of the mesh

src/helpers/radiolib/RadioLibWrappers.h
```diff
@@ -38,6 +39,7 @@ class RadioLibWrapper : public mesh::Radio {
   }
 
   virtual float getCurrentRSSI() =0;
```
weebl2000, line 41: Consider extracting the repeated ternary to a helper to keep the 4 call sites DRY:
```suggestion
  virtual uint8_t getSpreadingFactor() const { return LORA_SF; }
  static uint16_t preambleLengthForSF(uint8_t sf) { return sf <= 8 ? 32 : 16; }
```

weebl2000 (commented):
Minor: the preamble expression is repeated 4 times — a small helper would reduce duplication.

weebl2000:
@OverkillFPV what do you think about 24 instead of 32.

jbrazio:
LGTM but why sf9 as the pivot point ?

OverkillFPV:
I tried 24, 32 and 64 and 32 had an improvement over 24 with only a slight improvement going up to 64. Sf9 was just a middle ground i picked. As it would require many many iterations to determine where the swing should be.

OverkillFPV:
> Minor: the preamble expression is repeated 4 times — a small helper would reduce duplication.

Possibly I'll have a look at simplifying it

OverkillFPV:
I'm currently playing around with a companion of 64 and repeaters at 32. As I am seeing if it helps with the random dropped messages

weebl2000:
Interested to hear the results!

OverkillFPV:
Early testing is looking to suggest the tag being a bit more reliable for tx and less rx biased. Only my first impressions though

Edit: hasn't actually had much of an affect when testing from known hard to reach spots. My early thoughts was most likely due to good conditions

OverkillFPV:
Longer than 32 had no benefit after doing more thorough testing. So this pull request is current as per my testing

src/helpers/radiolib/RadioLibWrappers.cpp
```diff
@@ -142,6 +144,11 @@ uint32_t RadioLibWrapper::getEstAirtimeFor(int len_bytes) {
 
 bool RadioLibWrapper::startSendRaw(const uint8_t* bytes, int len) {
   _board->onBeforeTransmit();
+  uint8_t sf = getSpreadingFactor();
```
robekl: At src/Dispatcher.cpp:329, send timing and timeout budgeting are determined before transmit starts, but src/helpers/radiolib/RadioLibWrappers.cpp:147 updates the preamble length only inside startSendRaw(), after budgeting has already happened, and src/helpers/radiolib/RadioLibWrappers.cpp:142 estimates airtime using the current radio state. That means the first packet after an SF change can be scheduled with stale airtime assumptions. Since timeout/budget logic depends on airtime, this can create hard-to-reproduce send failures or premature timeout handling.

everett1992:
How can I test this - what stats or measurements are you talking?

OverkillFPV:
> How can I test this - what stats or measurements are you talking?

I was testing this by sending a message with the standard 16 preamble and seeing the heard repeats fluctuate, and also complete inability to ping a repeater very close or quite far. Then on 32 the heard repeats were very consistant, and i could ping and login to a repeater within 5m that I could not with the shorter preamble.

src/helpers/radiolib/RadioLibWrappers.cpp
OverkillFPV: Ah yes, I overlooked that. I'll have a look at getting the change in earlier to allow correct packet airtime calculation

OverkillFPV: Added in an update inside getestairtime to prevent incorrect airtime calcs

pelgraine:
Thought I'd add my two cents - so I just flashed this on my Station G2 repeater and it's a vast improvement when using my T-Deck Pro companion with its poor internal antenna. I've gone from about 4-6/10 timeouts when using repeater admin functions to 1/10 timeouts.

Calvario:
I'm testing it in Portugal 868 with T1000-e (client), Heltec wsl3 (RPT) and Seed P1 (RPT) with a good result too. It's astonishing if you try many 0hop ping from the client to the repeater. Still doing some tests and would update later my feedback.

robekl:
My AI said this, it seems like a reasonable analysis but does it matter in practice?

`getEstAirtimeFor()` now mutates the live radio by calling `setPreambleLength(...)` when SF changes. That looks risky because this method is used from the RX path as well as TX scheduling, and in `recvRaw()` the radio has already been put back into receive mode before `Dispatcher::checkRecv()` calls `_radio->getEstAirtimeFor(len)`. So after a runtime SF change, the first airtime estimate can reconfigure LoRa settings while the radio is actively receiving, which can create a receive blind spot or drop packets depending on how the underlying RadioLib driver handles `setPreambleLength()` in RX mode.

I think the safer approach is to keep `getEstAirtimeFor()` side-effect free and only apply the preamble update in explicit radio reconfiguration/TX setup paths, ideally at the same point where SF itself is changed.

OverkillFPV:
> My AI said this, it seems like a reasonable analysis but does it matter in practice?
> 
> `getEstAirtimeFor()` now mutates the live radio by calling `setPreambleLength(...)` when SF changes. That looks risky because this method is used from the RX path as well as TX scheduling, and in `recvRaw()` the radio has already been put back into receive mode before `Dispatcher::checkRecv()` calls `_radio->getEstAirtimeFor(len)`. So after a runtime SF change, the first airtime estimate can reconfigure LoRa settings while the radio is actively receiving, which can create a receive blind spot or drop packets depending on how the underlying RadioLib driver handles `setPreambleLength()` in RX mode.
> 
> I think the safer approach is to keep `getEstAirtimeFor()` side-effect free and only apply the preamble update in explicit radio reconfiguration/TX setup paths, ideally at the same point where SF itself is changed.

Can just revert back to my previous method of having it only in start send raw. As worst case it means the first tx will have an incorrect airtime calculation, but after that they'll be fine? The main reason I did it this way is to do it where the radio settings are changed needs to be done in many spots

OverkillFPV:
Now i changed it to set the preamble correctly when the radio settings are changed (The way i should have done it)

terminalvelocity23:
So far it's working very well for our dense mesh set at SF7 62.5 kHz. It's especially noticeable in repeater UI, it became much snappier with less timeouts.

OverkillFPV:
> So far it's working very well for our dense mesh set at SF7 62.5 kHz. It's especially noticeable in repeater UI, it became much snappier with less timeouts.

Yeah so far only upsides with only a slight increase of airtime as a downside

Calvario:
In the last 20 days, with the help of multiple volunteers in Portugal, we tested this PR on various clients and repeaters (868) across several cities with challenging topologies.

I can safely say that, in Portugal, this change does make a difference, particularly at very short and long distances.

On devices with already weak TX (such as the T1000-e or any repeater with a small antenna), the difference is more noticeable.

OverkillFPV:
> In the last 20 days, with the help of multiple volunteers in Portugal, we tested this PR on various clients and repeaters (868) across several cities with challenging topologies.
> 
> I can safely say that, in Portugal, this change does make a difference, particularly at very short and long distances.
> 
> On devices with already weak TX (such as the T1000-e or any repeater with a small antenna), the difference is more noticeable.

Thank you heaps! Hopefully this gains more attention and gets widely implemented

palachzzz:
Why this PR is not merged yet?

liamcottle:
Thanks for the work on this @OverkillFPV and to all who have extensively tested it.

Based on feedback this seems to be of great benefit. I'll merge in now.

@ripplebiz  is planning to refactor the `radio_set_params` which is duplicated across all of the `target.cpp` for each variant, so it'll be good to get this PR merged in before it goes stale and needs more work.

JDat:
Still no technical answer with quotes form Semtech datasheets why this is working. So far only vibe and feelings "it works for me so great".  This is not physics and technology, this is vibe bull*it.

recrof:
> Still no technical answer with quotes form Semtech datasheets why this is working. So far only vibe and feelings "it works for me so great". This is not physics and technology, this is vibe bull*it.

are you upset about anything in particular? I believe this was tested and it's not been reported to break any communication between radios. not everything needs to be documented to be feasible.

OverkillFPV:
Unfortunately it's a bit like choosing the spreading factor, there isn't really a 'right' answer. Just testing of what works well

JDat:
The problem are simple:
1) "it works better for me" instead of researching and explaining what technically changes and why.
2) In cases if not sure what and why working, bad thing is to just burn new parameters in firmware instead of adding somewhere in advanced config additional command to configure parameters by hand if something goes wrong.

Now it become burned in firmware instead of configurable. Do not forget that "your case" or "your experience" is not the same in all environment over the world.

It would be nice if you make preamble configurable by user.

recrof:
> "it works better for me" instead of researching and explaining what technically changes and why.

preamble is very known and documented part of lora packet. longer preamble means higher chance of radio capturing distant packets.

> It would be nice if you make preamble configurable by user.

we want to introduce rx duty cycle later and this is critical step before that. if it was configurable, it would break radios that have rx duty cycle and shorter preamble. it doesn't really make sense to have it configurable. it's better to have it the same because of older chipsets like sx1276 that can struggle with different preamble length. it can be painful transition for older chips, but benefits outweight drawbacks. testing was critical to know if any other issues come to light - as said not everything is (and can) be documented in datasheets.

PR #1956, opened 2026-03-07
fix: enable GPIO on NFC pins for RAK4631 and WisMesh Tag variants

got-root:
## Summary

Adds `-D CONFIG_NFCT_PINS_AS_GPIOS=1` build flag to the RAK4631 and WisMesh Tag base
variant configs, fixing a bug where the user button (P0.09) becomes permanently
unresponsive after a full flash erase.

## Problem

On the nRF52840, pins P0.09 and P0.10 are dual-function: they serve as NFC antenna
pins by default, and can only be used as GPIO if the UICR `NFCPINS` register is
reprogrammed to release them. Both the RAK4631 and WisMesh Tag use P0.09 as the user
button (`PIN_BUTTON1`).

Under normal circumstances, the UICR is already correctly configured by a prior
bootloader or firmware, so the button works. However, performing a full flash erase
(e.g. via the MeshCore web flasher's "Erase Flash" step) resets the UICR to factory
defaults, re-enabling NFC on P0.09. After this, `digitalRead()` on the button pin
returns a fixed value and the button is completely unresponsive — no press gesture
(single, double, triple, long) produces any effect.

The Nordic SDK provides a mechanism to fix this automatically: when
`CONFIG_NFCT_PINS_AS_GPIOS` is defined, `SystemInit()` checks the UICR on every boot
and reprograms it if NFC is still enabled, followed by a one-time automatic reset.
This flag was not set for either variant.

## Scoping Decision

We considered three scopes for this fix:

| Scope | Affected builds |
|---|---|
| `rak_wismesh_tag` only | 5 WisMesh Tag environments |
| `rak_wismesh_tag` + `rak4631` | 5 + 8 = 13 environments |
| `nrf52_base` (all nRF52) | All nRF52 variants globally |

We chose `rak_wismesh_tag` + `rak4631` because:
- Both variants define `PIN_BUTTON1 = 9` (P0.09 / NFC1) and are directly affected
- No device in the MeshCore codebase uses NFC functionality, so the flag is safe
- The flag is a no-op on devices where the UICR is already correctly configured
- Broader `nrf52_base` scope could be done separately if desired, but these are the
  two variants we could directly validate

## Changes

Two files, one line each:

- `variants/rak_wismesh_tag/platformio.ini` — added `-D CONFIG_NFCT_PINS_AS_GPIOS=1`
  to `[rak_wismesh_tag]` base build flags
- `variants/rak4631/platformio.ini` — added `-D CONFIG_NFCT_PINS_AS_GPIOS=1` to
  `[rak4631]` base build flags

## Testing

Tested on a RAK WisMesh Tag (companion radio BLE build) that had previously undergone
a full flash erase via the web flasher, rendering the button unresponsive.

After flashing with this fix:
- **Triple press** — buzzer toggle: **PASS** (audible beep on enable, silence on disable)
- **Double press** — flood advert: **PASS** (advert received by other radios on the mesh)
- **Long press (3s)** — shutdown: **PASS** (device powers off, LEDs dark, button press to wake)

First boot after flash showed the expected one-time automatic reset as `SystemInit()`
reprogrammed the UICR. Subsequent boots were normal with no extra reset.

### DIFF
--- a/variants/rak4631/platformio.ini
+++ b/variants/rak4631/platformio.ini
@@ -19,6 +19,7 @@ build_flags = ${nrf52_base.build_flags}
   -D LORA_TX_POWER=22
   -D SX126X_CURRENT_LIMIT=140
   -D SX126X_RX_BOOSTED_GAIN=1
+  -D CONFIG_NFCT_PINS_AS_GPIOS=1
 build_src_filter = ${nrf52_base.build_src_filter}
   +<../variants/rak4631>
   +<helpers/sensors>
--- a/variants/rak_wismesh_tag/platformio.ini
+++ b/variants/rak_wismesh_tag/platformio.ini
@@ -25,6 +25,7 @@ build_flags = ${nrf52_base.build_flags}
   -D PIN_BUZZER=21
   -D PIN_BOARD_SDA=PIN_WIRE_SDA
   -D PIN_BOARD_SCL=PIN_WIRE_SCL
+  -D CONFIG_NFCT_PINS_AS_GPIOS=1
 build_src_filter = ${nrf52_base.build_src_filter}
   +<../variants/rak_wismesh_tag>
   +<helpers/ui/MomentaryButton.cpp>


got-root:
**Closing this PR as unnecessary.**

After further testing, the root cause of the reported button issue was identified as **flashing the wrong firmware variant**, not a missing GPIO configuration.

**Findings:**

1. **WisMesh Tag devices should use `RAK_WisMesh_Tag_*` builds**, which correctly define `PIN_USER_BTN=9` in the variant headers. The `RAK_4631_*` builds are intended for different hardware configurations and lack button support in the base variant.

2. **The official 1.14.0 firmware works correctly** on fresh WisMesh Tag devices even after flash erase, suggesting the bootloader already configures UICR.NFCPINS appropriately.

3. **Our `CONFIG_NFCT_PINS_AS_GPIOS=1` addition provides no functional benefit** for correctly-flashed devices, as the NFC pins are already available for GPIO use.

While this change could serve as a defensive/robustness measure, it adds unnecessary complexity without solving an actual bug. Users experiencing button issues should verify they're flashing the correct variant (`RAK_WisMesh_Tag_companion_radio_ble` not `RAK_4631_companion_radio_ble`).

PR #1961, opened 2026-03-07
Add get/set dutycycle command

weebl2000:
Allow users to simply set dutycycle percentage instead of having to deal with converting to airtime factor.

Old command still works but shows it's deprecated in favor of dutycycle command. We store internally all the same - this avoids breaking any prefs and works well enough internally.

Build it [HERE](https://mcimages.weebl.me?commitId=dutycycle-command)

image


jbrazio:
I'm not so sure if we should keep the old command.
"Duty Cycle" is the technical [and legal term] for this, "air time factor" is more about the actual time a signal uses the airwaves. I prefer the DRY principle here. :-P

weebl2000:
> I'm not so sure if we should keep the old command. "Duty Cycle" is the technical [and legal term] for this, "air time factor" is more about the actual time a signal uses the airwaves. I prefer the DRY principle here. :-P

I say keep it for one release, users will see it's deprecated, remove in the release after.

DRY is nice, but confusing users by disappearing commands instead of deprecating is also bad.

src/helpers/CommonCLI.cpp
```diff
@@ -283,8 +283,13 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
...
       const char* config = &command[4];
-      if (memcmp(config, "af", 2) == 0) {
-        sprintf(reply, "> %s", StrHelper::ftoa(_prefs->airtime_factor));
+      if (memcmp(config, "dutycycle", 8) == 0) {
+        float dc = 100.0f / (_prefs->airtime_factor + 1.0f);
+        int dc_int = (int)dc;
+        int dc_frac = (int)((dc - dc_int) * 10.0f + 0.5f);
+        sprintf(reply, "> %d.%d%%", dc_int, dc_frac);
+      } else if (memcmp(config, "af", 2) == 0) {
+        sprintf(reply, "> %s (deprecated, use 'get dutycycle')", StrHelper::ftoa(_prefs->airtime_factor));
```
4np: 👍

src/helpers/CommonCLI.cpp
```diff
@@ -436,10 +441,25 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
...
+          sprintf(reply, "OK - %d.%d%%", a_int, a_frac);
+        }
+      } else if (memcmp(config, "af ", 3) == 0) {
         _prefs->airtime_factor = atof(&config[3]);
         savePrefs();
-        strcpy(reply, "OK");
+        float actual = 100.0f / (_prefs->airtime_factor + 1.0f);
+        int a_int = (int)actual;
+        int a_frac = (int)((actual - a_int) * 10.0f + 0.5f);
+        sprintf(reply, "OK - %d.%d%% (deprecated, use 'set dutycycle')", a_int, a_frac);
```
4np: 👍

towerviewcams:
> > I'm not so sure if we should keep the old command. "Duty Cycle" is the technical [and legal term] for this, "air time factor" is more about the actual time a signal uses the airwaves. I prefer the DRY principle here. :-P
> 
> I say keep it for one release, users will see it's deprecated, remove in the release after.
> 
> DRY is nice, but confusing users by disappearing commands instead of deprecating is also bad.

Keep yes, very much. Soft transition is always best.

ViezeVingertjes:
We could call it an `alias` and call it the day haha.
But i think this approach is nice, most people will know this percentage 'if' its applicable to their country. 👍🏼

docs/cli_commands.md
```diff
@@ -419,15 +419,22 @@ This document provides an overview of CLI commands that can be sent to MeshCore
 
 ---
 
-#### View or change the airtime factor (duty cycle limit)
```
liamcottle, line 503: please revert the deletion of previous command docs. we need to keep these around as older firmware, which is the majority of the mesh, will have no documentation for the existing commands.

we should keep airtime factor docs and add a comment saying it's deprecated as of firmware version x.x.x, with a link to the section for the newly added command.

which should also prob have a note saying which firmware version it was added in.

src/helpers/CommonCLI.cpp
```diff
@@ -283,8 +283,13 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
...
       const char* config = &command[4];
-      if (memcmp(config, "af", 2) == 0) {
-        sprintf(reply, "> %s", StrHelper::ftoa(_prefs->airtime_factor));
+      if (memcmp(config, "dutycycle", 9) == 0) {
+        float dc = 100.0f / (_prefs->airtime_factor + 1.0f);
+        int dc_int = (int)dc;
+        int dc_frac = (int)((dc - dc_int) * 10.0f + 0.5f);
+        sprintf(reply, "> %d.%d%%", dc_int, dc_frac);
+      } else if (memcmp(config, "af", 2) == 0) {
+        sprintf(reply, "> %s (deprecated, use 'get dutycycle')", StrHelper::ftoa(_prefs->airtime_factor));
```
liamcottle: we prob shouldn't change the format of the existing response, incase it was being parsed by clients. it's unlikely, but would be a breaking change due to extra info at the end. should probably just document that it's deprecated in the docs as well as release notes.

for example, the app actually parsed the response text as a number and put it into the airtime factor field. however this field was never shown in the production app.

src/helpers/CommonCLI.cpp
```diff
@@ -436,10 +441,25 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
      */
     } else if (memcmp(command, "set ", 4) == 0) {
       const char* config = &command[4];
-      if (memcmp(config, "af ", 3) == 0) {
+      if (memcmp(config, "dutycycle ", 10) == 0) {
+        float dc = atof(&config[10]);
+        if (dc < 10 || dc > 100) {
```
liamcottle: is there a reason why we aren't allowed to set a duty cycle limit lower than 10%?

src/helpers/CommonCLI.cpp
```diff
@@ -436,10 +441,25 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
...
+          sprintf(reply, "OK - %d.%d%%", a_int, a_frac);
+        }
+      } else if (memcmp(config, "af ", 3) == 0) {
         _prefs->airtime_factor = atof(&config[3]);
         savePrefs();
-        strcpy(reply, "OK");
+        float actual = 100.0f / (_prefs->airtime_factor + 1.0f);
+        int a_int = (int)actual;
+        int a_frac = (int)((actual - a_int) * 10.0f + 0.5f);
+        sprintf(reply, "OK - %d.%d%% (deprecated, use 'set dutycycle')", a_int, a_frac);
```
liamcottle: same comment as above about the adde deprecation text being added

src/helpers/CommonCLI.cpp
weebl2000: You're right, I've changed it to go down to 1% (which is EU dutcycyle limit on some portions of the band. I don't think going below 1% is very useful.

src/helpers/CommonCLI.cpp
weebl2000: Revert to original response.

src/helpers/CommonCLI.cpp
weebl2000: Reverted, update docs to favor dutycycle. 👍🏽

weebl2000:
@liamcottle all comments addressed

docs/cli_commands.md
```diff
@@ -500,18 +500,35 @@ This document provides an overview of CLI commands that can be sent to MeshCore
...
+
 #### View or change the airtime factor (duty cycle limit)
+> **Deprecated** as of firmware v1.15.0. Use [`get/set dutycycle`](#view-or-change-the-duty-cycle-limit) instead.
+
 **Usage:**
 - `get af`
 - `set af <value>`
 
 **Parameters:**
-- `value`: Airtime factor (0-9). After each transmission, the repeater enforces a silent period of approximately the on-air transmission time multiplied by the value. This results in a long-term duty cycle of roughly 1 divided by (1 plus the value). For example:
```
liamcottle, line 509: Shouldn't these docs be kept for repeaters on older firmware? Sometimes repeaters will not upgrade for some time, and if airtime factor is the only option for them, having this info could be useful.

src/helpers/CommonCLI.cpp
```diff
@@ -451,7 +456,19 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
      */
     } else if (memcmp(command, "set ", 4) == 0) {
       const char* config = &command[4];
-      if (memcmp(config, "af ", 3) == 0) {
+      if (memcmp(config, "dutycycle ", 10) == 0) {
```
liamcottle, line 459: This is unlikely to be an issue, but I wanted to point it out just in case...

- Values such as 17, 18 and 19 are saved as 20% duty cycle
- Values such as 21 are saved as 25% duty cycle
- Values such as 26 are saved as 33% duty cycle
- Values such as 51+ are saved as 100% duty cycle

The numbers provided above aren't standard duty cycle values I've ever seen used by radio regulations. This issue looks to be a side affect of storing/using the original airtime factor value in preferences, rather than the raw duty cycle provided by the user.

The firmware will output the actual duty cycle that was saved to preferences, calculated from the stored airtime factor. So the user will be aware anyway, of the value that was actually saved, if it differs from what they entered.

It looks like these are the only possible duty cycle values:

- `1`
- `2`
- `3`
- `4`
- `5`
- `6.3`
- `7.1`
- `8.3`
- `9.1`
- `10`
- `11.1`
- `12.5`
- `14.3`
- `16.7`
- `20`
- `25`
- `33.3`
- `50`
- `100`

liamcottle (approved):

src/helpers/CommonCLI.cpp
weebl2000: `af` is stored as float, should work fine?

weebl2000: Screenshot_20260324-031140

liamcottle: If you try some of the values I mentioned, you'll see what I mean

weebl2000: Old version btw, let me build the latest to make sure.

docs/cli_commands.md
weebl2000: Fixed, I made a mistake restoring the edits and/or rebasing to dev when conflicts appeared because someone else updated the text heh.

src/helpers/CommonCLI.cpp
weebl2000: With current PR code:

Screenshot_20260324-031929

liamcottle: Lmao, my bad... I was testing this externally to firmware :') copied the logic into cpp tool, and had stored af as an integer... Ignore me! Good to see it is working as intended then!

weebl2000: Always good to doublecheck! ✅ ✅

docs/cli_commands.md
weebl2000: Done

PR #1972, opened 2026-03-08
Fix timewrap bug

weebl2000:
I don't think this one would actually occur in the field. But @ripplebiz managed to point me in the right direction by setting clock 5 mins prior to the 49 day wraparound.

First half of the 49 day wrap period the check works, but the second half it didn't because of the wraparound bug.

Good to have it fixed for any future changes that might impact it.

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -78,7 +78,7 @@ public:
   }
 
   void poll() override {
-    if (millis() >= dismiss_after) {
+    if (millis_passed(dismiss_after)) {
       _task->gotoHomeScreen();
     }
   }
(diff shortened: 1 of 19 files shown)


JDat:
Looks OKayish! Need to add documentation so everyone, who is adding/writing Meshcore code can easy find function named millis_passed() without looking/listing tons of headers.

weebl2000:
> Looks OKayish! Need to add documentation so everyone, who is adding/writing Meshcore code can easy find function named millis_passed() without looking/listing tons of headers.

Added a doxygen comment, do we want docs in more places?

JDat:
Is doxygen documentation generated automatically? That would be nice.

Btw, problem, theory and fix on millis() 49 day rollover problem is here.

[https://www.norwegiancreations.com/2018/10/arduino-tutorial-avoiding-the-overflow-issue-when-using-millis-and-micros/](url)

PR #1976, opened 2026-03-08
isLooped checks for any duplicate prefixes

everett1992:
previous implemenation only checked for duplicates self_id. A node would happily forward a packet that had signs it was already looped. Since not all nodes in a network will enable loop detection I want configured nodes to drop these packets.

This is a quadratic loop, but it's upper bound is 64. It can be improved with a better data structure but I wanted to keep it simple.

The package doesn't have a test framework so I wrote a small harness to double check the behavior.  All the 'self' tests pass with the previous impl.

```
runTest(1, {0x11, 0x22, 0x33}, max_loop_strict, false, "no dups");
runTest(1, {0xAA, 0x22}, max_loop_strict, true, "dup with self");
runTest(1, {0x11, 0x11, 0x22, 0x33}, max_loop_strict, true, "dup in path");
runTest(2, {0xAA, 0x22, 0xBB, 0x44}, max_loop_strict, false, "2-byte no dup");
runTest(2, {0xAA, 0xBB, 0xCC, 0xDD}, max_loop_strict, true, "2-byte dup with self");
runTest(2, {0x00, 0x11, 0x00, 0x11}, max_loop_strict, true, "2-byte dup in path");
runTest(3, {0xAA, 0xBB, 0xCC}, max_loop_strict, true, "3-byte dup with self");
runTest(3, {0x11, 0x22, 0x33, 0x11, 0x22, 0x33}, max_loop_strict, true, "3-byte dup in path");

runTest(1, {0x11, 0x11, 0x11, 0x11}, max_loop_minimal, false, "4 repeats: under minimal limit");
runTest(1, {0x11, 0x11, 0x11, 0x11, 0x11}, max_loop_minimal, true, "5 repeats: over minimal limit");
runTest(1, {0xAA, 0xAA, 0xAA}, max_loop_minimal, false, "3 repeats and self: under minimal limit");
runTest(1, {0xAA, 0xAA, 0xAA, 0xAA}, max_loop_minimal, true, "4 repeats and self: over minimal limit");
```

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -402,15 +402,21 @@ static uint8_t max_loop_strict[] =   { 0, /* 1-byte */  1, /* 2-byte */  1, /* 3
 
 bool MyMesh::isLooped(const mesh::Packet* packet, const uint8_t max_counters[]) {
   uint8_t hash_size = packet->getPathHashSize();
-  uint8_t hash_count = packet->getPathHashCount();
-  uint8_t n = 0;
+  uint8_t path_byte_len = packet->getPathByteLen();
+  uint8_t limit = max_counters[hash_size];
   const uint8_t* path = packet->path;
-  while (hash_count > 0) {      // count how many times this node is already in the path
-    if (self_id.isHashMatch(path, hash_size)) n++;
-    hash_count--;
-    path += hash_size;
+
+  for (uint8_t i = 0; i < path_byte_len; i += hash_size) {
+      uint8_t n = 0;
+
+      if (self_id.isHashMatch(&path[i], hash_size)) n++;
+
+      for (uint8_t j = i + hash_size; j < path_byte_len; j += hash_size) {
+          if (memcmp(&path[i], &path[j], hash_size) == 0) n++;
+      }
+      if (n >= limit) return true;
   }
-  return n >= max_counters[hash_size];
+  return false;
 }
 
 bool MyMesh::allowPacketForward(const mesh::Packet *packet) {


devurandomhex:
Ahh, I should have rechecked PRs before I implemented my own [very similar version](https://github.com/devurandomhex/MeshCore/commit/6d09f28a18ff0f635ea5d273ecaf0c28ec9035c0) of this. Agree that it's absolutely needed though, especially in the case where the misbehaving repeater is not running vanilla firmware - this has exhausted the airtime for our region multiple times in the last month.

I made one potential variant [here](https://github.com/devurandomhex/MeshCore/commit/142b825484b013d7259555529bfbad32b9d4f219) after considering things a bit further. In particular, I wanted to get around a potential blackhole scenario where long hauls might have a prefix collision on one byte and any strict loop detection setting would drop packets. Think `[aa]->[bb]->|other hops|->[bb]->[cc]` being caught out by the `bb`. Thus, falling back to sequential duped path bytes since that's by and large what we've seen when individual repeaters go off.

Both still do tighten the parameters around prefix-collisions within earshot of each other (especially on strict), but I feel like that's a little more expected/easy to admin your way out of (mostly have seen this when someone is flashing a bunch of nodes at home with vanity prefixes).

In general I prefer the counting-full-path method since it also catches things like bitflips which result in a U-turn packet, but wanted to lay out the case for "this might change current behavior slightly" if it lands as-is. Nice work!

examples/simple_repeater/MyMesh.cpp
```diff
@@ -402,15 +402,21 @@ static uint8_t max_loop_strict[] =   { 0, /* 1-byte */  1, /* 2-byte */  1, /* 3
...
-    hash_count--;
-    path += hash_size;
+
+  for (uint8_t i = 0; i < path_byte_len; i += hash_size) {
+      uint8_t n = 0;
+
+      if (self_id.isHashMatch(&path[i], hash_size)) n++;
+
+      for (uint8_t j = i + hash_size; j < path_byte_len; j += hash_size) {
+          if (memcmp(&path[i], &path[j], hash_size) == 0) n++;
```
robekl, line 415: loop detection will treat any repeated path hash segment as a loop, so equal 1-byte hashes can be considered loops if they collide and will cause healthy packets to be dropped. Is that okay?

everett1992: That's been the case since loop detection was added. There's an argument that this change will increase the false positive rate because we're checking all prefixes, but I think of it as asking "If that node in the path had my loop detection policy would it have repeated it?"

```
# Node AA00 with max_loop_strict would drop this packet.
# Node AA11 with max_loop_minimal would forward.
AA, BB

# Before: Node CC00 hearing this from AA11 would forward this packet regardless of it's loop detection policy
# After: Node CC00 with max_loop_strict would drop this packet, as if the 3rd hop had enforced max_loop_strict 
# Node CC11 with max_loop_minimal would still forward.
AA, BB, AA
```

These two tests are passing with original code and this change. An ID has to appear 4 times, and match your ID to be dropped.
```
assertForward(1, {0xAA, 0xAA, 0xAA}, max_loop_minimal, false, "3 repeats and self: under minimal limit");
assertDrop(1, {0xAA, 0xAA, 0xAA, 0xAA}, max_loop_minimal, true, "4 repeats and self: over minimal limit");
```

These tests pass with my change, the id must appear 5 times. 
```
assertForward(1, {0x11, 0x11, 0x11, 0x11}, max_loop_minimal, false, "4 repeats: under minimal limit");
assertDrop(1, {0x11, 0x11, 0x11, 0x11, 0x11}, max_loop_minimal, true, "5 repeats: over minimal limit");
```

In both cases the threshold accounts for the current node's ID - loops in other nodes need one more duplicate to breach the threshold.

Assuming random distribution of node ids there's a 0.15% chance to drop a full 64 hop packet with max_loop_minimal.
> About 0.15% — roughly 1 in 664 chance that when you pick 64 random numbers from 1–254, at least one 
number shows up 5 or more times.

skyepn:
> That's been the case since loop detection was added. There's an argument that this change will increase the false positive rate because we're checking all prefixes, but I think of it as asking "If that node in the path had my loop detection policy would it have repeated it?"

Makes sense, I think it should be merged as we're seeing pretty frequent corrupt packet storms in large meshes, it would make `loop.detect` a lot more useful for trunk repeater admins.

PR #1977, opened 2026-03-09
Update docs to align with implementation

robekl:
- Synced CLI documentation with the implemented command set and argument names.
- Corrected payload documentation to match the implemented request/response formats.
- Corrected companion protocol documentation to match the implemented BLE UUIDs, command layouts, response layouts, and channel secret handling.
- Corrected KISS modem documentation to match the implemented crypto behavior and packet-format reference.

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -294,7 +294,7 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 
 **Serial Only:**
 - `get prv.key`: Yes
-- `set prv.key`: Yes
+- `set prv.key`: No
 
 **Note:** Requires reboot to take effect after setting
 


docs/cli_commands.md
```diff
@@ -288,7 +294,7 @@ This document provides an overview of CLI commands that can be sent to MeshCore
 
 **Serial Only:**
 - `get prv.key`: Yes
-- `set prv.key`: No
+- `set prv.key`: Yes
```
liamcottle: `set prv.key` is allowed over LoRa as of v1.12.0+ firmware, added in https://github.com/meshcore-dev/MeshCore/pull/1457

robekl: updated

PR #1978, opened 2026-03-09
Perf/UI allocation reduction

robekl:
## Summary

This branch reduces avoidable UI-time heap allocation in two hot paths:

- Reuses a persistent scanline buffer in the ST7789 flush path instead of allocating/freeing a temporary pixel buffer on each row update.
- Rewrites `OLEDDisplay::drawString()` to scan the source string directly instead of copying it with `strdup()` and splitting it with `strtok()`.

### ST7789 scanline buffer reuse

What changed:
- Added a persistent `_pixbuf` scratch buffer and `_pixbuf_capacity` to `ST7789Spi`.
- Allocates the scratch buffer in `connect()` when needed and reuses it across `display()` calls.
- Removes per-iteration `rtos_malloc()` / `rtos_free()` calls from both ST7789 `display()` code paths.
- Frees the scratch buffer in the destructor.
- Preserves the existing buffer if a resize/allocation attempt fails.

Why:
- `display()` was doing repeated heap allocation in a render hot path.
- Reusing a single buffer reduces allocator churn and fragmentation risk on embedded targets.

### OLED drawString heap removal

What changed:
- Replaced `strdup()` + `strtok()` in `OLEDDisplay::drawString()` with direct pointer-based scanning over `strUser.c_str()`.
- Keeps vertical centering logic intact.
- Preserves the previous behavior of skipping empty newline-only segments.

Why:
- `drawString()` previously allocated and copied the full input string on every call.
- Rendering now avoids heap allocation entirely for this path.

## Behavior

Expected behavior is unchanged aside from allocation strategy:
- ST7789 rendering still produces the same converted scanline data.
- `OLEDDisplay::drawString()` still renders the same logical lines and alignment behavior as before.

## Benefit

These changes remove unnecessary heap allocation from two UI rendering paths.

- Lower runtime allocation churn during display updates.
- Reduced risk of heap fragmentation on long-running embedded devices.
- More predictable rendering performance, especially on frequent redraw paths.
- Same behavior with less transient memory traffic.

### DIFF
--- a/src/helpers/ui/OLEDDisplay.cpp
+++ b/src/helpers/ui/OLEDDisplay.cpp
@@ -619,11 +619,8 @@ uint16_t OLEDDisplay::drawStringInternal(int16_t xMove, int16_t yMove, const cha
 
 uint16_t OLEDDisplay::drawString(int16_t xMove, int16_t yMove, const String &strUser) {
   uint16_t lineHeight = pgm_read_byte(fontData + HEIGHT_POS);
-
-  // char* text must be freed!
-  char* text = strdup(strUser.c_str());
-  if (!text) {
-    DEBUG_OLEDDISPLAY("[OLEDDISPLAY][drawString] Can't allocate char array.\n");
+  const char* text = strUser.c_str();
+  if (text == nullptr) {
     return 0;
   }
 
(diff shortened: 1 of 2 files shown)


MGJ520:
Awesome, I am also adapting my screen.

PR #1983, opened 2026-03-09
Support for RAK ethernet module

rgregg:
This pull request adds Ethernet support to the `companion_radio`, `simple_repeater`, and `simple_room_server` examples for nRF52-based devices, enabling TCP CLI access and mesh command handling over Ethernet. It also updates the build workflow to include new Ethernet-enabled targets. The main changes are grouped below by theme.

Everything is compile-time gated by the ETH_ENABLE flag which is only enabled for a new environment profile for RAK firmware + Ethernet support.

### Ethernet support for examples

* Added Ethernet initialization, hardware setup, and DHCP management to `examples/simple_repeater/main.cpp` and `examples/simple_room_server/main.cpp`, including FreeRTOS task for background handling and unique MAC address generation. [[1]](diffhunk://#diff-10a0092562d43e5bb73a25560e519d6106a0d09b143204ed36ae90d8a3ad24aaR11-R127) [[2]](diffhunk://#diff-7d800f8b6cb1ad04c198fc903e3a8f600151216d62c62bfc2265207395c95b11R6-R113)
* Implemented TCP server and client handling for CLI access via Ethernet, including command parsing and reply formatting for both repeater and room server examples. [[1]](diffhunk://#diff-10a0092562d43e5bb73a25560e519d6106a0d09b143204ed36ae90d8a3ad24aaR11-R127) [[2]](diffhunk://#diff-10a0092562d43e5bb73a25560e519d6106a0d09b143204ed36ae90d8a3ad24aaR262-R296) [[3]](diffhunk://#diff-7d800f8b6cb1ad04c198fc903e3a8f600151216d62c62bfc2265207395c95b11R6-R113) [[4]](diffhunk://#diff-7d800f8b6cb1ad04c198fc903e3a8f600151216d62c62bfc2265207395c95b11R232-R266)
* Added command handling logic for Ethernet CLI, including support for the "eth" command to report connection status and IP address. [[1]](diffhunk://#diff-10a0092562d43e5bb73a25560e519d6106a0d09b143204ed36ae90d8a3ad24aaR11-R127) [[2]](diffhunk://#diff-10a0092562d43e5bb73a25560e519d6106a0d09b143204ed36ae90d8a3ad24aaR250-R253) [[3]](diffhunk://#diff-7d800f8b6cb1ad04c198fc903e3a8f600151216d62c62bfc2265207395c95b11R6-R113) [[4]](diffhunk://#diff-7d800f8b6cb1ad04c198fc903e3a8f600151216d62c62bfc2265207395c95b11R220-R223)

### Integration and initialization

* Updated `setup()` and `loop()` functions in all examples to start Ethernet tasks, maintain connections, and handle mesh commands from both serial and Ethernet interfaces. [[1]](diffhunk://#diff-10a0092562d43e5bb73a25560e519d6106a0d09b143204ed36ae90d8a3ad24aaR208-R210) [[2]](diffhunk://#diff-10a0092562d43e5bb73a25560e519d6106a0d09b143204ed36ae90d8a3ad24aaR220-R231) [[3]](diffhunk://#diff-7d800f8b6cb1ad04c198fc903e3a8f600151216d62c62bfc2265207395c95b11R181-R183) [[4]](diffhunk://#diff-7d800f8b6cb1ad04c198fc903e3a8f600151216d62c62bfc2265207395c95b11R193-R196) [[5]](diffhunk://#diff-7d800f8b6cb1ad04c198fc903e3a8f600151216d62c62bfc2265207395c95b11R128-R130) [[6]](diffhunk://#diff-7d800f8b6cb1ad04c198fc903e3a8f600151216d62c62bfc2265207395c95b11R220-R223) [[7]](diffhunk://#diff-7d800f8b6cb1ad04c198fc903e3a8f600151216d62c62bfc2265207395c95b11R232-R266) [[8]](diffhunk://#diff-10a0092562d43e5bb73a25560e519d6106a0d09b143204ed36ae90d8a3ad24aaR250-R253) [[9]](diffhunk://#diff-10a0092562d43e5bb73a25560e519d6106a0d09b143204ed36ae90d8a3ad24aaR262-R296)
* Modified `examples/companion_radio/main.cpp` to support conditional inclusion of Ethernet interface and initialization, including debug output and error handling. [[1]](diffhunk://#diff-cb110bbd265c3b5e7b496073da4f75e83b1b56d61776d2692d4b4a800c4b46f8R79-R93) [[2]](diffhunk://#diff-cb110bbd265c3b5e7b496073da4f75e83b1b56d61776d2692d4b4a800c4b46f8R164-R180) [[3]](diffhunk://#diff-cb110bbd265c3b5e7b496073da4f75e83b1b56d61776d2692d4b4a800c4b46f8R254-R263) [[4]](diffhunk://#diff-cb110bbd265c3b5e7b496073da4f75e83b1b56d61776d2692d4b4a800c4b46f8R15-R16)

### Build workflow updates

* Added new Ethernet-enabled build targets for nRF52 devices in `.github/workflows/pr-build-check.yml` to ensure CI coverage for the new features.

These changes collectively enable robust Ethernet-based CLI and mesh communication for nRF52 devices, improving flexibility and connectivity options.


rgregg:
Updated to address all the issues in the Copilot feedback.

src/helpers/nrf52/SerialEthernetInterface.cpp
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  return 0;
+}
+
+bool SerialEthernetInterface::isConnected() const {
+  return deviceConnected;  //pServer != NULL && pServer->getConnectedCount() > 0;
+}
+
+void SerialEthernetInterface::generateDeviceMac(uint8_t mac[6]) {
```
liamcottle: this function looks to be duplicated in a few places, can this class version be used in the other places to avoid duplication?

src/helpers/nrf52/SerialEthernetInterface.h
```diff
@@ -0,0 +1,85 @@
+#pragma once
+
+#include "helpers/BaseSerialInterface.h"
+#include <SPI.h>
+#include <RAK13800_W5100S.h>
+
+#ifndef ETH_TCP_PORT
+  #define ETH_TCP_PORT 5000
```
liamcottle: please refactor all references of `ETH_...` to `ETHERNET_...` so it's more explicit

src/helpers/nrf52/SerialEthernetInterface.h
```diff
@@ -0,0 +1,85 @@
...
+    void maintain();
+
+private:
+  void generateDeviceMac(uint8_t mac[6]);
+};
+
+
+#if ETH_DEBUG_LOGGING && ARDUINO
+  #include <Arduino.h>
+  #define ETH_DEBUG_PRINT(F, ...) Serial.printf("ETH: " F, ##__VA_ARGS__)
```
liamcottle: same with these, some may have no idea what `ETH` means, `ETHERNET` is more explicit

variants/rak4631/platformio.ini
```diff
@@ -46,6 +46,26 @@ build_src_filter = ${rak4631.build_src_filter}
   +<helpers/ui/SSD1306Display.cpp>
   +<../examples/simple_repeater>
 
+[env:RAK_4631_repeater_eth]
```
liamcottle: prob good to go with `RAK_4631_repeater_ethernet` as well, same with other variants

variants/rak4631/platformio.ini
```diff
@@ -46,6 +46,26 @@ build_src_filter = ${rak4631.build_src_filter}
...
+extends = rak4631
+build_flags =
+  ${rak4631.build_flags}
+  -D DISPLAY_CLASS=SSD1306Display
+  -D ADVERT_NAME='"RAK4631 Repeater ETH"'
```
liamcottle: same here `RAK4631 Repeater Ethernet`, else just drop it entirely

variants/rak4631/platformio.ini
```diff
@@ -46,6 +46,26 @@ build_src_filter = ${rak4631.build_src_filter}
...
+  -D ADVERT_LAT=0.0
+  -D ADVERT_LON=0.0
+  -D ADMIN_PASSWORD='"password"'
+  -D MAX_NEIGHBOURS=50
+  -D ETH_ENABLED=1
```
liamcottle: `ETHERNET_ENABLED=1`

.github/workflows/pr-build-check.yml
```diff
@@ -23,8 +23,11 @@ jobs:
           - Heltec_v3_room_server
           # nRF52
           - RAK_4631_companion_radio_ble
+          - RAK_4631_companion_radio_eth
```
liamcottle: prefer to have full name `RAK_4631_companion_radio_ethernet`

examples/companion_radio/main.cpp
```diff
@@ -12,19 +12,20 @@ static uint32_t _atoi(const char* sp) {
...
     DataStore store(InternalFS, QSPIFlash, rtc_clock);
   #else
-  #if defined(EXTRAFS)
-    #include <CustomLFS.h>
-    CustomLFS ExtraFS(0xD4000, 0x19000, 128);
-    DataStore store(InternalFS, ExtraFS, rtc_clock);
-  #else
-    DataStore store(InternalFS, rtc_clock);
-  #endif
+    #if defined(EXTRAFS)
```
liamcottle, line 21: not sure what happened with indentation here?

examples/companion_radio/main.cpp
```diff
@@ -74,13 +75,21 @@ static uint32_t _atoi(const char* sp) {
   #ifdef BLE_PIN_CODE
     #include <helpers/nrf52/SerialBLEInterface.h>
     SerialBLEInterface serial_interface;
+  #elif defined(ETH_ENABLED)
```
liamcottle: `ETH_ENABLED` -> `ETHERNET_ENABLED`

examples/companion_radio/main.cpp
```diff
@@ -225,4 +248,8 @@ void loop() {
   ui_task.loop();
 #endif
   rtc_clock.tick();
+
+#ifdef ETH_ENABLED
+  serial_interface.maintain();
```
liamcottle: should prob call it `.loop` so it's consistent with the others?

examples/simple_repeater/main.cpp
```diff
@@ -8,6 +8,123 @@
...
 #endif
 
+#ifdef ETH_ENABLED
+  #include <SPI.h>
+  #include <RAK13800_W5100S.h>
+
+  #define PIN_SPI1_MISO (29)
+  #define PIN_SPI1_MOSI (30)
+  #define PIN_SPI1_SCK  (3)
+  SPIClass ETH_SPI_PORT(NRF_SPIM1, PIN_SPI1_MISO, PIN_SPI1_SCK, PIN_SPI1_MOSI);
```
liamcottle: `ETH_SPI_PORT` -> `ETHERNET_SPI_PORT`

examples/simple_repeater/main.cpp
```diff
@@ -8,6 +8,123 @@
...
+    #define ETH_TCP_PORT 23  // telnet port for CLI access
+  #endif
+
+  #define ETH_RETRY_INTERVAL_MS 30000
+
+  static EthernetServer eth_server(ETH_TCP_PORT);
+  static EthernetClient eth_client;
+  static volatile bool eth_running = false;
+
+  static void generateDeviceMac(uint8_t mac[6]) {
```
liamcottle: `generateDeviceMac` seems to be used in a few places, would be good to refactor to one class somewhere

examples/simple_repeater/main.cpp
```diff
@@ -8,6 +8,123 @@
...
+  #define PIN_ETHERNET_SS    26
+
+  #ifndef ETH_TCP_PORT
+    #define ETH_TCP_PORT 23  // telnet port for CLI access
+  #endif
+
+  #define ETH_RETRY_INTERVAL_MS 30000
+
+  static EthernetServer eth_server(ETH_TCP_PORT);
+  static EthernetClient eth_client;
```
liamcottle: `eth_client` -> `ethernet_client`

examples/simple_repeater/main.cpp
```diff
@@ -119,6 +247,10 @@ void loop() {
     Serial.print('\n');
     command[len - 1] = 0;  // replace newline with C string null terminator
     char reply[160];
+    reply[0] = 0;
+#ifdef ETH_ENABLED
+    if (!eth_handle_command(command, reply))
```
liamcottle: looks to be an `if` check, with no body?

examples/simple_room_server/main.cpp
```diff
@@ -3,6 +3,114 @@
...
+  #define PIN_SPI1_MOSI (30)
+  #define PIN_SPI1_SCK  (3)
+  SPIClass ETH_SPI_PORT(NRF_SPIM1, PIN_SPI1_MISO, PIN_SPI1_SCK, PIN_SPI1_MOSI);
+
+  #define PIN_ETH_POWER_EN  WB_IO2
+  #define PIN_ETHERNET_RESET 21
+  #define PIN_ETHERNET_SS    26
+
+  #ifndef ETH_TCP_PORT
+    #define ETH_TCP_PORT 23
```
liamcottle: `ETH_TCP_PORT` -> `ETHERNET_TCP_PORT`

examples/simple_room_server/main.cpp
```diff
@@ -99,6 +217,10 @@ void loop() {
   if (len > 0 && command[len - 1] == '\r') {  // received complete line
     command[len - 1] = 0;  // replace newline with C string null terminator
     char reply[160];
+    reply[0] = 0;
+#ifdef ETH_ENABLED
+    if (!eth_handle_command(command, reply))
```
liamcottle: seems like another empty if body?

src/helpers/nrf52/SerialEthernetInterface.cpp
```diff
@@ -0,0 +1,303 @@
...
+      mac[3],
+      mac[4],
+      mac[5]);
+  ETH_DEBUG_PRINTLN("Init");
+  ETH_SPI_PORT.begin();
+  Ethernet.init(ETH_SPI_PORT, PIN_ETHERNET_SS);
+
+  // Hardcode IP address for now
+  #ifdef STATIC_IP
+  IPAddress ip(192, 168, 8, 118);
```
liamcottle: could prob use the build flag to define the static ips, such as the following.

- `ETHERNET_IP=192.168.1.2`
- `ETHERNET_GATEWAY=192.168.1.1`
- `ETHERNET_SUBNET_MASK=255.255.255.0`
- `ETHERNET_DNS=192.168.1.1`

variants/rak4631/platformio.ini
```diff
@@ -108,6 +128,26 @@ build_src_filter = ${rak4631.build_src_filter}
   +<helpers/ui/SSD1306Display.cpp>
   +<../examples/simple_room_server>
 
+[env:RAK_4631_room_server_eth]
```
liamcottle: `RAK_4631_room_server_eth` -> `RAK_4631_room_server_ethernet`

variants/rak4631/platformio.ini
```diff
@@ -108,6 +128,26 @@ build_src_filter = ${rak4631.build_src_filter}
...
+extends = rak4631
+build_flags =
+  ${rak4631.build_flags}
+  -D DISPLAY_CLASS=SSD1306Display
+  -D ADVERT_NAME='"Test Room ETH"'
```
liamcottle: prob good to remove "Test" and make it similar to existing room server names

liamcottle (changes requested):
Howdy! Thanks for the PR. I've added some comments :)

src/helpers/nrf52/SerialEthernetInterface.cpp
rgregg: Yep - thanks for pointing this out. Fixed.

src/helpers/nrf52/SerialEthernetInterface.h
rgregg: Ack - makes sense.

examples/companion_radio/main.cpp
rgregg: Corrected the intendentation.

examples/companion_radio/main.cpp
rgregg: Fair enough - the ethernet driver sample code used maintain, but we should be consistent.

examples/simple_repeater/main.cpp
rgregg: A bit too smart here, cleaned this up to be more readable.

variants/rak4631/platformio.ini
rgregg: The other names for the RAK firmware have "Test" in them - but I agree w/ you so I'm removing it here.

rgregg:
Thanks @liamcottle for the review and the feedback - I've updated the PR to reflect your feedback, I appreciate your time!

docs/cli_commands.md
```diff
@@ -881,3 +882,25 @@ region save
 **Default:** Varies by board
 
 ---
+
+### Ethernet (when Ethernet support is compiled in)
+
+Ethernet support is available on RAK4631 boards with a RAK13800 (W5100S) Ethernet module. Use the `_eth` firmware variants (e.g. `RAK_4631_repeater_eth`) to enable this feature.
```
liamcottle: missed this one ;) `_eth` -> `_ethernet`.

docs/faq.md
```diff
@@ -878,3 +879,27 @@ For companion radios, you can set these radios' transmit power in the smartphone
...
+
+**Hardware required:**
+- RAK4631 WisBlock Core
+- RAK19007 or RAK19018 WisBlock Base Board (with an available IO slot)
+- RAK13800 WisBlock Ethernet module
+- Ethernet cable connected to a network with a DHCP server
+
+**Firmware:**
+Flash one of the Ethernet-enabled firmware variants:
+- `RAK_4631_repeater_eth` - Repeater with Ethernet CLI access
```
liamcottle: Docs update for new `_ethernet` suffix

examples/companion_radio/main.cpp
```diff
@@ -152,6 +160,17 @@ void setup() {
...
+#elif defined(ETHERNET_ENABLED)
+  Serial.print("Waiting for serial to connect...\n");
+  time_t timeout = millis();
+  while (!Serial) {
+    if ((millis() - timeout) < 5000) { delay(100); } else { break; }
+  }
+  Serial.println("Initializing Ethernet adapter...");
+  if (!serial_interface.begin()) {
+    Serial.println("ETH: Init failed, halting");
+    halt();
```
liamcottle: I'm wondering if the firmware should still run as normal if ethernet fails to come up? That way remote management over LoRa would still work, and it could be rebooted remotely over the mesh?

Otherwise if firmware was to shutdown due to low voltage, and ethernet failed to come up on reboot, maybe due to still having low power, the unit would be bricked until physically rebooted.

Not sure if this would cause issues reading/writing commands to non initialised ethernet driver...

examples/simple_repeater/main.cpp
```diff
@@ -8,6 +8,116 @@
...
+    vTaskDelete(NULL);
+  }
+
+  static void ethernet_start_task() {
+    xTaskCreate(ethernet_task, "eth_init", 1024, NULL, 1, NULL);
+  }
+
+  // Format ethernet status into reply buffer. Returns true if command was handled.
+  static bool ethernet_handle_command(const char* command, char* reply) {
+    if (strcmp(command, "eth") != 0) return false;
```
liamcottle: could we switch the logic, so instead of returning false for a non match for the only command, wrap the status response inside an if block, and return false from ethernet_handle_command by default to indicate it didn't handle it. More so for when you add new commands later, this doesn't need a refactor. e.g:

```
static bool ethernet_handle_command(const char* command, char* reply) {
  if (strcmp(command, "eth.status") == 0) {
    if (!ethernet_running) {
      strcpy(reply, "not connected");
    } else {
      IPAddress ip = Ethernet.localIP();
      sprintf(reply, "IP: %u.%u.%u.%u:%d", ip[0], ip[1], ip[2], ip[3], ETHERNET_TCP_PORT);
    }
    return true;
  }
  
  // not handled
  return false;
}
```

Maybe we could rename it to `eth.status` or `eth.ip`. I'll leave that one up to you.

examples/simple_repeater/main.cpp
```diff
@@ -8,6 +8,116 @@
...
+    } else {
+      IPAddress ip = Ethernet.localIP();
+      sprintf(reply, "ETH: %u.%u.%u.%u:%d", ip[0], ip[1], ip[2], ip[3], ETHERNET_TCP_PORT);
+    }
+    return true;
+  }
+
+  // Check for new TCP client connections
+  static void ethernet_check_client() {
+    if (ethernet_client && ethernet_client.connected()) return;
```
liamcottle: I think this check would prevent new clients from being able to connect until the existing client disconnects. The WiFi firmware allows new connections to kick off old connections. We should allow the same here.

examples/simple_repeater/main.cpp
```diff
@@ -8,6 +8,116 @@
...
+
+    auto newClient = ethernet_server.available();
```
liamcottle: This logic below allows new connections to replace older connections, but due to that connected check above, it would never allow it.

examples/simple_repeater/main.cpp
```diff
@@ -119,14 +240,57 @@ void loop() {
...
+      ethernet_client.println();
+      char reply[160];
+      reply[0] = 0;
+      if (!ethernet_handle_command(ethernet_command, reply)) {
+        the_mesh.handleCommand(0, ethernet_command, reply);
+      }
+      if (reply[0]) {
+        ethernet_client.print("  -> "); ethernet_client.println(reply);
+      }
+      ethernet_client.print("> ");
```
liamcottle: I'm not sure this is done by the standard serial interface? We should prob make sure it behaves same way. I've not verified.

docs/faq.md
```diff
@@ -878,3 +879,27 @@ For companion radios, you can set these radios' transmit power in the smartphone
...
+- `RAK_4631_room_server_eth` - Room server with Ethernet CLI access
+- `RAK_4631_companion_radio_eth` - Companion radio over Ethernet (replaces BLE)
+
+**Connecting:**
+- The device obtains an IP address via DHCP automatically on boot.
+- For repeaters and room servers, connect to the device on TCP port 5000 using any TCP client (e.g. `nc <ip> 5000` or PuTTY in raw mode). This gives you the same CLI available over serial/USB.
```
liamcottle: I saw you were using different ports, 23 for cli and 5000 for companion (which is standard for wifi as well). Prob need to document that here, as it currently says to use 5000 for repeaters and room servers, but code is actually using 23.

examples/simple_room_server/main.cpp
```diff
@@ -3,6 +3,107 @@
...
+    }
+    vTaskDelete(NULL);
+  }
+
+  static void ethernet_start_task() {
+    xTaskCreate(ethernet_task, "eth_init", 1024, NULL, 1, NULL);
+  }
+
+  static bool ethernet_handle_command(const char* command, char* reply) {
+    if (strcmp(command, "eth") != 0) return false;
```
liamcottle: same comments as repeater feedback

examples/simple_room_server/main.cpp
```diff
@@ -3,6 +3,107 @@
...
+      strcpy(reply, "ETH: not connected");
+    } else {
+      IPAddress ip = Ethernet.localIP();
+      sprintf(reply, "ETH: %u.%u.%u.%u:%d", ip[0], ip[1], ip[2], ip[3], ETHERNET_TCP_PORT);
+    }
+    return true;
+  }
+
+  static void ethernet_check_client() {
+    if (ethernet_client && ethernet_client.connected()) return;
```
liamcottle: same comments as repeater feedback about return preventing new clients connecting

examples/simple_room_server/main.cpp
```diff
@@ -99,14 +210,57 @@ void loop() {
...
+    reply[0] = 0;
+#ifdef ETHERNET_ENABLED
+    if (!ethernet_handle_command(command, reply)) {
+      the_mesh.handleCommand(0, command, reply);
+    }
+#else
     the_mesh.handleCommand(0, command, reply);  // NOTE: there is no sender_timestamp via serial!
+#endif
     if (reply[0]) {
       Serial.print("  -> "); Serial.println(reply);
```
liamcottle, line 132: I was originally wondering if the reply is meant to be going back out of the ethernet_interface instead of the Serial, but noted further down the reply is sent out as expected. I guess this just means the reply is sent out on main serial _and_ on ethernet. Not sure if that's an issue.

examples/simple_room_server/main.cpp
```diff
@@ -99,14 +210,57 @@ void loop() {
...
+      ethernet_client.println();
+      char reply[160];
+      reply[0] = 0;
+      if (!ethernet_handle_command(ethernet_command, reply)) {
+        the_mesh.handleCommand(0, ethernet_command, reply);
+      }
+      if (reply[0]) {
+        ethernet_client.print("  -> "); ethernet_client.println(reply);
+      }
+      ethernet_client.print("> ");
```
liamcottle: Same comment as noted on repeater firmware

src/helpers/nrf52/EthernetMac.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <Arduino.h>
+
+static inline void generateDeviceMac(uint8_t mac[6]) {
```
liamcottle: could maybe rename it so it's clearer? `generateEthernetMac`? Haven't looked to see what `NRF_FICR->DEVICEID` does, but I'm assuming we are generating an ethernet mac address, from the existing hardware mac address or device id.

src/helpers/nrf52/SerialEthernetInterface.cpp
```diff
@@ -0,0 +1,292 @@
...
+}
+
+bool SerialEthernetInterface::isWriteBusy() const {
+  return false;
+}
+
+size_t SerialEthernetInterface::checkRecvFrame(uint8_t dest[]) {
+  // check if new client connected
+  if (client && client.connected()) {
+    // Avoid polling for new clients while an active connection exists.
```
liamcottle: we should allow new clients to connect and kick off the old client

liamcottle (changes requested):
Awesome, thanks for the adjustments. I've gone through again and have some extra comments. Once these are sorted, would be good if you can test it's all working before I ask Scott to do a code review.

JDat:
Can this be made to work with W5500-EVB-PICO (Wiznet5500 + RPi Pi Pico RP2040) board because RAK is no the only toy with Ethernet.

neopiccolorat:
This PR comes at the right time!
I have this RAK13800 and I'm currently using with Meshtastic, but I wanted to switch this node to Meshcore and still being able to connect to it by its ip.

examples/companion_radio/main.cpp
rgregg: Good point - I think we can guard for that issue. I'll see how difficult that would be.

rgregg:
> Can this be made to work with W5500-EVB-PICO (Wiznet5500 + RPi Pi Pico RP2040) board because RAK is no the only toy with Ethernet.

Should be possible to extend. I don't have that hardware to test/validate with though.

docs/cli_commands.md
rgregg: Fixed - also updated the port number in the example output and notes while I was in there.

docs/faq.md
rgregg: Fixed - also corrected the port numbers: port 23 for repeater/room server CLI, port 5000 for companion radio.

examples/companion_radio/main.cpp
rgregg: Good call. Changed to log the failure and continue without the Ethernet interface - the mesh runs normally over LoRa so the unit can be recovered remotely.

examples/simple_repeater/main.cpp
rgregg: Refactored as suggested and went with `eth.status`.

examples/simple_repeater/main.cpp
rgregg: Fixed - removed the early return guard so `server.available()` is always called. New connections now replace existing ones.

examples/simple_repeater/main.cpp
rgregg: Fixed - both this and the guard above are resolved.

examples/simple_repeater/main.cpp
rgregg: Checked the serial interface - it doesn't send a prompt. Dropped the `> ` prompt from the ethernet CLI for consistency.

docs/faq.md
rgregg: Fixed in both faq.md and cli_commands.md: port 23 for repeater/room server CLI, port 5000 for companion radio.

examples/simple_room_server/main.cpp
rgregg: Fixed.

examples/simple_room_server/main.cpp
rgregg: Fixed.

examples/simple_room_server/main.cpp
rgregg: The channels are already separate - serial commands reply only to Serial, ethernet commands reply only to ethernet_client. No change needed.

examples/simple_room_server/main.cpp
rgregg: Fixed - dropped the prompt.

src/helpers/nrf52/EthernetMac.h
rgregg: Renamed to `generateEthernetMac`. It uses `NRF_FICR->DEVICEID` (the chip's unique hardware ID) to derive a stable MAC address, with the locally-administered bit set so it's valid for private use without conflicting with real hardware MACs.

src/helpers/nrf52/SerialEthernetInterface.cpp
rgregg: Fixed - removed the guard that skipped `server.available()` while a connection was active. New connections now replace existing ones.

JDat:
> > Can this be made to work with W5500-EVB-PICO (Wiznet5500 + RPi Pi Pico RP2040) board because RAK is no the only toy with Ethernet.
> 
> Should be possible to extend. I don't have that hardware to test/validate with though.

I have this board and I am OK to solder/wire LoRA module with board, and do some tests.

rgregg:
All three Ethernet-enabled firmwares have been tested on RAK4631 hardware with the RAK13800 W5100S Ethernet module and are working as expected:

- **RAK_4631_repeater_ethernet** — DHCP + CLI over telnet (port 23) ✅
- **RAK_4631_room_server_ethernet** — DHCP + CLI over telnet (port 23) ✅
- **RAK_4631_companion_radio_ethernet** — DHCP + companion protocol (port 5000) ✅

rgregg:
@liamcottle  - let me know if there is anything else, otherwise I think this is ready to merge.

JDat:
For what reason Ethernet header sit inside nRF52 helper directory? What if Ethernet need to be added to ESP32 and RP2040 microcontrollers. What a crazy platform dependent b**t! If you want Ethernet support, make it platform independent instead or only board laying on your table!

rgregg:
> For what reason Ethernet header sit inside nRF52 helper directory? What if Ethernet need to be added to ESP32 and RP2040 microcontrollers. What a crazy platform dependent b**t! If you want Ethernet support, make it platform independent instead or only board laying on your table!

This follows the existing pattern of SerialBLEInterface which lives in the esp32 and nrf52 folders. Firmware development is often very hardware specific - if it turns out we can reuse the same code, it'll be easy enough to refactor it accordingly when someone with the hardware can verify it works. There's no need to be rude about it.

JDat:
> > For what reason Ethernet header sit inside nRF52 helper directory? What if Ethernet need to be added to ESP32 and RP2040 microcontrollers. What a crazy platform dependent b**t! If you want Ethernet support, make it platform independent instead or only board laying on your table!
> 
> This follows the existing pattern of SerialBLEInterface which lives in the esp32 and nrf52 folders. Firmware development is often very hardware specific - if it turns out we can reuse the same code, it'll be easy enough to refactor it accordingly when someone with the hardware can verify it works. There's no need to be rude about it.

Why not to refactor things correctly as soon as they are developed? Is this future of vibe coding or what?

rgregg:
Ok - I have this running on a repeater now and seems to be stable and working as expected.

vaizki:
> Ok - I have this running on a repeater now and seems to be stable and working as expected.

I have also been running your PR fork for 2 days now on my repeater, seems fine. Thank you for making this!

drewzh:
Sorry to add to the noise, but does anyone have any idea when this getting merged?

Hiltronix:
I was also wondering when it was going to be merged, I was looking in the release notes for this in the release that came out this past weekend.  Which brought me back here to see it's still waiting approval.

rgregg:
Updated for the changes in the 1.15.0 release.

examples/companion_radio/main.cpp
```diff
@@ -12,19 +12,20 @@ static uint32_t _atoi(const char* sp) {
...
-  #else
-    DataStore store(InternalFS, rtc_clock);
-  #endif
+    #if defined(EXTRAFS)
+      #include <CustomLFS.h>
+      CustomLFS ExtraFS(0xD4000, 0x19000, 128);
+      DataStore store(InternalFS, ExtraFS, rtc_clock);
+    #else
+      DataStore store(InternalFS, rtc_clock);
+    #endif
```
oltaco, line 27: Is there still an indent issue here?

examples/companion_radio/main.cpp
```diff
@@ -74,13 +75,21 @@ static uint32_t _atoi(const char* sp) {
...
     #include <helpers/ArduinoSerialInterface.h>
     ArduinoSerialInterface serial_interface;
   #endif
 #elif defined(STM32_PLATFORM)
-  #include <helpers/ArduinoSerialInterface.h>
-  ArduinoSerialInterface serial_interface;
+  #ifdef ETHERNET_ENABLED
+    #include <helpers/nrf52/SerialEthernetInterface.h>
+    SerialEthernetInterface serial_interface;
+  #else
```
oltaco: I think this was probably unintended?

variants/rak4631/platformio.ini
```diff
@@ -48,6 +48,26 @@ build_src_filter = ${rak4631.build_src_filter}
...
+  -D ADMIN_PASSWORD='"password"'
+  -D MAX_NEIGHBOURS=50
+  -D ETHERNET_ENABLED=1
+;  -D MESH_DEBUG=1
+build_src_filter = ${rak4631.build_src_filter}
+  +<helpers/ui/SSD1306Display.cpp>
+  +<../examples/simple_repeater>
+lib_deps =
+  ${rak4631.lib_deps}
+  # renovate: datasource=github-tags depName=RAK13800-W5100S packageName=RAKWireless/RAK13800-W5100S
```
oltaco: I don't think we're using renovate anywhere else, so these can probably go?

oltaco (commented):
Nice work, very cool 😎

variants/rak4631/platformio.ini
rgregg: Indeed, that shouldn't have been there. Removed.

examples/companion_radio/main.cpp
rgregg: Yeah, this was left over from a previous iteration. Thanks for flagging. Now removed.

examples/companion_radio/main.cpp
rgregg: Everything appears properly intended in the full file view to me, with each #if/elif/else being intended by 2 spaces.

rgregg:
Thanks for the review @oltaco! I've updated the PR based on your feedback.

cwichura:
@rgregg Would you consider adding support for redirecting `MESH_PACKET_LOGGING` to the Ethernet session (at least for non-companions)?  This would allow mctomqtt (with the TCP PR waiting on it applied) to act as an observer to a device with PoE Ethernet as its hardwired connectivity.  Currently, `MESH_PACKET_LOGGING` writes to `Serial.`, which goes to the USB serial console output.  Thanks.

src/helpers/nrf52/SerialEthernetInterface.cpp
```diff
@@ -0,0 +1,277 @@
...
+    if (client) {
+      ETHERNET_DEBUG_PRINTLN("Closing previous client");
+      client.stop();
+    }
+    _state = RECV_STATE_IDLE;
+    _frame_len = 0;
+    _rx_len = 0;
+    client = newClient;
+    ETHERNET_DEBUG_PRINTLN("Switched to new client");
+  }
```
mcode6726, line 149: Has this actually been tested with Android/Iphone/Meshcore-CLI for companion? 
As a compile and test results in an immediate disconnect after first packet of data sent to the companion and it seems to be because of the code here.   This doesn't seem consistant with the RAK examples.   As any packet is received newClient will be created, and will point to same exact same socket#  as the previous client object already in memory that was created from previous packet, so you hit the code that called newClient.stop(0 which actually wiill close newClient AND client sockets. Which disconnects the companion.

I think I have been able to rework the logic to make it work, as at least my the changes I have made to the logic flow seem to have fixed it.  Will try to put a pull request in over on rgregg:rak-ethernet

cwichura:
@rgregg As an alternative to MESH_PACKET_LOGGING, add support for the Ethernet module to the KISS firmware?  Then we could run pyMC_repeater in a container/VM and have it connect remotely over Ethernet to the LoRa modem.

Also, do you have builds of this in its current state available online anywhere?  If so, I'd be happy to try testing it on my RAK Ethernet MQTT Gateway kit.   Thanks.

rgregg:
@mcode6726 — Thanks for catching the companion disconnect, I've mostly tested the repeater firmware which didn't have the same bug. I've adjusted both paths to use `server.accept()` instead as a better call pattern.

@liamcottle — all of your round-2 feedback plus oltaco's nits have been addressed in earlier commits, and this latest fix resolves the companion bug. Ready for another look when you have time.

AcidWizard:
Whats needed for this to be implemented? I want to use something similar to this but with an esp32-Olimex-POE board so using this as a base would be great, but if it will not be merged than I will continue on looking at a Raspberry Pi + normal MC device combination to put up on the roof.

If programming help is needed I am available as well, I however do not own any RAK modules right now. If needed I can acquire some.

rgregg:
@AcidWizard - I'm not really sure tbh. I've been running this firmware on my own repeater and companion nodes for a while, everything has been working great for me. I'd love to get this merged, it seems like there is sufficient interest in supporting it, but I haven't from the maintainers in several months.

Drasli-Tana:
So I've tested a bit with the repeater firmware, on a RAK4631microcontroller paired with a RAK 13800 Ethernet module. Everything seems to be working fine for now using DHCP, commands work the same way the CLI does.

It might be useful to later store network configuration in node memory instead of relying on values set at compile time when using a static IP, to avoid having to reflash everything if that config has to change afterwards.

rgregg:
@liamcottle — another nudge on this when you have time. Since your last review, all of your round-2 feedback is in (the `ETH_*` → `ETHERNET_*` rename, `serial_interface.loop()` naming, the build-env names), along with oltaco's nits and the companion-disconnect fix using `server.accept()` on both paths.

I've also merged up to current `upstream/dev`, so it's mergeable with no conflicts. It's had some outside testing now too — @Drasli-Tana confirmed the repeater firmware working on RAK4631 + RAK13800 over DHCP, and I've been running it on my own repeater and companion nodes for a few months without issues.

Your CHANGES_REQUESTED is the only thing holding up the review decision at this point. Happy to fix anything else you spot, but I think it's ready for another look.

rgregg:
@ripplebiz I am hoping to get your read on this one when you have a moment -- or any other maintainer.

Quick status: all of @liamcottle's review feedback from the earlier rounds is in (the `ETHERNET_*` rename, the `serial_interface.loop()` naming, and the build-env changes), CI is green, and the branch is mergeable. It's also had independent hardware validation beyond my own nodes, @Drasli-Tana tested the repeater firmware on a RAK4631 + RAK13800 over DHCP and reported everything working.

The main reason I'm tagging you: there are now **two overlapping Ethernet PRs on the bench** and I'd like a steer on which one you'd prefer to take forward so we're not duplicating effort:

- **#1983** (this PR):  base RAK Ethernet module support for repeater + companion, which I've been running in production for a while.
- **#2679** (@1sthandy): RAK4631/RAK10720 PoE + Ethernet companion support, which covers similar hardware and would arguably sit cleanly on top of this base once it lands.

I'm happy to go whichever direction makes the most sense to you, either get this merged as the foundation and coordinate #2679 on top, consolidate into one, or adjust scope. Just let me know your preference and I'll run with it. Thanks!

1sthandy:
@rgregg thanks for teeing this up so clearly. From my side I'm fully on board with your suggested direction: land #1983 as the Ethernet foundation, and I'll rebase the PoE-specific parts of #2679 on top of it.

The PoE handling (W5100S power/reset sequencing, no-sleep, boot-voltage bypass, deferred bring-up) is largely transport-independent, so it should port cleanly onto your base — and it would apply to the repeater/room-server variants too, not just the companion.

Happy to coordinate whenever #1983 merges. @ripplebiz — a steer here would be much appreciated.

ripplebiz (commented):
Happy to merge, if you can attend to merge conflicts.

rgregg:
Thanks @ripplebiz  - I've pushed a fix for the merge conflict and validated the PR functions as expected.

enigmaspb:
Hello @rgregg need some fix for you PR please see https://github.com/meshcore-dev/MeshCore/issues/2985

PR #1984, opened 2026-03-10
Add nrf52 power management for RAK3401

samschlegel:
## Summary
- Enable nrf52 power management for RAK3401, ported from the RAK4631 configuration

## Test plan
- [x] Verify boot protection triggers below 3300mV
- [x] Verify LPCOMP wake from SYSTEMOFF when voltage recovers
- [x] Verify normal boot path is unaffected above threshold

### DIFF
--- a/variants/rak3401/RAK3401Board.cpp
+++ b/variants/rak3401/RAK3401Board.cpp
@@ -3,6 +3,31 @@
 
 #include "RAK3401Board.h"
 
+#ifdef NRF52_POWER_MANAGEMENT
+// Static configuration for power management
+// Values set in variant.h defines
+const PowerMgtConfig power_config = {
+  .lpcomp_ain_channel = PWRMGT_LPCOMP_AIN,
+  .lpcomp_refsel = PWRMGT_LPCOMP_REFSEL,
+  .voltage_bootlock = PWRMGT_VOLTAGE_BOOTLOCK
+};
+
+void RAK3401Board::initiateShutdown(uint8_t reason) {
+  // Disable SKY66122 FEM (CSD+CPS LOW = shutdown, <1 uA)
+  digitalWrite(SX126X_POWER_EN, LOW);
+
+  // Disable 3V3 switched peripherals and 5V boost
+  digitalWrite(PIN_3V3_EN, LOW);
+
+  if (reason == SHUTDOWN_REASON_LOW_VOLTAGE ||
+      reason == SHUTDOWN_REASON_BOOT_PROTECT) {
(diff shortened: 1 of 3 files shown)


samschlegel:
Will get a verified commit pushed up shortly

EDIT: Done!

variants/rak3401/platformio.ini
```diff
@@ -124,4 +125,4 @@ build_flags =
   ;-D MESH_DEBUG=1
 build_src_filter = ${rak3401.build_src_filter}
   +<helpers/ui/SSD1306Display.cpp>
-  +<../examples/simple_sensor>
```
samschlegel, line 127: can remove if desired. I have a pretty minimal VSCode setup and I'm pretty sure this is just the default so maybe work keeping?

variants/rak3401/RAK3401Board.cpp
```diff
@@ -31,18 +56,11 @@ void RAK3401Board::begin() {
...
+  // Boot voltage protection check (may not return if voltage too low)
+  // We need to call this after we configure SX126X_POWER_EN as output but before we pull high
+  checkBootVoltage(&power_config);
+#endif
   digitalWrite(SX126X_POWER_EN, HIGH);
   delay(1);  // SKY66122 turn-on settling time (tON = 3us typ)
 }
-
-#ifdef NRF52_POWER_MANAGEMENT
-void RAK3401Board::initiateShutdown(uint8_t reason) {
```
samschlegel, line 39: had to move the block for power_config to be defined before begin and figured it made sense to just move it all

linsomniac:
I gather this fixes the issue where the RAK 3401 goes into deep sleep when the battery voltage drops below 3.3v and then never recovers when the battery comes back up above that?  I'm currently following the Austin Mesh guidelines using a USB battery pack that shuts off when it drops below 5v (the Voltaic brand packs), but being able to use a battery pack directly connected to the RAK would be great.

samschlegel:
> I gather this fixes the issue where the RAK 3401 goes into deep sleep when the battery voltage drops below 3.3v and then never recovers when the battery comes back up above that?  I'm currently following the Austin Mesh guidelines using a USB battery pack that shuts off when it drops below 5v (the Voltaic brand packs), but being able to use a battery pack directly connected to the RAK would be great.

Not sure actually. As I understand it, continuous monitoring hasn't been implemented and what exists right now is an on boot check that sets up a hook for when the battery comes back up that will wake things back up. I'm not sure if that would trigger for the case you're seeing unless the battery is dropping low enough that it reboots. I tested with a power supply and wasn't able to trigger any kind of shutoff until I hit like 2.9v

PR #1991, opened 2026-03-10
Prevent auto-restarting BLE when disabling it on nRF52

### LINKED ISSUE #1933: Bluetooth toogle doesn't work on BLE companion for T114
The toggle on the companion does nothing, the connection from the phone is briefly interrupted but it immediately reconnects, and if the phone is disconnected manually, it's possible to discover and reconnect to the device. 
I'm testing this on the 1.13 main firmware.
Interestingly enough, it's working properly for V4.

weebl2000:
If client is still connected, client would automatically reconnect immediately thus keeping BLE on. When disabling bluetooth, stop advertising first to prevent clients keeping the connection open.

fixes #1933

Try this PR by [building here](https://mcimages.weebl.me?commitId=prevent-auto-restart-ble-nrf52)

Tested & works fine on T1000E too.

### DIFF
--- a/src/helpers/nrf52/SerialBLEInterface.cpp
+++ b/src/helpers/nrf52/SerialBLEInterface.cpp
@@ -246,6 +246,7 @@ void SerialBLEInterface::enable() {
   clearBuffers();
   _last_health_check = millis();
 
+  Bluefruit.Advertising.restartOnDisconnect(true);
   Bluefruit.Advertising.start(0);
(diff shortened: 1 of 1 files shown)


weebl2000:
Confirmed working by 1 user so far.

liamcottle:
Tested on my Wio Tracker L1 Pro, seems to be behaving as expected now.

PR #2001, opened 2026-03-11
round af/rx-/txdelay to 3 decimal places

weebl2000:
Minor nit to fix this:

image

7 decimals precision is required for frequency get/set but not so relevant for the delays/af etc.


weebl2000:
@liamcottle thoughts?

PR #2008, opened 2026-03-12
Fix RAK4631 SX1262 hardware pin config

weebl2000:
image

image

Confirmed against https://github.com/meshtastic/firmware/blob/develop/variants/nrf52840/rak4631/variant.h#L212 too.

### DIFF
--- a/variants/rak4631/variant.h
+++ b/variants/rak4631/variant.h
@@ -147,7 +147,7 @@ extern "C"
 // LoRa radio module pins for RAK4631
 #define  P_LORA_DIO_1 (47)
 #define  P_LORA_NSS (42)
-#define  P_LORA_RESET (-1)
+#define  P_LORA_RESET (38)
 #define  P_LORA_BUSY (46)
 #define  P_LORA_SCLK (43)
 #define  P_LORA_MISO (45)


MGJ520:
I also noticed this problem. I found it when I was adapting the machine. I don't know why it wasn't added, but I don't have permission to merge.

fschrempf:
FYI: I already provided the same fix weeks ago in this PR: https://github.com/meshcore-dev/MeshCore/pull/1565.

robekl:
Looking through the history, this fix seems justified.

The incorrect reset mapping predates the move into `variants/rak4631/variant.h`: before `f4187242` moved the pin definitions, the older `RAK4631Board.h` history already carried the same effective mismatch as `P_LORA_RESET  RADIOLIB_NC   // 38`, and after the move the board still had `P_LORA_RESET (-1)` alongside the rest of the SX1262 pins (`DIO1=47`, `NSS=42`, `BUSY=46`, `SCLK=43`, `MISO=45`, `MOSI=44`, `SX126X_POWER_EN=37`). I went back through the git history looking for a rationale for leaving reset disconnected and did not find one.

That makes this change look like a correction of a long-standing board-definition error rather than a behavior change. The authoritative RAK4631 datasheet documents SX1262 `NRESET -> P1.06`, and in this BSP `P1.06` maps to digital pin `38`, so changing `P_LORA_RESET` to `38` is consistent with the rest of the already-correct SX1262 pin block.

Authoritative source: https://docs.rakwireless.com/product-categories/wisblock/rak4631/datasheet/

weebl2000:
@liamcottle one-liner

PR #2016, opened 2026-03-13
Add sanity build for LR1110 and SX1276 too

weebl2000:
We had all the arches covered, but [here](https://github.com/jbrazio/MeshCore/pull/2) I noticed that LR1110 not building wasn't being caught. Add builds for both LR1110 and SX1276 chips.


weebl2000:
@liamcottle sanity builds for LR1110 and SX1276 chips so we catch any errors early.

PR #2022, opened 2026-03-14
Increase MAX_FRAME_SIZE by 4 bytes

weebl2000:
When a region is set, transport codes add 4 bytes to the wire format. For longer flood messages (~133+ chars for 14 char username), this pushes 1-hop repeat packets past the logRxRaw() limit of MAX_FRAME_SIZE - 3 (169 bytes). The raw packet gets silently dropped, so the companion app never sees repeats and can't show "heard repeats" for those messages.

Shorter messages and messages without a region set are unaffected because they stay under the limit.

Fix: bump MAX_FRAME_SIZE from 172 to 176 (the size of the transport code field). All frame buffers and size checks reference this constant so nothing else needs to change.

Memory cost is ~1 KB worst case across all queue buffers — negligible on all platforms.

This is the easiest fix for now without dropping region scoped messages by 10-16 bytes at worst.

Confirmed this fixes it for companions. First message is 1 byte over the limit with 173 bytes, second message is at the limit with 172 bytes, third message is after flashing with this fix and stays under 176 byte limit.

image

### DIFF
--- a/src/helpers/BaseSerialInterface.h
+++ b/src/helpers/BaseSerialInterface.h
@@ -2,7 +2,7 @@
 
 #include <Arduino.h>
 
-#define MAX_FRAME_SIZE  172
+#define MAX_FRAME_SIZE  176   // +4 for transport codes (region scoping)
 
 class BaseSerialInterface {
 protected:


weebl2000:
@liamcottle thoughts? We talked about this on Discord 2 weeks ago and this seems to fix the issue without causing any problems.

liamcottle:
> @liamcottle thoughts? We talked about this on Discord 2 weeks ago and this seems to fix the issue without causing any problems.

- Would need to test building it on the more constrained boards we support to make sure the firmware still fits.
- Would need to test that BLE still works on Android/iOS/Windows/Web/Mac across ESP32 and nRF52 to make sure there's no weird issues with BLE MTU being increased. Since this value is passed to the BLE stack as well.

I doubt there will be any issues, and it is indeed a very minimal change, but I'm hesitant to merge without it actually being tested properly due to the wide effect it will have. We already have a bunch of BLE issues, don't wanna introduce more :)

weebl2000:
I've tested on Android 16. Works fine there. Can get my hands on an iPhone later to test if needed.

Magalex2x14:
Hello!
Among the supported devices, are there one or two that cause concern, with the most limited resources?

RikoDEV:
I encountered this issue while implementing the scope in ha-meshcore. When sending a longer message with scope, it was not delivered to the RPT. Now I know, that's just MC firmware issue.

weebl2000:
> I've tested on Android 16. Works fine there. Can get my hands on an iPhone later to test if needed.

Tested on iPhone with latest iOS and it works fine there too.

liamcottle:
Tested and working on the following devices:
- Samsung Galaxy S5, Android 5.0, MeshCore App v1.43.0
- Samsung Galaxy S8, Android 9.0, MeshCore App v1.36.0
- Samsung Galaxy S24, Android 15, MeshCore App v1.44.0
- iPod Touch 6th Gen, iOS 12.5.8, MeshCore App v1.44.0
- Apple iPhone X, iOS 16.7.8, MeshCore App v1.44.0
- Apple iPhone 15, iOS 18.7.7 MeshCore App v1.44.0

PR #2035, opened 2026-03-15
Bugfix: auto-adverts don't support multi-byte prefixes

jourdant:
It was highlighted in the Sydney channel of the MeshCore Australia discord community by 👽 wazmac when a node configured with 2 byte hash mode (path.hash.mode 1 = 2 byte), that automatic adverts were still being sent out as single byte.

I checked out the repeater, room_server and sensor code and found the bug in repeater and room_server. This was already fixed for sensor. In the `MyMesh::loop()` function, sendFlood was not passing the byte path preference. I've submitted minor fixes in this PR.

It's also worth noting this is a latent bug in the `BaseChatMesh` class. companion_radio inherits this base class but overrides with the correct prefs injected, but simple_secure_chat does have this issue. In this case simple_secure_chat doesn't have a preference for path.hash.mode. I have used one of the unused bytes in NodePrefs to store the path_hash_mode and fixed the bug in this class too.

### DIFF
--- a/examples/simple_secure_chat/main.cpp
+++ b/examples/simple_secure_chat/main.cpp
@@ -67,7 +67,8 @@ struct NodePrefs {  // persisted to file
   double node_lat, node_lon;
   float freq;
   int8_t tx_power_dbm;
-  uint8_t unused[3];
+  uint8_t path_hash_mode;
+  uint8_t unused[2];
 };
 
 class MyMesh : public BaseChatMesh, ContactVisitor {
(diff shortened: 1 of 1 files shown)


ripplebiz:
Thanks for surfacing this. I have pushed a change to fix this directly, as there were problems with this PR (compilation errors, and wrong base branch -- should be 'dev').

jourdant:
Apologies on wrong base branch. Thanks for addressing so quickly!

PR #2044, opened 2026-03-16
Bugfix: BaseChatMesh has inconsistent use of multi-byte paths

jourdant:
I submitted another PR (#2035) to fix a bug with auto-adverts not being published with multi-byte paths on repeaters and room servers. My original PR setup was bad, so @ripplebiz applied the patch directly to dev branch instead. 

There was some code left behind in the PR that wasn't merged related to classes inheriting `BaseChatMesh` (Companion, Simple Secure Chat etc). The base function `sendFloodScoped` was defaulting to a single byte path. If this function was not overridden by subclasses, then the behaviour would persist unknowingly into those classes. In this case, only Simple Secure Chat was affected, Companion had overridden the behaviour but had the `path_hash_mode + 1` convention mentioned in 5 places.

This PR introduces a new virtual helper function `getPathHashSize()` in the `BaseChatMesh` class so that all subclasses have to properly implement the multi-byte support. I have updated the subclasses to be in-line with this approach too. Hopefully it prevents any future bugs by obscurity and standardises the implementation for future subclasses too.

I'm open to commentary from community/reviewers on whether this `getPathHashSize()` function makes sense to be in all the firmwares, because there are 7 occurences of the `path_hash_mode + 1` function being used elsewhere. At least building out the helper function reduces the risk of not being consistent across the codebase when future changes happen. Thoughts?

### DIFF
--- a/examples/simple_secure_chat/main.cpp
+++ b/examples/simple_secure_chat/main.cpp
@@ -67,12 +67,15 @@ struct NodePrefs {  // persisted to file
   double node_lat, node_lon;
   float freq;
   int8_t tx_power_dbm;
-  uint8_t unused[3];
+  uint8_t path_hash_mode;
+  uint8_t unused[2];
 };
 
 class MyMesh : public BaseChatMesh, ContactVisitor {
   FILESYSTEM* _fs;
   NodePrefs _prefs;
+
+  uint8_t getPathHashSize() const override { return _prefs.path_hash_mode + 1; }
   uint32_t expected_ack_crc;
   ChannelDetails* _public;
   unsigned long last_msg_sent;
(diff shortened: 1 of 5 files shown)


jourdant:
If there is interest in generalising the helper function across the codebase (IE, for classes that don't use the BaseChatMesh class such as repeater/room/sensor), I can add another commit to this PR to standardise it all.

examples/simple_secure_chat/main.cpp
```diff
@@ -411,7 +414,7 @@ class MyMesh : public BaseChatMesh, ContactVisitor {
       int len = strlen((char *) &temp[5]);
       auto pkt = createGroupDatagram(PAYLOAD_TYPE_GRP_TXT, _public->channel, temp, 5 + len);
       if (pkt) {
-        sendFlood(pkt);
+        sendFlood(pkt, 0, getPathHashSize());
```
ripplebiz: Pretty sure this won't compile. Zero is treated the same as NULL (pointer), so makes for ambiguous method signature. (is a weird quirk of C++)

jourdant: @ripplebiz I had successful builds and the PR build check matrix on my local environment before submitting the PR, all the standard builds worked OK. I can update that to be 0u if you would prefer? that way there's no ambiguity with a nullpointer etc.

jourdant:
@ripplebiz @liamcottle is there anything else I can do to help with this one?

PR #2056, opened 2026-03-17
Fix Heltec V3 LoRa reset pin

robekl:
## Summary
- set P_LORA_RESET to 12 for the Heltec V3 variant
- align the SX1262 reset wiring with Heltec's published V3 board mapping
- leave the rest of the Heltec V3 configuration unchanged

## Why this is the right fix
Heltec's WiFi LoRa 32 V3 pin map and datasheet both document the SX1262 reset line on GPIO12 as LoRa_RST. MeshCore already passes P_LORA_RESET into the Heltec V3 radio Module constructor, so defining that pin as disconnected was inconsistent with the published hardware wiring.

Setting P_LORA_RESET to 12 matches the vendor mapping already used by the closely related Heltec V4 variant and corrects the V3 board definition without changing any other board behavior.

Authoritative sources:
- Heltec WiFi LoRa 32 V3 pin map: https://resource.heltec.cn/download/WiFi_LoRa_32_V3/Wi-Fi_LoRa32_V3.2_Pinmap.png
- Heltec WiFi LoRa 32 V3 datasheet: https://resource.heltec.cn/download/WiFi_LoRa_32_V3/HTIT-WB32LA_V3%28Rev1.1%29.pdf

## Validation
- built Heltec_v3_repeater successfully after the pin correction
- reviewed the repo for other Heltec V3 files that encode the reset pin and did not find additional board-specific references that needed updating

## Testing Firmware


robekl:
Looking through the history, this also seems justified.

The incorrect value was introduced in `c4a2b139` ("moved HeltecV3Board.h to variant folder"), where `variants/heltec_v3/platformio.ini` gained an explicit `P_LORA_RESET=RADIOLIB_NC` while the rest of the SX1262 pin block was already the documented Heltec layout (`DIO1=14`, `NSS=8`, `BUSY=13`, `SCK=9`, `MISO=11`, `MOSI=10`). I checked the subsequent history and did not find any commit or PR discussion explaining why V3 should intentionally ignore a wired reset line.

That makes this PR look like a straightforward board-definition correction. Heltec's published WiFi LoRa 32 V3 pin map and datasheet show `LoRa_RST` on `GPIO12`, and the target code already passes `P_LORA_RESET` into the SX1262 `Module` constructor, so using `12` here is consistent with both the vendor docs and the existing driver wiring.

Authoritative sources:
- https://resource.heltec.cn/download/WiFi_LoRa_32_V3/Wi-Fi_LoRa32_V3.2_Pinmap.png
- https://resource.heltec.cn/download/WiFi_LoRa_32_V3/HTIT-WB32LA_V3%28Rev1.1%29.pdf

PR #2057, opened 2026-03-17
Fix Heltec Tracker LoRa reset pin

robekl:
## Summary
- set P_LORA_RESET to 12 for the Heltec Wireless Tracker variant
- align the SX1262 reset wiring with Heltec's published tracker pin map
- leave the rest of the tracker board definition unchanged

## Why this is the right fix
Heltec's Wireless Tracker documentation shows the SX1262 reset line on GPIO12 as LoRa_RST. MeshCore already passes P_LORA_RESET into the Heltec Tracker radio Module constructor, so leaving this variant at RADIOLIB_NC treated a documented hardware reset line as disconnected.

Setting P_LORA_RESET to 12 brings the tracker variant in line with the published board wiring instead of introducing any new board behavior.

Authoritative source:
- Heltec Wireless Tracker datasheet: https://resource.heltec.cn/download/Wireless_Tracker/Wireless_Tacker1.1/HTIT-Tracker_V0.5.pdf

## Validation
- built Heltec_Wireless_Tracker_repeater successfully after the pin correction
- reviewed the tracker variant for related board-specific pin references and did not find any additional LoRa reset definitions to update

## Testing Firmware


robekl:
Looking through the history, this fix also seems justified.

The mismatch entered the tree in `5b7d7386` ("fix building issues with heltec wireless paper and heltec tracker"), which set `P_LORA_RESET=RADIOLIB_NC` in `variants/heltec_tracker/platformio.ini`. I reviewed the later history and did not find any explanation for why the tracker variant should leave SX1262 reset disconnected even though the target code passes `P_LORA_RESET` straight into the radio `Module` constructor.

So this looks like the same class of board-definition bug as the Heltec V3 issue rather than a deliberate behavior choice. Heltec's tracker documentation shows `LoRa_RST` on `GPIO12`, so changing the variant to use `12` is consistent with the vendor pin map and the way the radio is already instantiated in the code.

Authoritative source:
- https://resource.heltec.cn/download/Wireless_Tracker/Wireless_Tacker1.1/HTIT-Tracker_V0.5.pdf

PR #2058, opened 2026-03-17
Fix Heltec Tracker GPS pin mapping

robekl:
## Summary
- map Heltec Tracker GPS reset to GPIO35
- expose the documented PPS line on GPIO36
- stop advertising a separate GPS enable pin that is not documented for this board

## Why this is the right fix
Heltec's Wireless Tracker documentation identifies GPIO35 as GNSS_RST and GPIO36 as PPS. The current variant defined GPIO35 as PIN_GPS_EN and GPIO36 as PIN_GPS_RESET, which inverted the documented control lines and omitted the PPS signal entirely.

This change moves PIN_GPS_RESET to 35, adds PIN_GPS_PPS on 36, and removes the undocumented dedicated GPS enable pin from the variant. That keeps the board definition aligned with the vendor pin map and matches how this target already shares peripheral power through PIN_VEXT_EN.

Authoritative source:
- Heltec Wireless Tracker datasheet: https://resource.heltec.cn/download/Wireless_Tracker/Wireless_Tacker1.1/HTIT-Tracker_V0.5.pdf

## Validation
- built Heltec_Wireless_Tracker_repeater successfully after the pin-map correction
- reviewed the target code that constructs the MicroNMEALocationProvider so the updated mapping and comments stay consistent

## Testing Firmware


robekl:
Looking through the history, this one also appears justified.

The incorrect GPS control mapping was introduced in `59d9770a` ("Add GPS support Heltec Wireless Tracker v1.x"). That commit set `PIN_GPS_EN=35`, `PIN_GPS_RESET=36`, and added the matching comment in `variants/heltec_tracker/target.cpp` describing GPIO35 as enable and GPIO36 as reset. I went back through the later history and did not find any follow-up discussion or correction that would justify treating those pins that way.

What makes this a strong fix is that the same commit appears to have simply misread the Heltec board docs: the official tracker pinout labels `GPIO35` as `GNSS_RST` and `GPIO36` as `PPS`. So this PR is not changing established intended behavior so much as correcting the original board-definition interpretation and exposing the documented PPS line that was previously omitted.

Authoritative source:
- https://resource.heltec.cn/download/Wireless_Tracker/Wireless_Tacker1.1/HTIT-Tracker_V0.5.pdf

PR #2065, opened 2026-03-17
Fix T1000-E sensor power pin mapping

robekl:
## Summary
- stop treating T1000-E `P0.04` as a sensor power-enable output
- remove the extra `SENSOR_EN` rail toggle from the T1000-E sensor path
- keep sensor power control on the documented `PIN_3V3_EN` / `P1.06` rail

## Issue
The current T1000-E variant defines two different sensor-power control signals:
- `PIN_3V3_EN = 38` (`P1.06`)
- `SENSOR_EN = 4` (`P0.04`)

The firmware actively drives both of them:
- `initVariant()` configures both as outputs and drives them low
- `t1000e_get_temperature()` and `t1000e_get_light()` drive both high before sampling
- `powerOff()` drives both low again

That is hard to reconcile with Seeed's newer official T1000-E pin table, which documents:
- `P1.06` as `Sensor VCC Enable`
- `P0.04` as `VCC voltage detect`

If that mapping is correct, the current variant is driving a voltage-detect input as if it were a rail-enable output.

## Authoritative source
Seeed Studio's newer official board page:
- https://wiki.seeedstudio.com/get_started_with_lorawan_tracker/

Relevant entries from the pin table:
- `P0.04` = `VCC voltage detect`
- `P1.06` = `Sensor VCC Enable`
- `P0.05` = `Charger insert detect`
- `P1.03` = `Charger State`

The older page at https://wiki.seeedstudio.com/t1000_e_intro/ is less complete and conflicts with this newer table.

## Fix
This change removes the stale `SENSOR_EN` definition and stops driving `P0.04` anywhere in the T1000-E implementation.

After this change:
- the board no longer configures `P0.04` as an output
- temp/light reads only toggle `PIN_3V3_EN`
- power-off no longer tries to drive an extra sensor-enable pin

## Why this fixes it
This keeps sensor rail control on the pin Seeed currently documents for that purpose (`P1.06`) and removes the most likely incorrect electrical behavior in the current tree: treating `P0.04` as an output control line.

It also avoids changing unrelated GPS, charging, or battery logic. The fix is intentionally narrow: remove the extra power-enable path that conflicts with the newer hardware documentation.

## Validation
- `pio run -e t1000e_companion_radio_ble`

### DIFF
--- a/variants/t1000-e/T1000eBoard.h
+++ b/variants/t1000-e/T1000eBoard.h
@@ -69,10 +69,6 @@ public:
     #ifdef PIN_3V3_ACC_EN
         digitalWrite(PIN_3V3_ACC_EN, LOW);
     #endif
-    #ifdef SENSOR_EN
-        digitalWrite(SENSOR_EN, LOW);
-    #endif
-
     // set led on and wait for button release before poweroff
     #ifdef LED_PIN
     digitalWrite(LED_PIN, HIGH);
--- a/variants/t1000-e/t1000e_sensors.cpp
+++ b/variants/t1000-e/t1000e_sensors.cpp
@@ -72,7 +72,6 @@ float t1000e_get_temperature(void) {
   unsigned int ntc_v, vcc_v;
 
   digitalWrite(PIN_3V3_EN, HIGH);
-  digitalWrite(SENSOR_EN, HIGH);
   analogReference(AR_INTERNAL_3_0);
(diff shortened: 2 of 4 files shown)


IoTThinks:
I am unable to find the schematic for T1000e.
P0.04 = VCC voltage detect

May I know what is P0.04 for?

PR #2073, opened 2026-03-17
feat: VLAN-style region tagging for untagged flood packets

### LINKED ISSUE #2041: Feature Request: Auto-assign default scope to messages without scope
**Problem**

In the Netherlands, the MeshCore network experiences congestion due to:
- Excessive “test” and “Good morning” messages
- Floods caused by short-interval adverts
- Messages sent **without a scope**, which are forwarded across the entire mesh

This causes repeaters to be overloaded and users to resend messages multiple times.

Current solutions (manual scopes, regions) help, but depend on users knowing how to configure scopes. Forcing denyF *: would reduce traffic, **but new users would not be able to send messages**.

**Proposed Solution**

Add a **repeater-level configuration** to automatically assign a **default scope** to messages received without a scope.

**Behavior:**

1. Only applies to public messages without a scope.
2. The repeater rewrites the message’s scope to a pre-configured default, e.g., the local city or province.
3. This prevents flooding the entire mesh while still allowing new users to send messages.
4. Repeaters on older firmware will forward the rewritten message from updated repeaters, reducing overall traffic.

**Example:**

Repeater located in Barendrecht (South Holland) has regions:
   eu            (Europe)
   nl             (Netherlands)
   nl-zh        (South Holland)
   nl-zh-brr (Barendrecht)

- A user sends a public message without a scope.
- The repeater automatically rewrites it to nl-zh.
- Message stays within the province instead of flooding the Netherlands.

**Benefits**

- Reduces network-wide flooding
- Ensures new users can still send messages
- Simplifies traffic management for repeater admins
- Compatible with repeaters not running the latest firmware

**Optional Implementation Detail**

- Config option could be called: route.unknown.scope or default_scope
- Admin sets it per repeater
- Messages are rewritten only if no scope exists

dale-ruane:
## Summary

Closes #2041

- Repeaters can now opt in to automatically tagging untagged flood packets (`ROUTE_TYPE_FLOOD`) with their home region's transport code, converting them to `ROUTE_TYPE_TRANSPORT_FLOOD`
- Auto-tagging is **off by default** and must be explicitly enabled via `set region.autotag on` — having a home region set alone does not activate it
- A configurable hop limit (`set region.autotag.max.hops <0-8>`, default `1`) prevents mis-tagging packets that have already traversed multiple repeaters, since the further a packet has travelled the less likely the tagging repeater's home region is correct
- Adds `TRANSPORT_CODE_ALL` (`0xFFFF`) as a reserved code that bypasses region filtering, allowing mesh-wide flooding when explicitly requested
- Fully backwards compatible: repeaters with auto-tagging off (the default) behave exactly as before

## New CLI commands

| Command | Description |
|---|---|
| `get/set region.autotag on\|off` | Enable/disable auto-tagging of untagged flood packets (default: `off`) |
| `get/set region.autotag.max.hops <0-8>` | Only auto-tag packets with a path hash count ≤ this value (default: `1`). `0` = zero-hop (direct) only |

## How it works

- **Untagged flood arrives** + auto-tagging on + home region set + hop count within limit → tagged with home region's transport code and forwarded
- **Untagged flood arrives** + auto-tagging off or no home region → forwarded as-is (existing behaviour)
- **Untagged flood arrives** + hop count exceeds `region.autotag.max.hops` → forwarded as-is, not tagged
- **Tagged flood with ALL (0xFFFF)** → always forwarded regardless of region config
- **Tagged flood with region code** → existing region matching (unchanged)

## Test plan

- [ ] Repeater with `region.autotag off` (default) forwards untagged floods as before, even with a home region set
- [ ] Repeater with `region.autotag on` and home region set tags untagged floods with correct transport code
- [ ] Packets with `getPathHashCount()` > `region.autotag.max.hops` are forwarded without tagging
- [ ] Already-tagged packets pass through unchanged
- [ ] `TRANSPORT_CODE_ALL` packets are forwarded by all repeaters
- [ ] `set region.autotag.max.hops` rejects values outside 0-8
- [ ] Builds clean on ESP32 and NRF52 platforms


awhite2000:
@weebl2000 @ViezeVingertjes could someone please review this? it's a very important improvement to the way scoping works, which allows for a gentler and more flexible adoption of regions.

ViezeVingertjes (commented):
Looks good, and imho it does make sense to have this. 

Just a sanity check; Did you check if any documentation requires an update for this change?

awhite2000:
It definitely needs good docs, it's a new feature. IMHO docs on regions need to be beefed up anyway, explaining the concepts and not just the config syntax.
If @dale-ruane can provide it in this PR then that's great, if not then I think, given the backlog and how long this has been sitting, this PR should be merged and a new PR for improving docs on regions should be created, which I can take if no one else does.

AI7NC:
I think this is the best way to bootstrap region use adoption.

skyepn:
It looks like this also solves PR1810 "Allow direct message paths when denyf * is set"
https://github.com/meshcore-dev/MeshCore/pull/1810

Because `recv_pkt_region` is being set in `filterRecvFloodPacket`? This would be a very good thing, PR1810 is currently the top upvoted PR and many people in large meshes are using custom repeater builds with that PR which is not an ideal solution because it breaks the semantics of "deny flood packets with unset scope"

skyepn:
Should there be a new repeater CLI setting boolean to control this behaviour? Even though not setting a home region is a defacto way of turning it off, it should probably be explicit.

skyepn:
Just thinking about this a little more, especially in the case where a "boundary" repeater can hear nodes in other regions it is essential that the choice of home region is correct so that unscoped flood packets are not tagged with an incorrect region.  

For this reason I think there should be a new repeater setting for this feature and the default being off, as the choice of home region requires the repeater admin to think carefully about the intended behaviour.

To that end, it will be important for the `region home` command documentation to describe the process for choosing very clearly, which I believe should be **"the most specific region in the local region hierarchy that includes ALL nodes that the repeater can possibly hear"**

Local example using the [Pacific Northwest MeshCore Region Strategy](https://gessaman.com/meshcore/regions/) and a repeater on Mt Baker, Whatcom County, WA USA.  The physical location of the repeater would be `bli` but it would also be able to hear other counties, so you move up the tree to `w-wa`. But it can also hear packets from across the border in BC, Canada so the correct home region would be `pnw` as that is the most specific level where it is not possible to hear unscoped packets from any neighboring regions at that level (specifically `ca` ie California)

Writing this now, it seems like it might actually be better to only apply this to a configurable number of hops to avoid tagging flood packets that originated in distant regions and were forwarded by repeaters that did not have unscoped tagging turned on (or older firmwares that wouldn't support it).  I see that `_prefs.flood_max` is checked at the top of `allowPacketForward` but this new tagging behaviour should probably have it's own setting, and actually maybe only apply to zero hop flood packets received? Haven't thought this part through completely, but NOT restricting to direct/zero hop flood packets would almost guarantee that some flood packets received would be tagged incorrectly as there would be no way to choose the home region properly.

In other words, the repeater admin must know what geographical area they will receive flood packets from in order to choose the correct home region.

AI7NC:
This definitely needs some thought, repeater operators opting in seems like a good idea outside just setting the home region. I myself would be against tagging traffic for a repeater between two regions as the pnw region. That's way to wide of a scope to auto-tag traffic, it may as well be * at that point.

I do think there should be a auto-tag max hop setting where a repeater operator can tone down the auto tagging depending on where the repeater is located and what it hears. If scoping at a sub-region or state level, going up to 3 or 4 hops may make sense, while other areas even a zero hop will have some mis-tag issues. Allowing the admin to tune for best affect I think is most practical. 

Despite the above controls, some situations will not be fully be able to control 100% accuracy of the auto tagging. At some point it'll need to boil down to If originators want their traffic tagged correctly, they need to configure the region scope for their traffic at origination.

skyepn:
Yeah, the more I think about it, a repeater admin can't be expected to know the geographical reach even 1 or 2 hops out, it's hard enough to keep track of the location your own (direct, zero hop) contacts over time.

I think it auto-scoping should only be applied if the new setting is on and `getPathHashCount() == 0`?

awhite2000:
I agree that this limit is useful but let's make it configurable with the default 1 or 2. I know that in my area's topology no harm would be done even with 2-3.

EDIT: and we have to document it really well :)

dale-ruane:
Great suggestions all - I've added the opt-in and hop-limit. Docs coming in the next commit. Would appreciate feedback on the Docs and commit above

docs/cli_commands.md
```diff
@@ -712,6 +712,36 @@ This document provides an overview of CLI commands that can be sent to MeshCore
...
+**Usage:**
+- `get region.autotag`
+- `set region.autotag <state>`
+
+**Parameters:**
+- `state`: `on` (enable) or `off` (disable)
+
+**Default:** `off`
+
+**Note:** When enabled, the repeater stamps its home region's transport code onto untagged flood packets (`ROUTE_TYPE_FLOOD`) it receives, converting them to `ROUTE_TYPE_TRANSPORT_FLOOD` before re-broadcast. This provides VLAN-style scoping of legacy / un-scoped traffic, but requires a home region to be configured (see `region home`). Because mis-tagging is possible when the repeater can hear traffic originating outside its home region, this feature is opt-in. See also `region.autotag.max.hops` to limit how far a packet may have travelled before becoming eligible for auto-tagging. The reserved transport code `0xFFFF` (TRANSPORT_CODE_ALL) is always forwarded regardless of local region configuration, allowing explicit mesh-wide flooding when a sender requests it.
```
awhite2000: I don't think referring to VLAN is useful because there are both similarities and important differences. This was useful perhaps during conceptual development but our docs should be self-sufficient without requiring network engineering degree ;)

docs/cli_commands.md
```diff
@@ -712,6 +712,36 @@ This document provides an overview of CLI commands that can be sent to MeshCore
...
+
+---
+
+#### View or change the max hop count for auto-tagging
+**Usage:**
+- `get region.autotag.max.hops`
+- `set region.autotag.max.hops <value>`
+
+**Parameters:**
+- `value`: Maximum path hash count (0-8). `0` means only auto-tag packets received directly (zero-hop); higher values also auto-tag packets that already traversed that many repeaters.
```
awhite2000: "auto-tag packets" maybe add "without scope" for clarity?

docs/cli_commands.md
```diff
@@ -712,6 +712,36 @@ This document provides an overview of CLI commands that can be sent to MeshCore
 **Parameters:**
 - `name`: Region name
 
+**Note:** The home region is used by the auto-tag feature (see `region.autotag` below) to stamp a transport code onto untagged flood packets this repeater receives, converting them from `ROUTE_TYPE_FLOOD` to `ROUTE_TYPE_TRANSPORT_FLOOD`. When selecting a home region, choose the **most specific region in the local region hierarchy that includes all nodes this repeater can possibly hear** (including via hops you intend to cover — see `region.autotag.max.hops`). Choosing a home region that is too narrow will cause unscoped packets originating from neighbouring regions to be tagged incorrectly; choosing one that is too broad defeats the purpose of scoping.
```
awhite2000, line 731: We should also explain what happens when the home region is not set.

src/Packet.h
```diff
@@ -16,6 +16,8 @@ namespace mesh {
 #define ROUTE_TYPE_DIRECT            0x02    // direct route, 'path' is supplied
 #define ROUTE_TYPE_TRANSPORT_DIRECT  0x03    // direct route + transport codes
 
+#define TRANSPORT_CODE_ALL           0xFFFF  // special transport code: forward to all regions
```
awhite2000, line 19: Client apps will need to support sending packets with this code, which is different from sending without a scope. @liamcottle

dale-ruane: Yeah I had hoped that maybe this could be implemented in the client apps. Perhaps a one-time toggle for a message which then reverts? Just to gently encourage people not to just scope every message to all

src/helpers/CommonCLI.cpp
```diff
@@ -117,6 +119,8 @@ void CommonCLI::loadPrefsInt(FILESYSTEM* fs, const char* filename) {
 
     // sanitise settings
     _prefs->rx_boosted_gain = constrain(_prefs->rx_boosted_gain, 0, 1); // boolean
+    _prefs->region_autotag = constrain(_prefs->region_autotag, 0, 1);   // boolean
+    _prefs->region_autotag_max_hops = constrain(_prefs->region_autotag_max_hops, 0, 8);
```
awhite2000: This limit should be documented.

src/helpers/CommonCLI.cpp
```diff
@@ -597,6 +607,19 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, const char* command, ch
         *dp = 0;
         savePrefs();
         strcpy(reply, "OK");
+      } else if (memcmp(config, "region.autotag.max.hops ", 24) == 0) {
+        int h = atoi(&config[24]);
+        if (h >= 0 && h <= 8) {
```
awhite2000: See above, docs.
(In general I dislike magic numbers sprinkled in multiple places, perhaps a #define would be better).

docs/cli_commands.md
```diff
@@ -738,7 +738,9 @@ This document provides an overview of CLI commands that can be sent to MeshCore
 - `set region.autotag.max.hops <value>`
 
 **Parameters:**
-- `value`: Maximum path hash count (0-8). `0` means only auto-tag packets without scope received directly (zero-hop); higher values also auto-tag packets without scope that already traversed that many repeaters.
+- `value`: Maximum path hash count. `0` means only auto-tag packets without scope received directly (zero-hop); higher values also auto-tag packets without scope that already traversed that many repeaters.
+
+**Range:** `0` to `8` (inclusive). Values outside this range are rejected by `set` and clamped to this range on load. The upper bound of `8` is intentionally well below the default `flood.max` of `64`, because auto-tagging packets from far across the mesh almost always produces incorrect region assignments — the limit exists to keep admins honest about the geographic scope they can actually account for.
```
awhite2000, line 759: +1, nice.

ripplebiz:
OK, the TRANSPORT_CODE_ALL doesn't make much sense, to me. It's yet another rule change that would take time to percolate around with firmware version updates. And there already a rule that covers it (well, repeaters that allow un-scoped). Also, strikes me as something that would be abused, ie. if repeaters will all allow 0xFFFF ?

Also, auto-assigning the home region for un-scoped, seems more like a work-around for the current problem of most clients sending un-scoped floods (eg. for DM's, logins, requests, etc). But, this should soon be solved with the 'default scope' feature. (currently in the default-scope branch)

AI7NC:
I think this feature is needed. It'll be much easier to get the large-coverage repeater operators to region tag untagged traffic than to get all the users to. Once the majority is tagged then we can move to start blocking * traffic more quickly to better control our airtimes. Without this feature we're at the mercy of every client in the PNW RF network to decide to start tagging their traffic.

There needs to be an incentive to do so and there's no way of doing that right now than blocking * which is much more impactful than tagging their * traffic as the local region.  I'm all for this feature.

awhite2000:
> But, this should soon be solved with the 'default scope' feature. (currently in the default-scope branch)

@ripplebiz Please explain how is that different from the auto-tagging that this PR implements? I briefly looked at your branch (which doesn't have a corresponding PR nor any writeup so it's hard to tell what's the end goal) and it seems less robust than this implementation.

src/Packet.h
AI7NC: I'm against '0xFFFF'; that can way to easily be abused to DDOS the whole network. I'd say a large region can decide on their own whether to agree on a region wide flood code for emergencies or not.

AI7NC:
@dale-ruane Can you edit your top comment to include the added hop control? Some folks are reading the top comment proposal and are missing that you're including a hop max include setting to keep repeaters from tagging traffic from outside their region based on hops.

docs/cli_commands.md
LitBomb: 8 hops max is very arbitrary.  why cap it at 8 hops when it is already configurable by repeater owners?  Let the repeater owners decide what max hop fits their needs.  there is no one number that works for all meshes.

It is OK to have some recommendations for repeater owners of what values to use based on some other mesh variables.

LitBomb:
8 hops max is very arbitrary. why cap it at 8 hops when it is already configurable by repeater owners? Let the repeater owners decide what max hop fits their needs. there is no one number that works for all meshes.

It is OK to have some recommendations for repeater owners of what values to use based on some other mesh variables.

skyepn:
> Also, auto-assigning the home region for un-scoped, seems more like a work-around for the current problem of most clients sending un-scoped floods (eg. for DM's, logins, requests, etc). But, this should soon be solved with the 'default scope' feature. (currently in the default-scope branch)

Keep in mind Liam’s app is not the only app, all the apps and bots will have to be updated, plus all the repeaters, plus there’s no guarantee any new app or new bot will implement it, so unscoped flood will still be around for a very long time., probably forever. 

Flood traffic control is a big problem now and this would allow key backbone repeater admins to add flood control immediately. I don’t really see any downside other than the possibility of applying the wrong scope to a few packets which we’ve discussed above. 

This is a needed workaround imho, repeater owners should be able to tag the traffic going through their own repeaters.

dale-ruane:
> @dale-ruane Can you edit your top comment to include the added hop control? Some folks are reading the top comment proposal and are missing that you're including a hop max include setting to keep repeaters from tagging traffic from outside their region based on hops.

All done.

Also resolved conflicts caused by new upstream release :)

PR #2074, opened 2026-03-17
fix(gps and sensor): fix extended uptime failures in GPS and sensor code

NickDunklee:
Updated to cover all GPS code paths that have the same variable math problems.

This covers fixing GPS looping bugs and also a divide by zero potential crash in the t1000-e. Was originally an attempt to help fix t1000-e uptime, and figured might as well fix the math in all the rest in the same go. If preferred, I can break out the t1000-e sensor fix into its own PR so this is just GPS stuff.

- Fix millis() wraparound (49-day) in GPS sensor loop() and MicroNMEALocationProvider::loop() by switching from absolute comparison (millis() > next_*) to elapsed-time math ((uint32_t)(millis() - last_*) >= interval) with uint32_t timestamps; previous pattern could stall GPS updates and time sync for up to ~24 days after 49-day-wraparound
- Fix divide-by-zero crash in get_heater_temperature() when ntc_volt == 0 (open circuit or bad ADC read); now returns 0.0f safely
- Fix out-of-bounds array access in get_heater_temperature(): loop exit at i == 0 caused ntc_res2[i-1] / ntc_temp2[i-1] at index -1; loop exhaustion at i == 136 caused ntc_res2[136] one-past-end read; clamped i to [1, 135] with early-exit returns for edge temperatures so it doesn't run out of bounds


NickDunklee:
Just note, haven't had a t1000e failure since this change.

NickDunklee:
#2937 fixes the GPS bug as well, so closing. Huzzah! 🎉  I'll check the other code vs 1.17.0 and open another PR if needed.

PR #2077, opened 2026-03-17
fix: align bridge.source CLI names

robekl:
## Summary
- make bridge.source accept the documented logRx and logTx values exactly
- keep rx and tx as legacy aliases for compatibility
- update the CLI docs to reflect the accepted values and aliases

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -901,6 +901,8 @@ region save
   - `logRx`: bridges received packets
   - `logTx`: bridges transmitted packets
 
+**Compatibility:** `rx` and `tx` are also accepted as legacy aliases for `logRx` and `logTx`
+
 **Default:** `logTx`
 
 ---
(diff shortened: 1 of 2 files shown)


jbrazio:
This that type of thing that only adds more code without any added value.
Just update the docs to use `rx` and `tx`.

liamcottle:
Howdy, please just update the docs as suggested by @jbrazio. I don't see any gain to adding a new command that does exactly the same thing, while still also accepting the old command...

robekl:
I'm used to the strategy of "when making backwards-incompatible interface changes, deprecate but continue to support the old functionality for at least one release cycle so clients can update to the regular/supported interface without loss of functionality". Anyway, [here](https://github.com/meshcore-dev/MeshCore/pull/2100) is a new PR to support only `rx`/`tx` as requested and docs updated.

@jbrazio @liamcottle

jbrazio:
I understand your point.. but this is such a niche feature that in the end was just wrong documented. Actually people need to understand what this does before messing with it.

PR #2089, opened 2026-03-19
fix(nrf52): add BLE security timeout and advertising watchdog

ipduffy:
Commit https://github.com/meshcore-dev/MeshCore/commit/25ea953cc3d744fde55b0cd78ffe222b4af5715f introduced a condition where a BLE connection that is established but not yet encrypted could result in a hung BLE stack under certain conditions.

If `onConnect` is called with a valid connection_handle, the `_isDeviceConnected` flag remains set to false. This prevents the watchdog loop on lines 336-346 from running until after the `onSecured` function is called and the `_isDeviceConnected` flag is set to true. 

For various reasons that are apparently more common in iPhones versus Android devices, the establishment of encryption is more likely to not complete, which would leave the nRF device BLE stack in a hung state.

To address this potential for a hung state I've implemented a timeout (BLE_SECURITY_TIMEOUT_MS) on the call to `onSecured` of 15 seconds, so that if a BLE connection is established via `onConnect` without the subsequent call to `onSecured` within that time, then `disconnect` is called, allowing the device to return to an advertising state and allow further connection attempts.

I have tested this patched firmware on a RAK 4631 board and it's been stable without any BLE issues for the past 24 hours.

### DIFF
--- a/src/helpers/nrf52/SerialBLEInterface.cpp
+++ b/src/helpers/nrf52/SerialBLEInterface.cpp
@@ -342,7 +342,18 @@ size_t SerialBLEInterface::checkRecvFrame(uint8_t dest[]) {
   if (_isEnabled && _conn_handle != BLE_CONN_HANDLE_INVALID && !_isDeviceConnected) {
     if (_conn_pending_since > 0 && (now - _conn_pending_since) >= BLE_SECURITY_TIMEOUT_MS) {
       BLE_DEBUG_PRINTLN("SerialBLEInterface: security timeout, forcing disconnect");
+      // Reset _conn_pending_since so we don't call disconnect more than once in a half-connected state
+      _conn_pending_since = 0;
+
+      // Disconnect the current connection
       disconnect();
+
+      // Force an advertising health check
+      _last_health_check = now;
+      if (!isAdvertising()) {
+        BLE_DEBUG_PRINTLN("SerialBLEInterface: advertising watchdog - advertising stopped, restarting");
+        Bluefruit.Advertising.start(0);
+      }
     }
   }
 
--- a/src/helpers/nrf52/SerialBLEInterface.cpp
+++ b/src/helpers/nrf52/SerialBLEInterface.cpp
@@ -6,6 +6,7 @@
 
 // Magic numbers came from actual testing
 #define BLE_HEALTH_CHECK_INTERVAL  10000  // Advertising watchdog check every 10 seconds
+#define BLE_SECURITY_TIMEOUT_MS    15000  // Max time to wait for pairing/security to complete
 #define BLE_RETRY_THROTTLE_MS      250    // Throttle retries to 250ms when queue buildup detected
 
 // Connection parameters (units: interval=1.25ms, timeout=10ms)
@@ -29,6 +30,7 @@ void SerialBLEInterface::onConnect(uint16_t connection_handle) {
   if (instance) {
     instance->_conn_handle = connection_handle;
     instance->_isDeviceConnected = false;
+    instance->_conn_pending_since = millis();
     instance->clearBuffers();
   }
 }
@@ -39,6 +41,7 @@ void SerialBLEInterface::onDisconnect(uint16_t connection_handle, uint8_t reason
     if (instance->_conn_handle == connection_handle) {
       instance->_conn_handle = BLE_CONN_HANDLE_INVALID;
       instance->_isDeviceConnected = false;
+      instance->_conn_pending_since = 0;
       instance->clearBuffers();
     }
   }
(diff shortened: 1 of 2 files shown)


src/helpers/nrf52/SerialBLEInterface.cpp
```diff
@@ -331,13 +335,23 @@ size_t SerialBLEInterface::checkRecvFrame(uint8_t dest[]) {
     return len;
   }
   
+  unsigned long now = millis();
+
+  // Security timeout: if a connection has been pending pairing for too long, force-disconnect
+  // and let the watchdog below restart advertising. This handles the iOS pairing stall case.
+  if (_isEnabled && _conn_handle != BLE_CONN_HANDLE_INVALID && !_isDeviceConnected) {
+    if (_conn_pending_since > 0 && (now - _conn_pending_since) >= BLE_SECURITY_TIMEOUT_MS) {
+      BLE_DEBUG_PRINTLN("SerialBLEInterface: security timeout, forcing disconnect");
```
robekl, line 344: I think this needs one more state guard around the timeout path.

Right now `_conn_pending_since` is only cleared in `onDisconnect()` / `onSecured()`. Once the timeout threshold is crossed, every subsequent `checkRecvFrame()` poll can call `disconnect()` again until the disconnect callback actually arrives. Since `checkRecvFrame()` is polled continuously, that means `sd_ble_gap_disconnect()` can be hit repeatedly for the same timed-out connection.

I’d suggest clearing `_conn_pending_since` before calling `disconnect()`, or adding a separate “disconnect requested” latch so the timeout path only fires once per connection attempt.

src/helpers/nrf52/SerialBLEInterface.cpp
```diff
@@ -331,13 +335,23 @@ size_t SerialBLEInterface::checkRecvFrame(uint8_t dest[]) {
     return len;
   }
   
+  unsigned long now = millis();
+
+  // Security timeout: if a connection has been pending pairing for too long, force-disconnect
+  // and let the watchdog below restart advertising. This handles the iOS pairing stall case.
```
robekl, line 341: One more recovery-path concern: this improves the half-connected stall, but the actual worst-case recovery still looks longer than the new 15s timeout.

If pairing stalls, the code waits 15s, then requests `disconnect()`. If advertising does not resume cleanly after that, the separate advertising watchdog still only runs on its normal 10s cadence. So the real recovery window can still be closer to ~25s worst case.

Would it make sense to force an immediate advertising recovery check after the timeout-triggered disconnect, or at least reset `_last_health_check` so the watchdog runs immediately once the handle drops?

src/helpers/nrf52/SerialBLEInterface.cpp
ipduffy: I think that setting `_conn_pending_since = 0` in the code is the cleanest and easiest way to address this rather than add another state flag to the struct. I will read your second comment before I make the change and push it up.

src/helpers/nrf52/SerialBLEInterface.cpp
ipduffy: I can also reset the _last_health_check in the half-connect stall scenario, which should force the code to enter the checks on lines 351-360. This makes sense and I'll submit the updated changes to the PR.

ipduffy: Rather than expiring the _last_health_check value, I just explicitly check and call the health check code. This I think is easier to read and understand and doesn't rely on code blocks that could be changed in the future.

src/helpers/nrf52/SerialBLEInterface.cpp
ipduffy: Reset _conn_pending_since on line 346

src/helpers/nrf52/SerialBLEInterface.cpp
ipduffy: Perform explicit advertising health check on lines 352-356

ipduffy:
Flashing this firmware to my test board now and I'll report back if there are any issues.

ipduffy:
BLE connection has been stable on my device for the last 3 days.

ipduffy:
Fixes #2213

ipduffy:
My companion node has been stable BLE-wise for three weeks. No issues.

dotspencer:
Any chance this could be reopened and considered for merging?

I've seen this failure on two different RAK 4631 companion nodes running v1.16.0 when connecting with iOS app. Same symptoms as #2213. Looks like the diff still applies cleanly to current HEAD.

@ipduffy did you hit any issues in testing, or was there some other reason this was closed?

ipduffy:
I didn't run into any issues and in fact I'm still running my modified firmware on my wisblock today. It's working great. I closed the PR because of the recent blog post about vibe coding and how the zeitgeist seemed to be anti-AI (https://blog.meshcore.io/2026/04/23/the-split). I used Claude code to find this bug and also to implement some of the fixes. I am a C/C++ developer with some embedded programming experience, mostly on ESP32 boards, but I fully understood the bug and the fix that Claude proposed and implemented. But if the team is against any AI related code then I thought it best to close the PR and let this issue be fixed more organically.

PR #2101, opened 2026-03-20
fix: restrict bridge baud to documented rates

robekl:
## Summary
- restrict `bridge.baud` to the documented RS-232 baud rates
- reject unsupported values instead of accepting any value in range
- align the CLI docs and inline comments with the implementation

## Testing
- `git diff --check`

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -918,6 +918,8 @@ region save
 **Parameters:**
 - `rate`: Baud rate (`9600`, `19200`, `38400`, `57600`, or `115200`)
 
+`set bridge.baud` returns an error for any other value.
+
 **Default:** `115200`
 
 ---
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -8,6 +8,19 @@
 #define BRIDGE_MAX_BAUD 115200
 #endif
 
+static bool isSupportedBridgeBaud(uint32_t baud) {
+  switch (baud) {
+    case 9600:
+    case 19200:
+    case 38400:
+    case 57600:
+    case 115200:
+      return true;
+    default:
+      return false;
+  }
+}
+
 // Believe it or not, this std C function is busted on some platforms!
 static uint32_t _atoi(const char* sp) {
   uint32_t n = 0;
@@ -107,7 +120,9 @@ void CommonCLI::loadPrefsInt(FILESYSTEM* fs, const char* filename) {
     _prefs->bridge_enabled = constrain(_prefs->bridge_enabled, 0, 1);
     _prefs->bridge_delay = constrain(_prefs->bridge_delay, 0, 10000);
     _prefs->bridge_pkt_src = constrain(_prefs->bridge_pkt_src, 0, 1);
-    _prefs->bridge_baud = constrain(_prefs->bridge_baud, 9600, BRIDGE_MAX_BAUD);
+    if (!isSupportedBridgeBaud(_prefs->bridge_baud)) {
+      _prefs->bridge_baud = BRIDGE_MAX_BAUD;
+    }
     _prefs->bridge_channel = constrain(_prefs->bridge_channel, 0, 14);
 
     _prefs->powersaving_enabled = constrain(_prefs->powersaving_enabled, 0, 1);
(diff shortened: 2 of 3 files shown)


kbrantley:
Where are these rates documented? I've been unable to find anything in the standard that says that you have to use one of those specific values.

robekl:
Thanks for catching this. I dug deeper into the actual RS-232 bridge implementation and the lower-level UART support across a non-trivial set of MeshCore bridge targets, and I don’t think the current enumerated-only restriction is well supported.

The key point is that MeshCore does not implement a board-specific baud table for RS-232 bridge mode. In [`RS232Bridge.cpp`](https://github.com/meshcore-dev/MeshCore/blob/dev/src/helpers/bridges/RS232Bridge.cpp), the configured value is passed straight through to the platform UART driver via `begin(_prefs->bridge_baud)` after platform-specific pin setup:

- ESP32: `HardwareSerial`
- nRF52: `Uart`
- RP2040: `SerialUART`
- STM32: `HardwareSerial`

I checked enabled or explicitly supported RS-232 bridge variants across several hardware families, including:

- ESP32:
  - [`variants/heltec_v3/platformio.ini`](https://github.com/meshcore-dev/MeshCore/blob/dev/variants/heltec_v3/platformio.ini)
  - [`variants/lilygo_tlora_v2_1/platformio.ini`](https://github.com/meshcore-dev/MeshCore/blob/dev/variants/lilygo_tlora_v2_1/platformio.ini)
- nRF52:
  - [`variants/rak4631/platformio.ini`](https://github.com/meshcore-dev/MeshCore/blob/dev/variants/rak4631/platformio.ini)
  - [`variants/promicro/platformio.ini`](https://github.com/meshcore-dev/MeshCore/blob/dev/variants/promicro/platformio.ini)
  - [`variants/rak3112/platformio.ini`](https://github.com/meshcore-dev/MeshCore/blob/dev/variants/rak3112/platformio.ini)
- RP2040:
  - [`variants/waveshare_rp2040_lora/platformio.ini`](https://github.com/meshcore-dev/MeshCore/blob/dev/variants/waveshare_rp2040_lora/platformio.ini)
  - [`variants/rak11310/platformio.ini`](https://github.com/meshcore-dev/MeshCore/blob/dev/variants/rak11310/platformio.ini)
- STM32:
  - [`variants/wio-e5-dev/platformio.ini`](https://github.com/meshcore-dev/MeshCore/blob/dev/variants/wio-e5-dev/platformio.ini)

From the lower-level platform documentation:

- ESP32 Arduino docs describe `Serial.begin(baud-rate, ...)` with supported baud rates from `300` to `5000000`.
  - <https://docs.espressif.com/projects/arduino-esp32/en/latest/api/serial.html>

- Arduino-Pico documents hardware serial as `begin(baudrate, ...)`, and Raspberry Pi’s Pico SDK documents `uart_set_baudrate()` / `uart_init()` as setting the UART to the nearest achievable baud.
  - <https://arduino-pico.readthedocs.io/en/stable/serial.html>
  - <https://www.raspberrypi.com/documentation/pico-sdk/hardware.html>

- Nordic documentation points to broader UART support than the five-value restriction here. For example, nRF52 documentation describes configurable UART/UARTE baud settings up to `1 Mbps`, not just `9600/19200/38400/57600/115200`.
  - <https://infocenter.nordicsemi.com/pdf/nRF52840_PDK_User_Guide_v1.2.pdf>
  - <https://infocenter.nordicsemi.com/pdf/nRF51_RM_v3.0.pdf>

- STM32 support also appears broader than the current five-value set. STM32duino’s serial API is generic `begin(baud)` rather than a fixed enumerated table, and the STM32 serial stack is documented as configurable rather than limited to those specific values.
  - <https://github.com/stm32duino/Arduino_Core_STM32/wiki/API>
  - <https://os.mbed.com/docs/mbed-os/v5.15/apis/serial.html>

Based on that, I think the current restriction is narrower than the actual lower-level UART support on at least ESP32 and RP2040, and likely also on nRF52 and STM32. So while the old docs/comments were too specific, changing the implementation to enforce only those five values appears to overcorrect.

I think the safer next step would be another PR that updates the docs/comments to match the actual lower-level support model, and verifies that baud-rate consumption in MeshCore is aligned with what the underlying platform drivers really support.

robekl:
Superseded by https://github.com/meshcore-dev/MeshCore/pull/2111, which updates bridge.baud to match the underlying UART pass-through behavior and refreshes the docs/comments accordingly.

PR #2103, opened 2026-03-20
docs: add radio.rxgain command reference

robekl:
## Summary
- document the `radio.rxgain` CLI command
- describe the accepted values and current board support

## Testing
- `git diff --check`

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -210,6 +210,20 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 
 ---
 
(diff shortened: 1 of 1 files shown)


fred777:
This one is supposed to work not only on repeaters but also on companions, right? 

And is it related to the external LNA on certain boards like Heltec v4 or does it only affect SX1276 internal gain?

I tried it with a USB v1.14.1 companion on a Heltec v4 board but it says "unknown variable".

recrof:
@liamcottle can be safely merged

Issue #1805, opened 2026-02-23
Thinknode M1 1.13.0 - Red Led Solid on


tgaguy:
Just updated from 1.9.0 and now the RED led is 99% solid on.  Pin 38 ownership issue?

#define PIN_SPI1_MISO           (38)
also
#define LED_RED                 (38)

also these duplicates
#define LED_BLUE (14)
#define PIN_NEOPIXEL (14)

#define PIN_GPS_PPS (36)
#define LED_GREEN   (36)

The Thinknode M1 doesnt have a visible Green Led.

Not sure GPS is working, cant seem to get a fix after many many minutes in clear sky. Says Sats are 0.

jakobmn:
The issue with the LED "solid" on is the polarity being defined incorrectly.
As part of verifying this fix, I discovered that there are quite a few discrepancies with the datasheet.
I will create a separate pull request for cleaning up these issues.

jakobmn:
> This issue is stale because it has been open for 60 days with no activity. Remove the stale label or add a comment if this issue is still relevant, otherwise this issue will automatically close in 7 days.

Still waiting for someone to react to the pull request that fixes this issue...

tgaguy:
updated to 1.16.0.  Red led it somewhat lit.

PR #2105, opened 2026-03-20
Fix LED polarity on Thinknode M1, and align target configuration with datasheet

jakobmn:
Fix polarity of LEDs on Thinknode M1
Saw that someone reported this as #1805 
I had the same issue on mine. After changing the polarity, the status LED is on for 20-200ms instead of ~4s on and 20-200ms off.

### DIFF
--- a/variants/thinknode_m1/ThinkNodeM1Board.cpp
+++ b/variants/thinknode_m1/ThinkNodeM1Board.cpp
@@ -12,7 +12,7 @@ void ThinkNodeM1Board::begin() {
 
 #ifdef P_LORA_TX_LED
   pinMode(P_LORA_TX_LED, OUTPUT);
-  digitalWrite(P_LORA_TX_LED, LOW);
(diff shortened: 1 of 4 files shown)


jakobmn:
Sorry. I'm not used to Githubs workflow, so I thought that I could make two different pull requests, but the second commit somehow ended up here even though I made it after the pull request.

In the second commit I clean up the target configuration for the Thinknode M1, and align it with the datasheet: https://www.elecrow.com/download/product/CIL12901M/ThinkNode%20M1_LoRa_Meshtastic_Transceiver_DataSheet.pdf

- No green LED on target - what was defined as the green LED was in fact the RED
- Blue LED defined incorrectly - explicitly disable it when used as Lora TX LED, and define it correctly otherwise
- button 2 defined incorrectly
- display MISO not connected
- GPS PPS not connected on target
- Use LED_STATE_ON define in thinknode m1 internal functions

tgaguy:
Updated to 1.16.0, Red LED on flashes OFF.

PR #2111, opened 2026-03-21
Make bridge baud a UART pass-through setting

robekl:
## Summary
- stop enforcing a MeshCore-specific baud list for `bridge.baud`
- store any positive parsed baud value and pass it through to the underlying UART library
- keep the default at `115200` when the stored preference is unset
- align CLI docs and comments with the actual pass-through behavior

## Why
MeshCore does not implement board-specific baud validation for the RS-232 bridge. The configured value is forwarded directly into the platform UART driver in `RS232Bridge::begin()` via `begin(_prefs->bridge_baud)`, so the lower-level UART implementation is the authority on what is supported or how a requested baud is approximated.

Relevant lower-level references:
- ESP32 Arduino serial API: https://docs.espressif.com/projects/arduino-esp32/en/latest/api/serial.html
- Arduino-Pico serial docs: https://arduino-pico.readthedocs.io/en/stable/serial.html
- Raspberry Pi Pico SDK UART docs: https://www.raspberrypi.com/documentation/pico-sdk/hardware.html
- Nordic nRF52840 PDK / UART documentation reference: https://infocenter.nordicsemi.com/pdf/nRF52840_PDK_User_Guide_v1.2.pdf
- STM32duino API reference: https://github.com/stm32duino/Arduino_Core_STM32/wiki/API

## Behavior
- `set bridge.baud <rate>` now accepts any positive integer parsed by the existing CLI style
- invalid or non-positive values still return an error
- MeshCore no longer clamps persisted baud values back into an artificial `9600..115200` window on load
- unsupported values remain the responsibility of the underlying platform UART stack

## Validation
- `git diff --check`
- `pio run -e Heltec_v3_terminal_chat`
- `pio run -e waveshare_rp2040_lora_repeater`
- `pio run -e wio-e5-repeater_bridge_rs232`
- `pio run -e RAK_4631_repeater_bridge_rs232_serial1`

## Testing Firmware

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -916,9 +916,13 @@ region save
 - `set bridge.baud <rate>`
 
 **Parameters:**
-- `rate`: Baud rate (`9600`, `19200`, `38400`, `57600`, or `115200`)
+- `rate`: Positive integer baud rate value passed to the underlying serial library
 
 **Default:** `115200`
+
+**Notes:**
+- MeshCore stores the configured baud rate and passes it to the platform UART library when the RS-232 bridge starts or restarts.
+- Supported values depend on the board/core UART implementation. Some platforms may reject unsupported values or round to the nearest achievable baud rate.
 
 ---
 
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -4,10 +4,6 @@
 #include "AdvertDataHelpers.h"
 #include <RTClib.h>
 
-#ifndef BRIDGE_MAX_BAUD
-#define BRIDGE_MAX_BAUD 115200
-#endif
-
 // Believe it or not, this std C function is busted on some platforms!
 static uint32_t _atoi(const char* sp) {
   uint32_t n = 0;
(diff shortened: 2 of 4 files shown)


robekl:
One likely review question here is how unsupported positive baud values are handled.

In the current implementation, MeshCore does not own that validation.  forwards  directly into the platform serial object via , so support is determined by the underlying UART stack on each platform.

That matches the lower-level APIs we are already using:
- ESP32 Arduino exposes  and documents configurable baud rates directly rather than a fixed MeshCore-owned table: https://docs.espressif.com/projects/arduino-esp32/en/latest/api/serial.html
- Arduino-Pico also uses , and the Pico SDK describes the UART layer selecting the nearest achievable baud rather than surfacing a generic validation error back through this call path: https://arduino-pico.readthedocs.io/en/latest/serial.html and https://www.raspberrypi.com/documentation/pico-sdk/hardware.html
- STM32duino similarly uses  without a uniform error return at the MeshCore call site: https://github.com/stm32duino/Arduino_Core_STM32/wiki/API

So the right split here is:
- MeshCore rejects non-positive values
- MeshCore stores positive values and passes them through unchanged
- the platform UART library decides whether the requested baud is supported or approximated

There is not a portable backend error for MeshCore to propagate here, because the APIs in this path are configure-style  calls rather than a cross-platform validated interface with a meaningful success/failure result.

PR #2114, opened 2026-03-21
Add repeater-only observability metrics

robekl:
## Summary
- add a first batch of repeater-only observability metrics
- expose the new counters through a new serial-only `stats-repeater` command
- add `active_nbrs` and `stable_nbrs` to repeater `stats-core`
- leave the remote repeater status payload unchanged for compatibility

## What changed
This adds a small repeater-local metrics block in `examples/simple_repeater/MyMesh` and increments it at the actual decision points for:
- direct ACK sends
- flood ACK sends
- admin request handling
- CLI command handling
- admin replies sent direct vs flood
- flood packets forwarded
- flood drops due to:
  - forwarding disabled
  - `flood_max`
  - missing allowed region
  - loop detection

It also tracks repeater neighbor activity using the existing neighbor table and exposes two derived counts:
- `active_nbrs`
- `stable_nbrs`

The new serial-only command is:
- `stats-repeater`

## Why serial-only first
These metrics are operationally useful for repeater tuning, but they do not need a protocol change yet.

Keeping them serial-only in the first step avoids changing the packed remote repeater status response and avoids breaking existing clients or parsers that already consume the current `REQ_TYPE_GET_STATUS` payload.

That keeps this change low risk while still making the new observability immediately available on deployed repeaters through the local admin/debug surface.

## Benefits
- gives repeater operators visibility into ACK behavior and admin traffic
- makes flood forwarding decisions and drop reasons observable
- adds a simple signal for local repeater neighborhood health
- keeps the implementation cheap and local to the repeater firmware
- avoids any wire-format or compatibility changes

## Tradeoffs
- the new counters are currently available only through serial CLI
- counters are cumulative from boot and reset with `clear stats`
- neighbor stability is intentionally simple in this first pass

## Validation
- `git diff --check`
- `pio run -e Heltec_v3_repeater`
- `pio run -e RAK_4631_repeater`
- `pio run -e waveshare_rp2040_lora_repeater`

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -119,6 +119,8 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 
 **Serial Only:** Yes
 
+**Repeater Note:** Repeater firmware also includes `active_nbrs` and `stable_nbrs` in this response.
+
 ---
 
 ### Radio Stats - Noise floor, Last RSSI/SNR, Airtime, Receive errors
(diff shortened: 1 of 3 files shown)


terminalvelocity23:
Looks simple and nice, a good addition to #1938.

PR #2116, opened 2026-03-21
Add nrf52 power management for Promicro/Faketec

ctrlaltca:
Enables power management for Promicro/Faketec boards.
Tested locally on 4 different boards:
- undervoltage (< 3300mV) boot prevention works and system gets shutdown; over 3300mV it boots normally;
- LPCOMP is configured and once the system reaches the vale defined by REFSEL, the system boots up again;
- on VBUS the board boots up again

Please note that the value of R1 and R2 resistors is not standard on faketec boards, so REFSEL may need tuning depending on the ratio used by the voltage divider. The provided value seems to work fine for R1=R2 (voltage divider ratio =2) and should be conservative for R1=1.5*R2 (another common ratio used).
In meshcore the ADC_MULTIPLER ratio is currently set to `1.815f`, that i assume it's meant to be equal to `(2 - losses)`.


Pacjunk:
This doesnt seem to take into account recent changes to this code. See #2088

ctrlaltca:
Thanks @Pacjunk , i saw that PR but I'll have to wait until it gets merged before rebasing mine on it.

oltaco:
I don't support enabling this by default for promicro. In my testing I found waking from LPCOMP to be very finicky, most likely dependent on size of vdiv resistors used.

For example if you use 1m/1m for the vdiv the voltage rise has to be very fast to trigger the wake from LPCOMP and it will likely never happen when on solar power.

ctrlaltca:
@oltaco I'm now using really lower values for vdiv resistors (10k/10k), but i still came to the conclusion that LPCOMP not always works reliably, mostly due to the internal voltage getting to a stable value (around 3v) only when battery voltage is at least 3.3v.
Still, I think this PR is still desiderable for a couple of reasons:
 * wake up ok VUSB never failed me, even when using an external solar power connected on USB-C
 * when the battery gets in the < 2.5v area the board enters a bootloop (restart -> die at first advert -> restart) that goes on until the flash gets fried, practically bricking the board until it gets erased and reflashed. The `checkBootVoltage()` avoids this.

ctrlaltca:
Rebased on 1.15.0

PR #2124, opened 2026-03-22
Make radio.rxgain true by default after upgrades

### LINKED ISSUE #2118: [Bug] RX Boosted Gain Mode switches to OFF when upgrading to v1.14.1
Already discussed in the thread of the PR: [first finding](https://github.com/meshcore-dev/MeshCore/pull/1653#issuecomment-4101341378) + [more testing](https://github.com/meshcore-dev/MeshCore/pull/1653#issuecomment-4104182803)

The problem only occurs in repeater and room-server, not in the companion firmware.

**Fresh install** of v1.14.1: 
RX Boosted Gain Mode = **ON**

**Upgrade** from v1.14.0 to v1.14.1: 
RX Boosted Gain Mode = **OFF** => unexpected / inconsistent

Steps to repeat the problem:

- Flash previous repeater firmware v1.14.0 using the web flasher, with 'Erase device' checked
- Connected to serial CLI
- `set radio 869.6179809,62.5,8,8`
- `reboot`
- Flash repeater firmware v1.14.1 using the web flasher, with Erase device NOT checked
- `get radio.rxgain` returns **off** => UNEXPECTED

In previous releases where the `radio.rxgain` setting was not availble, the setting was always ON.
So when you upgrade to 1.14.1 and don't think about out, it suddenly switches to OFF.

#### Fix if you want to keep the old behaviour:
After upgrading to v1.14.1 execute in CLI: `set radio.rxgain on`

weebl2000:
fixes #2118

Try here: https://mcimages.weebl.me?commitId=fix-radio-rxgain-true-by-def

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -55,7 +55,7 @@ void CommonCLI::loadPrefsInt(FILESYSTEM* fs, const char* filename) {
     file.read((uint8_t *)&_prefs->tx_power_dbm, sizeof(_prefs->tx_power_dbm));        // 76
     file.read((uint8_t *)&_prefs->disable_fwd, sizeof(_prefs->disable_fwd));          // 77
     file.read((uint8_t *)&_prefs->advert_interval, sizeof(_prefs->advert_interval));  // 78
-    file.read((uint8_t *)&_prefs->rx_boosted_gain, sizeof(_prefs->rx_boosted_gain));  // 79
+    file.read(pad, 1);                                                                // 79 : 1 byte unused (was rx_boosted_gain in v1.14.1, moved to end for upgrade compat)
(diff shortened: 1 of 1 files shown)


jeroenvermeulen:
I have built the `dev` branch and tested the upgrade from 1.14.0 to 8009cf0d on Heltec HTIT Tracker.
I can confirm the problem is SOLVED.

renormalist:
@jeroenvermeulen Also when upgrading from 1.14.1?

jeroenvermeulen:
@renormalist As I have explained in my [comment](https://github.com/meshcore-dev/MeshCore/issues/2118#issuecomment-4106910487) on Bug #2118: Yes when you upgrade from 1.14.1 to a newer version the RX Boosted Gain Mode `radio.rxgain` will be ON.

Issue #2123, opened 2026-03-22
rx_delay, tx_delay and direct.tx_delay defaults are too Low


KPrivitt:
The current default values are:
  _prefs.rx_delay_base = 0.0f;   // turn off by default, was 10.0;
  _prefs.tx_delay_factor = 0.5f; // was 0.25f
  _prefs.direct_tx_delay_factor = 0.3f; // was 0.2

These control the Random Backoff behavior of a repeater, which reduces the probability of a collision between neighbor repeaters.

txdelay and rxdelay address different collision modes:
•	txdelay randomizes when nodes transmit, directly reducing simultaneous airtime occupancy.
•	rxdelay randomizes when nodes listen, preventing synchronized receive windows, ACK storms, and deterministic hop timing patterns.

tx_delay
•	Reduces collisions when multiple repeaters re-send a Flood messages that are received at the same moment by spreading the transmissions across a number of Slots that are randomly selected.

direct.txdelay is not designed to mitigate collisions between copies of the same message. Instead, it reduces contention between independent direct transmissions, such as:
•	Multiple nodes sending direct messages at similar times
•	A single node with several direct messages queued close together
•	Bidirectional or interactive traffic where both ends transmit concurrently.

Collisions are reduced by a similar mechanism: spreading the transmissions/receptions across a number of Slots that are randomly selected.

Backoff is essential in avoiding collisions.  Many repeater owners are not aware of this capability and do not change them, this leaves the mesh under configured in order to aviod collisions. **The default values are too low and are not providing any significant backoff for the Mesh.**

The default value for rx_delay = 0, which is off.  So, this mechanism is not being utilized.  A recommended minimum value for **rxdelay is 3**

The default value for direct.x_delay = 0.3, which is provides 2 backoff slots, which has a high probability of a collision.    A recommended minimum value for **direct.x_delay is 1** which will provide 5 backoff slots and reduces the probability of a collision.

The default value for tx_delay = 0.5, which is provides 2 backoff slots, additional backoff slots will reduce the probability of a collision.    A recommended minimum value for **txdelay is 1.6** which provides 8 backoff slots and reduces the probability of a collision.

These are all recommended minimum values and are not the optimum values for a particular repeater and its number of neighboring repeaters.  But it is better than the current settings and will reduce the collisions that occur.  The optimum values vary according to the environment (the number of neighbor repeaters) of an individual repeater and are higher than these minimum recommendations.

[...]

PR #2125, opened 2026-03-22
Autotune of delays based on number of neighbors

stachuman:
Based on discussion: [Algorithm to Automatically Adjust Repeater Parameters](https://github.com/meshcore-dev/MeshCore/discussions/2053)

Implements automatic tuning of txdelay, direct.txdelay, and rxdelay parameters based on active neighbor count, as described in White Paper 1 (Section 8.0). Delay factors are consolidated into the Mesh base class, eliminating duplicated override code across examples.

- Adds a lookup table (DelayTuning.h) indexed by active neighbor count (SNR > 0, heard within 7 (define value) days) with 13 entries ranging from sparse (0 neighbors) to regional (12+)
- Moves getRetransmitDelay, getDirectRetransmitDelay, and calcRxDelay from per-example overrides into the Mesh base class, driven by three configurable float members

- Repeater: counts active neighbors and recalculates on neighbor update, every 5 minutes (define value), and on CLI changes
- Room server / Sensor: use 0-neighbor baseline (no neighbor table available)
- Adds auto_tune_delays persisted preference (on by default for new installs), with CLI: get autotune, set autotune on/off
- Manually setting rxdelay, txdelay, or direct.txdelay via CLI auto-disables autotune

New variable (set/get):
autotune - with values on/off

The default - after upgrade - is autotune ON

This change is backward compatible.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -59,6 +59,8 @@
 #define CLI_REPLY_DELAY_MILLIS      600
 
 #define LAZY_CONTACTS_WRITE_DELAY    5000
+#define NEIGHBOUR_ACTIVE_MAX_AGE    (7 * 24 * 3600)  // 7 days
+#define AUTOTUNE_INTERVAL_MILLIS    (5 * 60 * 1000)  // 5 minutes
 
 void MyMesh::putNeighbour(const mesh::Identity &id, uint32_t timestamp, float snr) {
 #if MAX_NEIGHBOURS // check if neighbours enabled
@@ -84,6 +86,7 @@ void MyMesh::putNeighbour(const mesh::Identity &id, uint32_t timestamp, float sn
   neighbour->advert_timestamp = timestamp;
   neighbour->heard_timestamp = getRTCClock()->getCurrentTime();
   neighbour->snr = (int8_t)(snr * 4);
+  recalcAutoTune();
 #endif
 }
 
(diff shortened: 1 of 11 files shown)


KPrivitt:
Are you planning to add a purge of stale neighbors? (remove all with a last heard greater than 10 days).

This is a necessary part of the algorithm to prevent neighbor count creep: when stale invalid neighbors are counted pushing up and possibly overflowing the index into the table.  Left unchecked the values will continually increase an unchecked false count.

KPrivitt:
One additional thing: my example table limited the txdelay to 2.0, which was the maximum value allowed by the CLI.  This may not be a valid limitation, but I have no data to back this up, my leaning would be to increase this to 2.5 (possibly 3.0) to allow additional slots for a regional repeater.  Additional slots is beneficial to regional repeaters since they have a set of non-conflicting values with the repeaters "below them", there is no possibility of a collision within this set.  Eliminating a collision is always beneficial vs the increased airtime  of the retransmission of failed message.  But a larger vakue does add additional delay to the end to end transmission time (for that hop) it seems that the additional delay is not much to pay for a non-collision transmission (but that's my opinion, I have no information that points to the potential detrimental effect of a delayed transmission) In terms of the human timeframe it is not really noticeable.

For what it is worth, this seems reasonable.

stachuman:
> Are you planning to add a purge of stale neighbors? (remove all with a last heard greater than 10 days).
> 
> This is a necessary part of the algorithm to prevent neighbor count creep: when stale invalid neighbors are counted pushing up and possibly overflowing the index into the table. Left unchecked the values will continually increase an unchecked false count.

counting active neighbors take into account two factors - 
SNR > 0 and LAST HEADR < NEIGHBOUR_ACTIVE_MAX_AGE

At the moment I've set it:
#define NEIGHBOUR_ACTIVE_MAX_AGE    (7 * 24 * 3600)  // 7 days

The way it works - every AUTOTUNE_INTERVAL_MILLIS (which now I set to 5 minutes - quite often, but this is fast function) I'm simply checking neighbors table and I'm counting number of neighbors according to above criteria.

stachuman:
> One additional thing: my example table limited the txdelay to 2.0, which was the maximum value allowed by the CLI. This may not be a valid limitation, but I have no data to back this up, my leaning would be to increase this to 2.5 (possibly 3.0) to allow additional slots for a regional repeater. Additional slots is beneficial to regional repeaters since they have a set of non-conflicting values with the repeaters "below them", there is no possibility of a collision within this set. Eliminating a collision is always beneficial vs the increased airtime of the retransmission of failed message. But a larger vakue does add additional delay to the end to end transmission time (for that hop) it seems that the additional delay is not much to pay for a non-collision transmission (but that's my opinion, I have no information that points to the potential detrimental effect of a delayed transmission) In terms of the human timeframe it is not really noticeable.
> 
> For what it is worth, this seems reasonable.

I have no data either. I went through the white paper (you are the author?), I've checked situation in my area (a lot of collision, I need to built better monitoring software to have more clear picture, at the moment I can only refer to my 3 companions and 3 nodes I have - which are located in different areas - rarely I'm able to make direct communication, most goes through flood).

The limit for delays - when setting in code - is of course non-existing.

weebl2000:
Some people asked if they could test this, for those who wnat to, I've pushed this PR to my fork and people can try it with https://mcimages.weebl.me?commitId=stachuman

stachuman:
> Some people asked if they could test this, for those who wnat to, I've pushed this PR to my fork and people can try it with https://mcimages.weebl.me?commitId=stachuman

Thank you!

I realised that get command (e.g. get rxdelay) will NOT return auto-adjusted value - added fix [get delays fix](https://github.com/meshcore-dev/MeshCore/pull/2125/commits/51dfbb2f6361e437f181a91e4a784f60bbc09aa8) - so could be that some of the test builds won't return value - even if internally it is used.

KPrivitt:
stachuman, Thank You! for implementing this.  After I sent my comment, I realized that you had checked the last heard and didn't use it in the count, so your implementation works just fine.

But stale neighbors will build up in the database and should be purged (but that does happens every time a new version comes out, so it's not permanent).

I posted a request to #2123 for repeater owners to post their current Error Rate [# Errors/(Total RX + #Errors)], which I suspect is mostly due to collisions.  This is some easily accessible data.

stachuman:
PR updated to the latests finding from the discussion: https://github.com/meshcore-dev/MeshCore/discussions/2053

## Measured performance at defaults  (same 4 topologies, 6 seeds each):

Density | Msg Delivery | Channel Delivery | ACK Delivery | Avg ACK Copies
-- | -- | -- | -- | --
sparse | 67.0% | 57.5% | 36.0% | 0.60
medium | 62.7% | 78.8% | 33.7% | 0.70
dense | 65.0% | 64.5% | 24.3% | 0.50
very_dense | 64.3% | 43.0% | 13.0% | 0.20
Mean | 64.8% | 60.9% | 26.8% | 0.50


## Proposed changes - autotune tx/direct tx delays:
Density | Msg Delivery Δ | ACK Delivery Δ | ACK Copies Δ
-- | -- | -- | --
sparse | −2.7pp | +11.0pp | +0.80
medium | −6.3pp | +1.6pp | +1.00
dense | −6.0pp | +13.7pp | +1.30
very_dense | −9.3pp | +17.0pp | +1.40
Mean | −6.1pp | +10.9pp | +1.13

stachuman:
Created new PR -> https://github.com/meshcore-dev/MeshCore/pull/2337 - closing this one

PR #2134, opened 2026-03-23
Fix default radio.rxgain for Station G2

jeroenvermeulen:
As @LitBomb pointed out in his [comment](https://github.com/meshcore-dev/MeshCore/issues/2118#issuecomment-4108168109) on #2118 RX Boosted Gain should not be enabled for the Station G2.

This change is a fix for #2124 to make the default of `radio.rxgain` to be OFF on the Station G2.

This restores the pre-1.14.1 behaviour with the only change being the user is now able to change the setting in the CLI.


LitBomb:
@jeroenvermeulen Thank you!!

PR #2138, opened 2026-03-24
Support for GAT562 Mesh Watch13 device

MGJ520:
Support for GAT562 Mesh Watch13 device
支持 GAT562 Mesh Watch13 设备

I have verified:  我已经核实过：
**Transmit  **发送
Receive  收到
Bluetooth  蓝牙
**

I have tested ALL firmware.
我已经测试了所有固件。

his one is based off of a combination of documentation from:
他的这份文件是基于以下来源的资料综合而成：
https://github.com/gat-iot/GAT562-family/tree/main

Purchase channels:  购买渠道：
https://www.aliexpress.com/item/1005010477549302.html
or  或者
https://item.taobao.com/item.htm?abbucket=8&id=966983316104&mi_id=0000X1yi3Av6rAM_ne6TA4jwLN4DGTHyXnNllHKNJsL2NGs&ns=1&skuId=6075618278611&spm=a21n57.1.hoverItem.7&utparam=%7B%22aplus_abtest%22%3A%22f231cfa66214eda3e16388c4a3c3be3a%22%7D&xxc=taobaoSearch
image
image

### DIFF
--- a/variants/gat562_mesh_watch13/platformio.ini
+++ b/variants/gat562_mesh_watch13/platformio.ini
@@ -28,37 +28,38 @@ lib_deps =
   ${sensor_base.lib_deps}
   adafruit/Adafruit SSD1306 @ ^2.5.13
 
-[env:GAT562_Mesh_Watch13_repeater]
-extends = GAT562_Mesh_Watch13
-build_flags =
-  ${GAT562_Mesh_Watch13.build_flags}
-  -D DISPLAY_CLASS=SSD1306Display
-  -D ADVERT_NAME='"GAT562 Repeater"'
-  -D ADVERT_LAT=0.0
-  -D ADVERT_LON=0.0
-  -D ADMIN_PASSWORD='"password"'
-  -D MAX_NEIGHBOURS=50
-;  -D MESH_PACKET_LOGGING=1
-;  -D MESH_DEBUG=1
-build_src_filter = ${GAT562_Mesh_Watch13.build_src_filter}
-  +<helpers/ui/SSD1306Display.cpp>
-  +<../examples/simple_repeater>
 
-[env:GAT562_Mesh_Watch13_room_server]
-extends = GAT562_Mesh_Watch13
-build_flags =
-  ${GAT562_Mesh_Watch13.build_flags}
-  -D DISPLAY_CLASS=SSD1306Display
-  -D ADVERT_NAME='"GAT562 Room"'
-  -D ADVERT_LAT=0.0
-  -D ADVERT_LON=0.0
-  -D ADMIN_PASSWORD='"password"'
-  -D ROOM_PASSWORD='"hello"'
-;  -D MESH_PACKET_LOGGING=1
-;  -D MESH_DEBUG=1
-build_src_filter = ${GAT562_Mesh_Watch13.build_src_filter}
-  +<helpers/ui/SSD1306Display.cpp>
-  +<../examples/simple_room_server>
+;[env:GAT562_Mesh_Watch13_repeater]
+;extends = GAT562_Mesh_Watch13
+;build_flags =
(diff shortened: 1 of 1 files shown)


variants/gat562_mesh_watch13/platformio.ini
```diff
@@ -0,0 +1,88 @@
...
+build_src_filter = ${nrf52_base.build_src_filter}
+  +<../variants/gat562_mesh_watch13>
+  +<helpers/ui/SSD1306Display.cpp>
+  +<helpers/ui/MomentaryButton.cpp>
+  +<helpers/sensors>
+lib_deps =
+  ${nrf52_base.lib_deps}
+  ${sensor_base.lib_deps}
+  adafruit/Adafruit SSD1306 @ ^2.5.13
+
```
ripplebiz, line 30: Not sure the repeater and room server targets are practical(?)
Could help out our release compute time by removing those two. Already so many variants that get built at each release. Good to not have spurious ones that aren't needed.

MGJ520: > Not sure the repeater and room server targets are practical(?) Could help out our release compute time by removing those two. Already so many variants that get built at each release. Good to not have spurious ones that aren't needed.

Hello, I have thought about this issue. The low-power design of the watch is very suitable for outdoor solar-powered message relay nodes.

liamcottle: > The low-power design of the watch is very suitable for outdoor solar-powered message relay nodes.

I would say you can keep these in the `platformio.ini` but you should comment them out. Using a watch as a solar powered repeater node doesn't sound like a very normal use case to me either. Anyone that wants this can uncomment the definitions and build it themselves :)

As @ripplebiz noted, this would cost us more build minutes and increase our already 1-hour-long build process when we do new releases.

I would probably say just keep the companion role for the watch. The companion firmware already supports companion repeat mode, if a relay node use case is required.

variants/gat562_mesh_watch13/platformio.ini
```diff
@@ -0,0 +1,88 @@
...
+[env:GAT562_Mesh_Watch13_repeater]
```
liamcottle: Please comment out the repeater role, as discussed in previous comment.

variants/gat562_mesh_watch13/platformio.ini
```diff
@@ -0,0 +1,88 @@
...
+  -D ADVERT_LON=0.0
+  -D ADMIN_PASSWORD='"password"'
+  -D MAX_NEIGHBOURS=50
+;  -D MESH_PACKET_LOGGING=1
+;  -D MESH_DEBUG=1
+build_src_filter = ${GAT562_Mesh_Watch13.build_src_filter}
+  +<helpers/ui/SSD1306Display.cpp>
+  +<../examples/simple_repeater>
+
+[env:GAT562_Mesh_Watch13_room_server]
```
liamcottle: Room server probably shouldn't be portable on someone's wrist either, as this would lead to pathing issues pretty quickly. The room server will be wasting airtime trying to deliver messages when it's moved somewhere else. I'd suggest commenting this one out and allowing someone to build manually if desired.

variants/gat562_mesh_watch13/platformio.ini
MGJ520: Hmm, it's a pretty good solution.

MGJ520:
> Thanks for the PR! Looks good to me. I've left some review notes for adjustment. Once those are resolved, should be good to merge :)
> 
> Btw, I ordered one of these nodes from the AliExpress store. Keen to try it out!顺便说一下，我是在 AliExpress 商店订购了这

I have already resolved it and submitted a PR.:)

PR #2141, opened 2026-03-24
Fix: Re-queue packet on send failure instead of dropping it

OverkillFPV:
Currently when the CAD timeout's due to high traffic, the packet is force sent after 4 seconds. If StartSendRaw fails due to the radio being busy (likely to be due to channel activity still) the packet is dropped. This change makes it so that instead of dropping the packet entirely, it is requeued with a delay and retried. This seems to be why some of my high traffic nodes with 20+ neighbours have become unreliable of late.

### DIFF
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -60,7 +60,7 @@ uint32_t Dispatcher::getCADFailRetryDelay() const {
   return 200;
 }
 uint32_t Dispatcher::getCADFailMaxDuration() const {
-  return 6000;   // 6 seconds
+  return 8000;   // 8 seconds
 }
 
 void Dispatcher::loop() {
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -332,9 +332,11 @@ void Dispatcher::checkSend() {
         logTxFail(outbound, outbound->getRawLength());
 
         // re-queue instead of dropping so the packet gets another chance
-        _mgr->queueOutbound(outbound, 0, futureMillis(getCADFailRetryDelay()));
+        int retry_delay = getCADFailRetryDelay();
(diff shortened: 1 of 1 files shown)


OverkillFPV:
I was also wondering about increasing the timeout to >4 seconds as for some nodes with 50 neighbours it's not too unheard of to have 4 seconds of traffic

michaeldale:
I'm just adding my +1 as a comment for visibility. As the Australian NSW mesh grows, ensuring reliable message transfer is very important.

stachuman:
It would be an interesting case to test with simulator I just built -> [simv2](https://github.com/stachuman/meshcore_simv2) - I'll be happy to help building the case.

OverkillFPV:
> It would be an interesting case to test with simulator I just built -> [simv2](https://github.com/stachuman/meshcore_simv2) - I'll be happy to help building the case.

It could be yea. I have been running this fix for a while now and have had improvements on packets never being sent by busy repeaters

stachuman:
I did the analysis of library + spec: https://cdn.sparkfun.com/assets/6/b/5/1/4/SX1262_datasheet.pdf - I might be wrong, but this is how I see it (reason - to make sure, that my radio sim works correct):

startSendRaw uses internally startTransmit - which uses directly setTx() - from SX1262 - if there is ongoing RX - simply interrupt RX and starts TX (no standby called in between)
If that returns false - it's kind of rare issue with radio, not again CAD, so - in most of the cases - if there will be interference, ongoing RX - this method will not return false.

Long story short, false will be returned only on hardware failure.

OverkillFPV:
Ah yes, i had the impression after reading it that it would fail if the radio is busy, probably a reasonable fix to have anyway, but also the 6s longer queue must have a much greater affect that i had originally planned for

PR #2146, opened 2026-03-24
fix(bme680): fix gas resistance variable to produce usable data

NickDunklee:
This fixes the gas resistance output of the BME680 sensor so the data can be usable. In its current state, the value can loop and go negative, and never actually show the correct values from the sensor.

- Gas resistance should be a Cayenne LPP generic sensor, type 100, uint32 BE.
- addAnalogInput overflows int16×0.01 for typical 10k–500k Ω measurements.
- Tested on RAK19003 with a BME680 sensor on board.


IoTThinks:
Hi,
So how will the gas resistance be shown in this PR?

NickDunklee:
Instead of looping improperly negative due to an overflowing integer, it shows a positive range of values. Attached a screenie from one of my dashboards after the change. These sensors are particular (read: annoying).
Screenshot 2026-03-31 at 22 05 20

NickDunklee:
I am going to write a consolidated bme680 PR that will make this and the other irrelevant, but will leave these both open in case people want to see these changes to utilize them in their own builds for now. Will close them both once I have that PR up.

NickDunklee:
Finally integrated my BME680 changes so closing this, commit for that is at https://github.com/meshcore-dev/MeshCore/pull/2634

PR #2149, opened 2026-03-24
fix(bme680): fixed bme680 loss of precision int math error in pressure and altitude measurements

NickDunklee:
BME680.pressure / 100 was integer division before being passed onward resulting in loss of sensor resoluiton for both altitude and pressure changes.

- Changed to float math to fix the value only moving in 1hPa steps
- Pressure/altitude should no longer appear "stuck" waiting to jump the next 1hPa change
- Tested on RAK19003+BME680
    
It looks like BME280, BMP280, etc. should not be affected as their readPressure() is a float. BMP085 may have a similar problem, but I do not have one to test with.


NickDunklee:
Screenshot 2026-03-24 at 20 38 37

Left side of graph was before change, right side after. Updated this comment with a better graph.

NickDunklee:
I am going to write a consolidated bme680 PR that will make this and the other irrelevant, but will leave these both open in case people want to see these changes to utilize them in their own builds for now. Will close them both once I have that PR up.

NickDunklee:
Finally integrated my BME680 changes so closing this, commit for that is at https://github.com/meshcore-dev/MeshCore/pull/2634

PR #2151, opened 2026-03-25
fix: add zombie radio recovery for TX fail reset and RX stuck reboot

KG7QIN:
## Problem

Two silent failure modes can leave the radio non-functional with no
recovery path:

1. **TX zombie:** `startSendRaw()` fails but the packet is silently
   dropped. The node appears to be running but nothing is transmitted.
2. **RX zombie:** The radio leaves RX mode and never returns.
   `ERR_EVENT_STARTRX_TIMEOUT` was already detecting this after 8s but
   no recovery action was taken.

### Dispatcher

On `startSendRaw()` failure, re-queue the packet instead of dropping it.
Increment `tx_fail_count` on each failure and call `onTxStuck()` when the
count reaches the configured threshold. Reset to 0 on any successful TX.

On `STARTRX_TIMEOUT`, increment `rx_stuck_count` and call `onRxStuck()`
each 8s window. When the count reaches the threshold, call
`onRxUnrecoverable()`. After each recovery attempt, reset
`radio_nonrx_start`, `prev_isrecv_mode`, `cad_busy_start`, and
`next_agc_reset_time` so the 8s window restarts cleanly. Reset
`rx_stuck_count` to 0 when the radio returns to RX.

### Virtual hook API

```cpp
virtual uint8_t getTxFailResetThreshold() const { return 3; }
virtual void    onTxStuck()                     { _radio->resetAGC(); }
virtual uint8_t getRxFailRebootThreshold() const { return 3; }
virtual void    onRxStuck()                     { _radio->resetAGC(); }
virtual void    onRxUnrecoverable()             { }
```

Defaults are conservative (threshold 3, AGC reset as first attempt,
no-op unrecoverable). Subclasses override to reboot or do a deeper reset.

### NodePrefs / CLI

Thresholds are persisted and adjustable at runtime:

```
get tx.fail.threshold       -> current value (default 3)
set tx.fail.threshold 5     -> "OK - tx fail reset after 5 failures"
set tx.fail.threshold 0     -> "OK - tx fail reset disabled"

get rx.fail.threshold       -> current value (default 3)
set rx.fail.threshold 5     -> "OK - reboot after 5 rx recovery failures"
set rx.fail.threshold 0     -> "OK - rx fail reboot disabled"
```

CommonCLI stores these at fields 291/292 (next free: 293).
Companion radio stores at its own DataStore fields 90/91 (next free: 92).
Both default to 3 on first boot or upgrade from older firmware.

### Example overrides

All four examples override the threshold getters to return the persisted
prefs value and implement `onRxUnrecoverable()` to call `board.reboot()`.

---

## Follow-up: error flags, counter fixes, and forceResetAGC

The initial fix was confirmed non-functional when a radio entered zombie
mode in the field. Investigation identified several bugs:

**Root cause of failed recovery:** `resetAGC()` guards on
`isReceivingPacket()`, which reads live hardware IRQ flags
(`PREAMBLE_DETECTED`, `HEADER_VALID` on SX126x/LR11x0;
`SIGNAL_DETECTED`/`SIGNAL_SYNCHRONIZED` on SX1276). These flags can be
permanently stuck from interference, causing every recovery attempt to
silently no-op.

**Counter bugs:** `rx_stuck_count` was never reset after a forced
recovery attempt because `onRxStuck()` sets `prev_isrecv_mode = true`,
which prevents the normal transition check from firing. Both `uint8_t`
counters could also wrap to 0 at 255, defeating the threshold check.
After calling `onRxUnrecoverable()`, resetting the counter to 0 caused
the next retry to wait `threshold * 8s` instead of 8s.

### Additional changes

**`forceResetAGC()`** — new method on `RadioLibWrapper` that bypasses
the `isReceivingPacket()` guard. The chip-specific reset sequence
(SX126x/LR11x0: sleep + standby + calibrate; SX1276: warm sleep) already
clears stuck IRQ bits as a side effect. A pending completed interrupt is
still respected to avoid discarding a packet already in the FIFO.
`onRxStuck()` now calls `forceResetAGC()` instead of `resetAGC()`.

**Counter fixes:**
- All three `uint8_t` counters saturate at `UINT8_MAX` instead of wrapping
- `rx_stuck_count` resets to `reboot_threshold` (not 0) after a reboot attempt, so retries fire every 8s
- `rx_reboot_count` tracks how many times `onRxUnrecoverable()` has been called without recovery
- `tx_fail_count` is zeroed when the threshold is disabled (0) to prevent stale accumulation

**Max reboot attempts:**

```cpp
virtual uint8_t getRxMaxRebootAttempts() const { return 0; }  // 0 = unlimited
```

**Error flags** — two new queryable flags in `_err_flags`, cleared on
`begin()` and `resetStats()`:

- `ERR_EVENT_RADIO_ZOMBIE` (bit 3) — set when the stuck count exceeds the
  reboot threshold; cleared on recovery
- `ERR_EVENT_RADIO_DEAD` (bit 4) — set when `getRxMaxRebootAttempts()` is
  exhausted; requires a power cycle

Exposed via CLI on repeater, room server, and sensor firmware:

```
get radio.zombie    -> 0 or 1
get radio.dead      -> 0 or 1
get radio.err       -> 0x0000  (all flags as hex)
```

Also exposed programmatically:

```cpp
bool     isRadioZombie() const;
bool     isRadioDead()   const;
uint16_t getErrFlags()   const;
```

### Documentation

`docs/cli_commands.md` updated to document all new CLI commands:
- `get radio.zombie` / `get radio.dead` / `get radio.err` with error flag bit table
- `get`/`set tx.fail.threshold` and `get`/`set rx.fail.threshold`

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -7,6 +7,7 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 - [Operational](#operational)
 - [Neighbors](#neighbors-repeater-only)
 - [Statistics](#statistics)
+  - [Radio diagnostics](#radio-diagnostics)
 - [Logging](#logging)
 - [Information](#info)
 - [Configuration](#configuration)
@@ -139,6 +140,37 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 **Usage:** `stats-packets`
 
 **Serial Only:** Yes
+
+---
+
+### Radio diagnostics - Zombie and error flag queries
+
+#### Query zombie radio state
+**Usage:** `get radio.zombie`
+
+**Note:** Returns `1` if the radio has entered a zombie state (stuck count exceeded the RX fail threshold and a reboot has been attempted), `0` otherwise. Cleared automatically when the radio recovers or on firmware reboot.
+
+---
+
+#### Query dead radio state
+**Usage:** `get radio.dead`
+
+**Note:** Returns `1` if all reboot attempts have been exhausted and the radio cannot self-recover. A physical power cycle is required. Cleared on firmware reboot.
+
+---
+
+#### Query all radio error flags
+**Usage:** `get radio.err`
+
+**Note:** Returns all radio error flags as a 4-digit hex value (e.g. `0x0018`). Flags accumulate since last boot or `clear stats`. Use `get radio.zombie` and `get radio.dead` for simple boolean checks.
+
+| Bit | Flag                        | Meaning                                   |
+|-----|-----------------------------|-------------------------------------------|
+| 0   | `ERR_EVENT_FULL`            | Packet pool exhausted                     |
+| 1   | `ERR_EVENT_CAD_TIMEOUT`     | Channel activity detection timeout        |
+| 2   | `ERR_EVENT_STARTRX_TIMEOUT` | Radio failed to return to RX within 8s    |
+| 3   | `ERR_EVENT_RADIO_ZOMBIE`    | Radio stuck; reboot attempted             |
+| 4   | `ERR_EVENT_RADIO_DEAD`      | Radio unrecoverable; power cycle required |
 
 ---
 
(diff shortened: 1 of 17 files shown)


KG7QIN:
This is a common problem with bot operators.  The radios will become zombies failing on TX or RX and causing problems.  

The only way to fix this is to power cycle the radio by pressing the reset button or pulling power.

This is an attempt at fixing this without having to have physical access to the radio.

PR #1873, opened 2026-02-28
sensecap solar: add poweroff/shutdown support and nrf52 PM parity

Specter242:
## Summary
Add SenseCAP Solar repeater power-management parity for software shutdown:

- Add board-level power management hooks for SenseCAP Solar (`initiateShutdown`, `powerOff`) with button-based wake/system-off behavior.
- Enable required SenseCAP Solar variant defines for NRF52 power-management.
- Add SenseCAP Solar LED/wiring parity updates:
  - update LoRa TX LED pin mapping (`P_LORA_TX_LED`) from `12` to `11` for board wiring parity,
  - explicitly switch `LED_GREEN` and `LED_BLUE` off in `powerOff()` before entering system-off.
- Add CLI aliases: `poweroff` and `shutdown`.
- Add stability/usability fixes for real device behavior:
  - configure wake button with pull-up in system-off to avoid floating/false wake,
  - add SenseCAP Solar repeater long-press user-button shutdown (hold ~1.5s).
- Update NRF52 power-management support docs matrix for SenseCAP Solar.

## Why
SenseCAP Solar repeater firmware lacked a practical software power-off path and CLI shutdown command. This improves battery-device operability and aligns behavior with expected NRF52 power-management support.

Fixes #1872
Related to #1145

## Validation
- `~/.local/bin/pio run -e SenseCap_Solar_repeater` (SUCCESS)
- Verified firmware links and artifacts are produced for the target environment.

PR #1689, opened 2026-02-13
sensecap_solar: Fixing LED definitions

Confituurke:
### Summary

Corrects the LED pin definitions for the SenseCap Solar P1 to match the actual hardware. The blue LoRa TX LED was incorrectly named green in the code, and the user LED is white (not blue).

### Problem

- **User LED (D11)** – White LED on pin 11 was defined as `LED_GREEN` on pin 12
- **LoRa TX LED (D12)** – Blue LED on pin 12 was defined as `LED_BLUE` on pin 11

### Changes

- `LED_GREEN` (12) → `LED_WHITE` (11) – User LED on D11
- `LED_BLUE` (11) → `LED_BLUE` (12) – LoRa TX indicator on D12

### Hardware mapping

| Pin | Function | Define |
|-----|----------|--------|
| D11 (P0.15) | User LED (white) | `LED_WHITE` |
| D12 (P0.19) | LoRa TX / breathing LED (blue) | `LED_BLUE` |

PR #2157, opened 2026-03-25
fix(sensecap_solar): correct TX LED after remap & power-button pin changes

Confituurke:
Updates the SenseCap Solar build flag `P_LORA_TX_LED` from `11` to `12` so LoRa TX activity is driven on the blue LED again, consistent with `variants/sensecap_solar/variant.h` (LED_BLUE / PIN_LED on pin 12).

SenseCap Solar LED GPIO definitions were corrected in #1689 . A later power-button–related pin mapping change e323755 , #1873 updated board wiring, but `P_LORA_TX_LED` changed to `11`, which maps to `LED_WHITE` in the current headers and caused drift between the forks. That made the white LED blink on TX, which looks like a white/blue swap even though the underlying problem was the TX LED build flag not matching the updated LED pin macros.

@ripplebiz & @liamcottle FYI and approval; so these don't get changed/overwritten in another future PR.

@Specter242 FYI; since you initiated PR #1873 

Thank you

### DIFF
--- a/variants/sensecap_solar/SenseCapSolarBoard.cpp
+++ b/variants/sensecap_solar/SenseCapSolarBoard.cpp
@@ -49,7 +49,7 @@ void SenseCapSolarBoard::begin() {
 
 #ifdef LED_WHITE
   pinMode(LED_WHITE, OUTPUT);
-  digitalWrite(LED_WHITE, HIGH);
+  digitalWrite(LED_WHITE, LOW);
 #endif
 #ifdef LED_BLUE
   pinMode(LED_BLUE, OUTPUT);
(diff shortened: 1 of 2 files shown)


Specter242:
Thanks, this seems like it works better.

Confituurke:
Rebased to version **1.15.0** `dev`-branch

oltaco (approved):

oltaco:
Something doesn't seem quite right with this, can youchange the white LED setup in ``SenseCapSolarBoard::begin()`` so that it's not permanently lit?
image

```
--- a/variants/sensecap_solar/SenseCapSolarBoard.cpp
+++ b/variants/sensecap_solar/SenseCapSolarBoard.cpp
@@ -49,7 +49,7 @@ void SenseCapSolarBoard::begin() {

 #ifdef LED_WHITE
   pinMode(LED_WHITE, OUTPUT);
-  digitalWrite(LED_WHITE, HIGH);
+  digitalWrite(LED_WHITE, LOW);
 #endif
 #ifdef LED_BLUE
   pinMode(LED_BLUE, OUTPUT);
   ```

oltaco:
Thanks for your work on this everybody 👍

PR #2161, opened 2026-03-26
devcontainer doesn't need --privileged

nextgens:
Remove ``--privileged`` from the docker image for devcontainer; we don't need it.

166 is the magic number that should work for everyone https://docs.kernel.org/admin-guide/devices.html
We may need to add others for JTAG/debuggers

```
vscode ➜ /workspaces/MeshCore-tests (unit_tests) $ ls -lh /dev/serial/by-id/
lrwxrwxrwx 1 root root 13 Mar 26 15:05 usb-Espressif_Systems_heltec_wifi_lora_32_v4__16_MB_FLASH__2_MB_PSRAM__9070699BE068-if00 -> ../../ttyACM1
```

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -10,15 +10,19 @@
 		}
 	},
 	"runArgs": [
-		"--privileged",
 		"--network=host",
-		"--device=/dev/bus/usb",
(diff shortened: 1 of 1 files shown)


liamcottle:
Requesting review from @jbrazio as they do most of the work on this file.

PR #2176, opened 2026-03-27
Update RAK 4631 entry in FAQ on new bootloader - removed "see note"

jschrempp:
Removed "see note" from RAK 4631 entry in FAQ. 

Looks like the text of supported hardware was copy/pasted and the "see note" text only makes sense when you are in the bootloader repo and the text is a link to more info about the RAK 4631.


oltaco:
Looks good to me 👍

PR #2177, opened 2026-03-28
RAK12035 Soil Moisture Sensor Driver

KPrivitt:
The RAK12035 Soil Moisture Sensor connects to a RAK12023 which provide three connectors for sensors.  This driver only supports one sensor attached with a default I2C address of 0x20.  Support can be added via the Set Sensor Address command (which is incorporated) for additional sensors (by changing the sensors I2C address to 0x21 or 0x22) but this is not included.

The sensor will be automatically detected and incorporated into the Telemetry and 
can be accessed via the Telemetry screen in a repeater or a companion.  The data is displayed on Channel 2.

This code also supports Calibration of a sensor via a custom version of the firmware.  See the RAK12035_SoilMoisture.cpp for instructions on the calibration procedure, which uses Channel 3.

### DIFF
--- a/lib/RAK12035/RAK12035_SoilMoisture.cpp
+++ b/lib/RAK12035/RAK12035_SoilMoisture.cpp
@@ -119,14 +119,14 @@ bool RAK12035_SoilMoisture::begin(uint8_t addr)
  *      Power = Current Dry calibration value
  *  
  *  4) Click refresh several times.  This will take a capacitance reading and if it is
- *     greater than the current Dry value it will storee it in the sensor
+ *     greater than the current Dry value it will store it in the sensor
  *     The value will bounce a little, and eventually (a few clicks)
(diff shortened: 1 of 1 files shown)


KPrivitt:
This has been tested on a RAK19007 with a RAK12023 and RAK12035 installed.  Repeater and Companion versions have been tested and operated correctly for 48 hours.

This has been tested with a RAK1901 and RAK1902 sensor which report their data on Channel 1.

The RAK12035 was calibrated and reports Soil Moisture percentage (0 - 100%) as well as the Soil Temperature.  The displayed % Moisture value changes continuously when inserted into water and taken out and dried.

KPrivitt:
Removing this PR and doing a new one based on meshcore-dev/MeshCore/dev

There were code changes that were not intended to be included.

PR #2178, opened 2026-03-28
fix: use max estimated airtime for duty cycle enforcement

fmoessbauer:
Previously the max required airtime to enter the next transmit slot was computed to be half the estimated time. However, the transmit logic uses 1.5 times the estimated time as maximum TX duration. By that, the duty cycle could still be violated as the actual transmit logic might send for longer than what we ensured to stay within the cycle.

We fix this by aligning both computations and always use 1.5 times the estimated airtime.

Xref: #817

### DIFF
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -271,19 +271,21 @@ void Dispatcher::processRecvPacket(Packet* pkt) {
   }
 }
 
-void Dispatcher::checkSend() {
-  if (_mgr->getOutboundCount(_ms->getMillis()) == 0) return;
-  
-  updateTxBudget();
-  
-  uint32_t est_airtime = _radio->getEstAirtimeFor(MAX_TRANS_UNIT);
+void Dispatcher::ensureTxDutyCycle(int tx_len) {
+  uint32_t est_airtime = _radio->getEstAirtimeFor(tx_len);
   if (tx_budget_ms < est_airtime / MIN_TX_BUDGET_AIRTIME_DIV) {
     float duty_cycle = 1.0f / (1.0f + getAirtimeBudgetFactor());
     unsigned long needed = est_airtime / MIN_TX_BUDGET_AIRTIME_DIV - tx_budget_ms;
     next_tx_time = futureMillis((unsigned long)(needed / duty_cycle));
-    return;
   }
-  
+}
+
+void Dispatcher::checkSend() {
+  if (_mgr->getOutboundCount(_ms->getMillis()) == 0) return;
+
+  updateTxBudget();
+  ensureTxDutyCycle(MAX_TRANS_UNIT);
+
   if (!millisHasNowPassed(next_tx_time)) return;
   if (_radio->isReceiving()) {
     if (cad_busy_start == 0) {
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -11,6 +11,8 @@ namespace mesh {
 #define MAX_RX_DELAY_MILLIS        32000  // 32 seconds
 #define MIN_TX_BUDGET_RESERVE_MS   100    // min budget (ms) required before allowing next TX
 #define MIN_TX_BUDGET_AIRTIME_DIV  2      // require at least 1/N of estimated airtime as budget before TX
+/// add some time to get the max allowed airtime for the est. airtime
+#define MAX_TX_AIRTIME_FOR_EST(est_airtime) (((est_airtime) * 3) / 2)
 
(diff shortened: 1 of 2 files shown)


fmoessbauer:
CCing @marcelverdult, @ViezeVingertjes, @fschrempf as you were involved in the original implementation and discussions.

I'm proposing this as draft as I did not yet perform extensive tests. While implementing I was a bit unsure where the original `tx_budget_ms < est_airtime / 2` came from as this looks like a typo which was intended to be  `tx_budget_ms < est_airtime*3 / 2`.

I further plan to move the time check further down to compute actual estimated airtime of the next packet instead of the maximal one (which is just a slight improvement).

ViezeVingertjes:
I will take a look at it too when i get home later. First thought is, if we do an estimate, can't we replace it? As we can calculate the exact airtime based on the LoRa setting and packet size there is no need for guessing right?

But might be misunderstanding the issue as i am on mobile and its hard to review on here. 😋 will get back to this. ✌️

fmoessbauer:
> I will take a look at it too when i get home later. First thought is, if we do an estimate, can't we replace it? As we can calculate the exact airtime based on the LoRa setting and packet size there is no need for guessing right?

Yes, that's what I plan to implement as well. This however requires us to pop a package from the queue and re-inject on error, whereby this logic is currently behind the `CAD busy max duration` handling which I honestly don't fully understand.

Anyways, I wanted to make the current implementation compliant with the duty cycle. Then we can work on improvements. Thanks for looking into it!

ViezeVingertjes:
> This however requires us to pop a package from the queue and re-inject on error,

I see what you mean; The more permanent solution would probably be introducing a `peek` in `StaticPoolPacketManager` and then `Dispatcher` probably needs some refactoring to straighten things out.

Any idea how much this will over-estimate (if it does)?
Because that might be a reason to go with one solution or another i think.

fmoessbauer:
> I see what you mean; The more permanent solution would probably be introducing a peek in StaticPoolPacketManager and then Dispatcher probably needs some refactoring to straighten things out.

Yes, but I would like to avoid mixing simple fixes and major refactoring. We still can later improve the code, but currently the limit used for the token-bucket algorithm is simply computed incorrectly ( div by 2 instead of mult by 1.5).

> Any idea how much this will over-estimate (if it does)?

Unfortunately we don't have telemetry data that gives us an indication, but based on meshcore analyzer data I assume the average packet size is ~120 bytes (vs. 255b * 1.5 max). ~That's a lot of tx opportunity we waste.~

> Because that might be a reason to go with one solution or another i think.

A simple solution would be to assume the worst case to make the decision to send or not and subtract that from the bucket. Then, once the transmission is done, we use adjust the bucket by the actual air time (here [1]). By that, we stay within legal limits, but still don't waste any budget. (i.e. we give back the delta between MAX and `t`). I'll push a commit that does this.

**Update**: This is actually already implemented like this. So we don't waste anything, but the changes here just fix the case we start transmitting close to 0 budget.

Apart from that, I'm wondering if the current implementation of the token bucket algorithm actually matches legal limits, especially considering the case that we spend all TX time at the beginning of the hour and then have to wait for the hour to pass (currently we already re-fill during that period). But that's another topic and this can hopefully be fixed by just adjusting the initial budget.

[1] https://github.com/meshcore-dev/MeshCore/blob/31007d9790d247f1f203a0c2222da116a019878f/src/Dispatcher.cpp#L87

ViezeVingertjes:
Keeping it small is fine too indeed. 

As for the calculation, in here duty cycle is a sliding window (for whole EU i suppose) so that should be right.

Been a while, but pretty sure its implemented that way, right?

fmoessbauer:
> As for the calculation, in here duty cycle is a sliding window (for whole EU i suppose) so that should be right.

The current implementation (Token Bucket) only guarantees the following properties:

- long term average stays below threshold (e.g. 10% on a 10% duty cycle) (due to refill rate)
- short term average (1h in our case) stays below `2*threshold` (remaining budget + refill)

The second property hereby is the problematic one, which becomes obvious when considering the case where we start with a full bucket: We can transmit until the bucket is empty and due to the refilling can transmit the same amount of data until the hour ends. By that, we used the double duty cycle during that exact hour.

ViezeVingertjes:
I haven’t seen any repeaters exceed 10% so far, not even close. This could be due to LoRa’s half duplex limitations though, as for every packet a repeater hears and retransmits, it likely misses many others, so it often cannot hear packets it should repeat.

Your explanation sounds plausible though, so it would still be good to see it fixed.

PR #2182, opened 2026-03-29
docs: add Swiss Mesh Pro to number allocations table

ssozonoff:


ripplebiz:
We have re-adjusted the data-type ranges. So, please use an ID starting from 0x0100

ssozonoff:
ok thats fixed thx

ripplebiz:
Could you add some info in the Contact column about where to get info about the project?

ssozonoff:
There is not much to add at the moment. I have some prototype stuff (weather station) which is currently running and was/is using Group Data before you guys further refined the protocol. Hence its not currently conforming to the official protocol and its my understanding that in order to confirm to the protocol I need to register and ID in the allocation table.

liamcottle:
> There is not much to add at the moment. I have some prototype stuff (weather station) which is currently running and was/is using Group Data before you guys further refined the protocol. Hence its not currently conforming to the official protocol and its my understanding that in order to confirm to the protocol I need to register and ID in the allocation table.

For now, I would recommend using any ID within the `0xFF00` to `0xFFFF` range. These are reserved for testing and development use. You have access to ~256 unique IDs in this range, so the potential for clashes with other users in your area is probably very low at this time.

Once you have a production app ready, for example something being released to the Apple App Store, or Google Play Store, you could then request a unique ID to be reserved by us to avoid clashes with other users/apps on public mesh networks.

But since you're just tinkering/testing, you don't really need a globally unique ID at this time :)

We will need to update the docs so this is more apparent.

ssozonoff:
OK I will just use whatever.

PR #2185, opened 2026-03-29
Add JOYSTICK_ROTATION build flag for joystick-equipped boards

sneznaovca:
## Motivation                                                                                                                                                                                                                                                        
  
  Some 3D-printed enclosures for the Wio Tracker L1 eInk (e.g.                                                                                                                                                                                                         
  [this one on Printables](https://www.printables.com/model/1420534-seeed-wio-tracker-l1-e-ink-enclosure))
  mount the display beside the joystick rather than above it. In this                                                                                                                                                                                                  
  orientation the physical up/down buttons align with left/right navigation                                                                                                                                                                                            
  on screen, making the default joystick mapping unusable.                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                       
  I tried working around this with `DISPLAY_ROTATION` but the current UI                                                                                                                                                                                               
  does not handle portrait orientation well — that would need more involved
  changes. This PR takes the simpler approach of remapping the joystick                                                                                                                                                                                                
  input to match the physical orientation, which is sufficient for normal
  use. (Happy to also look into proper portrait UI support if that's of                                                                                                                                                                                                
  interest to maintainers.)                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                       
  ## Changes                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                       
  Adds a `JOYSTICK_ROTATION` build flag following the same `0`–`3`                                                                                                                                                                                                     
  convention as `DISPLAY_ROTATION`:
                                                                                                                                                                                                                                                                       
  | Value | Effect |                                                                                                                                                                                                                                                   
  |-------|--------|
  | `0` (default) | no change — left/right navigate left/right |                                                                                                                                                                                                       
  | `1` | 90° CW — up/down navigate right/left |                                                                                                                                                                                                                       
  | `2` | 180° — left/right are swapped |
  | `3` | 270° — up/down navigate left/right |                                                                                                                                                                                                                         
                                                            
  Variants that define `JOYSTICK_UP`/`JOYSTICK_DOWN` pins                                                                                                                                                                                                              
  (`wio-tracker-l1`, `gat562_mesh_tracker_pro`, `gat562_30s_mesh_kit`)
  now instantiate `joystick_up` and `joystick_down` `MomentaryButton`                                                                                                                                                                                                  
  objects so they are available when rotation 1 or 3 is selected.                                                                                                                                                                                                      
  A compile-time `#error` catches misconfigured builds (rotation 1/3                                                                                                                                                                                                   
  without the required pin definitions).                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                       
  To enable on Wio Tracker L1 eInk, uncomment in `platformio.ini`:                                                                                                                                                                                                     
  ```ini                                                                                                                                                                                                                                                               
  -D JOYSTICK_ROTATION=1                                                                                                                                                                                                                                               
 ```
                                                           
## Testing

  Tested on physical Wio Tracker L1 eInk hardware with the enclosure                                                                                                                                                                                                   
  linked above.

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -29,6 +29,11 @@
   #define PRESS_LABEL "long press"
 #endif
 
+#if defined(JOYSTICK_ROTATION) && (JOYSTICK_ROTATION == 1 || JOYSTICK_ROTATION == 3) \
+    && (!defined(JOYSTICK_UP) || !defined(JOYSTICK_DOWN))
+  #error "JOYSTICK_ROTATION 1/3 requires JOYSTICK_UP and JOYSTICK_DOWN pin definitions"
+#endif
+
 #include "icons.h"
 
 class SplashScreen : public UIScreen {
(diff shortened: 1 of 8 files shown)


ripplebiz:
A much simpler way would be in variant.h, just put these around the JOYSTICK_ pin defs:
#ifndef JOYSTICK_LEFT
#define JOYSTICK_LEFT           PIN_BUTTON4
#endif

Then in your custom platformio.ini just have:
   -D JOYSTICK_LEFT  PIN_BUTTON2
etc
ie. just remap the pins

PR #2190, opened 2026-03-30
Add missing methods in ESPNOWRadio()

brianwiddas:
ESP-NOW radios (ie, Generic_ESPNOW_* variants) do not compile due to missing methods

Changes in January 2026 (019bbf74) to add additional stats (receive errors) to CMD_GET_STATS was not implemented in the ESPNOWRadio() class

Changes in March 2026 (9a95e25e) to add setRxBoostedGainMode to all devices rather than just SX1262/SX1268 were not applied to the ESPNowRadio() driver

Specifically, this change adds the following to ESPNOWRadio()
* getPacketsRecvErrors()    - always returns 0
* getRxBoostedGainMode()    - always returns false
* setRxBoostedGainMode()    - does nothing

### DIFF
--- a/src/helpers/esp32/ESPNOWRadio.h
+++ b/src/helpers/esp32/ESPNOWRadio.h
@@ -4,10 +4,10 @@
 
 class ESPNOWRadio : public mesh::Radio {
 protected:
-  uint32_t n_recv, n_sent;
+  uint32_t n_recv, n_sent, n_recv_errors;
 
 public:
-  ESPNOWRadio() { n_recv = n_sent = 0; }
+  ESPNOWRadio() { n_recv = n_sent = n_recv_errors = 0; }
 
   void init();
   int recvRaw(uint8_t* bytes, int sz) override;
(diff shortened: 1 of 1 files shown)


romainreignier:
@brianwiddas Thanks for the fix, I have just faced the same issue using the main branch and was about to provide a fix too.
But, out of curiosity, did you manage to use the `Generic_ESPNOW_comp_radio_usb` target? I can build it and flash it to an ESP32C3 but I cannot communicate with it using meshcore-cli:
```
meshcli -s /dev/ttyACM0 -d "" infos
```

brianwiddas:
This is a copy of my Discord post, for reference

It seems to be down to the definitions in the platformio.ini file. I had the same problem, using a Xiao S3.

The default platformio.ini (`variants/generic_espnow/platformio.ini`) has the board set up for `esp32-c3-devkitm-1`, and I had changed it to `esp32-s3-devkitm-1` (which is commented out below it) to compile for the S3 chip. However, that board defintion does not include the necessary config to get the serial port up and running. You also need

```
  -D ARDUINO_USB_MODE=1
  -D ARDUINO_USB_CDC_ON_BOOT=1
```
Which are commented out in the platformio.ini file

I also found I needed to unplug and replug the board from my laptop once I'd flashed it, in order to have it accept the connection (in my case, from the MeshCore browser app) 

The other thing that worked, for me at least, was changing the board type to what it actually was. In this case, `seeed_xiao_esp32s3`. That included a board definition file which did the same thing as uncommenting the definitions above

romainreignier:
Thanks a lot @brianwiddas !
These lines were also needed to use ESP32C3 USB Serial port.

PR #2191, opened 2026-03-30
Added Sensor RAK12035 Soil Moisture and Temperature

KPrivitt:
A RAK12035 Soil Moisture Sensor will be automatically detected and incorporated into the Telemetry and
can be accessed via the Telemetry screen in a repeater or a companion. The sensor data is displayed on Channel 2.

This code also supports Calibration of a sensor via a custom version of the firmware. See the RAK12035_SoilMoisture.cpp for instructions on the calibration procedure, which uses Channel 3.  Calibration must be done to a sensor; each sensor will generate different capacitance output values.  Calibration will ensure the reported Moisture will return a value between 0 and 100%.

The RAK12035 Soil Moisture Sensor connects to a RAK12023 which provide three connectors for sensors. This driver only supports one sensor attached with a default I2C address of 0x20. Support can be added for additional sensors via the Set Sensor Address command (which is incorporated).  The sensors attached must all have a different I2C address (this can be done by changing an additional sensors I2C address to 0x21 or 0x22) and this code will need to include the initialization of each of these sensors (but this is not included).

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -134,6 +134,7 @@ build_flags =
   -D ENV_INCLUDE_VL53L0X=1
   -D ENV_INCLUDE_BME680=1
   -D ENV_INCLUDE_BMP085=1
+  -D ENV_INCLUDE_RAK12035=1
 lib_deps =
   adafruit/Adafruit INA3221 Library @ ^1.0.1
   adafruit/Adafruit INA219 @ ^1.2.3
--- a/src/helpers/sensors/RAK12035_SoilMoisture.cpp
+++ b/src/helpers/sensors/RAK12035_SoilMoisture.cpp
@@ -1,0 +1,528 @@
+/*----------------------------------------------------------------------*
+ * RAK12035_SoilMoistureSensor.cpp - Arduino library for the Sensor     *
+ * version of I2C Soil Moisture Sensor version from Chrirp              *
+ * (https://github.com/Miceuz/i2c-moisture-sensor).                     *
+ *                                                                      *
+ * Ingo Fischer 11Nov2015                                               *
+ * https://github.com/Apollon77/I2CSoilMoistureSensor                   *
+ *                                                                      *    
+ * Ken Privitt 8Feb2026                                                 *
+ * Adapted for MeshCore Firmware Stack                                  *
+ *                                                                      *                        
+ * MIT license                                                          *
+ *                                                                      *
+ * This file contains a collection of routines to access the            *
+ * RAK12035 Soil Moisture Sensor via I2C. The sensor provides           *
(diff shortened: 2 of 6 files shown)


KPrivitt:
This has been tested on a RAK19007 with a RAK12023 and a single RAK12035 (default I2C 0x20) sensor installed. Repeater and Companion versions have been tested and operated correctly for 48 hours.

This has been tested with a RAK1901 and RAK1902 sensor installed which report their data on Channel 1.  This sensor reports its data on Channel 2 (there is no conflict)

A RAK12035 sensor was calibrated, and it now reports Soil Moisture percentage (0 - 100%) as well as the Soil Temperature. The displayed % Moisture value changes continuously when inserted into water and taken out and dried.  The Temperature changes if it is placed in warm or cold water or held in your hand to warm it up.

recrof:
you're using main as source branch, can you change it to dev?

src/helpers/sensors/RAK12035_SoilMoisture.cpp
```diff
@@ -0,0 +1,528 @@
...
+ *                                                                      *
+ *----------------------------------------------------------------------*/
+
+#include "RAK12035_SoilMoisture.h"
+
+#include <Wire.h>
+
+#if MESH_DEBUG && ARDUINO
+  #include <Arduino.h>
+  #define MESH_DEBUG_PRINT(F, ...) Serial.printf("DEBUG: " F, ##__VA_ARGS__)
```
liamcottle, line 26: The `MESH_DEBUG_` logging functions are defined in `src/MeshCore.h`, we probably shouldn't redefine them in a sensor class.

liamcottle (changes requested):

platformio.ini
```diff
@@ -134,6 +134,7 @@ build_flags =
   -D ENV_INCLUDE_VL53L0X=1
   -D ENV_INCLUDE_BME680=1
   -D ENV_INCLUDE_BMP085=1
+  -D ENV_INCLUDE_RAK12035=1
```
liamcottle, line 137: The `ENV_INCLUDE_RAK12035` will probably need to be moved to the RAK4631 variant platformio.ini instead of global sensors, as it causes compilation errors for other devices, due to RAK WisBlock specific pin defines not existing.

```
src/helpers/sensors/RAK12035_SoilMoisture.cpp:246:18: error: 'WB_IO2' was not declared in this scope
```

You can see these errors in the automated tests run against this PR.
https://github.com/meshcore-dev/MeshCore/actions/runs/23731296245/job/69516637482?pr=2191

liamcottle (changes requested):

PR #2202, opened 2026-03-30
Move sendNodeDiscoverReq() from private to public to enable post-boot neighbor discovery

terminalvelocity23:
Currently `sendNodeDiscoverReq()` is private, which prevents calling it from setup() after reboot. This limits the ability to proactively discover neighbors without waiting for CLI command.

This simple change moves the method to public, allowing:
- Automatic neighbor discovery after reboot (faster network integration)
- External callers to refresh neighbor table on demand

No functional changes, just visibility.

### DIFF
--- a/examples/simple_repeater/MyMesh.h
+++ b/examples/simple_repeater/MyMesh.h
@@ -119,7 +119,6 @@ class MyMesh : public mesh::Mesh, public CommonCLICallbacks {
 #endif
 
   void putNeighbour(const mesh::Identity& id, uint32_t timestamp, float snr);
-  void sendNodeDiscoverReq();
   uint8_t handleLoginReq(const mesh::Identity& sender, const uint8_t* secret, uint32_t sender_timestamp, const uint8_t* data, bool is_flood);
   uint8_t handleAnonRegionsReq(const mesh::Identity& sender, uint32_t sender_timestamp, const uint8_t* data);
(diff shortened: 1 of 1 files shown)


terminalvelocity23:
Tested on Heltec V4 — after reboot repeater sends discovery request successfully.

liamcottle:
Hey, thanks for the PR.

Happy merge this to make the method public so you can use it in your custom firmware etc, but I don't think we would be willing to accept a follow up PR that makes nodes perform discovery requests on reboot automatically, the way you intend to use it.

This would just lead to more unnecessary traffic on the mesh. Repeater admins need to login to the node and fetch the neighbours list when they want to see it, so I don't see an issue with them tapping the button to send a discovery request while they're there.

If we have repeater nodes boot looping due to bad solar setups etc, it could lead to packet spam from repeated discovery requests over and over. Although these discovery requests are zero hop, it could still lead to packet loss for other network users.

terminalvelocity23:
@liamcottle No, I've had no intention to do that. It's intended for those who can write the neccessary code by himself and, ostensibly, properly configure his hardware.

liamcottle:
> @liamcottle No, I've had no intention to do that. It's intended for those who can write the neccessary code by himself and, ostensibly, properly configure his hardware.

Good stuff, you should be good to go then. Has been merged :)

PR #2205, opened 2026-03-30
Heltec v3 without display repeater

MeetTheTree:
Added a heltec v3 repeater variant **without display**. Can be useful if the display is broken or for saving of battery.

### DIFF
--- a/variants/heltec_v3/target.h
+++ b/variants/heltec_v3/target.h
@@ -8,20 +8,14 @@
 #include <helpers/AutoDiscoverRTCClock.h>
 #include <helpers/SensorManager.h>
 #include <helpers/sensors/EnvironmentSensorManager.h>
-// ---------------------------------------------------------
-// Display handling
-// If DISPLAY_CLASS is defined AND not NullDisplayDriver,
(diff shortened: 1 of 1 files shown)


IoTThinks:
I think this may not be required.
When the display is not avaialble, I2C will fail and no power will be used.

MeetTheTree:
Many thanks for the comment. I have two heltec v3 with broken display / no display. After installing standard firmware I2C fails and the nodes are stuck, it is not possible to connect via console or browser to configure them. This pull request solves the problem and the nodes can be configured via cmdline or browser. Best regards.

IoTThinks:
Ok, it is fine.
You may want to do a PR to merge to dev branch instead.

MeetTheTree:
Hi, thanks for the feedback, i will try to create a PR to merge to dev branch. Best regards.

MeetTheTree:
Closed and submitted a new PR to merge to the dev branch under [https://github.com/meshcore-dev/MeshCore/pull/2250](https://github.com/meshcore-dev/MeshCore/pull/2250)

Issue #2119, opened 2026-03-21
[FEATURE REQUEST] Increased BLE Transmit Strength for Repeaters using BLE during OTA


KPrivitt:
Can the default value be changed to 8 to increase the BLE signal strength for repeaters that utilize the nRF52840 during an OTA firmware update?

The nRF52840 supports a transmit power level of up to 8 maximum.   
bluefruit.h
    // Supported tx_power values depending on mcu:
    // - nRF52832: -40dBm, -20dBm, -16dBm, -12dBm, -8dBm, -4dBm, 0dBm, +3dBm and +4dBm.
    // - nRF52840: -40dBm, -20dBm, -16dBm, -12dBm, -8dBm, -4dBm, 0dBm, +2dBm, +3dBm, +4dBm, +5dBm, +6dBm, +7dBm and +8dBm.

In the file NRF52Board.cpp OTA is initialized with a BLE power of 4 {line 328} 

  // Set max power. Accepted values are: -40, -30, -20, -16, -12, -8, -4, 0, 4
  Bluefruit.setTxPower(4);

CODE CHANGE:

#if defined(NRF52832_XXAA)
   Bluefruit.setTxPower(4);
#elif defined(NRF52840_XXAA) || defined(NRF52833_XXAA)
  Bluefruit.setTxPower(8);
#endif

BLE is normally disabled in a repeater.  In OTA mode BLE is enabled to allow for the transfer of the file during the "Over the Air" firmware update.  Updating a repeater using BLE OTA eliminates the need to physically connect to the repeater (bring it down, reflash and return it, which can be "risky" on a highly slanted roof).  The increased power can increase the range of the BLE connection (it may be high up on your roof).

The power utilization impact of the increased transmit signal is minimal since this mode is only used during a Firmware update.  A low BLE signal level can slow the speed of the data transfer, an increase in transmit signal can improve speed and the success rate of an OTA update.

PR #2214, opened 2026-04-01
Increased BLE TXPower during Repeater OTA for nRF52833 or nRF52840 

KPrivitt:
[FEATURE REQUEST] Increased BLE Transmit Strength for Repeaters using BLE during OTA #2119

The nRF52840 supports a transmit power level of up to +8dBm maximum.
// - nRF52840: -40dBm, -20dBm, -16dBm, -12dBm, -8dBm, -4dBm, 0dBm, +2dBm, +3dBm, +4dBm, +5dBm, +6dBm, +7dBm and **+8dBm**.

This changes the BLE Transmit power from +4dBm to +8dBm only for repeaters using an nRF52833 or nRF52840.  For other devices it remains at the previous +4dBM.

BLE is normally disabled in a repeater. In OTA mode BLE is enabled to allow for the transfer of the file during the "Over the Air" firmware update. Updating a repeater using BLE OTA eliminates the need to physically connect to the repeater (bring it down, reflash and return it, which can be "risky" on a highly slanted roof). The increased power can increase the range of the BLE connection (it may be high up on your roof).

The power utilization impact of the increased transmit signal is minimal since this mode is only used during a Firmware update. A low BLE signal level can slow the speed of the data transfer, an increase in transmit signal can improve speed and the success rate of an OTA update.


KPrivitt:
Please approve.

oltaco:
To be clear, this would only change tx power of advertising the bootloader DFU service, the change still has to be made to the bootloader.

I have the change ready to go and I'm hoping to push a new version of the OTAFIX Bootloader this weekend.

KPrivitt:
I will look for the new bootloader.  I have updated to the UF2 Bootloader 0.9.2-OTAFIX2.1-BP1.2 for all of my repeaters except for the one up on the roof which is still at 1.14.0 and the old bootloader.

For testing should I bring down the repeater, load your new bootloader but keep the 1.14.0 version.  Reinstall it up on the roof and update to 1.14.1 via OTA.

I can use the BLE Serial nRF52 tool to verify if the Power Level is now set to 8.  But, as you stated this is only the advertising power level, how can I verify the transfer power level?
(there is a power "bar" that is shown during the transfer, is there a better place to verify this?)

recrof:
> For testing should I bring down the repeater, load your new bootloader but keep the 1.14.0 version. Reinstall it up on the roof and update to 1.14.1 via OTA.
> 
you can flash the bootloader via DFU app as well - there is minimal risk, as it uses app partition to write new bootloader, crc checks it and only after that replaces old bootloader. you will then need to send repeater zip image via DFU, as it was erased by bootloader update.

KPrivitt:
oltaco,
Is the new bootloader ready?
What is the new version number I should be looking for?

recrof:
> oltaco, Is the new bootloader ready? What is the new version number I should be looking for?

https://github.com/oltaco/Adafruit_nRF52_Bootloader_OTAFIX/releases/tag/0.9.2-OTAFIX2.2-BP1.3

KPrivitt:
**Testing Completed:**

**1)** Loaded new bootloader (over USB through copy to attached drive)
UF2 Bootloader 0.9.2-OTAFIX2.2-BP1.3 lib/nrfx (v2.0.0) lib/tinyusb (0.12.0-145-g9775e7691) lib/uf2 (remotes/origin/configupdate-9-gadbb8c7)
Model: WisBlock RAK4631 Board
Board-ID: WisBlock-RAK4631-Board
Date: Apr 13 2026
SoftDevice: S140 6.1.1

**2)** With 1.14.1 467959c started OTA and verified BLE Power = 4 (-56db) with BLE Serial nRF52 iOS App
20260413_204923000_iOS

**3)** Generated custom compiled Firmware with PR #2214 using Visual Studio Code and installed with VSCode (USB)

**4)** Started OTA verified BLE Power = 8 (-39db) with BLE Serial nRF52 iOS App
20260413_211226000_iOS

**5)** Started and successfully completed OTA update to official release 1.14.1 ~5min.

**6)** With reinstalled 1.14.1 467959c started OTA and verified BLE Power = 4 (-41db) with BLE Serial nRF52 iOS App
20260413_212640000_iOS

**Successful testing and verification. This is ready to start acceptance workflow.**

recrof:
there is another issue that needs to be addressed - we only want to enable 8dBm if OTAFIX 2.2 bootloader is installed. else there is big chance that user will be able to connect and switch board to BLE DFU bootloader mode, however it will be set to 4dBm, so link will degrade - that might mean inability to update the firmware.

KPrivitt:
I'm not sure how to check for the version of the bootloader.  I believe there is a CLI to get the version of the bootloader and that could be added to the conditional check that is already there.  Is there someone that would know what that "variable" is to check to verify it is >= to 0.9.2-OTAFIX2.2-BP1.3?  With this added then this should be ready.

One good thing is the I was able to verify that the power increase worked.

recrof:
> Is there someone that would know what that "variable" is to check to verify it is >= to 0.9.2-OTAFIX2.2-BP1.3?

I believe that @oltaco will be able to help with this.

KPrivitt:
I found a function to get the Bootloader version in the same file NRF52Board.cpp 

But I'm not sure which part of the retrieved data will indicate the current version
(I assume the current version const char* ver should be compared with the compatible version which is ??):

bool NRF52Board::getBootloaderVersion(char* out, size_t max_len) {
    static const char BOOTLOADER_MARKER[] = "UF2 Bootloader ";
    const uint8_t* flash = (const uint8_t*)0x000FB000; // earliest known info.txt location is 0xFB90B, latest is 0xFCC4B

    for (uint32_t i = 0; i < 0x3000 - (sizeof(BOOTLOADER_MARKER) - 1); i++) {
        if (memcmp(&flash[i], BOOTLOADER_MARKER, sizeof(BOOTLOADER_MARKER) - 1) == 0) {
            const char* ver = (const char*)&flash[i + sizeof(BOOTLOADER_MARKER) - 1];
            size_t len = 0;
            while (len < max_len - 1 && ver[len] != '\0' && ver[len] != ' ' && ver[len] != '\n' && ver[len] != '\r') {
                out[len] = ver[len];
                len++;
            }
            out[len] = '\0';
            return len > 0; // bootloader string is non-empty
        }
    }
    return false;
}

KPrivitt:
I'm not sure your feedback about requiring a check of the revision of the bootloader is correct.  oltaco stated that "To be clear, this would only change tx power of advertising the bootloader DFU service, the change still has to be made to the bootloader."

In order to have the new firmware transferred by the bootloader at an increased power the bootloader would also need to increase the power.  So, he was indeed referring to an older bootloader.  The probability of the transfer failing, is exactly the same if it is advertised at a lower power or a higher power.

I know I attempted an OTA with the prior fixed bootloader and it failed, (which is why I posted this PR). But with the fixed/updated bootloader it no longer would hang, it fell back to DFU and allowed me to reappempt the OTA.

An OTA update is still fragile with a lower tx power and any improvement is better than the current situation. A bootloader check is not necessary. oltaco do you agree?

oltaco:
> An OTA update is still fragile with a lower tx power and any improvement is better than the current situation. A bootloader check is not necessary. oltaco do you agree?

I think a bootloader check is probably still a good idea, because in marginal situations where setting BLE tx power to +8 gives you enough reception that you can start the ota process there would be the potential that when it reboots into the BLE bootloader and the tx power is set to +4 you might not be able to complete the update. I know it would be a very small chance and it would be extremely bad luck but it would be best to avoid that situation.

Starting from OTAFIX 2.2 (released last week) the BLE tx power is now set to +8 for NRF52840 devices, so a bootloader version check could test for OTAFIX2.2 to be safe.

edit: corrected OTAFIX version number

KPrivitt:
You are asking me to make significant changes (this is literally a single bit change to the firmware) in order to degrade BLE performance so that a condition what will not and cannot be avoided is addressed.  A much simpler change is to cancel and close this PR (the BLE power will remain low).

Your request would require scanning the bootloader string, pulling out the version and comparing it then reducing the tx power.

Repeaters have no clue what power level is being transmitted to them, the RF environment (walls, roofs, etc) have a much larger effect than a 4dB power change, in fact on a low signal connection (further away) simply turning your phone 90o has more effect and can cause the OTA to fail.  BLE is designed for this power level, the fact that other things can cause an OTA failure has no bearing on the power level being used.

Your request also seems to violate the guidelines:


• Smaller PRs are better

• Elaborate or large-scale revisions can delay the review process. Keep changes limited to what is absolutely necessary for the task or issue.

• Consistency and necessity matter more than general elaboration

• Make changes that align with the existing codebase’s style and purpose the least intrusive way.

This change as it stands will improve some future OTA updates; not making the change will degrade some future OTA updates and both still have the possibility of failing.

1nerdherder:
It seems there is a good opportunity to improve the resilience of the OTA updates here.
Improving the power output in one direction (from node to mobile device/laptop) can't hurt.
What I would recommend is that if the upload fails, and the node falls back into the DFU mode - that it only remain in DFU mode for a set period of time (let's say, 5 minutes), after which it goes back to running the firmware that is still in the node.

recrof:
> It seems there is a good opportunity to improve the resilience of the OTA updates here. Improving the power output in one direction (from node to mobile device/laptop) can't hurt. What I would recommend is that if the upload fails, and the node falls back into the DFU mode - that it only remain in DFU mode for a set period of time (let's say, 5 minutes), after which it goes back to running the firmware that is still in the node.

you can't return to firmware after failed update - the flash is marked "dirty" and it needs to be reflashed properly. btw you're talking about bootloader in wrong repo.. issues/feature reqs of bootloader are best to be suggested in oltaco's otafix github issues.

KPrivitt:
With the new 1.15.0 code I was able to test your theory.  My last testing replaced my code with the standard 1.14.1 (tx power = 4) so the BLE advertisement was at low power.  The bootloader was at the higher power.

I intentionally started the process at a further distance to the repeater.  OTA was started (it has a very high SNR and logs in quickly as expected). OTA was started and gave an IP address, the advertisement was at the lower power.  I used BLE Tools selected 1.15.0 and 4631_OTA was advertised and connected (again this was at low power on the repeater).  I clicked start but it failed (it hung and the transfer never started) the bootloader was at the higher power.  I retried and it failed again.  This can be expected with a marginal BLE link.

I then moved closer (half the distance) and restarted the procedure.  But this time it was advertising 4631_DFU, so the first attempts had fallen back to DFU mode and the boatloader (at the higher power) was controlling the update which completed then successfully.

I believe this shows that with a lower signal strength a failure can occur and does occur but this is normal and should be expected. Having a higher power does partially compensate for a degraded link and is always a good thing and can only improve the outcome. Intentionally degrading the link will degrade the outcome.  The fallback to DFU allows for a retry and the lower strength can be addressed by getting closer and a successful OTA can be achieved.

Please approve this PR, otherwise I will be withdrawing and closing this PR.

By increasing the probability (even in the presence of failure) makes it easier for users to update and can improve the acceptance of MeshCore and most importantly it will reduce **injuries** from users that must go up to the roof to retrieve their repeaters.

recrof:
> I believe this shows that with a lower signal strength a failure can occur and does occur but this is normal and should be expected.

On the contrary. User expects that if the link quality at the start is reported as favorable, it will work whole time. If you set it to 8dBm and bootloader will still be 4dBm, this will not be the case.

KPrivitt:
Well my test shows you are wrong.  I had signal to connect and it failed.  A marginal signal can complete successfully and it can also fail.

I know because had the OTA fail multiple times even before I proposed this change. So a connection does NOT means a successful update.

PR #2216, opened 2026-04-01
Fix for default preset EU/UK (Narrow)

formtapez:
In #1820 the default preset was changed to `EU/UK (Narrow)`, but without setting `LORA_CR` to 8.
The coding rate actually defaults to 5, so that it doesn't match the preset.
These settings are compatible to each other, but the lower coding rate is not very reliable.


Confituurke:
**+1 for this**

Right now, a lot of tooling classifies nodes on stock defaults as a **custom** preset (`869.618 / 62.5 / SF8 / CR5`) instead of matching EU/UK (Narrow) (`869.618 / 62.5 / SF8 / CR8`), which makes maps and presets misleading.

For a mesh we can rely on, every node on that channel should use the same LoRa modem settings (frequency, bandwidth, spreading factor, and coding rate - not only the first three).

We are already seeing fragmented/degraded meshes in Western Europe because of this split, and it was hard to debug because a CR mismatch is easy to miss next to the correct frequency, BW, and SF.

I am surprised this was not raised sooner.

Unless there is evidence that CR5 is deliberately better for that EU narrow air setting, we should ship this fix in the next release, otherwise the honest alternative is to change the official EU/UK (Narrow) preset (and API copy) to CR5 so documentation, apps, and firmware all match + sending out a message to the whole MeshCore community.

recrof:
> For a mesh we can rely on, every node on that channel should use the same LoRa modem settings (frequency, bandwidth, spreading factor, and coding rate - not only the first three).

How is lower CR degrading your performance? CR 5 and CR8 are compatible. there is no asymmetry introduced.

> Unless there is evidence that CR5 is deliberately better for that EU narrow air setting

There is. we save around 40% of airtime, just look at the LoRa calculator: https://www.semtech.com/design-support/lora-calculator
set payload to 200 bytes, and compare 62.5/8/8 vs 62.5/8/5.. there is noticable difference.

Confituurke:
So should we change the default to be CR5 instead of 8?

formtapez:
With CR5 you loose more packets due to transmission errors, which are not correctable anymore. (With CR8 they are)
When i try to reach a repeater far away, i have no chance with CR5. Switching to CR8 solved the problem and the link is stable.

Confituurke:
@recrof I understand and agree to your standpoint on airtime efficiency. However, the core issue here is simply aligning the preset definition and the firmware default and vice versa. I guess; to either fix the firmware default to CR8 (this PR), or officially update the EU/UK (Narrow) preset to CR5.
Updating the default preset to EU/UK Narrow was the subject of the first PR #1820

---

Besides that, I am open to discuss the pros and cons of choosing the best standard between CR8 and CR5.
Worth noting for example: CR5 provides virtually no forward error correction, while CR8 does. For accessibility and reliability to MeshCore, CR8 makes more sense as the default - users with a well-established mesh can always opt to switch to CR5 for the airtime savings.

---

What do we do with this PR?

efiten:
I would set SF8 & CR8 like it was before, so it's still clear to everyone. In my case, for a big part of Belgium, switching to CR5 would lose connections, because the nodes are not dense enough.

Issue #917, opened 2025-10-11
Set WiFi ssid and password via serial cli


mikecarper:
Make it not need a build, allowing for a 3rd type to be used from the web flasher

recrof:
@mikecarper I see you pretty active on the issues department, how about some PRs? :)

mikecarper:
Point me where this change would go and I'll give it a go

mrlmuk:
Surely this isn't possible untill the firmware for usb/bt/wifi are combined onto one? I think that's planned for the future.

MarekWo:
I'm a Heltec V4 user running a companion radio over WiFi/TCP. Currently the only way to keep my SSID/password through a firmware upgrade is a local `git stash` → `git pull` → `git stash pop` → `pio run -t upload` workflow every release, since the credentials live as `-D WIFI_SSID` / `-D WIFI_PWD` build flags in `variants/heltec_v4/platformio.ini`. Runtime configuration would be a big quality-of-life improvement, especially for users flashing prebuilt binaries from the web flasher.

Worth noting that two open PRs already implement most of this:
- #2217 (dz0ny) — persists `wifi_ssid` / `wifi_pwd` to `NodePrefs` + `DataStore`, exposed via CLI. Linked to this issue.
- #2720 (NickDunklee) — smaller CLI-rescue approach (`wifi_ssid`, `wifi_pwd`, `wifi_commit`, `wifi_clear`).

It would be great if a maintainer could review either of them.

liamcottle:
Now that we have companion CLI, this is likely to be implemented by https://github.com/meshcore-dev/MeshCore/pull/3357

MarekWo:
Thanks @liamcottle, that's great news — and it makes sense now that `CMD_RUN_CLI_COMMAND` is in.

For anyone landing here later: #3357 adds `set wifi.ssid` / `set wifi.pwd` / `set wifi.enabled` / `set wifi.clear` (plus `get wifi.ssid|enabled|status|ip`), persisted to `NodePrefs`, with the build-time `WIFI_SSID` / `WIFI_PWD` flags kept as a fallback. Reachable both from the serial rescue CLI and from the app over the companion CLI.

Happy to test it on a Heltec V4 companion over WiFi/TCP once it lands in `dev`.

Still relevant, so please keep this one off the stale pile. 🙂

PR #2217, opened 2026-04-01
feat: Add companion WiFi CLI

dz0ny:
This is a small PR that enables configuring wifi over CLI. 

ref: https://github.com/meshcore-dev/MeshCore/issues/917


petrkr:
Maybe move static char from stack to heap, because stack is already very full by contacts, messages and channels and it is on almost its limits, so another 100B is too much I guess while most of wifi ssid/psk would be only 30B on heap if you will use it dynamic allocated.

What you think?

Issue #2051, opened 2026-03-16
[Feature Request / Bug] int.thresh is hardcoded to 0 in Companion firmware— expose _prefs.interference_threshold


jirogit:
Summary

In Companion firmware, getInterferenceThreshold() unconditionally returns 0, ignoring _prefs.interference_threshold entirely. As a result, RSSI-based carrier sensing (LBT: Listen Bore Transmission) is effectively disabled on Companion nodes, with no way for users to enable it — even though the preference field exists in NodePrefs and set int.thresh is fully functional on Repeater and Room Server firmware.

Root Cause

In examples/companion_radio/MyMesh.cpp:

int MyMesh::getInterferenceThreshold() const {
  return 0; // disabled for now, until currentRSSI() problem is resolved
}

The comment references an unresolved currentRSSI() problem. By contrast, the Repeater implementation in examples/simple_repeater/MyMesh.h correctly reads from _prefs:

int getInterferenceThreshold() const override {
  return _prefs.interference_threshold;
}

The Companion firmware already holds a NodePrefs _prefs instance (declared in MyMesh.h) which contains the interference_threshold field — it is simply never used in this path.

Current Behavior

Firmware
getInterferenceThreshold()
User-configurable

Repeater
returns _prefs.interference_threshold
✅ via set int.thresh

Room Server
returns _prefs.interference_threshold
✅ via set int.thresh

Companion
always returns 0
❌ hardcoded

Proposed Fix

Step 1 — examples/companion_radio/MyMesh.cpp

// Before:
int MyMesh::getInterferenceThreshold() const {
  return 0; // disabled for now, until currentRSSI() problem is resolved
}

// After:
int MyMesh::getInterferenceThreshold() const {
  return _prefs.interference_threshold;
}

Step 2 — expose set int.thresh / get int.thresh to Companion (firmware side)

Currently set int.thresh is handled only in CommonCLI.cpp and is not accessible from Companion firmware. As a first step, adding int.thresh handling to the Companion's configuration command path (serial/BLE CLI) would allow users to configure it without recompiling — consistent with how it is already handled in Repeater and Room Server firmware.

Step 3 — mobile app UI (separate task, separate repository)

Once the firmware accepts set int.thresh on Companion nodes, exposing interference_threshold as a field in the MeshCore iOS/Android app's radio settings screen would make this accessible to the widest range of users. As this requires changes to the app repository, it is noted here as a follow-up task rather than part of this issue.

[...]

PR #2218, opened 2026-04-01
Add Japan LoRa LBT compliance (ARIB STD-T108) with runtime frequency detection

### LINKED ISSUE #2079: Proposal: Japan-Compliant JP_STRICT Regulatory Mode
Summary

Introduce a JP_STRICT regulatory mode that enforces Japan (ARIB STD-T108) compliance at the firmware level, providing a compliant baseline instead of relying entirely on user configuration.

Background

Current MeshCore configuration allows full user control over frequency, TX power, airtime, and carrier sensing. This flexibility conflicts with Japanese regulations, where compliance depends on a fixed and tested configuration required for Giteki certification (技術基準適合証明 / 工事設計認証 — Japan's mandatory radio equipment approval).
As a result:
Users can easily operate outside legal limits
Giteki-certified hardware may become non-compliant after configuration changes or firmware updates

Regulatory Target Class

ARIB STD-T108 defines multiple operating modes for the 920 MHz band. JP_STRICT explicitly targets the following class:
Target: 特定小電力無線局 (Specified Low Power Radio Station), 20 mW class, LBT mode

Parameter
Target value

Output class
≤ 20 mW (≤ 13 dBm)

Operating mode
LBT (Listen Before Talk) — carrier sense ≥ 5 ms

Frequency zone
920.6–922.2 MHz (standard LBT zone)

Problem

Japan (ARIB STD-T108) requires:

Parameter
Requirement

Frequency
920.5–923.5 MHz; sub-band rules apply

TX power
≤ 13 dBm (20 mW)

Transmission time
≤ 4s per TX (920.6–922.2 MHz zone); 
≤ 400ms per TX (922.4–923.4 MHz zone)

Carrier sense (LBT)
Channel activity check required before TX; minimum sensing duration varies by zone and method: 
≥ 5 ms (920.6–922.2 MHz, standard LBT); 
≥ 128 µs (922.4–923.4 MHz, short CS); 

920.6 MHz to 923.4 MHz, radio equipment shall stop its transmission of radio wave for 4 s or less after it starts to emit radio wave. And the radio
equipment shall wait for 50 ms or more before the following transmission.
On condition that this re-transmission starts within 4 s after the first transmission, the
radio equipment can make the re-transmission without waiting for 50 ms.

ARIB requires channel activity detection based on energy sensing (typically implemented via RSSI); LoRa hardware CAD (which detects LoRa preambles only) does not satisfy this requirement on its own

Duty cycle
No explicit hourly duty cycle limit when LBT conditions are met (920.6–922.2 MHz zone);
 ≤ 10% (360 s) per hour (922.4–923.4 MHz zone with short CS)

These constraints are not enforced at the firmware level and rely on user configuration.

Proposal: JP_STRICT Mode

Compliance enforcement must happen at the firmware level — this is the only layer that actually controls radio transmission. App-side restrictions alone can be bypassed via CLI or meshcore-cli and carry no weight under radio law. 

The recommended approach is therefore two-phase:

Phase 1 — Firmware (primary, Giteki-relevant)

Add a dedicated region mode with compliant defaults enforced in firmware:

Region: JP_STRICT

Parameter
Behavior

Frequency
Default to compliant sub-band (e.g. 920.875 MHz)

TX power
Default to ≤ 13 dBm; warn if user attempts to exceed

Max TX duration
Enforce zone-specific limits:
 ≤ 4s per transmission (920.6–922.2 MHz);
 ≤ 400ms per transmission (922.4–923.4 MHz)

Airtime control
Zone-CS method specific (922.4–923.4 MHz zone with short CS) duty-cycle enforcement

Carrier sense (LBT)
Enable energy-based sensing (int.thresh) by default — ARIB requires channel activity detection based on energy sensing (typically implemented via RSSI); LoRa hardware CAD (which detects LoRa preambles only) does not satisfy this requirement on its own; sensing duration must meet zone requirements (≥ 5 ms for 920.6–922.2 MHz; ≥ 128 µs for 922.4–923.4 MHz)

Implementation Gap: CS Duration vs. CS Threshold

int.thresh controls how strong a signal must be to consider the channel busy — it is an RSSI intensity threshold, not a time-based measurement. ARIB STD-T108 requires both:
Intensity — RSSI must be below threshold ← int.thresh addresses this
Duration — channel must remain clear for ≥ 5 ms continuously ← not currently implemented
The current isChannelActive() performs a single instantaneous RSSI snapshot. This does not satisfy the ARIB 5 ms continuous sensing requirement, regardless of the int.thresh value set.
A compliant JP_STRICT implementation would require a sensing loop that samples RSSI continuously for ≥ 5 ms and defers transmission if the threshold is exceeded at any point during that window. This is a separate implementation task from int.thresh and should be tracked as a distinct requirement.

Override behavior

Warn user when changing regulated parameters; do not hard-lock (users can always recompile)
This is the only implementation location that is meaningful from a Giteki perspective, as it governs the actual radio transmissions regardless of which client tool is used (app, CLI, or meshcore-cli).

Phase 2 — Mobile App UI

Once firmware enforces JP_STRICT, the MeshCore iOS/Android app should expose this as a visible compliance state:
Show a "JP_STRICT" indicator when the connected device is running in this mode
Warn the user explicitly when attempting to change regulated parameters (frequency, TX power)
Prevent the app from sending configuration commands that would put the device outside JP_STRICT bounds
This phase requires changes to the app repository and is noted here as a next task. It improves user experience but does not replace firmware-level enforcement.

Known Limitation

This mode does not resolve the fundamental Giteki certification issue: Giteki approval is granted for a specific hardware and configuration combination (including firmware behavior). Any firmware update — including updates to JP_STRICT itself — may require re-certification to maintain full legal compliance. This is a structural limitation of applying open-source firmware to regulated radio hardware, and is outside the scope of this proposal.
JP_STRICT mitigates the risk of accidental non-compliance for typical users, but does not constitute a Giteki certification guarantee.

Optional Enhancements

Display firmware version and build hash in device info (useful for Giteki compliance documentation)

Benefits

Reduces risk of illegal operation in Japan
Aligns with Giteki certification expectations
Improves trust for commercial and community deployments
Makes devices usable by non-expert users without requiring knowledge of ARIB STD-T108

Key Insight

Open configuration and regulatory compliance are fundamentally in tension. JP_STRICT resolves this by enforcing a compliant baseline in software — while being honest that user-level overrides cannot be fully prevented in open-source firmware.

This is not just a regional preset — it is a compliance mode. Japan is a strong use case, but this approach could generalize to other regulated regions (e.g. EU duty cycle enforcement, Korean power limits).

References

ARIB STD-T108 v1.3: https://www.arib.or.jp/english/html/overview/doc/5-STD-T108v1_3-E1.pdf

MeshCore int.thresh Companion firmware exposure (companion issue) #2051

Proposed Japan Frequency Preset
https://github.com/meshcore-dev/MeshCore/issues/460#issuecomment-4080531481

jirogit:
Closes #2079

## Update (2026-08-15)

Radio parameters migrated from SF12/CR4:8 to SF10/CR4:6 (jp-v1.8.0). As a result:

- The dynamic MAX_TEXT_LEN table below is currently a no-op. At SF10/CR4:6,
  even a 255-byte packet (LoRa's max) takes 2854ms, leaving 1146ms of margin
  under the 4-second ARIB limit — the CR-dependent length limit is no longer
  the binding constraint. See comments below for details.

- Re: @4np's concern about false-positive activation on AS923-2 deployments
  (Indonesia/Vietnam) that share the 920.8/921.0/921.2 MHz frequencies:
  isAS923_1_JP() remains frequency-only. A BW125 check was considered, but
  125kHz is the common default bandwidth across AS923 deployments generally,
  so it would not meaningfully narrow false positives. As noted in the
  original review response below, Vietnam's current default (920.250 MHz)
  does not match any of the three checked frequencies, and Indonesia has no
  defined default frequency in MeshCore today. Real-world false-positive
  risk remains low but not zero; a more general regional-compliance
  mechanism (e.g. an explicit `radio_law` setting) would be a more robust
  long-term fix, left for future discussion.

- Folded in #3103 (SF10 backoff parameter tuning) as an additional commit
  on this branch. base_ms 2000→500ms, max_backoff 16000→4000ms, reflecting
  the same SF12→SF10 airtime reduction noted above. #3103 is closed as
  superseded.

- Field usage: currently running as jp-v1.9.0, in use by the Japan - LoRa & Mesh Discord community (204 members).

- Frequency set updated 2026-08-27: 921.0 / 921.2 / 921.4 MHz (previously 920.8 / 921.0 / 921.2 MHz), to avoid conflict with Meshtastic JP's default allocation. Comments below predating this change refer to the old frequency set.

---
 
(Original PR description below is left unedited for continuity.)

---

## Summary

This PR implements Listen-Before-Talk (LBT) for Japan LoRa operation, as required by ARIB STD-T108 / Giteki (技適) certification. Activation is automatic based on operating frequency — no build flags required.

## Changes

**Radio accessor methods (RadioLibWrappers.h/cpp, Wrapper classes):**
- `getCodingRate()` — returns current LoRa coding rate at runtime
- `getFreqMHz()` — returns operating frequency in MHz
- `getMaxTextLen()` — returns dynamic DM message length limit based on CR
- `getMaxGroupTextLen()` — returns dynamic channel message length limit based on CR
- `isJapanMode()` ---> 'isAS923_1_JP()' — returns true when operating on JP LoRa frequencies (921.000 / 921.200 / 921.400 MHz)

**Japan LBT (BaseChatMesh.cpp, RadioLibWrappers.cpp, Dispatcher.h):**
- 5ms continuous RSSI sensing before TX (−80dBm absolute threshold, ARIB STD-T108 §4.4)
- Exponential backoff when channel busy (base 2000ms, max 16000ms)
- Small jitter on channel-free detection (0–500ms) to desynchronize simultaneous TX
- 50ms post-TX wait
- Override `getCADFailMaxDuration()` to `UINT32_MAX` in Japan mode — prevents forced TX during LBT backoff (Dispatcher default is 4s, shorter than max backoff of 16s)
- Dynamic message length limits per coding rate, derived from ARIB STD-T108 4-second maximum TX time limit, verified by measuring `getTimeOnAir()` on RAK WisMesh Tag (SF12/BW125):

| CR | DM limit | airtime | Channel limit | airtime |
|------|-------------|-----------|----------------|-----------|
| 4/5 | 64 bytes | 3874ms | 64 bytes | 3710ms |
| 4/6 | 48 bytes | 3874ms | 48 bytes | 3678ms |
| 4/7 | 32 bytes | 3678ms | 39 bytes | 3907ms |
| 4/8 | 24 bytes | 3547ms | 29 bytes | 3809ms |

DM and channel packets differ by 1 byte overhead, warranting separate limits for CR4/7 and CR4/8. Non-JP returns default bytes unchanged.

**Build:**
- Fix build for non-FreeRTOS and non-SX1262 platforms (`YIELD_TASK()` macro)

## Note on commit history

Early commits include CAD-based channel detection based on weebl2000's work (PR #1727). This was subsequently replaced with RSSI-based LBT, as ARIB STD-T108 requires energy detection, not LoRa preamble detection.

As of the current implementation (rebased onto upstream/dev which includes merged #1727), RSSI LBT and CAD operate independently. JP nodes run RSSI sensing first (ARIB requirement), then fall through to CAD if _cad_enabled. 
Non-JP nodes use the existing _threshold and CAD path unchanged. 
The commit history reflects the earlier exploration path before #1727 was merged.

## Testing

Devices: RAK WisMesh Tag (SX1262), Wio Tracker L1 Pro (SX1262)
Settings: SF12 / BW125 / CR4-8 (JP standard)
- 16-byte simultaneous DM: 3/3 success
- 48-byte simultaneous DM: 3/3 success

- Channel messages: all CR rates confirmed (CR4/5–CR4/8).
  Effective text limit is (channel_limit - 2 - sender_name_length) bytes,
  where 2 bytes are consumed by the ": " separator.
  Example (8-byte sender name): CR4/5=54B, CR4/6=38B, CR4/7=29B, CR4/8=19B

## Known limitations

- In Japan mode, the UI does not reflect the reduced message length limits. Channel messages longer than the CR-dependent limit will be silently truncated at send time. DM messages exceeding the limit will fail to send. A UI-level fix requires exposing `getMaxTextLen()`/`getMaxGroupTextLen()` to the UI layer and is left for a follow-up.
- TX power limit (≤13dBm, ARIB STD-T108) is not enforced — relies on user configuration.
- If the ambient noise floor exceeds −80 dBm, transmission will be blocked indefinitely. No forced TX occurs — this is intentional, as ARIB STD-T108 prohibits transmission while the channel is busy. Users can monitor the noise floor via the MeshCore companion app: tap the three-dot menu (⋮) in the top right corner → Tools → Noise Floor.


src/helpers/radiolib/RadioLibWrappers.h
```diff
@@ -31,14 +32,43 @@ class RadioLibWrapper : public mesh::Radio {
...
     if (isReceivingPacket()) return true;
 
     return isChannelActive();
   }
 
   virtual float getCurrentRSSI() =0;
+  virtual uint8_t getCodingRate() const { return 8; }   // default CR4/8, override in subclass
+  virtual float getFreqMHz() const { return 0.0f; }     // default unknown, override in subclass
+						      //
+  bool isJapanMode() const {
```
4np: The function name doesn't match the frequencies being checked. 920.8, 921.0, and 921.2 MHz are the default channels for AS923-2, which is the plan for Indonesia and Vietnam, not Japan. Japan uses AS923-1 with default channels at 923.2 and 923.4 MHz. This means the function will return false for most Japanese devices and true for Indonesian/Vietnamese ones, a false positive for an entirely different region. Suggest renaming to isAS923_2Mode() or correcting the frequencies to 923.2/923.4 MHz depending on the intended behavior.

jirogit: > The function name doesn't match the frequencies being checked. 920.8, 921.0, and 921.2 MHz are the default channels for AS923-2, which is the plan for Indonesia and Vietnam, not Japan. Japan uses AS923-1 with default channels at 923.2 and 923.4 MHz. This means the function will return false for most Japanese devices and true for Indonesian/Vietnamese ones, a false positive for an entirely different region. Suggest renaming to isAS923_2Mode() or correcting the frequencies to 923.2/923.4 MHz depending on the intended behavior.

Thank you for the review. You are correct that 920.8/921.0/921.2 MHz overlap with AS923-2 (Indonesia/Vietnam in LoRaWAN terms). However, these frequencies are legal under Japanese radio law (ARIB STD-T108, 920.6–928.0 MHz) and are within the band permitted for use in Japan.

Regarding the false positive concern for Indonesian/Vietnamese users: MeshCore does not currently have a defined default frequency for Indonesia. Vietnam's current default is 920.250 MHz, which does not match any of the three frequencies checked by isJapanMode() (920.800 / 921.000 / 921.200 MHz). In practice, the risk of unintended activation on non-Japanese nodes is low.

That said, the concern is valid in principle and points toward a broader architectural question: as MeshCore expands into regions with different regulatory requirements — duty cycle limits in the EU, LBT variants in other Asian countries, TX power caps, airtime limits — a more general regional compliance framework would be valuable. 

A future "set radio_law / get radio_law" command (distinct from the existing "set region") could allow users to explicitly select their regulatory domain (e.g. JP, EU, FCC), enabling the firmware to apply the appropriate constraints automatically. 

The current frequency-based isJapanMode() is intentionally minimal and pragmatic given the near-zero Japanese node count today, but we recognize it is not a scalable long-term solution. We are open to adding such a mechanism if the maintainer prefers a more structured approach, and would welcome guidance on the preferred direction for regional compliance in MeshCore going forward.

jirogit:
Related RFC Discussion: https://github.com/meshcore-dev/MeshCore/discussions/2285

dreirund:
Question from https://github.com/meshcore-dev/MeshCore/pull/1727#issuecomment-4352266813:

How does it differ from the current `int.thresh` implementation?

Regards!

jirogit:
> How does it differ from the current `int.thresh` implementation?

Thank you very much for your question!

This PR does not use `int.thresh`.

The `int.thresh`-based mechanism uses a relative noise floor threshold (`noise_floor + threshold`) as a voluntary interference avoidance heuristic (currently hardcoded to 0 in Companion firmware — see #2051).

ARIB STD-T108 LBT requires an absolute −80 dBm threshold, a minimum 5ms continuous RSSI sensing window before each TX, and a defined backoff protocol — these are legal requirements, not preferences. The `getCADFailMaxDuration()` override to `UINT32_MAX` is also needed to prevent the Dispatcher's default 4s forced-TX timeout from firing during a legitimate LBT backoff.

So the two mechanisms are complementary: `int.thresh` applies generally as a configurable courtesy; JP LBT applies when on Japan frequencies as a non-negotiable legal obligation with specified parameters.

jirogit:
One more note on the jitter mechanism in this PR, in context of the `txdelay` discussion in #1727:

The 0–500ms random jitter added on channel-free detection serves a similar purpose to `txdelay`, but targets a different point in the transmission chain:

| | `txdelay` | JP LBT jitter |
|---|---|---|
| Purpose | Stagger repeater relay TX after receiving a packet | Stagger any TX attempt the moment a node detects the channel as free |
| Target | Repeater relay behavior | All nodes, every TX attempt |
| Value | Proportional to airtime (default factor 0.5) | Random (0–500ms) |
| Configuration | `set txdelay` CLI | Internal to JP LBT |

Both mechanisms address the same root problem — synchronized transmission from multiple nodes — at different points in the chain. txdelay acts on the repeater _receive side_ (after a packet arrives); JP LBT jitter acts on the _transmit side_ (just before each TX attempt by any node, when the channel is free).

syssi's video in #1727 clearly demonstrates what happens without staggering: with `txdelay 0`, all repeaters synchronize and collide simultaneously. The same collision pattern occurs at TX origination without jitter.

jirogit:
Update: airtime-scaled jitter (f2aad641)

Replaced fixed random(0, 500) jitter with airtime-scaled random(0, airtime_ms / JP_LBT_JITTER_DIVISOR) (divisor=32).

Jitter upper bound by setting:
SF12/BW125/CR4/8 (JP standard): ~245ms (was 500ms)
SF7/BW62.5/CR4/5 (USA/Canada preset): ~12ms

Tested divisors 16 and 32 at SF12/BW125/CR4/8 — both 3/3 success with no measurable difference in delivery time. Divisor 32 adopted as it halves the unnecessary wait without observable downside.
The divisor is a named constant (JP_LBT_JITTER_DIVISOR) to make future tuning straightforward.

jirogit:
I've renamed isJapanMode() to isAS923_1_JP() to address @4np's concern about AS923-2 frequency overlap.
Rationale: LoRaMac-node uses the same sub-plan naming (AS923_1_JP) to distinguish Japan's ARIB STD-T108 LBT requirement from AS923 in general — AS923 itself has no LBT mandate; LBT here is Japan-specific. isAS923_1_JP() makes that explicit.

jirogit:
Added commit 4e09f6a: JP LBT txdelay suppression.

**Problem**

In JP LBT mode (SF12/BW125/CR4/8), the standard `txdelay` formula (`random(0, 5 × airtime × factor)`) produces delays up to ~8.9s with default `tx_delay_factor=0.5`. Urban Japan (Akihabara being the design target) has a high noise floor that makes SF7/BW125 ("LongFast") impractical — SF12/BW125 is required for reliable communication. At SF12, airtime is ~29× longer than SF7, which makes the standard txdelay formula produce multi-second delays. In this environment, LBT backoff is already doing the heavy lifting for channel access; txdelay on top is pure latency penalty.

**Why txdelay has limited value in JP LBT**

The purpose of `txdelay` is to stagger retransmit timing across multiple repeaters to reduce collisions. In JP LBT mode, this role is already covered by two mechanisms:

- **LBT backoff**: exponential backoff (1–16s) when RSSI > −80 dBm — handles environmental noise and active transmissions
- **Jitter** (added in f2aad64): `random(0, airtime/32)` applied after a clean LBT sense — handles the case where multiple nodes clear simultaneously

With very few MeshCore nodes currently deployed in Japan, repeater-to-repeater collision is not a practical concern.

**Fix**

When `isAS923_1_JP()` is true, `getRetransmitDelay()` and `getDirectRetransmitDelay()` now return `random(0, jitter_max)` instead of the full txdelay window — matching the jitter scale used in `isChannelActive()`:

```
SF12/BW125/CR4/8:  jitter_max ≈ 111ms
  before: txdelay max ≈ 8,900ms, average ≈ 4,400ms
  after:  txdelay max ≈   111ms, average ≈    56ms
```

Collision probability analysis (4 simultaneous repeaters, 1ms detection threshold):
- jitter alone: ~10.4% chance of true collision
- jitter + txdelay at jitter-scale: ~7.0% (33% improvement)
- average added latency: 56ms vs 4,400ms

This scales naturally as CR decreases over time (CR4/8 → CR4/5): `jitter_max` shrinks with airtime, so txdelay tracks it automatically without any config change.

`JP_LBT_JITTER_DIVISOR` is promoted from a local `static const` in `RadioLibWrappers.cpp` to a `public static constexpr` in `RadioLibWrappers.h` so the examples can reference it directly.

Applied to `simple_repeater`, `simple_room_server`, and `simple_sensor`. Also aligns `simple_sensor`'s non-JP path to use `nextInt(0, 5*t+1)` (was `nextInt(0, 6)*t`).

jirogit:
Rebased onto upstream/dev (includes #1727)

This branch has been completely rebuilt on top of the current upstream/dev (which includes the merged CAD LBT from #1727). 

The previous history had #1727's CAD commits mixed in, causing conflicts across RadioLibWrappers.cpp/.h. Rather than resolving those conflict-by-conflict, I created a clean branch from upstream/dev and re-implemented the JP LBT changes on top.

isAS923_1_JP()  LBT and CAD now operate independently:

JP mode: isAS923_1_JP() is active when frequency is set to 920.8 / 921.0 / 921.2 MHz, CAD optional — runs after RSSI clear if set cad on

Non-JP:  isAS923_1_JP() not active, CAD optional, existing behavior unchanged

JP nodes run RSSI sensing first (ARIB requirement), then fall through to CAD if _cad_enabled. No existing behavior changes for non-JP nodes.

Also recovered two orphaned commits (f2aad64, 4e09f6a) that were not on any branch:

Airtime-scaled jitter (random(0, airtime/32)) instead of fixed 500ms
txdelay suppressed to jitter-scale in JP mode to avoid adding ~4400ms latency on top of LBT backoff

Hardware tested: RAK WisMesh Tag (SX1262) + LilyGo T1000-E (LR1110), companion firmware, SF12/BW125/CR4/8 at 920.8 MHz. DM and group messages at maximum text length delivered successfully.

usrflo:
One thing I wanted to check on the 50 ms post-TX delay(50) in `onSendFinished()`: as far as I can see in RadioLib 7.7.1, `SX126x::finishTransmit()` ends in `standby()`, so the chip is in STANDBY — not RX — right after TX. RX is only re-armed from the main loop (recvRaw(), when state != STATE_RX), and that loop is blocked inside this delay(50). So the radio is effectively not receiving for 50 ms after every TX. ARIB STD-T108 §3.4.1 mandates a minimum spacing between transmissions; it doesn't appear to require the receiver to be muted during that gap. Is keeping RX closed during the wait intentional, or would re-arming RX immediately after `finishTransmit()` (letting the 50 ms spacing elapse while still receiving) be feasible?

Related: the busy-channel backoff seemts to block up to 16 s inside `checkSend()`, during which `loop()` can't service incoming packets - was a non-blocking, timestamp-based backoff considered so RX handling keeps running?

jirogit:
Good catch on both points — appreciate the close read of RadioLib's actual state transitions.

1. You're right, and it's not intentional. `finishTransmit()` leaves the radio in STANDBY, `recvRaw()` only re-arms RX when `state != STATE_RX`, and the blocking `delay(50)` sits in between, so RX stays closed for the duration. Since ARIB STD-T108 §3.4.1 only requires spacing between transmissions and not RX muting, I'll switch this to re-arm RX (`startReceive()`) immediately after `finishTransmit()` and enforce the 50ms spacing as a non-blocking timestamp check before the next TX attempt instead.

2. Also correct. The busy-channel backoff already avoids a hard `delay()` (that was changed earlier specifically to stop it blocking BLE/UI), but `checkSend()` still doesn't return until the backoff elapses, so packet handling is deferred for up to 16s. I'll convert this into a proper non-blocking check: store the backoff deadline and have `checkSend()` return immediately when busy, re-checking on the next loop iteration rather than spin-waiting inside the call.

Will have an updated version of both within the next few days.

jirogit:
Thanks for the detailed review — both points below.

**Post-TX 50ms delay(50):** You're right, and confirmed by testing —
`delay(50)` does block RX from resuming for that window, since
`finishTransmit()` leaves the radio in STANDBY and `recvRaw()` only
re-arms RX on the next `loop()` pass. I looked into replacing this
with a non-blocking timestamp check, but decided against it for this
PR: the potential gain (avoiding a 50ms RX gap once per TX) is small
relative to the busy-channel backoff below, and getting a
timestamp-based implementation of the ARIB §3.4.1 spacing requirement
wrong would be a silent compliance issue rather than a visible bug.
`delay(50)` is simple and unambiguous, so I'd rather leave it as-is
unless there's a stronger case for changing it.

**Busy-channel backoff blocking up to 16s:** Converted to a
non-blocking state machine (`_lbt_backoff_active` / `_lbt_deadline`).
On busy detection, the backoff deadline is recorded and the function
returns immediately; `Dispatcher::checkSend()`'s existing retry
mechanism re-polls on the next `loop()` pass instead of spin-waiting.
`loop()` (and RX servicing) is no longer blocked during backoff.
The initial 5ms RSSI sense and the post-clear jitter wait are left
blocking, since both are short relative to the up-to-16s backoff this
targets.

Worth noting: 16s is the cap for a single backoff cycle, not the
total worst case. On repeated busy detections, `_busy_count` keeps
growing and each new cycle re-rolls a fresh 16s-capped window, so
under sustained channel activity the total wait is effectively
unbounded (by design -- `getCADFailMaxDuration()` returns
`UINT32_MAX` in JP mode, since ARIB STD-T108 prohibits forcing a
transmission through). In testing under real electrical interference
I saw a single message take ~4 minutes to clear; once the interference
source was removed, delivery was back to normal (single-digit
seconds). The important change here is that this multi-minute wait
no longer blocks the Dispatcher loop -- RX keeps working throughout.

Hardware-tested (Lilygo T3S3 sx1262 / Wio Tracker L1 Pro), including bidirectional
simultaneous-send collision scenarios -- backoff resolves correctly
and RX keeps working throughout.

jirogit:
Rebased onto upstream/dev (now includes v1.17.0). One conflict in CustomLR1110.h (upstream's startReceive()/isReceiving() changes vs. this PR's getCodingRate() addition — both preserved, no logical overlap). Build-verified on T1000-E, XIAO nRF52 + Wio-SX1262, and T3S3 (all clean).

jirogit:
Superseded by #3383, which generalises this into a CLI-configurable mechanism with no radio-regulation-specific code.

PR #2223, opened 2026-04-01
Added RAK12035 Soil Moisture and Temperature Sensor

KPrivitt:
A RAK12035 Soil Moisture Sensor will be automatically detected and incorporated into the Telemetry and can be accessed via the Telemetry screen in a repeater or a companion. The sensor data is displayed on Channel 2.

This code also supports Calibration of a sensor via a custom version of the firmware. See the RAK12035_SoilMoisture.cpp for instructions on the calibration procedure, which uses Channel 3. Calibration must be done to a sensor; each sensor will generate different capacitance output values. Calibration will ensure the reported Moisture will return a value between 0 and 100%.

The RAK12035 Soil Moisture Sensor connects to a RAK12023 which provide three connectors for sensors. This driver only supports one sensor attached with a default I2C address of 0x20. Support can be added for additional sensors via the Set Sensor Address command (which is incorporated). The sensors attached must all have a different I2C address (this can be done by changing an additional sensors I2C address to 0x21 or 0x22) and this code will need to include the initialization of each of these sensors (but this is not included).

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -134,6 +134,7 @@ build_flags =
   -D ENV_INCLUDE_VL53L0X=1
   -D ENV_INCLUDE_BME680=1
   -D ENV_INCLUDE_BMP085=1
+  -D ENV_INCLUDE_RAK12035=1
 lib_deps =
   adafruit/Adafruit INA3221 Library @ ^1.0.1
   adafruit/Adafruit INA219 @ ^1.2.3
--- a/platformio.ini
+++ b/platformio.ini
@@ -134,7 +134,6 @@ build_flags =
   -D ENV_INCLUDE_VL53L0X=1
   -D ENV_INCLUDE_BME680=1
   -D ENV_INCLUDE_BMP085=1
-  -D ENV_INCLUDE_RAK12035=1
 lib_deps =
   adafruit/Adafruit INA3221 Library @ ^1.0.1
   adafruit/Adafruit INA219 @ ^1.2.3
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -735,7 +735,7 @@ void EnvironmentSensorManager::start_gps() {
   _location->begin();
   _location->reset();
 
-#ifndef PIN_GPS_RESET
+#ifndef PIN_GPS_EN
   MESH_DEBUG_PRINTLN("Start GPS is N/A on this board. Actual GPS state unchanged");
 #endif
 }
@@ -759,6 +759,9 @@ void EnvironmentSensorManager::loop() {
   static long next_gps_update = 0;
 
   #if ENV_INCLUDE_GPS
+  if (gps_active) {
+    _location->loop();
+  }
   _location->loop();
   if (millis() > next_gps_update) {
 
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -103,7 +103,7 @@ static Adafruit_VL53L0X VL53L0X;
 
 #if ENV_INCLUDE_RAK12035
 #define TELEM_RAK12035_ADDRESS 0x20      // RAK12035 Soil Moisture sensor I2C address
-#include "RAK12035_SoilMoisture.h"
+#include <RAK12035_SoilMoisture.h>
 static RAK12035_SoilMoisture RAK12035;
 #endif
 
@@ -290,7 +290,7 @@ bool EnvironmentSensorManager::begin() {
     INA260_initialized = true;
   } else {
     INA260_initialized = false;
-    MESH_DEBUG_PRINTLN("INA260 was not found at I2C address %02X", TELEM_INA219_ADDRESS);
+    MESH_DEBUG_PRINTLN("INA260 was not found at I2C address %02X", TELEM_INA260_ADDRESS);
   }
   #endif
 
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -718,8 +718,6 @@ bool EnvironmentSensorManager::gpsIsAwake(uint8_t ioPin){
     gps_detected = true;
     return true;
   }
-  
-  
   pinMode(ioPin, INPUT);
   MESH_DEBUG_PRINTLN("GPS did not init with this IO pin... try the next");
   return false;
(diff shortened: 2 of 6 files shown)


KPrivitt:
Changed PR #2191 to use dev as the base.  Removed platformIO.ini and added
-D ENV_INCLUDE_RAK12035=1 to variants/platformIO.ini

It still compiles correctly for the RAK4631 versions but, is it still getting errors from the Heltec_v3 due to the mismatched pin numbering.

Since it is not possible to connect a RAK12035 to a Heltec (they do not have the connector), it will never be detected, and this code will not be utilized.

I can conditionalize it, but I'm not that familiar with the Heltec V3.  I found the pin they use to control their 3v3 supply.  All RAK boards use IO2 to control 3v3_S, but the EnvironmentSensorManager rakGPSInit() is incorrectly is using IO4 and IO5.  These are actually the *reset signal for the RAK12500 GNSS GPS Location unit.  The RAK12035 uses the IO4 as its *reset input.

#if defined(ARDUINO_RAK4631) || defined(PLATFORM_RAK_WISBLOCK)
    #define SENSOR_PWR_PIN   WB_IO2
    #define SENSOR_RST_PIN   WB_IO4

#elif defined(ARDUINO_HELTEC_WIFI_LORA_32_V3) || defined(HELTEC_V3)
    // Heltec V3 uses Vext for 3V3 peripheral power
    #define SENSOR_PWR_PIN   Vext
    // Heltec V3 does not have a reset pin for sensors
    #define SOIL_RST_PIN   GPIO4   // I believe this is a General Purpose I/O and is not tied to anything

#else
    #error "Unsupported board: define SOIL_PWR_PIN and SOIL_RST_PIN"

OR should I conditionalize each use of the WB_IO2, WB_IO4 with
#ifdef RAK_WISBLOCK_GPS
   pinMode(ioPin, OUTPUT);
   digitalwrite(ioPin, HIGH);
#endif

OR I can comment all of them out, the purpose of this code is to turn off the 3v3_S when the sensor is put to sleep and to turn the power back on when it is woken up.
But MeshCore does not have any power management for sensors yet.  I put this in for a possible sensor power management controller that would signal sensors to power on and reinitialize themselves.

KPrivitt:
This now compiles correctly for Heltec v3 versions.

I commented out all the uses of pinMode() and digitalWrite() and included comments on what the intended functionality was for.  This was intended for when a future sensor power management is implemented.  None of the routines affected were utilized, so this did not affect the functionality.

The unutilized routines were added to provide the following functionality:
1) To put the sensor into sleep mode, optionally to turn the 3v3_S power off to obtain even lower power draw.
2) To turn the power back on, properly reset the sensor and verify it will respond to commands

These can be uncommented and potentially used with future sensor power management firmware. At that time, it will require modifications to utilize the pinMode() and digitalWrite() functions with a Heltec v3.  The errors will return when the comments are removed.

KPrivitt:
Should I close PR #2191

.github/workflows/github-pages.yml
```diff
@@ -31,6 +31,6 @@ jobs:
         uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          cname: docs.meshcore.nz
+          cname: docs.meshcore.io
```
ripplebiz: This has already been update in the main branch

ripplebiz:
> Should I close PR #2191

Yes.   This PR looks fine. Ready to merge once that unneeded yaml change is reverted.

.github/workflows/github-pages.yml
KPrivitt: Do I need to remove this?  I did not make this change.  I believe it appeared when I moved to the dev branch from main.  How do I fix this?

KPrivitt:
#2191 has been closed.

This is my first PR, how do I revert this?

KPrivitt:
Here are screen shots of Channel 1, which has two other sensor cards that are plugged in and their data shown 

image

Channel 2, which is this new sensor and its data shown

image

Soil Moisture and Soil Temperature

.github/workflows/github-pages.yml
liamcottle: These changes are probably being pulled in because @KPrivitt made their changes against the `main` branch in their local environment, and then changed the base branch to `dev` in the GitHub UI.

Ideally, users should check out the `dev` branch locally, and make changes against that.

You should be able to fix this PR by performing a `rebase` of your local `2026/RAK12035` branch against the `dev` branch and then force push it.

The reason the extra commits show up, is because your local branch is based on `main` which has these new commits that don't yet exist in the dev branch.

It's not really a big deal, and we could merge as is, it would just create duplicate commits, but it's cleaner for the history if you rebase the branch.

KPrivitt (commented):
I believe the rebase worked.  It appears that it is no longer included but it is now shown as a comment.  Let me know if I still need to do something.

PS C:\Users\kenpr\Documents\MeshCore\MeshCore-dev> git rebase  --continue
>>
[detached HEAD e7b9401b] RAK12035: Rebase to resolve YAML changes
 1 file changed, 33 insertions(+), 2 deletions(-)
dropping 0423729b1f61c9b41b9bccf6902ccc6ded0d7eb9 Remomve redefinition of MESH_DEBUG_PRINT/LN -- patch contents already upstream
Successfully rebased and updated refs/heads/2026/RAK12035.
PS C:\Users\kenpr\Documents\MeshCore\MeshCore-dev> git push --force-with-lease
>>
Everything up-to-date

KPrivitt:
There are now only the five that were updated or changed, and it is now pointing to meshcore-dev:dev

This was a disaster, the rebase failed and I had to basically start over.  All the changes I had
previously done to get it to compile clean were lost and were regenerated.  Well at least
the comments are now better.

Please recheck, I verified that it compiles for the RAk4631 and Heltec v3 
So, everything should now be ready to merge.

liamcottle:
Thanks for the rebase, tests appear to be passing now that the include flag has moved to the RAK variant.

The commit log is now 18 commits long, so I'll squash these during merge so there's only 1 commit in the project history.

Thanks for contributing!

KPrivitt:
FYI, I moved the include flag to RAK_Variant but that did not fix the build errors. But this seems to be a better place for it since this sensor is essentially only for RAK boards (it has a custom connector and can only plug into one of them).

The fix was removing the code that accessed the pins (WB_IO2 and WB_IO4 that control the power and reset).  The Heltec v3 uses different pin names (vEXT to turn control the sensor power).  This code was included to support turning the RAK12035 power ON/OFf (to save power) and to toggle its *reset pin (to recover from the power shutdown).  I commented it out and added some verbiage to explain what it was for, its future use and the Heltec v3 issue.  This fixed the errors.

PR #2234, opened 2026-04-03
Allows us to have custom PlatformIO envs and configs

andyshinn:
This allows us to keep a `platformio_override.ini` in the project root that we can use for custom envs or config without needing to dirty git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,4 @@ cmake-*
 compile_commands.json
 .venv/
 venv/
-platformio_override.ini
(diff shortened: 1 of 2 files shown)


liamcottle:
Sounds good to me.

Can you please rename this to `platformio.local.ini`?

This will allow it to show underneath `platformio.ini` in alphabetical order, and also notes that it's a local only configuration.

andyshinn:
Sure, done.

PR #2235, opened 2026-04-03
fix: enable radio.rxgain CLI command for LR1110 (T1000-E)

jirogit:
## Problem

`radio.rxgain` get/set commands were broken on T1000-E (LR1110).

The CLI branches in `CommonCLI.cpp` were guarded by
`#if defined(USE_SX1262) || defined(USE_SX1268)`, but `USE_LR1110` was
not defined in the T1000-E build, so both branches were compiled out.

This caused two different failure modes:

- `get radio.rxgain` → returned the full radio config string (e.g.
  `> 921.000,125.0,10,5`) because `memcmp("radio", 5)` matched the
  next branch
- `set radio.rxgain on` → returned `unknown config: radio.rxgain on`
  because no branch matched

## Fix

- Add `-D USE_LR1110` to `variants/t1000-e/platformio.ini`
- Add `USE_LR1110` to the `#if` guard on both get and set branches

`CustomLR1110Wrapper` already implements `setRxBoostedGainMode()` and
`getRxBoostedGainMode()` via the `RadioLibWrapper` virtual interface —
no radio-layer changes required.

## Testing (T1000-E)

- `get radio.rxgain` → `> off` ✅
- `set radio.rxgain on` → `OK` ✅  
- `get radio.rxgain` after reboot → `> on` ✅

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -547,7 +547,7 @@ void CommonCLI::handleSetCmd(uint32_t sender_timestamp, char* command, char* rep
     _prefs->disable_fwd = memcmp(&config[7], "off", 3) == 0;
     savePrefs();
     strcpy(reply, _prefs->disable_fwd ? "OK - repeat is now OFF" : "OK - repeat is now ON");
-#if defined(USE_SX1262) || defined(USE_SX1268)
+#if defined(USE_SX1262) || defined(USE_SX1268) || defined(USE_LR1110)
   } else if (memcmp(config, "radio.rxgain ", 13) == 0) {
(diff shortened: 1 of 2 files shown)


jirogit:
This also partially addresses #2118. 
The root cause of that issue (rxgain defaulting to OFF on upgrade) 
is separate, but on T1000-E (LR1110) the CLI command was completely 
non-functional, so users had no way to restore the setting even manually.
This fix enables `set radio.rxgain on` to work on T1000-E.

jirogit:
Rebased onto current upstream/dev. Resolved conflict: get/set commands were refactored into handleGetCmd()/handleSetCmd() in dev. Applied USE_LR1110 guard to both new functions.

PR #2236, opened 2026-04-03
fix(RegionMap): compact storage /regions3, fix save/load return values, add paged region CLI output

### LINKED ISSUE #1880: region name memory space max out

### LINKED ISSUE #2166: region CLI command output truncated to ~5-10 entries due to packet payload limit
region command output is silently truncated when region names are long, because the reply is limited to a single LoRa packet payload.

In examples/simple_repeater/MyMesh.cpp:

uint8_t temp[166];
char *reply = (char *)&temp[5];  // reply = 161 bytes max

region_map.exportTo(reply, 160);  // hard-coded 160 byte limit

With encryption overhead (`CIPHER_MAC_SIZE=2` + `CIPHER_BLOCK_SIZE=16`), 161 bytes is already the physical maximum for a single packet. At 30-character names, only ~5 entries fit before truncation.

Proposed fix

Paging support for the `region` command:

region        # page 1
region 2      # page 2
region 3      # page 3

Or a denser `region list` output using the existing `exportNamesTo()` which uses comma separators (1 byte) instead of newlines with indentation.

Related

- Blocks full testing of #1880 fix (32 regions × 30 chars)

jirogit:
### Problem

On nRF52840 (T1000-e), the `/regions2` binary format used 128 bytes of padding per record:

```
2 (id) + 2 (parent) + 31 (name) + 1 (flags) + 128 (pad) = 164 bytes/record
32 entries = 5,258 bytes total
```

This exceeds the available space in nRF52 InternalFS when shared with preferences and identity data. `save()` would fail silently (hardcoded `return true`), and on next boot `load()` would read a truncated file, losing region names beyond ~11 entries.

Additionally, the `region` CLI command was silently truncated to ~5 entries due to the single-packet reply buffer limit (~161 bytes), making it impossible to verify how many regions were actually stored.

### Changes

**`src/helpers/RegionMap.cpp`**

- New `/regions3` format: 64-byte padding per record (100 bytes/record, 3,210 bytes for 32 entries = 78% of one 4096B LittleFS block)
- `load()` prefers `/regions3`; falls back to `/regions2` (128-byte pad) for automatic migration on next `save()`
- After successful `save()` to `/regions3`, deletes `/regions2` if it exists
- Fixed `save()`: hardcoded `return true` → `return success`
- Fixed `load()`: hardcoded `return true` → `return num_regions > 0 || success`

**`examples/simple_repeater/MyMesh.cpp`**

- `region` command now supports paged output:
  - `region` — first page (wildcard `*` + entries from index 0)
  - `region N` — entries from index N onward
  - Footer `>N` at end of each page indicates the next start index
  - No footer = last page; `-none-` = no entries at all
- Fixes silent truncation that limited display to ~5 entries regardless of how many were stored
- Note: the mobile app UI still shows only ~5 regions due to a separate app-side limitation; that is outside the scope of this fix

### Storage size comparison

| Format | bytes/record | 32 entries total | nRF52 4096B block usage |
|---|---|---|---|
| `/regions2` (old) | 164 | 5,258 B | **1.28×** ← over limit |
| `/regions3` (new) | 100 | 3,210 B | 0.78× |

### Testing

Tested on Seeedstudio T1000-e:

- ✅ 32 entries with 29-character names saved and reloaded after reboot
- ✅ `region save` returns `OK`
- ✅ `region` / `region 4` / `region 8` ... paging works correctly with `>N` footer
- ✅ Last page shows no footer
- ⚠️ Migration from `/regions2` → `/regions3` not tested on physical hardware (code path verified by review only)

Fixes #2166

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -1180,8 +1180,50 @@ void MyMesh::handleCommand(uint32_t sender_timestamp, char *command, char *reply
 
     const char* parts[4];
     int n = mesh::Utils::parseTextParts(command, parts, 4, ' ');
-    if (n == 1) {
-      region_map.exportTo(reply, 160);
+    if (n == 1 || (n >= 2 && parts[1][0] >= '0' && parts[1][0] <= '9')) {
+      // Paged region listing. Argument is the start index into the named-region
+      // array (0-based). 'region' or 'region 0' starts from the beginning.
+      // The first page always prepends the wildcard (*).
+      // The reply footer shows the next start index, e.g. ">5" meaning
+      // run 'region 5' to continue. No footer = last page.
+      int start = (n >= 2) ? atoi(parts[1]) : 0;
+      if (start < 0) start = 0;
+
+      int total = region_map.getCount();   // named regions only (excludes *)
+
+      char *dp = reply;
+      char *limit = reply + 148;  // leave 12 bytes for ">NN\0" footer + safety
+
+      if (start == 0) {
+        // always show wildcard on first page
+        const RegionEntry& wc = region_map.getWildcard();
+        int wlen = snprintf(dp, (size_t)(limit - dp), "*%s\n",
+                            (wc.flags & REGION_DENY_FLOOD) ? "" : " F");
+        if (wlen > 0 && dp + wlen < limit) dp += wlen;
+      }
+
+      int i = start;
+      for (; i < total; i++) {
+        const RegionEntry* r = region_map.getByIdx(i);
+        const char* name = r->name[0] == '#' ? r->name + 1 : r->name;
+        int rlen = snprintf(dp, (size_t)(limit - dp), "%s%s\n",
+                            name,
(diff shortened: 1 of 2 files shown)


jirogit:
Closing this PR. The region CLI has been significantly refactored into CommonCLI::handleRegionCmd() in upstream (commit 4131a455), making the MyMesh.cpp changes in this PR obsolete. The paged region output feature will be resubmitted as a separate focused PR targeting CommonCLI::handleRegionCmd().
Note: the save/load bug (save() always returning true on write failure) does not appear to have been fixed in upstream — regions are lost after reboot when the LittleFS block limit is exceeded.

PR #2240, opened 2026-04-03
SX1276 used bad Sync word, caused radio init error -2

petrkr:
Bad sync word to initialize SX1276. This fixes that.

### DIFF
--- a/src/helpers/radiolib/CustomSX1276.h
+++ b/src/helpers/radiolib/CustomSX1276.h
@@ -39,7 +39,7 @@ class CustomSX1276 : public SX1276 {
       if (spi) spi->begin(P_LORA_SCLK, P_LORA_MISO, P_LORA_MOSI);
     #endif
   #endif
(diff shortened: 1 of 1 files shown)


liamcottle:
Hello, thanks for the PR!

From what I can tell, `RADIOLIB_SX126X_SYNC_WORD_PRIVATE` and `RADIOLIB_SX127X_SYNC_WORD` both resolve to `0x12`.

The error code `-2` you're receiving is defined as  `RADIOLIB_ERR_CHIP_NOT_FOUND` in `RadioLib/src/TypeDef.h`.

I don't see how this PR would resolve the issue you're facing, since it's telling you the radio chip was not found.

petrkr:
Me either, maybe there was also some HW glitch on duponts and I just did not expand both definitions to check.

But also it would be atleast more right atleast by names of definitions.

Also had error -13 which is something with tx power, so I lowered it TX POWER to 20 and it's working.

I am using this HW since TTN network, HNT Heiium Mestastick, ExpLora and now MeshCore. So it's old enough to had some HW issues too.

petrkr:
Fixed?

PR #2242, opened 2026-04-03
Implement group message blocking and management in MyMesh class

RichardPar:
Block Clients hash based on Routing -  This looks at Originator only. 

CLI Commands - 

set blocked 77FFFF - blocks 0x77****
set blocked 77ABFF - blocks 0x77AB**
set unblock 77FFFF - removes block
get blocked - shows all blocked addresses as 6-digit hex

Currently there is no persistence and I appreciate comments -


liamcottle:
Hey there, thanks for the PR. Unfortunately this isn't the approach we are looking to take right now, so will close out.

shadowdemon9956:
> Hey there, thanks for the PR. Unfortunately this isn't the approach we are looking to take right now, so will close out.

Could we have a solution for this Liam? we were told this would be available by AK on release and it has never happened. We are having issues with rouge users spamming abuse in the UK.

PR #2250, opened 2026-04-04
Heltec v3 without display repeater.

MeetTheTree:
I have two heltec v3 without display or with broken display. When flashing with the default firmware it is not possible to configure the nodes using the browser or webclient, because the nodes are stuck in the boot process. This PR solves the problem by omitting all display related stuff. So that the nodes can start and be configured even if the display is missing or there is a fault during the initialization of the I2C bus.


recrof:
just use WSL3 firmware for display-less v3.

PR #2251, opened 2026-04-04
shell.nix: provide adafruit-nrfutil

mweinelt:
This is marked unfree in nixpkgs upstream, which is something we need to explicitly allow in our shell.

Then we need to permit CVE-2024-23342 in ecdsa, a timing security issue which upstream does not plan to resolve, but isn't relevant to the use of adafruit-nrfutil at all.

The switch to `mkShellNoCC` makes the shell closure smaller, since it does not include the nixpkgs stdenv.

Renamed from default.nix, as shell.nix is the cleaner name for when the entrypoint is a nix shell. This does not require changes to the direnv configuration.

This increases the shell closure from 21.55 MiB to 21.66 MiB.

cc @neutralinsomniac


neutralinsomniac:
Thanks for the CC. I'll test this later today when I'm home :)

neutralinsomniac:
Looks good to me and worked fine!

PR #2259, opened 2026-04-05
fix: expose radio.rxgain CLI to Companion firmware

jirogit:
## Problem
`get/set radio.rxgain` commands were only available on Repeater and 
Room Server firmware via CommonCLI. Companion firmware returned 
"unknown config" for these commands.

This was noted as a follow-up in PR #1653 by jbrazio:
> "That could probably be added in a follow up PR"

## Changes
- Added `get radio.rxgain` to Companion CLI rescue mode
- Added `set radio.rxgain on/off` to Companion CLI rescue mode
- Setting is persisted via `savePrefs()` and applied immediately 
  via `radio_driver.setRxBoostedGainMode()`

## Testing
Tested on WisMesh Tag (nRF52840 + SX1262):
- `get radio.rxgain` → returns current state
- `set radio.rxgain on` / `set radio.rxgain off` → confirmed working
- Setting persists after reboot

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -2046,9 +2046,20 @@ void MyMesh::checkCLIRescueCmd() {
         _prefs.ble_pin = atoi(&config[4]);
         savePrefs();
         Serial.printf("  > pin is now %06d\n", _prefs.ble_pin);
+
+      } else if (memcmp(config, "radio.rxgain ", 13) == 0) {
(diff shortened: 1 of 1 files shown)


jirogit:
Additionally tested on Heltec T114 (nRF52840 + SX1262, without display):
- `get radio.rxgain` → returns current state
- `set radio.rxgain on` / `set radio.rxgain off` → confirmed working
- Setting persists after reboot

jirogit:
Rebased onto current dev, no conflicts.

PR #2264, opened 2026-04-06
Elegant OTA library update

petrkr:
This removes hard-coded old version of AsyncElegantOTA package and replaces it by official main version.

Also it updates AsyncWebServer

It also fixes ESP32-C6 router builds with OTA as new SDK is supported and there is no longer problem with missing esp_wdt.h as in old OTA version.

It is one of prerequest to bump ESP32 SDK to IDF 5.x and Arduino SDK to 3.x, which is also prerequest to support SPI Ethernet by ESP SDK.

Tested on T-Beam 1.2 SX1276 repeater


recrof:
there is very good reason we did not update the ElegantOTA - it changed license. one we use is the last version that is not dual licensed.

petrkr:
So then we need to fix it, because it is not compatible with new SDK and new Async web server

petrkr:
Well thanks to Michal on Telegram I made new version of this PR #2306 if that will be more suitable for you @recrof 

Tested on SDK 2.0.17 and on SDK 3.3.7 on T-Beam

liamcottle:
Hi @petrkr, thanks for the PR.

As @recrof noted above, we can't go with the library proposed in this PR since the license is GPLv3.

However, we will likely merge in your secondary PR #2306 as that bumps the existing library which is LGPLv3 licensed.

PR #2265, opened 2026-04-06
ESP32 Arduino SDK 3.x fixes

petrkr:
Here I will fix incompatible things with SDK 3.x IDF 5

- first thing is mentioned in #2264 
- far for now this contain fix of ESPNow! call back as it changed prototype

TODO for SDK 3 full support:
- Something reinitialize I2C, my guess was XPowerLibs on Tbeam, but if I will comment it out, it still does not works well (OLED for example)
- also PMU seem to works during board  initialization, but later in app is missing bat voltage/percent

```
DEBUG: Warning: Failed to find AXP2101 power management
DEBUG: AXP192 PMU init succeeded, using AXU...
        Found AXP192/AXP2101 PMU
        Found SSD1306/SH1106 display
Scan for devices is complete.

GPS RX pin: 12 GPS TX pin: 34

isCharging:YES
isDischarge:NO
isVbusIn:YES
getBattVoltage:4193mV
getVbusVoltage:5049mV
getSystemVoltage:4949mV
getBatteryPercent:100%

/// and since this one... So something deinitalized I2C or using different instance

[   792][E][esp32-hal-i2c.c:307] i2cSetClock(): bus is not initialized
[   798][E][esp32-hal-i2c.c:193] i2cWrite(): bus is not initialized
```

### DIFF
--- a/src/helpers/esp32/ESPNOWRadio.cpp
+++ b/src/helpers/esp32/ESPNOWRadio.cpp
@@ -1,5 +1,6 @@
 #include "ESPNOWRadio.h"
 #include <esp_now.h>
+#include <esp_mac.h>
 #include <WiFi.h>
 #include <esp_wifi.h>
 
--- a/src/helpers/esp32/ESPNOWRadio.cpp
+++ b/src/helpers/esp32/ESPNOWRadio.cpp
@@ -16,7 +16,11 @@ static void OnDataSent(const uint8_t *mac_addr, esp_now_send_status_t status) {
   ESPNOW_DEBUG_PRINTLN("Send Status: %d", (int)status);
 }
 
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0)
+static void OnDataRecv(const esp_now_recv_info_t *info, const uint8_t *data, int len) {
+#else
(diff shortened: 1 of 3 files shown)


petrkr:
Found what causing I2C issues.
Here
https://github.com/meshcore-dev/MeshCore/blob/dev/src/helpers/ui/SSD1306Display.h#L10-L12

id hard-coded PIN 21 if pin is not passed by board.

But on TBeam (and almost all others ESP32 boards) is PIN 21 used for I2C_SDA signal.

I suggest change it to -1 (same as for example here https://github.com/meshcore-dev/MeshCore/blob/dev/src/helpers/ui/SH1106Display.h#L10 )

Because now there are just two solutions (which I know)

1st is change it to -1 and OLED which will need it will pass it by -D compilation parameter (my guess there will be nothings like that, never saw OLED wit I2C and RST pin)

2nd keep it 21, but change every OLED board platformio.ini to -D PIN_OLED_RESET to -1

petrkr:
Rebased to latest dev

petrkr:
I also noticed there are some boards, where ESPNow sources are excluded due new SDK.. This should fix those boards too and they should support ESPNow too... Namely all c6 boards.

petrkr:
Rebased to latest dev

petrkr:
Rebase to latest dev

PR #2268, opened 2026-04-06
techo_lite: fixed RF switch

mlinkg:
This fixes the TX/RX switching for LilyGo_T-Echo-Lite. Without this modification, the attenuation of the TX path is approximately 15 dB.


oltaco:
Sorry we must have missed this PR, and it has just been superceded by https://github.com/meshcore-dev/MeshCore/pull/2511

PR #2020, opened 2026-03-13
Add support for Sensirion SHT3X

MKBastler:
Adds support for Sensirion SHT-3X sensors. Tested with SHT-31 and a Heltec-V3 sensor:
Screenshot_20260313_230001_MeshCore

This PR should address #565 and #1849

PR #2273, opened 2026-04-07
Added SHT3x temp sensor support

petrkr:
Added support for temperature sensor SHT3x

Tested on real HW

Tried to use SHT4x or SHT3HC library, but that was not compatible, so I added support for this one.


petrkr:
Hehe,

I did not noticed someone already did it too :)


#2020

So closing this one as implementation is basically same.

PR #2284, opened 2026-04-09
Fix the deadhang on the P1 Pro which could occur in rare cases

winnieXY:
Fix a deadhang bug which results in a unresponsive repeater after a initially successful login. 

Please note, that I wrote the patch without being able to test it yet - I'll build and test it this evening.  Please consider it currently untested.


winnieXY:
I'll provide a cleaner one.

PR #2287, opened 2026-04-10
Fixes #1183 — T-Echo Lite incorrect battery voltage

### LINKED ISSUE #1183: LilyGO T-Echo Lite displaying incorrect battery voltage
I have a T-Echo Lite Core (screenless version) flashed with the Meshcore v1.11 companion BLE firmware for the T-Echo Lite. It displays as 6.00v when a battery is connected, and also when no battery is connected and it's on USB C mains power. I'm using the iOS app. This error occurs when I've tried the Makerfocus 3.7v 1100mah battery, a generic li-on 3.7v 2000mah battery, and a Makerhawk 3.7v 3000mah battery. 

image
image
image

I've also tried changing the adcvalue in this bit of techoboard.cpp to 0.61 or 0.62 and it still reads as a 6v battery input regardless.
image

pelgraine:
The T-Echo Lite variant files were copied from the T-Echo without adapting to the Lite's different hardware. Three issues fixed, plus build environments added for the screenless Core variant.

**Battery (TechoBoard.h/cpp)**: Added PIN_VBAT_MEAS_EN (P0.31) — the T-Echo Lite has a gated voltage divider that must be enabled before reading. Also added pinMode(PIN_VBAT_READ, INPUT) before each ADC read to reclaim P0.02 from other peripherals. Pin definitions hardcoded from LilyGo's t_echo_lite_config.h.

**I2C (variant.h)**: Corrected SDA/SCL from P0.04/P0.02 to P1.04/P1.02 per LilyGo's IIC_1_SDA/IIC_1_SCL. The old P0.02 mapping conflicted with the battery ADC pin.

**GPS (variant.h)**: Corrected all five GPS pin assignments to match LilyGo's config — UART TX/RX, wake, PPS, and power enable were all scrambled.

Tested on T-Echo Lite Base (screenless):
H747-01_01

Battery readings now match Heltec V4 comparison reference within 10mV - Heltec v4 read partially discharged battery as 3.51v, screenshots of T-Echo Lite readings below:

IMG_2110
IMG_2111
IMG_2112

### DIFF
--- a/variants/lilygo_techo_lite/TechoBoard.cpp
+++ b/variants/lilygo_techo_lite/TechoBoard.cpp
@@ -8,24 +8,47 @@
 void TechoBoard::begin() {
   NRF52Board::begin();
 
+  // Configure battery measurement control BEFORE Wire.begin()
+  // to ensure P0.02 is not claimed by another peripheral
+  pinMode(PIN_VBAT_MEAS_EN, OUTPUT);
+  digitalWrite(PIN_VBAT_MEAS_EN, LOW);
+  pinMode(PIN_VBAT_READ, INPUT);
+
   Wire.begin();
 
   pinMode(SX126X_POWER_EN, OUTPUT);
   digitalWrite(SX126X_POWER_EN, HIGH);
-  delay(10);   // give sx1262 some time to power up
+  delay(10);
 }
 
 uint16_t TechoBoard::getBattMilliVolts() {
-  int adcvalue = 0;
-
+  // Use LilyGo's exact ADC configuration
   analogReference(AR_INTERNAL_3_0);
   analogReadResolution(12);
-  delay(10);
 
-  // ADC range is 0..3000mV and resolution is 12-bit (0..4095)
-  adcvalue = analogRead(PIN_VBAT_READ);
-  // Convert the raw value to compensated mv, taking the resistor-
-  // divider into account (providing the actual LIPO voltage)
-  return (uint16_t)((float)adcvalue * REAL_VBAT_MV_PER_LSB);
+  // Enable battery voltage divider (MOSFET gate on P0.31)
+  pinMode(PIN_VBAT_MEAS_EN, OUTPUT);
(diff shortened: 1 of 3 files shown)


jirogit:
Hi @pelgraine — a few questions to help clarify which variant this PR 
actually targets, since LilyGO's naming is confusing many people 
(including me):

1. **Which hardware do you have?** The PR title says "Core" but the 
   photos in here looks Non-Shell. Could you confirm 
   whether your device is:
   - T-Echo Lite Core (castellated module, no USB, no display)
   - T-Echo Lite Non-Shell (has USB-C, no display)
   - T-Echo Lite (full version with ePaper display)

2. **Naming convention**: The new base env `LilyGo_T-Echo-Lite-Core` 
   uses a hyphen before "Core", while the rest of the codebase uses 
   underscores (e.g. `Heltec_t114_without_display_companion_radio_ble`).
   Also, `_base` suffix is used for PlatformIO base envs elsewhere 
   (e.g. `Heltec_E213_base`), while `core` does not appear anywhere 
   in the codebase. Worth aligning with maintainer preference.

I ask because my PR #2353 targets the Non-Shell variant specifically,
and I want to make sure we cover all three variants correctly without
overlap or gaps.

pelgraine:
@jirogit 

Hi, yes, I shouldn't have put the platformio changes in. This fix is really just about the battery voltage problem detailed in issue #1183, not the headless build problem. The one I worked with was actually the T-Echo Lite Non-Shell (has USB-C, no display) version. When I have a spare minute I'll fix the PR to remove those.

pelgraine:
Updated to remove platformio changes and references to headless env.

Also tested with my new T-Echo Lite with shell & screen with the T-Echo Lite companion radio BLE build and can confirm that displays as expected.

IMG_2413

IMG_2414

pelgraine:
@jirogit all updated!

IoTThinks:
It works on my T-Echo lite.
Perfect.

Thanks a lot. 👯‍♂️

PR #2290, opened 2026-04-11
Add LilyGo T-Beam 1W SX1262 board support

mintylinux:
Add variant definition, board JSON, and TBeamBoard integration for the LilyGo T-Beam 1W with SX1262 radio and 1W PA.

Includes:
- Board variant with pin mappings for SX1262, GPS, OLED, fan control
- AXP2101 PMU power rail configuration for T-Beam 1W
- ADC-based battery voltage fallback (2S Li-ion, 7.4V nominal)
- Temperature-based fan control for 1W PA cooling
- GPIO fallback for GPS power when PMU is not present
- Companion BLE, Companion USB, Repeater, Room Server, and ESPNow Bridge environments


IoTThinks:
@mintylinux How is this PR compared to the current T-Beam 1W in MeshCore 1.16?
Is this an improvement or alternative to the current T-Beam 1W?
Thanks a lot.
https://github.com/meshcore-dev/MeshCore/tree/main/variants/lilygo_tbeam_1w

PR #2295, opened 2026-04-12
Fixes #2294 : RX path not enabled on Waveshare RP2040

### LINKED ISSUE #2294: Waveshare RP2040 : RX path not enabled (very poor sensitivity)
Hello,

On all Waveshare RP2040-LoRa boards, reception is severely degraded while transmission works correctly. The board receives almost nothing, or only very close/strong signals.

The board uses a PE4259 RF switch with two complementary control lines:

Signal	Pin	Driven by
CTRL	SX1262 DIO2	SX1262 internally via SX126X_DIO2_AS_RF_SWITCH
!CTRL	GPIO17	not configured in MeshCore
The PE4259 requires both pins to be driven in complementary logic. Without !CTRL being driven:

TX: DIO2=HIGH forces the TX path — works correctly
RX: DIO2=LOW but !CTRL floats — the switch never selects the LNA/RX path → reception is completely degraded

Fix : 
Add SX126X_RXEN=17 to variants/waveshare_rp2040_lora/platformio.ini. This tells RadioLib to drive GPIO17 as the complement of DIO2, correctly switching the PE4259 to the LNA path during receive.

Additionally, SX126X_DIO3_TCXO_VOLTAGE should be 1.8 (not 0): the board uses an active TCXO on DIO3 at 1.8V according to the hardware schematic.

-D SX126X_DIO2_AS_RF_SWITCH=true
-D SX126X_RXEN=17                   ← add this
-D SX126X_DIO3_TCXO_VOLTAGE=1.8     ← fix from 0
References :
Waveshare RP2040-LoRa schematic: https://files.waveshare.com/wiki/RP2040-LoRa/Rp2040-lora-sch.pdf
PE4259 datasheet page 4 (complementary-pin control logic)

We have tested this patch on multiple RP2040 nodes, and the result is very good. We start from -10SNR to something like +9SNR. Depending on each  node conditions.
Example after patching (around 12:00) : 

Image

The PR is coming very soon.

TheMalkavien:
Fixes #2294

Summary
Add SX126X_RXEN=17 to configure the complementary RF switch control pin (GPIO17 = !CTRL of the PE4259)
Fix SX126X_DIO3_TCXO_VOLTAGE from 0 to 1.8 to correctly power the on-board TCXO

The Waveshare RP2040-LoRa board uses a PE4259 RF switch with two complementary control lines:

Pin	Signal	Driven by
DIO2	CTRL	SX1262 internally (via SX126X_DIO2_AS_RF_SWITCH)
GPIO17	!CTRL	was not configured
Without !CTRL driven, the RF switch never properly selects the LNA/RX path — reception is completely degraded. TX is unaffected because DIO2=HIGH alone activates the TX path. Adding SX126X_RXEN=17 lets RadioLib drive GPIO17 complementary to DIO2, correctly routing the switch to the LNA on receive.

The TCXO fix corrects a secondary issue: the board has an active TCXO on DIO3 requiring 1.8 V; the previous value of 0 disabled DIO3 TCXO control entirely.

Reference
Waveshare RP2040-LoRa schematic: https://files.waveshare.com/wiki/RP2040-LoRa/Rp2040-lora-sch.pdf

It has been tested on multiple nodes (exemple in the issue)


TheMalkavien:
I'v selected the wrong branch sry... Another PR is coming

PR #2298, opened 2026-04-12
Fixes #2294 : RX path not enabled on Waveshare RP2040

### LINKED ISSUE #2294: Waveshare RP2040 : RX path not enabled (very poor sensitivity)

TheMalkavien:
Fixes #2294

The Waveshare RP2040-LoRa board has an RF switch with two complementary control lines:
- DIO2 (CTRL)  -- driven automatically by the SX1262 chip: HIGH on TX, LOW on RX
- GPIO17 (!CTRL) -- was not configured in MeshCore, so it floated and the switch never properly selected the LNA/RX path

Without this pin driven, the RF switch never switches to RX mode, resulting in completely degraded reception. TX was unaffected because DIO2 alone is sufficient to activate the TX path.

Adding SX126X_RXEN=17 lets RadioLib drive GPIO17 as the complement of DIO2, so the switch correctly routes the signal to the LNA on receive.

Reference: https://files.waveshare.com/wiki/RP2040-LoRa/Rp2040-lora-sch.pdf


variants/waveshare_rp2040_lora/platformio.ini
```diff
@@ -20,6 +20,7 @@ build_flags = ${rp2040_base.build_flags}
   -D P_LORA_MOSI=15
   -D P_LORA_TX_LED=25
   -D SX126X_DIO2_AS_RF_SWITCH=true
+  -D SX126X_RXEN=17
```
recrof, line 23: checked and works.. can be merged.

recrof (approved):

PR #2299, opened 2026-04-13
Add WiFi reconnection to companion radio

### LINKED ISSUE #2300: Bug: WiFi companion radio does not reconnect after losing AP connectivity
## Description

The WiFi companion radio firmware (`heltec_v4_companion_radio_wifi`) calls `WiFi.begin()` once at startup with no reconnection handling. If the AP goes down or the device loses WiFi connectivity after the initial connection, it never reconnects.

## Root Cause

The project uses `espressif32@6.11.0` which ships with ESP32 Arduino core 2.x (ESP-IDF 4.4). In this version, the built-in `WiFi.setAutoReconnect()` is unreliable — it doesn't persist through certain disconnect reasons like AP reboots, beacon loss, or DHCP timeouts.

## Expected Behavior

The device should periodically detect WiFi disconnection and attempt to reconnect automatically.

## Fix

#2299

swannman:
## Summary
- Adds periodic WiFi health check (every 60s) in `SerialWifiInterface::checkRecvFrame()`
- Only attempts reconnect after WiFi has successfully connected at least once (won't interfere with initial connection on startup)
- Fixes issue where companion radio WiFi firmware never rejoins the AP after losing connectivity

Fixes #2300

## Context
On ESP32 Arduino core 2.x (espressif32@6.11.0), the built-in `WiFi.setAutoReconnect()` is unreliable. This adds an explicit reconnection loop as a more dependable alternative.

## Test plan
- [x] Build `heltec_v4_companion_radio_wifi` target
- [x] Flash and confirm WiFi connects on startup (IP shown on display)
- [x] Disable/reboot the AP, wait >60s, re-enable AP
- [x] Confirm device reconnects and IP reappears on display
- [x] Cold boot with AP unavailable — confirm no reconnect spam

### DIFF
--- a/src/helpers/esp32/SerialWifiInterface.cpp
+++ b/src/helpers/esp32/SerialWifiInterface.cpp
@@ -53,6 +53,15 @@ void SerialWifiInterface::resetReceivedFrameHeader() {
 }
 
 size_t SerialWifiInterface::checkRecvFrame(uint8_t dest[]) {
+  // periodic WiFi health check (every 60s, only after first successful connection)
+  if (WiFi.status() == WL_CONNECTED) {
+    _wifi_was_connected = true;
+  } else if (_wifi_was_connected && millis() >= _wifi_check_time) {
+    WIFI_DEBUG_PRINTLN("WiFi disconnected, attempting reconnect...");
+    WiFi.reconnect();
+    _wifi_check_time = millis() + 60000;
+  }
+
   // check if new client connected
   auto newClient = server.available();
   if (newClient) {
--- a/src/helpers/esp32/SerialWifiInterface.h
+++ b/src/helpers/esp32/SerialWifiInterface.h
@@ -8,6 +8,8 @@ class SerialWifiInterface : public BaseSerialInterface {
   bool _isEnabled;
   unsigned long _last_write;
   unsigned long adv_restart_time;
+  bool _wifi_was_connected;
+  unsigned long _wifi_check_time;
 
   WiFiServer server;
   WiFiClient client;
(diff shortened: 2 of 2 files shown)


petrkr:
Is this needed? ESP IDF should take care about reconnects itself..
Atleast it works on other projects very well even if AP is down for days.

Would be better update SDK than making workarounds.

#2265

swannman:
Hi @petrkr - agreed that ESP-IDF has built-in reconnect support. We actually tried the simpler approach first (`WiFi.mode(WIFI_STA)` + `WiFi.setAutoReconnect(true)` before `WiFi.begin()`), but it didn't resolve the issue on this board with the current platform version (espressif32@6.11.0 / Arduino core 2.x).

Upgrading the SDK would be ideal long-term, but it's a significant change — the Arduino core 2.x to 3.x migration has breaking API changes across WiFi, BLE, SPI, etc., and would affect every ESP32 build target in the project. Felt too risky for this one fix.

petrkr:
Only 1 change with import, nothing breaking.. claude said this to you? It is wrong. Check code as human programmer first, then you will not write those missleading informations

swannman:
Hi @petrkr, could you clarify what you're referring to as "1 change with import"? 

I don't think you're suggesting that I fix this by bumping `espressif32@6.11.0` in `platformio.ini` to `6.13.0`, since that isn't where the WiFi reconnect behavior is implemented.

Are you suggesting that I take on the larger update to get us onto Arduino core 3.x with the ESP-IDF 5.x WiFi stack? If so, based on your PR #2265 this looks like a larger change with back-compat impact.

petrkr:
One import is in elegant ota #2306 

And for 3.0 upto 3.4 it's just 1 line in espnow
For 3.5 it's 2nd line in espnow

Everything is done in that PR and it is backward compatible

swannman:
Thanks for the context @petrkr. It sounds like your Arduino 3.x upgrade is still in progress and blocked on the OTA issue (#2304). When that lands, I agree that we should re-test this scenario without the changes from this PR and revert them if no longer needed.

PR #2306, opened 2026-04-14
SDK3.x: Fix OTA includes, bump Async version

petrkr:
Fix build issues with SDK 3.x,

Removed dead code includes in old ElegantOTA

This may replace #2264 update. It will stay on old version, but it works with SDK 3.x aswell

### DIFF
--- a/arch/esp32/AsyncElegantOTA/src/AsyncElegantOTA.h
+++ b/arch/esp32/AsyncElegantOTA/src/AsyncElegantOTA.h
@@ -15,8 +15,6 @@
     #include "WiFi.h"
     #include "AsyncTCP.h"
     #include "Update.h"
-    #include "esp_int_wdt.h"
-    #include "esp_task_wdt.h"
 #endif
 
 #include "Hash.h"
--- a/platformio.ini
+++ b/platformio.ini
@@ -66,7 +66,7 @@ build_src_filter = ${arduino_base.build_src_filter}
 
 [esp32_ota]
 lib_deps =
-  me-no-dev/ESPAsyncWebServer @ ^3.6.0
+  ESP32Async/ESPAsyncWebServer @ 3.10.3
   file://arch/esp32/AsyncElegantOTA
 
 ; esp32c6 uses arduino framework 3.x
--- a/variants/lilygo_tlora_c6/platformio.ini
+++ b/variants/lilygo_tlora_c6/platformio.ini
@@ -46,7 +46,7 @@ build_flags =
 ;  -D MESH_DEBUG=1
 lib_deps =
   ${tlora_c6.lib_deps}
-;  ${esp32_ota.lib_deps}
+  ${esp32_ota.lib_deps}
 
 [env:LilyGo_Tlora_C6_room_server_]
 extends = tlora_c6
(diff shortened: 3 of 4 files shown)


recrof:
@ripplebiz this seems as good addition - esp32c6 boards were cut-off from OTAs because of incompatibility with Arduino 3.x

liamcottle:
Tests we have done:

- @liamcottle tested building `Heltec_v2_repeater` to ensure it didn't cause any issues with ram overflow.
- @oltaco tested an OTA update with `Heltec_v3_repeater` to ensure the library change works as expected.

Will merge, thanks! :)

fizzyfuzzle:
@liamcottle was it also tested on this new firmware to update over OTA to another new firmware?

Currently I am running the **Dev branch** from **May 11th** and trying to do an OTA to the **current Dev branch** which does not seem to work.

Device: Heltec V4.3

- 'start ota' works
- uploading firmware.bin through the access point webpage also works
- device reboots
- **still previous firmware version**

So basically everything works, except that it stays on the previous firmware version.

petrkr:
> @liamcottle was it also tested on this new firmware to update over OTA to another new firmware?
> 
> Currently I am running the **Dev branch** from **May 11th** and trying to do an OTA to the **current Dev branch** which does not seem to work.
> 
> Device: Heltec V4.3
> 
> * 'start ota' works
> * uploading firmware.bin through the access point webpage also works
> * device reboots
> * **still previous firmware version**
> 
> So basically everything works, except that it stays on the previous firmware version.

This is something what I saw on different project.

Maybe there is meanwhile something what is initialized before setup() so it will 'eats' bootup routine, which calls 'ota succesfully'.

https://github.com/richonguzman/LoRa_APRS_iGate/pull/405

recrof:
> So basically everything works, except that it stays on the previous firmware version.

are you sure you're not uploading the merged version of firmware? that's the most common problem when it refuses to update

fizzyfuzzle:
@recrof 100% sure ;)

I did some troubleshooting tests and it seems that code before this PR does still work, so if I upload the official v1.15.0 firmware it will update again.
But when I then want to upload the latest Dev code it will NOT update again.

So I guess @petrkr might be onto something with this 'setup()' not being called in latest code?

It seems uploading the firmware works, but actually updating fails after this PR. But running the latest Dev still allows to update to code BEFORE this PR it seems.

- Running code BEFORE this PR -> Upgrade -> Code AFTER this PR = :x: 
- Running code AFTER this PR -> Upgrade -> Code BEFORE this PR = :heavy_check_mark:

petrkr:
> @recrof 100% sure ;)
> 
> I did some troubleshooting tests and it seems that code before this PR does still work, so if I upload the official v1.15.0 firmware it will update again. But when I then want to upload the latest Dev code it will NOT update again.
> 
> So I guess @petrkr might be onto something with this 'setup()' not being called in latest code?
> 
> I seems uploading the firmware works, but actually updating fails after this PR. But running the latest Dev still allows to update to code BEFORE this PR it seems.
> 
> * Running code BEFORE this PR -> Upgrade -> Code AFTER this PR = 🔴
> * Running code AFTER this PR -> Upgrade -> Code BEFORE this PR = ✔️

Then try return

    #include "esp_int_wdt.h"
    #include "esp_task_wdt.h"

lines. But those are not compatible with SDK3.x+ So it must be solved somehow differently in  iGate, because there they are using official latest Elegant OTA, so they may solved it by different way.

Then we must also test if it affects only SDK 2.x, if so, then FIX will be put this ti #ifdef version block

fizzyfuzzle:
Hmm it seems like it was because of my patched Dev branch for some reason.

Firmware v1.15.0 already got this PR merged, so I dont think this PR has anything to do with the problem after all :sweat_smile: 

So for now you can just ignore my comments :nerd_face:

PR #2323, opened 2026-04-17
Add support for Companion BLE OTA updates on nRF devices

liamcottle:
This PR adds support for BLE OTA updates to MeshCore companion devices running on nRF based boards.
Thanks to @txkbaldlaw for the original changes. I've opened the PR on your behalf so you get credits for the commits.

When I originally looked into BLE OTA updates for companion devices a few months ago, I had made the same changes in this PR. However, I ran into several issues. One was mainly due to the outdated bootloader that comes preflashed by the factory on most devices sold by manufacturers.

The outdated bootloader would fail to update firmware with the _default settings_ provided by the nRF DFU app for Android and iOS, and because it pre-wipes the flash when the BLE OTA process is started, it was not possible to reboot the device back into MeshCore companion firmware without a full reflash via USB.

Another issue is that if you had already paired with the device on Android _without_ the changes in this PR, you would need to forget the device in Android Bluetooth settings, otherwise the OTA would fail with an error saying the device doesn't not support nRF DFU. This is most likely due to the Android device caching the BLE services/characteristics at pairing time. So an unpair/repair is needed to discover that the device supports the nRF BLE OTA service...

Please see the following list of things to keep in mind when using BLE OTA for companion devices:
- These changes will likely be in MeshCore Companion firmware v1.15.0+
- You must flash MeshCore companion firmware that contains this patch before BLE OTA is supported.
- You must unpair the MeshCore device in Android BLE settings if nRF DFU app says it doesn't support nRF DFU.
- Most nRF boards come with an outdated bootloader, which usually fails OTA and requires reflashing via USB.
- It's recommended to update to the [OTAFIX bootloader](https://github.com/oltaco/Adafruit_nRF52_Bootloader_OTAFIX) for the best experience.
- If you don't update to OTAFIX bootloader, you must set packet receipts to `4` in the nRF DFU app for OTA to work, otherwise it fails 100% of the time due to this bug: https://github.com/adafruit/Adafruit_nRF52_Bootloader/issues/306

Please see our blog posts for more information on how to update the bootloader and perform a BLE OTA update:
- https://blog.meshcore.io/2026/04/06/otafix-bootloader
- https://blog.meshcore.io/2026/04/02/nrf-ota-update

These changes have been tested on the WioTracker L1 Pro, with both the outdated bootloader and the OTAFIX bootloader.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -801,7 +801,8 @@ Allow the browser user on it:
 13. If it fails, try turning off and on Bluetooth on your phone.  If that doesn't work, try rebooting your phone.  If you keep getting failures at the "Enabling Bootloader" step, try forgetting the NRF board in your IOS or Andriod device's bluetooth settings and re-pair it through the DFU app.
 14. Wait for the update to complete.  It can take a few minutes.   
 15. It is strongly recommended that you install and use the OTAFIX bootloader at https://github.com/oltaco/Adafruit_nRF52_Bootloader_OTAFIX. 
-16. Please see the Meshcore Blog for additional information on OTA firmware flashing: 
+16. To update a companion node over OTA, it must be running companion firmware v1.15 or greater.   
+17. Please see the Meshcore Blog for additional information on OTA firmware flashing: 
     - https://blog.meshcore.io/2026/04/06/otafix-bootloader
     - https://blog.meshcore.io/2026/04/02/nrf-ota-update
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -89,7 +89,7 @@ A list of frequently-asked questions and answers for MeshCore
   - [6.7. Q: My RAK/T1000-E/xiao\_nRF52 device seems to be corrupted, how do I wipe it clean to start fresh?](#67-q-my-rakt1000-exiao_nrf52-device-seems-to-be-corrupted-how-do-i-wipe-it-clean-to-start-fresh)
   - [6.8. Q: WebFlasher fails on Linux with failed to open](#68-q-webflasher-fails-on-linux-with-failed-to-open)
 - [7. Other Questions:](#7-other-questions)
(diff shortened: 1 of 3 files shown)


txkbaldlaw:
Thanks @liamcottle!   I appreciate you and @recrof's help working through this.  

I have experienced the same need to forget the BLE device to get OTA updates to work on IOS devices running the Nordic DFU app,  even prior to the changes in this PR.   I have noticed that it would fail immediate when executing the step to try enabling the bootloader.  If I forgot the device in BLE settings, and then re-paired through the DFU app, it worked fine, and continues to work going forward, both in the DFU app and the MeshCore app.  

I have also tested this PR on both  RAK4631 and RAK3401 boards and it works as intended.  

Using NRF Connect for Mobile, the Legacy DFU service now appears in the BLE stack where it did not before.   

IMG_8084

txkbaldlaw:
Updated the FAQ to reference Companion OTA.

LitBomb (commented):
Thanks for updating the FAQ to match the new functionality.  

there is a gap between 1.14.1 and the new FAQ information.  If I understand it correctly from the discussion, one suggestion I have is to add some info to remind users that this feature only works with 1.15 firmware or newer, and that 1.15 needs to be already flashed onto the companion radio before BLE OTA can work.

txkbaldlaw:
> Thanks for updating the FAQ to match the new functionality.
> 
> there is a gap between 1.14.1 and the new FAQ information. If I understand it correctly from the discussion, one suggestion I have is to add some info to remind users that this feature only works with 1.15 firmware or newer, and that 1.15 needs to be already flashed onto the companion radio before BLE OTA can work.

Additional language added as requested.   Thanks!

LitBomb (approved):
Thanks for the update!

Issue #2222, opened 2026-04-01
RAK board WisBlock 3V3_S power rail initialization


benders:
[ Expanding the discussion from #1874 and #2215 ]

On RAK boards, the `WB_IO2` pin also controls the 3V3_S power rail, which powers the WisBlock slots. That can include environmental sensors, GNSS, as well as the OLED screen. To operate properly we need to ensure that the power is brought up before probing modules, and that modules that need initialization (like the OLED display) are re-initialized anytime that power is cycled.

Some partial fixes for this already exist in the RAK3401 variant, but not in the RAK4631 variant. A proper fix should create a common path for all RAK boards, including cleanup to GPS module probing, and making sure that environment sensor modules work correctly in any slot configuration.

There is also the question of future power management improvements. If none of the connected modules are needed at any given time, the 3V3_S rail could be shutdown to save power. But doing that maintainably means creating some common control points and ensuring that all modules that need it get re-initialized after power comes back.

KPrivitt:
There are numerous variant.cpp files.  Do you know which one is used for the RAK190007,..., series boards?

I believe it's one of these two.

/variants/RAK4631/variant.cpp
/variants/rak_wismesh_tag/variant.cpp

I presume we don't need to update all the various variants, just one for these RAK boards.

Are you going to make the changes to this file to turn on IO2, IO4 & IO5?

I have a updated version of EnvironmentalSensorManager() as part of my driver, which I could modify to fix the cycling of IO2, IO4 & IO5 and remove the pinMode change to input. But we aren't supposed to mix issues (one PR per one fix).  I can make the change and send you the pieces of the two updated functions (rakGPSInit() and isGPSAwake())

With these two pieces (updated initVariant() & updated rakGPSInit()...) it should be ready to PR, to fix the bug that 3v3_S is not being properly initialized.

The architecture of sensor power management would be a future modification (do you have the Meshtastic firmware, they have implemented this and it could be an example of one way to approach this, it would need to be studied to understand how they coordinate turning the power on/off with the numerous sensors)

recrof:
rak_wismesh_tag is not standard wisblock - it has fixed configuration that you can't change.. would not touch it if something is broken.
please refer to https://store.rakwireless.com/collections/wisblock-core-fr to see which wisblock cores are able to support various RAK sensors/addons - we support few of them in meshcore codebase.

KPrivitt:
[...]

PR #2327, opened 2026-04-18
fix(sensors): improve sensor initialization and handling to prevent hangs and handle growth

NickDunklee:
This is a medium-ish refactor to attempt to clean up sensor handling logic both for board stability and future potential growth before the code becomes all spaghetti and meatballs.

I'd be curious to see if anyone running sensors out there that knows how to build and flash MeshCore code could give this a try and see how it behaves. It is working fine on my end on multiple nodes.

PR notes are gigantic because it is a fundamental behavior repair for sensors, so I wanted to over-explain. Also, if it hadn't been mentioned previously, push-back is always welcome. I'm just spending my time trying to clean up / fix / enhance this corner of the firmware, and want to contribute my improvements back to the project.

**Problem:**

Current MeshCore code makes no attempt to see what sensors are actually available on the I2C bus at startup and blindly tries to interact with sensors. This has some very bad side-effects, like if a sensor that is unsupported, or has a weird initialization process, the MeshCore node will just hang at boot and never successfully start up and ostensibly looks bricked, or the INA226 and SHT4X both sharing the same address and the code just silently fighting.

The current implementation also gloms sensor readouts from the MCU and environment sensors onto the same telemetry channel, with some arbitrary exceptions for incrementing channels based on certain behavioral situations. The MCU temperature and external temperature sensors would appear on channel 1, and it wouldn't be possible to tell which sensor the temperature value was coming from.

Per [CayenneLPP](https://github.com/myDevicesIoT/CayenneLPP): *Data Channel: Uniquely identifies each sensor in the device across frames, eg. “indoor sensor”* So this channel division implementation falls inline with what CayenneLPP intended. There are up to 256 channels available. So I tried to model this change in that behavioral style.

**Proposed Improvement:**

This implementation scans the I2C bus for what devices are present, sets each sensor to its own CayenneLPP channel, and keeps MCU telemetry on channel 1 only. So Channel 1 is always "self" and no confusion can result.

Details:
  - Channel 1 is always the MCU and things about it, so you always know that telemetry is from the board itself. Exception is GPS, GPS stays on channel 1 as well since it is "about the board" even though it's a bit gray-area as GPS can often be a secondary chip.
  - Each sensor board is allocated to a dedicated CayenneLPP channel, so if you are reading from that channel, you know the data is from that sensor only. (Sensors emitting more than one of the same type of measurement are exceptions.)
  - `scanI2CBus()` probes addresses 0x08–0x77 with raw `beginTransmission`/`endTransmission`. No sensor library is touched until after this completes. This will prevent sensor-based boot hangs, unknown or unresponsive devices never reach a library init call.
  - Created `SENSOR_TABLE` a compile-time array that is gated by the existing `ENV_INCLUDE_*` macros. A sentinel `{ 0, nullptr, nullptr, nullptr }` at the end keeps the array non-empty regardless of which sensors are enabled, avoiding zero-length array warnings.
  - When `begin()` is called, scan first, then loop: skip if address not detected, skip if `init()` returns 0, otherwise register one ActiveSensor entry per sub-channel.
  - `querySensors()` I replaced the entire #ifdef chain with a 3-line loop.
  - T1000-E has its own T1000SensorManager, so it should be completely unaffected by this change.
  - SHT4X quirky initialization behavior is retained.
  - MLX90614  - git commits around this didn't have any notes as to why it is reporting ambient temperature on a separate channel as well as the object temperature, as the ambient temperature is used internally to compute the object temperature and not really needed for the sensor's purpose - just the same, kept the existing behavior of reporting the ambient temperature one channel above the channel assigned to the sensor
  - All `bool *_initialized` fields are gone, replaced with `ActiveSensor _active_sensors[16]` (query function pointer and sub-channel index) and `_active_sensor_count. SensorDef` lives entirely in the `.cpp` so the header has no dependency on it.
  - Details on the INA226 and SHT4X: both default to address 0x44, the old code had a bug and would have both begin() calls fire and they would just fight each other silently. In the new code, the respective sensor code is only called if the device is actually present, however, if both were present simultaneously, SHT4X comes first in the table and would win, and INA226 would return false and be skipped. The INA226 has 16 possible addresses that are configurable in the hardware itself, so in a potential scenario where both sensors would be present, the person implementing that design could take that into account.
  - BME680 gas resistance will now transmit on the same channel as the rest of BME680 telemetry which is inline with CayenneLPP standards. Coupling this PR with https://github.com/meshcore-dev/MeshCore/pull/2146 streamline the whole sensor telemetry, and with https://github.com/meshcore-dev/MeshCore/pull/2149 will overall improve BME680 handling. The gas resistance sensor actually has a binary library to make it more useful, calibration, accounting for age of sensor, and other improvements, but since that adds more flash consumption, I have omitted that in PRs thus far.
 - RAK12035 and other current upstream dev branch changes integrated.

### DIFF
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -6,6 +6,10 @@
 #define TELEM_WIRE &Wire  // Use default I2C bus for Environment Sensors
 #endif
 
+// ============================================================
+// Sensor library includes and static driver instances
+// ============================================================
+
 #ifdef ENV_INCLUDE_BME680
 #ifndef TELEM_BME680_ADDRESS
 #define TELEM_BME680_ADDRESS 0x76
@@ -31,7 +35,7 @@ static Adafruit_AHTX0 AHTX0;
 #ifndef TELEM_BME280_ADDRESS
 #define TELEM_BME280_ADDRESS    0x76      // BME280 environmental sensor I2C address
 #endif
-#define TELEM_BME280_SEALEVELPRESSURE_HPA (1013.25)    // Athmospheric pressure at sea level
+#define TELEM_BME280_SEALEVELPRESSURE_HPA (1013.25)    // Atmospheric pressure at sea level
 #include <Adafruit_BME280.h>
 static Adafruit_BME280 BME280;
 #endif
@@ -40,7 +44,7 @@ static Adafruit_BME280 BME280;
 #ifndef TELEM_BMP280_ADDRESS
 #define TELEM_BMP280_ADDRESS    0x76      // BMP280 environmental sensor I2C address
 #endif
-#define TELEM_BMP280_SEALEVELPRESSURE_HPA (1013.25)    // Athmospheric pressure at sea level
+#define TELEM_BMP280_SEALEVELPRESSURE_HPA (1013.25)    // Atmospheric pressure at sea level
 #include <Adafruit_BMP280.h>
 static Adafruit_BMP280 BMP280(TELEM_WIRE);
 #endif
@@ -51,7 +55,7 @@ static Adafruit_SHTC3 SHTC3;
 #endif
 
 #if ENV_INCLUDE_SHT4X
-#define TELEM_SHT4X_ADDRESS 0x44  //0x44 - 0x46
+#define TELEM_SHT4X_ADDRESS 0x44
 #include <SensirionI2cSht4x.h>
 static SensirionI2cSht4x SHT4X;
 #endif
@@ -63,7 +67,7 @@ LPS22HBClass LPS22HB(*TELEM_WIRE);
 
 #if ENV_INCLUDE_INA3221
 #ifndef TELEM_INA3221_ADDRESS
-#define TELEM_INA3221_ADDRESS   0x42      // INA3221 3 channel current sensor I2C address
+#define TELEM_INA3221_ADDRESS     0x42    // INA3221 3 channel current sensor I2C address
 #endif
 #ifndef TELEM_INA3221_SHUNT_VALUE
 #define TELEM_INA3221_SHUNT_VALUE 0.100 // most variants will have a 0.1 ohm shunts
@@ -88,9 +92,9 @@ static Adafruit_INA260 INA260;
 #endif
 
 #if ENV_INCLUDE_INA226
-#define TELEM_INA226_ADDRESS    0x44
+#define TELEM_INA226_ADDRESS     0x44
 #define TELEM_INA226_SHUNT_VALUE 0.100
-#define TELEM_INA226_MAX_AMP 0.8
+#define TELEM_INA226_MAX_AMP     0.8
 #include <INA226.h>
 static INA226 INA226(TELEM_INA226_ADDRESS, TELEM_WIRE);
 #endif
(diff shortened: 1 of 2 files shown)


NickDunklee:
Examples from some nodes:
Node 1:
node1-2
node1-1

Node 2:
node2-2
node2-1

KPrivitt:
This is better, but with multiple sensors attached, it would be nice to display the sensor name so as you scroll down the possible 256 channels you can find the one you are looking for (what if it is 233, how do you know it's not 232 which looks identical) If this isn't possible this is not really making it better from a user perspective.

From my understanding CayenneLPP doesn't provide a simple text string output.  I couldn't find one.  I've been told CayenneLPP is not part of the MeshCore source code and is not modifiable here (done through BlueFruit??)

A secondary "issue" is sensor power management.  On RAK boards most sensors are on a common power plane, by knowing what sensors are present and having each sensor have its own initialization the power plane could be controlled to save power and brought up/down as needed.  This would be a significant improvement.

As you noted Sensor management is currently a mess.

Last note, not all sensors are on the I2C, some are on the serial interface and there is more than one I2C interface (wire and/or wire1, depending on the board used)

Good luck in getting this past the devs...

NickDunklee:
I had a thought to map sensor classes by channel banks to at least filter via inference, like assign channels 2-5 for environment, 6-9 to electrical, and such, so at least you would know if a sensor is in that range you could infer it is probably the one you want. It seems CayenneLPP doesn't have a "standard" for channel banks, though, so I figured going simple/generic for now would at least be a nudge in a more organized direction.

Not trying to step on toes either, just trying to kick code I'm working on back upstream in case it is useful to the mesh in general, and trying to adhere to a "simple, not too big" change format, although this one obviously became larger. I want to maintain optimum small code and ideally minimal power consumption too, while still trying to adhere to some design best practices. I have too much time to spend on this as I've been laid off for a bit, so I'm sure my active PRs will reduce to a trickle if I can fix _that_ problem sooner than later, haha.

Since I already had issues with bringing a previous sensor online due to how sensor initialization happens, I put some effort into plumbing that out in the same PR, as it kills both birds, and makes future sensor development easier, as when deploying test code, you get a usable core instead of what seems like bricking the whole board while trying to bring new sensors online. Seems good for new users too, if a new sensor is attached and the system doesn't support it, it should detect it, and ideally not have any code to execute and go on its merry, instead the board again appearing bricked.

I have downstream software ingesting the telemetry, which is what inspired the reorganization when I realized there is no way to tell temperatures (or anything, if duplicates) apart. One can have an environment sensor removed from the mainboard, exposed to the elements reading things, and I know that data is not coming from the mainboard's sensors. Proximity to the boards can throw off readings, like a RAK is charging its battery, suddenly even the "external" temp sensor shows a jump in temperature, (VOC too) so I'm trying to separate out the telemetry so a sensor could be moved away from the board via ribbon cable and not be affected by what's going on in the board.

Thanks for the notes around sensor wire and power planes too, reading up the RAK spec, I noticed there's some sleep interval that can be employed, that would probably be an area for further investigation! Cool to see your soil moisture sensor add-on too! I was hoping to dig into some other env sensors next.

KPrivitt:
My first sensor the RAK12035 was a learning experience and I finally got it working after several attempts.  But along the way I discovered that it would "brick" just like you too discovered. Mine was when the sensor was plugged in and on a cold power on.  Once it was running it was rock solid and worked, reliable detection and reliable readings.

My brick issue was traced down to the switched 3v3_S not being turned on properly (not intentionally turned on and not fully driven) Discussion #2222. I found that turning it on early resolved my issue.

My next sensor will be a different Soil Moisture sensor that uses an RS-485 link and the serial interface (not I2C).  The RS-485 allows a much longer cable run to the sensor and it can also can be daisy chained.

The next will be a very low cost sensor ($1.17 each) that is analog.  The issue I'll first have to deal with is detection (I can easily read the value, but how do I know that it is indeed my sensor not some other analog sensor).

NickDunklee:
Whoa, let me know if you need any help debugging the RS-485, that was one of my targets a little further down the line as well and I'm down to collab.

KPrivitt:
I'm gathering the new hardware, so it will be a while.  I'll be using the RAK5802 for the RS-485 interface.

omeg:
I think this is a good first step.

I'm working on on-device telemetry logging framework and the current sensor handling is pretty bad on multiple fronts. This PR at least fixes i2c initialization issues and problems with one source overwriting another.

LPP is really annoying in that you can't query the telemetry data once it's written without parsing it yourself. I'd like to at least have a public interface for getting the total number of channels in use and maybe metadata for what data types a particular channel contains (but that's probably a thing for later).

Good work anyway, I'm currently testing it on one xiao node with a combo bmp280+aht20 sensor, no issues so far. I have a bunch of other sensors in a box for the future, including some more exotic stuff like a magnetometer. Too few hours in a day for everything ;)

ripplebiz:
Good. I like it. Will need as many peeps as possible to give it a good shake tho

KPrivitt:
I am curious how this PR will interact with the RAK12035 sensor and other sensors all installed in a single node.

The RAK12035 uses these telemetry variables -

Temperature (Soil Temperature)
Percentage (Soil Moisture %)
Frequency (Raw Capacitance)
Temperature (Dry Calibration)
Power (Wet Calibration)

It sounds like you will group a single sensors data all under one channel.  But the RAK12025 uses two channels...

Other sensors that have temperature values will not be mixed with the Soil Temperature & Dry Calibration data?

I have found that CayenneLLP overwrites multiple data values and only displays one (I don't know if it is the first one or the last one called, but that is why I had to put the RAK12035 data on it own separate channels, it wasn't getting displayed, I did test with a temperature sensor on Channel 1 and it worked)

Channel 2 uses:
Temperature
Percentage

Channel 3 uses:
Frequency 
Temperature 
Power

Lastly the RAK_4631_Repeater failed to compile, this is an important platform:

A key usage model is where a repeater in the field (aka Forest) can read the local soil moisture and be used to assist in predicting Fire Danger at that location. A sea of repeaters can  monitor drought susceptible areas. (and extend the mesh gaps into mountainous areas, expanding the overall mesh which is a benefit to us all)

KPrivitt:
I walked through your code changes, and I now understand what you've done, and I like the significant improvements you have made.  Good Job! and Thanks!  (it would have been easier to add a sensor if this had been in place). Unfortunately, I no longer have the RAK12035 sensor setup and cannot test this to verify it still operates as intended.  I do not like the way contributors are treated by the "devs" and their lack of responsiveness, curt interactions and lack of a vision or direction on where MeshCore needs to go.  If they are the arbiter of what is included and what is not, they need to provide some direction on where this is going, so us contributors are not left wasting our efforts.  This is not a collaborative effort.

Therefore, I am no longer contributing to the MeshCore development. 

That said, there is an additional issue that you might want to clean up with EnvironmentSensorManager.  The GPS initialization is also causing some of the lockups.  It toggles IO2 which on RAK12001/3/7 devices is the 3v3_S enabling and control signal (it is turning all sensors on/off) and it does not leave the IOx signals in a proper state.  IO2 is left as an INPUT which is not fully driving the FET to enable the sensor power rail.  Lines 810-815 should be removed (this will be moved to RAK4631Board, the current code is stomping on the 3v3_S power rail) and line 850 should be removed (this the worst offending code) to leave IO2 as an OUTPUT (it's prior state).   The 3v3_S should be initialized earlier, and the initial reset of the various potential sensor boards should be done there as well:

Here's code and comments that explains the move.  This new code is placed at the end of RAK4631Board::Begin()


void RAK4631Board::begin() {
  NRF52BoardDCDC::begin();
  pinMode(PIN_VBAT_READ, INPUT);
#ifdef PIN_USER_BTN
  pinMode(PIN_USER_BTN, INPUT_PULLUP);
#endif
#ifdef PIN_USER_BTN_ANA
  pinMode(PIN_USER_BTN_ANA, INPUT_PULLUP);
#endif

#if defined(PIN_BOARD_SDA) && defined(PIN_BOARD_SCL)
  Wire.setPins(PIN_BOARD_SDA, PIN_BOARD_SCL);
#endif

  Wire.begin();

  pinMode(SX126X_POWER_EN, OUTPUT);
#ifdef NRF52_POWER_MANAGEMENT
  // Boot voltage protection check (may not return if voltage too low)
  // We need to call this after we configure SX126X_POWER_EN as output but before we pull high
  checkBootVoltage(&power_config);
#endif

  // 
  // Enable the 3V3_S Power Rail
  // The RAK19001/19003/19007 switched 3.3 Volt add-in sensor card power supply controlled by IO2
  //

  pinMode(WB_IO2, OUTPUT);        // Get power pin ready to be turned on
  digitalWrite(WB_IO2, HIGH);     // Turn on 3v3_3 power for the sensors
  delay(10);                      // Short delay to ensure power is stable

  //
  // Assert then De-assert *reset for the various RAK Sensor Modules
  //
  // ***NOTE*** RAK sensor slots do not all use the same IOx for the *reset pin
  //
  // RAK 40 pin I/O slots have a dedicated *reset on pin 13 controlled by the Reset Switch
  //
  // Some 40 pin I/O Sensor Modules use an IOx pin for a controlled *reset (IO4, IO5 or IO6)
  // RAK 24 pin Sensor Slots (A/B/C/D/E/F) use pin 10 for a controlled *reset (IO1, IO2, IO4, or IO6)
  //
  // ***NOTE*** the *reset of the GPS sensors in EnvorpnmentSensorManager has been moved to here:
  //    to address issues with the late 3v3_S power turn on and unify the initial reset of the sensors.
  //    Sensor code may initiate and control it's own reset for power on/off, or other management reasons
  //    This is here to centralize the INITIAL *reset sequence only.
  //

  //
  // This section asserts *reset (LOW) then de-asserts *reset (HIGH)
  // for IO1, IO2, IO4, IO5 and IO6 to reset any of the possible sensors attached
  //

  pinMode(WB_IO1, OUTPUT);        // Get *reset pin ready to be turned on
  digitalWrite(WB_IO1, HIGH);     // Drive IO1 HIGH to deassert *reset
  delay(10);                      // Short delay de-asserting *reset
  digitalWrite(WB_IO1, LOW);      // Drive IO1 LOW to assert *reset
  delay(10);                      // Hold *reset low for a moment to ensure proper reset

  digitalWrite(WB_IO2, HIGH);     // Drive IO2 HIGH to de-assert *reset
  pinMode(WB_IO2, OUTPUT);        // Get *reset pin ready to be turned on
  digitalWrite(WB_IO2, HIGH);     // Drive IO2 HIGH to deassert *reset
  delay(10);                      // Short delay de-asserting *reset
  digitalWrite(WB_IO2, LOW);      // Drive IO2 LOW to assert *reset
  delay(10);                      // Hold *reset low for a moment to ensure proper reset
  digitalWrite(WB_IO2, HIGH);     // Drive IO2 HIGH to de-assert *reset

  pinMode(WB_IO4, OUTPUT);        // Get *reset pin ready to be turned on
  digitalWrite(WB_IO4, HIGH);     // Drive IO4 HIGH to deassert *reset
  delay(10);                      // Short delay de-asserting *reset
  digitalWrite(WB_IO4, LOW);      // Drive IO4 LOW to assert *reset
  delay(10);                      // Hold *reset low for a moment to ensure proper reset
  digitalWrite(WB_IO4, HIGH);     // Drive IO4 HIGH to de-assert *reset

  pinMode(WB_IO5, OUTPUT);        // Get *reset pin ready to be turned on
  digitalWrite(WB_IO5, HIGH);     // Drive IO5 HIGH to deassert *reset
  delay(10);                      // Short delay de-asserting *reset
  digitalWrite(WB_IO5, LOW);      // Drive IO5 LOW to assert *reset
  delay(10);                      // Hold *reset low for a moment to ensure proper reset
  digitalWrite(WB_IO5, HIGH);     // Drive IO5 HIGH to de-assert *reset

  pinMode(WB_IO6, OUTPUT);        // Get *reset pin ready to be turned on
  digitalWrite(WB_IO6, HIGH);     // Drive IO6 HIGH to deassert *reset
  delay(10);                      // Short delay d-easserting *reset
  digitalWrite(WB_IO6, LOW);      // Drive IO6 LOW to assert *reset
  delay(10);                      // Hold *reset low for a moment to ensure proper reset
  digitalWrite(WB_IO6, HIGH);     // Drive IO6 HIGH to de-assert *reset
} 

I have posted my investigation of the "lockup", including scope pictures of the 3v3 turn on in another PR (search for 3v3).  Moving this to RAK4631Board fixed the issue we ran into, it also fixed another issue with an OLED display (the GPS code stomped on it with it's IO2 treatment).

I am posting this here in the hope that you might be able to make some further improvements.  I have given up with dealing with the "devs". I no longer have any respect for them and their shortsighted vision.

Good Luck!

NickDunklee:
Hey, thanks for that, really appreciate it. I'll look through that and check it out. I've noticed some sensors have been occasionally halt/restarting myself and was hoping to investigate, this might give me a head start where to hunt. I'll hopefully have a RAK12035 at some point to be able to re-validate it still works correctly. Good luck to you as well!

martenumberto:
Hey!
Just installed the Firmware v1.16.0 which should include your PR. Thanks for your eye on that!
Is it possible, that the changes dont apply to the Repeater Firmware? I only got Channel 1 (Battery and Temperture) and no Telemetry from the BME680. Im running a RAK 4630 and a RAK 3400.

NickDunklee:
Check telemetry after a bit - the BME680 should show up on channel 2 once it comes online. I did a lot of my testing on a BME680+repeater and BME680+sensor on a RAK19003 and RAK19007, so it definitely should be showing up, but it won't show up at first. Normally within 5 minutes.

martenumberto:
Thanks for your fast reply! The Repeater is up for 30min now.
I only can do some testing with the 1W 3400 RAK. Could it be possible, that these changes dont apply to it? It must be a RAK19007 as Baseboard

NickDunklee:
The change should be applying on that board as well. There is also a BME680 change that will use the Bosch sensor library to activate the full abilities of the sensor. I'll check the release later today and see if something funky is going on. Silly question, but is your mobile app up to date?

martenumberto:
Okay! Im looking forward to see your results!
Just tested it on an RAK3460 Room Server with the same result. All Boards have an BME680 attached.
Yes. Mobile App is up to date!

cwichura:
FWIW, I have two 4631+19007 systems with BME680 (RAK1906 module) and both are showing the new BSEC-based BME680 data on channel 2.

One got to accuracy of 3 in under an hour, while the other is still on accuracy 1 at nearly two hours.

cwichura:
@NickDunklee I suppose an enhancement/feature request is to add a cli command that lists all discovered sensors and their assigned channel numbers (and maybe their i2c address?).  You have the text name for the sensor in the array already, but no way to actually see that (the only thing currently using that is the debug statements, and in release firmware, those become no-ops).

martenumberto:
Thanks for your results!
Do you use the original RAK BME Sensor Board or do you solder a BME to the Board? I have soldered the BME to the J12 Header of the 19007.

KPrivitt:
Nick,
Regarding the 3v3_S an “easy” way to see if the GPS is stomping on the sensors (assuming it is the bad IO2 code) on a RAK19001/3/7 based system, is to just turn off the GPS (it is likely to be sensed on the first pass and it remembers which IOx it was detected on, which would be IO2).  So, if it was detected on IO2 and you turn GPS off, it will turn all sensors off even though you had detected them earlier on in your code and logged them as present, they will all just disappear (a bad design).  Who knows what problems that will now cause.

FYI,
I asked my friend (whom I made the RAK12035 sensor firmware for) and who now has the hardware setup, to test 1.16.0 on it and let you know if it works properly.  He can only test the non-calibration part since a custom version is needs for that, but if the basic sensor works it is likey that the calibration will too (but if you want it tested we can make a version to test that too).

NickDunklee:
@cwichura that's a real cool feature idea.

@martenumberto do all of your RAK boards have GPS onboard or did you disable the GPS run loop? If you get into the CLI of the repeater and type `gps` does it respond `on, deactivated, no fix, 0 sats`? You may want to try `gps on` if not, and then reboot and see if the sensors show up. I am still looking through board/schematics as well to see if something else might be the cause as well.

@KPrivitt that's a good observation, it looks exactly like GPS off can stomp on the 3v3_S and killing power to sensors.

RAK3401/variant.h even has a comment that isn't in RAK4631:
>// IO2 is hooked to GPS RESET = gpio 34, but it can not be used to this because IO2 is ALSO used to control 3V3_S power (1 is on).
>// Therefore must be 1 to keep peripherals powered

It looks like in the case of GPS and power:

```
  //search for the correct IO standby pin depending on socket used
  if(gpsIsAwake(WB_IO2)){
  }
```

then

```
>bool EnvironmentSensorManager::gpsIsAwake(uint8_t ioPin){
```

...

```
gpsResetPin = ioPin;
```

...

```
  gps_active = false;
  #ifdef RAK_WISBLOCK_GPS
    pinMode(gpsResetPin, OUTPUT);
    digitalWrite(gpsResetPin, LOW);
```

If that is the case, that pin gets assigned as the GPS reset pin, and the entire sensor power rail can be turned off and/or reset.

KPrivitt:
Nick,
Since you are going through the schematics (I used the WisBlock-IO-Pin-Mapper to find all the resets used in all the various devices that are included in the spreadsheet, I searched for "reset").  This includes their GPS sensor offerings.

For the main three base modules RAK19001/3/7 you will find that IO2 is used to control 3v3_S.  For the the RAK12500 GPS sensors the *reset fortunately also has an R/C circuit to ensure it is reset when power is turned on. So, it isn't really necessary for the firmware to do a controlled reset for this.  

If power is ever turned off (possibly in future sensor power management firmware that intentionally is turning everything thing off to save power), then the initial reset code I sent (RAK4631Board.cpp code) will turn on the power and reset everything and the hardware R/C reset will cover the RAK12500.

A "long term" solution could be to restrict/remove control of IO2 from all sensor code and the future sensor power management code could handle/coordinate the power on and reset sequencing. If any sensor needs it own reset control, it can use one of the other IOx pins.  The only one I am aware of that uses IO2 is the RAK12500 and it has the R/C.  I know that some sensors have low power modes: like the RAK12035 and I left routines in place that can be used, but didn't utilize and some things were commented out.  Everything was tested and it all worked...

It's unfortunate that RAK conflated the power and reset on the same pin, but it is what it is.

1nerdherder:
@NickDunklee - I can verify that having the sensor plugged in to the 12035 board will not allow it to boot.
If the sensor is removed, the board does boot.   If you personal message an address to me - I can send you a sensor and adapter board to assist your debugging.

martenumberto:
@NickDunklee: All of the Boards dont have a GPS Module attached. I haven't touched any GPS Settings in the Config/Cli. When i type "gps" it respond with "on, deactivated, no fix, 0 sats".
"gps on" or "gps off" doenst work. "gps toggle not found"

EDIT: 3,3V can be measured on the J12 i2c header, where the BME is connected to.
EDIT2: Maybe this helps? The latest Firmware works fine, but some Measured Values are wired. https://github.com/IoTThinks/MeshCore/commit/2cbe0c039872ec7fe28de3efed2f0f47954e0ac9
EDIT3: It seems that the RAK BME uses a different I2C Adress than mine. I have changed the I2C Adress to 0x77 and Channel 2 pops up! Maybe we need to make this changeable via Cli or make the Firmware also look for 0x77 BME680

```
DEBUG: BME680+BSEC not detected at I2C address 76
DEBUG: BMP085 found at 77 but failed to initialize
```

NickDunklee:
@martenumberto Oooh very good find. That gives something to parse along with previous comments about cli commands and such.

@1nerdherder as luck would have it, I think that was a recent sensor I wanted to add to my home test environment so I may have it, will let you know, and appreciate it. I have a feeling what you are seeing might not actually be the sensor code but a deeper meshcore issue in the loop stack I encountered initially while working on RAK12047 code where I saw similar "won't boot" behavior. It seems the loop stack is currently only 4KB, and the ed25519 code used for encryption consumes much of the stack. I was working on code for a RAK ambient light sensor last week (slightly different pattern, but I think same root cause) and external adverts were causing the board to "crash" - when it loaded the encryption code into the loop and ran out of loop stack. It looks like it was apparently going into some kind of infinite loop and not even trying to soft-restart as the way that loop is handled. Board had both RAK1903 and a RAK12047 onboard, and they apparently ate the rest of the loop stack. Along the way, it also appears this loop's memory issues could be causing silent corruptions that go unnoticed for some time depending on where they take place in memory. Been a deep dig to try and sort out. Although I've been busy with trying to fix the unemployment situation, hah, so that all went on pause for a moment.

I am testing a 3-layered fix for the loop stack issue on that same loaded sensor board to let it soak before breaking it into PRs, which will also hopefully help all sensors survive better, and maybe even fix some potential node instability. Hoping to have something later this week if time allows. 🤞

omeg:
Hi, I just wanted to chime in that I have a working VEML7700 implementation. The base loop was written by me, later pseudo-float representation and auto-tuning was mostly by LLM. I was thinking about upstreaming parts of it (the float representation is specific to my FW) but didn't have the time yet.
Feel free to use it in any way you like :)
https://github.com/omeg/MeshCore/blob/omeg/dev/src/helpers/sensors/EnvironmentSensorManager.cpp

E: oh, it seems your RAK module uses a different sensor. Anyway, I didn't see any crashes on my RAK4631 running this yet (active sensors: VEML7700, RAK12005 (rain), INA226, BME280).

NickDunklee:
Two PRs here based on feedback, trying to finish up a third.

CLI to query sensors:
https://github.com/meshcore-dev/MeshCore/pull/2830

CLI to get/set sensor address config:
https://github.com/meshcore-dev/MeshCore/pull/2831

Third one I am changing the runloop position so sensors are scanned before GPS starts, and protecting the 3v3_S from being touched by GPS, although a bit more involved to figure out.

KPrivitt:
Nick,
Not to add to your firmware list.

But, the RAK12035 Soil Moisture driver has a function to change a sensors I2C address.  I only partially addressed three sensors in my original driver.  Your sensor set address_RAK12035 could use the set I2C address function in the RAK12035 driver to move a default sensor (0x40) to a different address.

The RAK12023 board has three connectors which allows up to three sensors to be attached, but each attached sensor needs to be at a different I2C address otherwise there would be a conflict. Three sensors could possibly be supported by using these functions.

A default sensor (0x40) is installed and would be detected and logged, the set command could then move it to 0x41.  A second default sensor 0x40 could then be plugged in and it would be detected and logged along with the first now at 0x41.  The second default sensor would then be moved to 0x42. Installing a third would have all three sensors installed and logged at 0x40, 0x41 and 0x42.

Food for thought...

NickDunklee:
Ooh thanks for that @KPrivitt ! I shall put that food in my brain and percolate. 👍

PR #2328, opened 2026-04-18
fix: remove sensors from xiao c3 companion because of bootloops

recrof:


recrof:
issue: https://github.com/meshcore-dev/MeshCore/issues/1471

IoTThinks:
So fast.
Let me test.

IoTThinks:
I have just tested with my Xiao C3 and Wio 1262.
It works.

So the culpit is -ENV_INCLUDE_GPS and ENV_INCLUDE_VL53L0X
A bit weird. But nevermind.

Thanks a lot.

recrof:
> So the culpit is -ENV_INCLUDE_GPS and ENV_INCLUDE_VL53L0X A bit weird. But nevermind.

The culprit is probably low RAM or sensor detection code. if you look closely you will notice I removed sensor support for companions completely.

IoTThinks:
This fix works but break again when I add in some extra libraries.

The root cause is likely 4MB flash size of Xiao C3.
This will "mitigate" the root cause and keep the sensors.
`board_build.partitions = min_spiffs.csv ; get around 4mb flash limit`

This code can be found in MeshCore for some other small boards.

PR #2329, opened 2026-04-18
Fix build script

vvarg229:
Minor fixes to the build.sh script to make it more secure and stable.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -49,7 +49,9 @@ EOF
 
 # get a list of pio env names that start with "env:"
 get_pio_envs() {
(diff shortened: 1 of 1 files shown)


mikecarper:
I did a big refactor that has a menu added as well to make it easier for new people to custom compile. See https://github.com/meshcore-dev/MeshCore/pull/2349

vvarg229:
> I did a big refactor that has a menu added as well to make it easier for new people to custom compile. See #2349

You did a great job, this is really cool!
But maybe we shouldn't write such complex logic in the shell?
Over time, the script will become more and more complex, and it will be impossible to maintain.

PR #2333, opened 2026-04-19
BLE Ghost connect fix (ESP32)

gjelsoe:
# Fix BLE ghost connection state after reset/DFU/Update
~More testing might be needed, hence the draft.~
 
## Problem
 
After a firmware update or reset, the device could get stuck believing it was still connected to a BLE client, preventing new connections.
 
## Root Causes
 
1. `onDisconnect()` did not reset `deviceConnected` unconditionally, leaving stale state if disconnect occurred before `enable()` was called.
2. `onAuthenticationComplete()` set `deviceConnected = true` even when `_isEnabled` was false, allowing a bonded phone to ghost-connect before the device was ready.
3. `deviceConnected`, `oldDeviceConnected` and `adv_restart_time` were not explicitly reset in `begin()`, leaving potential stale RAM state.
## Fix
 
- Always reset `deviceConnected = false` in `onDisconnect()` regardless of `_isEnabled`
- Guard `deviceConnected = true` in `onAuthenticationComplete()` behind `_isEnabled`, disconnecting the peer if not yet enabled
- Explicitly reset connection state variables in `begin()`
## Changes
 
- `SerialBLEInterface.cpp` — `onDisconnect()`, `onAuthenticationComplete()`, `begin()`

## Test Device
- Heltec v4.2

### DIFF
--- a/src/helpers/esp32/SerialBLEInterface.cpp
+++ b/src/helpers/esp32/SerialBLEInterface.cpp
@@ -28,6 +28,10 @@ void SerialBLEInterface::begin(const char* prefix, char* name, uint32_t pin_code
   BLEDevice::setSecurityCallbacks(this);
   BLEDevice::setMTU(MAX_FRAME_SIZE);
 
+  deviceConnected = false;
+  oldDeviceConnected = false;
+  adv_restart_time = 0;
+  
   BLESecurity  sec;
   sec.setStaticPIN(pin_code);
   sec.setAuthenticationMode(ESP_LE_AUTH_REQ_SC_MITM_BOND);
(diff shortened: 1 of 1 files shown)


Giamma94:
@gjelsoe Has this fix been tested? This bug hates me!

gjelsoe:
It's as far i know only been tested by me.

Giamma94:
> It's as far i know only been tested by me.

Okay, I updated the firmware with your code added to this pull request. 
I can confirm it's been working correctly since yesterday. I did a lot of testing using two different Heltec v4 modules and multiple Android smartphones from various brands.
Everything works perfectly. 
Until your code is officially merged, I'll create my own custom firmware with your modification integrated.
Thanks for your work!

gjelsoe:
Thanks for testing.

IoTThinks:
So after reset or firmware update, what does get stuck mean?
I would love to test it.

gjelsoe:
Thinking it's already connected by BLE when not. Therefor you can't connect to it.

IoTThinks:
I hit the BLE issue quite often when connecting to new ESP32 BLE companions.
Let me test today if this PR helps me.

IoTThinks:
Seems less issue after this PR but not fix 100%
Tested on iPhone + Heltec v4.2.

Test Case 1: Already companions and connected
Flash new version 
Reconnect to app fine. No need to key in PIN.
=> Perfect.

Test Case 2: Already companions. Erased devices.
Flash new version.
Need to "Forget the BLE device" in Phone. It is fine.
Connect to app and key in BLE PIN. Still ask me to forget BLE device again in MC app. => Why?
I need to use iPhone to force close the MC app and connect again. I can connect without key in BLE PIN above again.
=> Same issue as before. 
=> I guess this is something to improve in MC app because forcing close MC app fixes the issue. I guess the MC app caches the old information of the old BLE device.

I will keep this PR and test with other ESP32 boards.
The logic in the PR looks correct and perfect to me.
@gjelsoe Thanks a lot for the PR. 👯‍♂️

IoTThinks:
Tested with Android.
The result is same as in iPhone.

I need to force close MeshCore app to pass Test case 2.

PR #2337, opened 2026-04-19
Delay auto-tuning v2

stachuman:
Improving ACK delivery and number of ACK received - theoretically ensuring that if DM is delivered - at least one ACK is received by sender (we speak of probability!)

This base on an extensive simulations (over 100k simulations done - dedicated simulator built (hopefully can be used also for other purposes)) - all details and road to this PR can be traced in this discussion https://github.com/meshcore-dev/MeshCore/discussions/2053

Proposal base on theoretical work of KPrivitt and my simulations (all source data used for simulations are available on my github page)

--

There are some differences comparing to the previous PR
1. change of variable to: auto.tune.delays (to make it more consistent)
 set auto.tune.delays on/off
2. wired recalc of parameters into onAdvertRecv - so no periodic recalc done, instead - automatic recalc called
3. auto.tune.delays by default is off

## Measured performance at defaults  (same 4 topologies, 6 rnd seeds each):

Density | Msg Delivery | Channel Delivery | ACK Delivery | Avg ACK Copies
-- | -- | -- | -- | --
sparse | 67.0% | 57.5% | 36.0% | 0.60
medium | 62.7% | 78.8% | 33.7% | 0.70
dense | 65.0% | 64.5% | 24.3% | 0.50
very_dense | 64.3% | 43.0% | 13.0% | 0.20
Mean | 64.8% | 60.9% | 26.8% | 0.50

## Proposed changes - autotune tx/direct tx delays:
Density | Msg Delivery Δ | ACK Delivery Δ | ACK Copies Δ
-- | -- | -- | --
sparse | −2.7pp | +11.0pp | +0.80
medium | −6.3pp | +1.6pp | +1.00
dense | −6.0pp | +13.7pp | +1.30
very_dense | −9.3pp | +17.0pp | +1.40
Mean | −6.1pp | +10.9pp | +1.13


KPrivitt:
In the prior PR the frequency of the neighbor count was every 5 min.  This is far too frequent and is consuming compute resources that can be utilized elsewhere.

While the SNR of a received Advert can vary several dB from message to message (pings can change on every one sent) and this can affect the neighbor count (for repeaters close to the 0dB SNR threshold), but the surrounding number of repeaters actually changes very slowly.  I believe the count should be done daily, twice a week or once a week.

1nerdherder:
I've been watching this work across the two pull request conversations.  This was a heavy lift, deserving of strong consideration amongst the devs.  At a minimum, the existing defaults are not optimal.  The power of the autotune algorithm approach is that it makes all repeaters "good neighbors" who will adapt their settings in harmony as the mesh evolves.

stachuman:
> In the prior PR the frequency of the neighbor count was every 5 min.  This is far too frequent and is consuming compute resources that can be utilized elsewhere.
> 
> While the SNR of a received Advert can vary several dB from message to message (pings can change on every one sent) and this can affect the neighbor count (for repeaters close to the 0dB SNR threshold), but the surrounding number of repeaters actually changes very slowly.  I believe the count should be done daily, twice a week or once a week.

Correct- on one hand calculating every couple of minutes is not a big burden, yet for the sake of clean code I have moved that to advert recp. Code.

terminalvelocity23:
Hello, I've tested your PR in our mesh, which is very dense and there's a lot of in-band noise. The algorithm has set the delays so high that the repeater effectively stopped functioning.
Also, it didn't return the delays to their original values after disabling.

изображение

stachuman:
> Hello, I've tested your PR in our mesh, which is very dense and there's a lot of in-band noise. The algorithm has set the delays so high that the repeater effectively stopped functioning. Also, it didn't return the delays to their original values after disabling.

-- 
Can you please elaborate on 'stopped functioning'? Was it one repeater with changed firmware or more? Creating wider network? Also - what do you mean - 'effectively stopped functioning'? There are delays - to limit number of collisions, but transmission is done.

For the last point - very valid point, let me update that.

terminalvelocity23:
> Can you please elaborate on 'stopped functioning'? Was it one repeater with changed firmware or more? Creating wider network? Also - what do you mean - 'effectively stopped functioning'? There are delays - to limit number of collisions, but transmission is done.

It was one repeater to test this feature. The delays were set so high it effectively stopped relaying packets, everything but its admin interface was handled by other repeaters around. It stopped showing up in outbound and inbound paths.

stachuman:
> > Can you please elaborate on 'stopped functioning'? Was it one repeater with changed firmware or more? Creating wider network? Also - what do you mean - 'effectively stopped functioning'? There are delays - to limit number of collisions, but transmission is done.
> 
> It was one repeater to test this feature. The delays were set so high it effectively stopped relaying packets, everything but its admin interface was handled by other repeaters around. It stopped showing up in outbound and inbound paths.

In fact - this is not bad thing what you observed, it’s in fact desired effect. The purpose of mesh network is NOT to ensure that every repeater is transmitting but to ensure effectivenes of the overall network. 
To be precise - in a dense network it is NOT recommended that ‚all repeaters’ transmit within the same time window - as this only increase probability of collision - failed transmission.

Not to go into details - ‚all the repeaters in the area carried on the transmission but your one was silent’ - if then - due to collisions - ‚all the other traffic would fails’ - your repeater will retransmit with a delay - giving the chance to deliver message, and opposite - if ‚all the other traffic’ will deliver, your one won’t be even required (it will kind of reduce density of network - what is a recommended thing).

And this is the purpose of PR - to increase overall probability of delivery (ACK) - it is NOT to increase single repeater number of transmissions. Effectiveness is not coming here from how quick re-transmission is done - but is coming from probability of evading collisions with other repeaters.

Hope- I’m clear in my explanation.

stachuman:
Here are theoretical results with 2 scenarios - 1. We address the busiest routers in an organized way, 2. We address randomly routers with auto delay function.
(0% - we use only default firmware, 30% - means - we use 30% of repeaters in auto-delay optimization mode)

### Degree Strategy (upgrade busiest nodes first)

| % Optimized | N nodes | Delivery | std | ACK | Channel | F_del | P_del | F_ack | P_ack | col/lost | ack/del |
|:-----------:|:-------:|:--------:|:---:|:---:|:-------:|:-----:|:-----:|:-----:|:-----:|:--------:|:-------:|
| 0% | 0 | 61.7% | 7.1 | 22% | 62% | 58% | 48% | 15% | 44% | 32.8 | 0.4 |
| 10% | 14 | 62.0% | 3.3 | 23% | 73% | 59% | 43% | 17% | 38% | 42.3 | 0.5 |
| 30% | 43 | 57.7% | 3.4 | 25% | 79% | 44% | 36% | 18% | 36% | 22.8 | 0.7 |
| **50%** | **72** | **55.0%** | **9.9** | **30%** | **75%** | **34%** | **32%** | **29%** | **34%** | **32.1** | **0.8** |
| **75%** | **107** | **52.3%** | **3.9** | **30%** | **84%** | **25%** | **24%** | **29%** | **31%** | **31.8** | **1.1** |
| 100% | 143 | 50.7% | 6.5 | 27% | 72% | 21% | 22% | 28% | 26% | 43.3 | 1.3 |

### Random Strategy (uncoordinated rollout - random repeaters uses auto-delay)

| % Optimized | N nodes | Delivery | std | ACK | Channel | F_del | P_del | F_ack | P_ack | col/lost | ack/del |
|:-----------:|:-------:|:--------:|:---:|:---:|:-------:|:-----:|:-----:|:-----:|:-----:|:--------:|:-------:|
| 0% | 0 | 61.7% | 7.1 | 22% | 62% | 58% | 48% | 15% | 44% | 32.8 | 0.4 |
| 10% | 14 | 58.3% | 8.0 | 27% | 74% | 51% | 46% | 21% | 38% | 34.3 | 0.6 |
| 30% | 43 | 56.0% | 4.2 | 29% | 80% | 43% | 33% | 25% | 35% | 38.4 | 0.8 |
| 50% | 72 | 51.3% | 7.8 | 22% | 79% | 42% | 23% | 23% | 21% | 26.7 | 0.8 |
| 75% | 107 | 46.3% | 5.7 | 26% | 79% | 30% | 22% | 30% | 26% | 33.9 | 1.2 |
| 100% | 143 | 50.7% | 6.5 | 27% | 72% | 21% | 22% | 28% | 26% | 43.3 | 1.3 |

### Radio Efficiency (collision-free RX ratio)

| % Optimized | Degree radio_eff | Degree ackpath_eff | Random radio_eff | Random ackpath_eff |
|:-----------:|:----------------:|:------------------:|:----------------:|:------------------:|
| 0% | 62.1% | 59.2% | 62.1% | 59.2% |
| 10% | 65.3% | 63.6% | 65.1% | 62.9% |
| 30% | 71.4% | 69.9% | 69.5% | 68.1% |
| 50% | 74.1% | 72.2% | 73.6% | 71.9% |
| 75% | 77.1% | 77.1% | 77.2% | 76.8% |
| 100% | 76.8% | 76.5% | 76.8% | 76.5% |

### 1. Mixed firmware outperforms full optimization on ACK delivery

The best ACK rate (30%) occurs at **degree 50-75%**, not at 100% (27%) - as it comes from the scenario 1 - it means - addressing the most busiest repeaters
My interpretation:
- Optimized nodes reduce collision pressure in dense clusters
- Default nodes relay faster, creating alternative paths and timing diversity
- The combination produces more successful ACK round-trips than either firmware alone

Note - around 75% we are reaching 1 ack delivered - so that's another reason to see that ass a sweet-spot setting.

### 2. Channel (broadcast) delivery peaks at degree 75%

Channel delivery reaches **84%** at degree 75% — a **+22pp improvement** over the 0% baseline (62%) and **+12pp over full optimization** (72%). 

### 3. Degree strategy is consistently better than random

| Metric (at 75%) | Degree | Random | Delta |
|:-----------------|:------:|:------:|:-----:|
| Delivery | 52.3% | 46.3% | +6.0pp |
| ACK | 30% | 26% | +4pp |
| Channel | 84% | 79% | +5pp |
| Std deviation | 3.9% | 5.7% | more stable |

1nerdherder:
Sorry for the late comment:
The earlier analysis showed the existing defaults to be flawed and actually making things worse, so should we not be selecting a new “default” rather than reuse the known defective one?

terminalvelocity23:
I'd say it's still too agressive. I've switched auto-tuning on two my repeaters pointed no the north and south of the high-rise I'm in, and dropped tx power on the companion, so nobody but them will hear it. 
After a few hours the delays have settled at 12.8 for flood and 38 and 40 for direct. I mean yeah, it makes for collision avoidance, but considering the fact that the max message length is 150/2 minus whatever bytes your name requires if you use non-English alphabet, conveying any complex thought requires a few messages in succession. And the delay of up to 40 seconds breaks the sequence.

stachuman:
> I'd say it's still too agressive. I've switched auto-tuning on two my repeaters pointed no the north and south of the high-rise I'm in, and dropped tx power on the companion, so nobody but them will hear it. After a few hours the delays have settled at 12.8 for flood and 38 and 40 for direct. I mean yeah, it makes for collision avoidance, but considering the fact that the max message length is 150/2 minus whatever bytes your name requires if you use non-English alphabet, conveying any complex thought requires a few messages in succession. And the delay of up to 40 seconds breaks the sequence.

Well... everything base on probability, not on feelings. However - I admit - scenario where sequences of messages is sent was not tested.
Would you like to propose a scenario?

terminalvelocity23:
@stachuman Idk about a scenario, but maybe capping the delays at maybe 20s max isn't a bad idea.

KPrivitt:
stachuman,
Thank you for your tremendous effort.  Your results validate my original supposition that as the density increases the need for additional backoff increases and density (which is an easy thing to measure using a neighbor count) is a valid metric to use to adjust and set the amount.

From a theoretical point of view this is easy to see: the probability of a collision drops off as the number of slots increases, however if the number of neighbors increases it counteracts that benefit.  Essentially, if the number of neighbors doubles the number of slots needs to double.

This does lead to high delay values in a very dense mesh.  But remember in my comments on "success" I did mention that as this increases it does introduce a delay and when that delay becomes humanly noticeable (40 seconds definitely exceeded that threshold) it would need to be capped.  A "balanced" or throttled setting is needed.

One thing is clear, an automatic tuning does improve mesh performance, and the current defaults are bad settings, in that they provide insufficient backoff.  The mesh works better when collisions are reduced and backoff reduces collisions.

**The question and discussion at hand is what should the table entries be for each neighbor count?**

The discussion also needs to be about what the "success criteria" should be: what metric should be used?  Just one?  A combination?  How should balancing be done.  This will likely be contentious and have many different and sometimes opposing views.  But the discussion is GOOD, all views should be entertained.  That usually will generate a better result.  I hope the dev experts will participate.

So, given the data we have: Can we start with a conservative table to enable automatic tuning and finalize the table later?  Lets take the low hanging fruit that is right in front of us.

One comment regarding some comments that this is a centralized approach.  It is not, it is decentralized since each repeater can have its own setting.  This can be disabled and any repeater owner can set the values they choose.  It is not forcing any particular setting, it allows choice and optimization.

The immediate value here is getting the defaults changed to a better setting (eliminate the majority of repeaters being set at the insufficient default values, we need to be good neighbors to each other, just one repeater setting it to better values will help their neighbor, but if the neighbors continue to stomp on you... well, it's best if we are all good neighbors) Plus having the ability to adjust based on density, "the mesh" can adjust for the future.

Optimization of the optimizer can come later.

My opinion: For what it is worth.. 

Last item: do we know why rxdelay does not affect the simulation results?  It should...

That can easily be added to the table as another column.  But with what settings...  Can it be added and left all 0's (the current default) until we get a future better table and understand what is going on and the "optimum" settings. rxdelay is a secondary benefit, but it is a tool in our chest why not use it.

stachuman:
> stachuman, Thank you for your tremendous effort. Your results validate my original supposition that as the density increases the need for additional backoff increases and density (which is an easy thing to measure using a neighbor count) is a valid metric to use to adjust and set the amount.
> 
...

I'm evaluating a different approach, as I have a feeling, that we address not enough. 

- Implementing routing protocol which would be back-compatible with existing network. It will take me some time to propose the approach (and maybe even do first tests).
I'll work on my fork - I'll share theory for discussion.

Magalex2x14:
Hello, guys! What do you think about reverse aproach? I mean "more positive SNR neighbours - lower delays". It would be interesting to look at the simulation results and compare.
We decided to try this reverse approach on our network, but we can’t evaluate the result; personally, I don’t see a difference in connection reliability, perhaps because not everyone adheres to the recommended settings.
The idea was picked up from here - https://meshsydney.com/wiki#repeater-profiles

stachuman:
I can help in case with setting the simulation, however, honestly, I lost faith in MeshCore :( - too much flooding, considering, we have 1% duty cycle in EU, it's killing our network.
I'm working on my own protocol.

1nerdherder:
@stachuman keep me posted - I can help test here in USA

stachuman:
> @stachuman keep me posted - I can help test here in USA

I will :) - Last full month I was designing and simulating, only now started to port to firmware

1nerdherder:
I suspect the detailed simulation you have run has surfaced optimizations that will really benefit the mesh.On May 29, 2026, at 11:09, Stani ***@***.***> wrote:﻿stachuman left a comment (meshcore-dev/MeshCore#2337)

@stachuman keep me posted - I can help test here in USA

I will :) - Last full month I was designing and simulating, only now started to port to firmware

OverkillFPV:
> Hello, guys! What do you think about reverse aproach? I mean "more positive SNR neighbours - lower delays". It would be interesting to look at the simulation results and compare.
> We decided to try this reverse approach on our network, but we can’t evaluate the result; personally, I don’t see a difference in connection reliability, perhaps because not everyone adheres to the recommended settings.
> The idea was picked up from here - https://meshsydney.com/wiki#repeater-profiles

Thanks for having a read of my wiki, from some not so scientific testing flipping the delays seemed to mainly improve repeaters with high neighbour counts and elevation not get stuck behind the 'wave' of repeats from lower neighbour count repeaters and would constantly max out the CAD delay. Would be interesting to see simulation results as it's hard to quantify the results that I've had.

Magalex2x14:
@OverkillFPV Yes, It seems that this is a dilemma. Directly with the switch to "reversed delays", I personally did not noticed a meaningful change in the reliability or in any other "feelings". However, over time, as the network grew, I began to notice jumps in "heards". That is, in public channels, the number of heard repeaters began to vary greatly from message to message. And for messages with unusually small number of heard repetitions, the last ones in paths were repeaters covering a large areas. My theory is that since repeaters covering large areas relay first, then a large number of other repeaters on the network will receive this transmission simultaneously, that is, their relay window starts at the same time, and many of them will try to transmit this packet at the same time, causing massive collision.

Magalex2x14:
I'm also trying to communicate with the CoreScope author about adding a metric that reflects the total number of repeaters that took part in relaying each flood routed packet - https://github.com/Kpa-clawbot/CoreScope/issues/1699#issuecomment-4733454798

OverkillFPV:
> @OverkillFPV Yes, It seems that this is a dilemma. Directly with the switch to "reversed delays", I personally did not noticed a meaningful change in the reliability or in any other "feelings". However, over time, as the network grew, I began to notice jumps in "heards". That is, in public channels, the number of heard repeaters began to vary greatly from message to message. And for messages with unusually small number of heard repetitions, the last ones in paths were repeaters covering a large areas. My theory is that since repeaters covering large areas relay first, then a large number of other repeaters on the network will receive this transmission simultaneously, that is, their relay window starts at the same time, and many of them will try to transmit this packet at the same time, causing massive collision.

Yes, I am hoping to try avoid my 'critical' repeaters not collide with each other. Am considering changing the delays in the meshsydney wiki to 0.5 for critical try hopefully help avoid that. Ideally I think we should work out whether reversed or non-reversed delays are the best as there isn't enough data to determine outcomes. However there is decent logic for both sides.

m0urs:
Question: 

auto.tune.delays ist setting my txdelay to a value of 4 although the CLI is not allowing me to set a value greater than 2? How could this be?

image

KPrivitt:
m0urs,
The reason is that this code (any code) has direct access to the variable (txdelay, etc) and can set it to any value as long as it conforms to its type. (One ex: the compiler won't let you write a 32 bit number into an 8 bit location, it would overwrite the neighbor variable and corrupt it).

However, recently there was a PR that was merged to force the CLI commands to match the documentation and its limits.  This didn't make much sense to me since typically expert users utilize these commands and do know what they are doing, but now the CLI code forces it to match the documentation which has no reasoning or explanation behind it.

KPrivitt:
@OverkillFPV 
The reverse approach is a viable solution, both approaches utilize using a "groups" assigned "slot number" to segregate the traffic.

Segregating the traffic can avoid/reduce collisions.

A packet that collides is destroyed and too many collisions destroys the traffic carrying capability and reliability of a transfer over the mesh.  Collision Bad, Delivery Good.

An interesting byproduct of the reverse approach is that the regional traffic has the lowest latency (it goes first, lower priority goes later)

**BUT** both of these have to contend with the fact that there is a default setting that is low and does not generate many slots (remember txdelay is multiplied by 5, so it takes a value of at least .2 to get to the first slot, .4 to get to the second, etc). More slots reduces collisions simply because of the probability of two choosing the same slot and then transmitting (same slot = collision)

Since most repeater owners do not know about these setting let alone how to set them, the majority of the repeaters out there have the default and are causing more collisions since they only have a couple slots to choose from, almost guaranteeing a collision and with only a few repeaters it becomes statistically guaranteed.

A key feature of AutoTune is that it is **automatic** and every repeater will have a optimized value (best for its location and number of neighbors).  By being automatic the entire mesh can be set to the appropriate value (reversed or non reversed).

As you have seen there isn't much of a visible much affect, since all the stock setting repeaters are all colliding together and preventing your message from propagating (you have to go through them). It just clogged by collisions.

I know this since I'm in a rather dense area and everyone around me has stock values.  So ACKs to my messages coming back are stomped on and I get a message fail.  50% of my repeaters packets are errors likely caused by collisions (so 1/3  of everything on the air is being corrupted, you can see this in the repeater stats).  So even if I use a much better value (I'm helping my neighbors traffic by not colliding) they are not helping mine and they continue to stomp on me in their oblivious ignorance (and bad setting).

Another interesting note is that we swept through the values (the first results said that 0, 0, 0 was the best setting) but this obviously is not correct, segregation and reduction of collisions is the fundamental purpose of them.  We found a bug and fixed it, and larger vakues improved delivery (both flood and direct routing).  If reverse was fundamentally better should it have skewed it to the lower end, this didn't happen.  But I do think both (reversed and non-reversed) are viable and we probably programmed in a "bias" with the way we constructed the table (we put regional and high density at the top).

I haven't had time to examine the detailed results to see what's going on.

The key pieces are:
1) Automatic (everyone gets an appropriate "good" setting)
2) Using the neighbor count and their SNR to understand the local density and use that to determine a groups settings
3) Using the segregation for regional repeaters (put them in their own group to reduce their collisions since they transport the long haul traffic and a lot of it, put everything else in another).
4) Implement a better default (one for a higher density rather than the lowest). Being automatic also fixes this.

Another idea I had was to change the computation (I think this helps, I don’t have the code in front of me but there is also a truncation in the original)

t = RND(txdelay * 5)

to

t = (RND((txdelay * 5)) + offset

Where the offset is a deterministic value based on the public ID that slightly skews the result.  Since each repeater has a unique ID this will slightly spread them all out.  The offset is a partial slot (bounded 0 to 1).  So one will transmit first, and since the second must listen before transmitting it will defer, it's the simultaneous transmission that is harmful.

So a collision may still occur (a partial one) but it won't be a full one by both picking the exact same start time and directly stomping on each other.  One good thing is that this change is fully compatible with the existing mesh.  Upgrading to newer firmware makes it better, but nothing breaks, existing repeaters still repeat. (Although some very badly...) no

KPrivitt:
I'm been thinking about this some more and reread the simulation results.

Collisions are a normal occurrence in a flood routed mesh.  They cannot be eliminated, but they can be possibly be reduced (or made worse with bad settings).

It was interesting that a 100% auto tune was not the optimum setting.  I now believe this is possibly because a combination of reversed and non-reversed is actually the best (do both within a group).

The purpose of these settings is to reduce collisions.

This is being done in two ways by the txdelay value.

1) segregating the repeaters, by setting the value based on "groups" with separate settings per group.

- Anecdotally regional repeaters have placed at or near the max txdelay setting and all others are set much lower.  This skews the regional traffic to transmit after the lower tiers finished.  This prevented collisions with the regional transmissions because all the others had already completed.  Theses can also be considered "high value" (they typically are a long distance and carry a lot of traffic) segregating them out gives a higher probability of getting to the destination.

2) using txdelay to spread out transmissions (the random delay) This is its basic purpose (and it used in other network protocols, like Ethernet) but it can also be used for segregation as above.
This can also be set in ranges (which is what my original proposal was based on neighbor count)

This will reduce collisions (doesn't prevent them) amoung local repeaters, by having them transmit a different times.  A larger txdelay value reduces the probability of a collision by spreading the transmissions over a larger time range (more random slots to choose from).

The simulations showed that a combination of defaults and autoTune was best amoung a group of repeaters.  Some set to low delay (the default, they transmit first) and others with higher values (based on the density) and regional at the top in its own group (this is essentially doing segregation).

Note that the simulations show that by not using all repeaters with defaults (using autotune) the simulations shows a 22% increase was achieved (AutoTune works). This definitely reveals that the defaults and having most repeaters with the exact same value is not optimum.  Something needs to be done.

The first thing I noticed about reversed is that the regional were at the bottom (low values) which does have an advantage of reducing the latency of the high value traffic.  A good thing and then the txdelay was increased for each: dense, medium, sparse.  Having regional here is an advantage.

- The question here is how to identify regionals: they typically have lots of neighbors (pushes them to the top, which is why autoTune placed them there).  So, reversed does have a value, but the best solution seems to be to manully set them because there are only a few of them and they are managed by experienced users and are well in hand.  For the regionals use a txvalue of 0, 0.2 and 0.4, there is more than one setting for in case where there are regionals that are close to each other (segregate and spread them out) do not set them all with identical values.

The reversed is essentially using mostly segregation and not utilizing the spreading which is necessary in a dense mesh.  Using a larger value is definitely better and dense should be at the high end and not the low end.

But the simulation results indicate to me that there is a benefit of using segregation within a group (the simulations with some defaults, this was done to see what happens as the mesh migrates to autotune).

The less dense groups are less sensitive (sparse by definition means there aren't many and could use any essentially any value, but lower does give a better latency)

So, using neighbor count of repeaters with an SNR > 0 (discard any that are distant too and quite likely have a higher error rate) to determine the group (sparse, medium, dense, very dense) seemed an appropriate strategy. 

But now  I realize that the segregation could be also be used within this group.  The SNR of each neighbor would be used to distribute the txdelays (possibly use the high values for high SNR and lower values for low SNR, weak signal and spread them across the range) leaving the regionals in their separate bin (we don't want to inject collisions by overlapping).

This would be more optimal.

But there is a key catch, for both reversed and non-reversed: this needs to be automatic (dynamic that adjusts as the mesh changes) and turned on by default.  Having most of the mesh at the default significantly reduces the effectiveness of both algorithms.  With "Auto" the mesh settings are adjusted all across the mesh.  This needs to be in the official release and the defaults removed.  Over time the number of repeaters running the defaults are reduced and eventually they no longer matter (still backwards compatible).

Having the majority of the repeaters set to a single poor default value is degrading the mesh.

So both are likely the **Best**

It would be interesting to see what a random assignment of txdelay (based on the repeaters fixed Public ID and spread across the entire range 0.5 to 4, reserving 0 to 0.4 for regional) would show...

This seems like it would definitely be better than a fixed default.

The only metric I am aware of that indicates collisions is the Received Packet Errors.

REQUEST:
Could all repeater owners gather their repeaters stats and calculate the percentage of error packets and post them here?

% = Received Packet Errors (Received Packet Errors + Packets Received)

Mine is 36% indicating a high level of collisions from the repeaters around me (I've already ruled out AGC reset as a source).  Any other ideas (or solutions) why I have such a high error percentage?

Are there any Maintainers that understand the routing protocol and purpose of the backoff mechanism that can explain the default selection and if there is a flaw in the logic above?  If you know one please forward this thread.

Magalex2x14:
I think that other factors may be the presence of the other signals on the channel, as well as a large number of signals received near SNR low limit.

RadiolibWrappers.cpp:
```
// Approximate SNR threshold per SF for successful reception (based on Semtech datasheets)
static float snr_threshold[] = {
    -7.5,  // SF7 needs at least -7.5 dB SNR
    -10,   // SF8 needs at least -10 dB SNR
    -12.5, // SF9 needs at least -12.5 dB SNR
    -15,  // SF10 needs at least -15 dB SNR
    -17.5,// SF11 needs at least -17.5 dB SNR
    -20   // SF12 needs at least -20 dB SNR
};
```
Personally, I’m used to looking not at the percentage of the total, but at the ratio of errors to received (it’s just easier to calculate in your head by superficially looking at the repeater statistics). In our network, for well-located nodes this ratio tends to 0.5. On my own repeater this ratio is 0.44 now, and equates to 31.4% (up to 40 neighbours with up to 19 SNR>0).

KPrivitt:
Magalex2x14,
I totally agree, I also just look at the ratio (Rx/Rx error) and it's always ~50% (this is much easier to mentally log, I used the full formula in the request to be "precise" regarding the percentage of errors).  I asked for this data to see if my situation is unique or if this is common (from the limited data I have this seems common).

Regarding the SNR, my repeater only has seven neighbors total, only one has an SNR below the -7.5dB threshold (WCMesh uses SF7) it's -8.5, the highest is +1.25dB, so if this was the cause I would expect a much smaller ratio (1/7 = 0.143), since of the seven packets in a flood only one is below the threshold, all the others are +2 to +8dB above the threshold.

FYI, I tested with the AGC reset interval at 0 and 500, no difference.

1nerdherder:
image

Magalex2x14:
My experience tells me that this ratio is common. Moreover, if it is low, then in our fairly dense network this may indicate that this node is deaf for some reason and may have some problems with reception. 

I wouldn't rely on neighbors to gauge this ratio, since they are fairly static, but the air is moving, it breathes. You can stop taking what you took an hour ago and vice versa. In your neighbours there is only what you successfully received once, that is, there are no these reception errors. I just meant that if there are repeaters at your location, the signal from which falls into the SNR threshold zone, then with a high probability its reception will end in error (as far as I understand, there is a high probability of failing the LoRa radio CRC check). For example, in your location there is only you and a repeater with an SNR level of about -8 for SF7 - in this case, I think the ratio will be quite large. You may not even receive a single packet from it, but the error counter will continue to grow. And no collisions 🤷‍♂️

KPrivitt:
Thanks, that is what I expected regarding the ratio.  But my main concern is not collisions around me but collisions of an ACK from a dense destinations DM.  (I'd like to understand my local error rate, I really only have one repeater out of the hole I'm in, so I'm essentially in a Sparse area, without my repeater up on the balcony, I receive **zero** messages).

Directed DMs are the most fragile type packet and do not get the reliability advantages of flood routing (it seems like most everything is send flood, which utilizes the most airtime).

When I send messages to a friend in Los Angeles (I'm in San Diego 120 miles away) he's in a very dense area, repeaters are everywhere   (converted Mestastic nodes).  It's an almost 100% probability that the directed paths ACK fails (sometimes the DM is actually delivered but most times it's not, in both cases I get a Failed, after the five tries).  But if I send it flood (it seems to be reliably received) I can get a Delivered within the three tries/retries (though sometimes the Delivered is delayed by a few seconds).  Since a single ACK is sent, collisions at a very dense destination can kill it.  This is what caused me to think about how to reduce collisions, especially in dense areas far from me (not like we have any control of where repeaters are placed or configured, which is why I advocate automating the settings).

With the path changing on every DM (it's based on ACKs returned, more is better) the DM is sensitive to getting killed and having more ACKs returned can help stabilize the resurrected path (it's not the "real" path, it's reconstructed, and long paths are inherently fragile), so any dense area (bottlenecks) along the way can reduce and even kill ACK returns (but being a flood it is taking multiple paths and inherently more reliable).  The original ACK teturned seems to be the most susceptible to collisions. (an aside, the simulations show that ACK delivery is improved with AutoTune).

Any help you can provide to push this along (prove or disprove it, I have no clue why it was designed and configured the way it is).  I educated myself from reading and the source code.  It would be nice to know the design criteria and selection criteria (so, can someone point me to where it's documented?) what's important is improving the transmission across the mesh.

PR #2344, opened 2026-04-20
feat: enhance command handling to support multiple commands in a sing…

OverkillFPV:
This change allows for sending multiple commands to a repeater split by a semicolon. The repeater then responds to each command with a semicolon seperated response.


liamcottle:
Hi, thanks for the PR.

Unfortunately this is not something we are looking to add at this time.

Will close out for now :)

PR #2346, opened 2026-04-20
FIX: Tbeam 1W Fan and LNA Sequncing

Robowarrior834:
This code improves the LNA sequencing and also impliments Fan control so the fan is only on when you transmit and 10 seconds after.

### DIFF
--- a/variants/lilygo_tbeam_1w/platformio.ini
+++ b/variants/lilygo_tbeam_1w/platformio.ini
@@ -146,6 +146,31 @@ lib_deps =
   ${LilyGo_TBeam_1W.lib_deps}
   densaugeo/base64 @ ~1.4.0
 
+; === LILYGO T-Beam 1W Companion Radio PS (BLE PS) ===
+[env:LilyGo_TBeam_1W_companion_radio_ble_ps]
+extends = LilyGo_TBeam_1W
+platform_packages =
+    framework-arduinoespressif32 @ symlink://D:/esp32-2.0.17
+build_flags =
+  ${LilyGo_TBeam_1W.build_flags}
+  -I examples/companion_radio/ui-new
+  -D MAX_CONTACTS=350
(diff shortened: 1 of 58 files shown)


variants/lilygo_tbeam_1w/platformio.ini
```diff
@@ -27,13 +28,20 @@ build_flags =
...
+  ; AGC reset interval default: 60 seconds (value * 4000ms).
+  ; Periodically recalibrates the SX1262 frontend to prevent progressive
+  ; deafness caused by the receiver getting stuck in a high-attenuation state
+  ; after repeated TX/RX transitions with the external PA+LNA module.
+  -D DEFAULT_AGC_RESET_INTERVAL=15
+
   ; Battery - 2S 7.4V LiPo (6.0V min, 8.4V max)
   -D BATT_MIN_MILLIVOLTS=6000
-  -D BATT_MAX_MILLIVOLTS=8400
+  -D BATT_MAX_MILLIVOLTS=7900
```
CullenShane, line 44: Why is this changed?  Can you justify it?

variants/lilygo_tbeam_1w/platformio.ini
```diff
@@ -154,7 +162,6 @@ build_flags =
   -I examples/companion_radio/ui-new
   -D MAX_CONTACTS=350
   -D MAX_GROUP_CHANNELS=40
-  -D OFFLINE_QUEUE_SIZE=256
```
CullenShane, line 157: Why have you removed the offline queue size definition?

variants/lilygo_tbeam_1w/platformio.ini
Robowarrior834: This was to calibrate my radio. For some reason my battery charges to a full voltage of 8v. Not sure if I just have a funky battery.

PR #2349, opened 2026-04-20
Pr/buildsh refactor menu

mikecarper:
This refactors `build.sh` to make the interactive target selection, env parsing, version selection, command dispatch, and artifact collection easier to follow and safer to maintain.

Running `bash build.sh` now opens a 2 tier menu and does decent guessing at the firmware version.

Also fixed various shellcheck issues.


mikecarper:
image
What happens when you run build.sh with no args. you get a menu with selection options

vvarg229:
Maintaining such large shell scripts is a nightmare.
Mb we should just rewrite it in Python?

mikecarper:
rewriting it in Python is an option; it would remove the need to call python to run this "PlatformIO exposes project config as JSON". I moved just about every string that could be changed to the top of the file which helps with maintaining this moving forward.

mikecarper:
Replacement PR opened from the renamed branch `buildsh-refactor`: #2489

Issue #2348, opened 2026-04-20
cli overflow will cause Repeater to Crash - verified


winnieXY:
Disclaimer: this bug was found with the help of Claude Opus 4.6 - however this bug was tested using a PoC (see attached python file):

`python bug7-cli-overflow.py -p /dev/ttyACM3 --dest "TestRepeater" --pwd "test"`

**I'm able to trigger reliable hangs on test repeaters via lora or serial connection.** 
Below is the description from our Code Review:
 
----
Unbounded `sprintf` Buffer Overflow in CLI Handler

**File:** `src/helpers/CommonCLI.cpp`  
**Impact:** Stack buffer overflow if an attacker sends a long CLI command; can corrupt memory or crash the device.

**Worst-Case Outcome:** Crash/reboot or potential code execution on repeater, room server, or sensor nodes. **Remote attack (BLE/radio):** An admin-authenticated client sends a 179-byte CLI command like `set AAAA...` (175-byte config). The firmware writes `"unknown config: "` (16 bytes) + 175 bytes = 191 bytes into a 161-byte `reply[]` buffer → **30-byte stack overflow**. On ARM Cortex-M4 (nRF52) this overwrites saved r4-r7 and the return address (LR) → HardFault or attacker-controlled PC. **Serial attack:** 155-byte config → 171 bytes into 160-byte buffer → 11-byte overflow. The `get owner.info` path writes up to 122 bytes char-by-char with zero bounds checking. All three firmware variants (repeater, room server, sensor) are affected.

**Description:**  
Multiple `handleCommand` / `handleSetCmd` code paths write attacker-controlled string data into the `reply` buffer using `sprintf` without any length check. The `reply` buffer is typically a fixed-size stack array in the calling firmware (often 160–256 bytes). A crafted long command string can overflow this buffer.

**Code (unknown config echo):**
```cpp
// CommonCLI.cpp:729 — handleSetCmd()
  } else {
    sprintf(reply, "unknown config: %s", config);
    // ← 'config' is attacker-controlled, no length limit
  }
```

**Code (password echo):**
```cpp
// CommonCLI.cpp:288 — handleCommand()
    } else if (memcmp(command, "password ", 9) == 0) {
      StrHelper::strncpy(_prefs->password, &command[9], sizeof(_prefs->password));
      savePrefs();
      sprintf(reply, "password now: %s", _prefs->password);
```

**Code (owner.info — char-by-char write without bounds):**
```cpp
[...]

PR #2350, opened 2026-04-20
prevent a oob via serial or lora - causing a crash and hardware interaction

winnieXY:
a possible attacker could crash a repeater/roomserver with a oob write. 

this minimal approch will prevent that in the future. This PR is a possible solution for #2348.


winnieXY:
This was fixed in commit db7baa7bd7a10a9f22af968b8e1f6abadf89f2bc and 3d982711a6014e318050c556c615916ebaa13391

No further need for this PR.

PR #2355, opened 2026-04-21
suspend the radio before blocking io

winnieXY:
acl.save() is blocking IO - during this time the nrf52 CPU is halted. If this happens in between a spi transaction with the sx1262 the bus is corrupted and the radio is stuck. the issue does _not_ occur on esp32 as esp32 has a separate flash controller. 

This behaviour is correctly detected by the repeater and the _err_flags are set:
`     _err_flags |= ERR_EVENT_STARTRX_TIMEOUT;
`
However - no action is taken upon that error yet.

There are two countermeasures - and one general countermeasure:
 - use non-blocking io (not trivial to near impossible on nrf52)
 - suspend the radio before writing the acls to prevent spi bus corruption
 - (restart the transmitter after a error was detected)

The last one is described within a separate PR - see #2347 - in order to have cleaner PRs.  This PR is a attempt to mitigate the spi bus corruption by suspending the radio before the blocking operation.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -1291,8 +1291,21 @@ void MyMesh::loop() {
 
   // is pending dirty contacts write needed?
   if (dirty_contacts_expiry && millisHasNowPassed(dirty_contacts_expiry)) {
-    acl.save(_fs);
-    dirty_contacts_expiry = 0;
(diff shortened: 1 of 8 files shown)


winnieXY:
Converting back to draft - there are more writes + Rationale behind the patch has changed. Let's rethink it a bit

PR #2356, opened 2026-04-21
Fix Direct respond on handleLoginReq

fizzyfuzzle:
This PR fixes the _handleLoginReq_ response (which is the first response when doing a login to remote admin), to use the **Direct** path if available.
Current firmware is ALWAYS responding the handleLoginReq with a **Flood** response. With this PR the response will be a **Direct** response if the path for the client is available.
Basically it just made it consistent with the other PAYLOAD_TYPE_ANON_REQ responses code, like _handleAnonRegionsReq_, _handleAnonOwnerReq_ and _handleAnonClockReq_.

This will prevent unnecessary flood messages from traveling all over the mesh, while a direct path was already available.
In the current meshes airtime is already very sparse, so every reduction in unnecessary flood messages is welcome ;)

P.S. No 'fix' needed for _simple-roomserver_, because the direct response was already working.

Initial PR was already done in #1485 by @oltaco, but code was outdated so this is a retry :sweat_smile:

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -130,6 +130,8 @@ uint8_t MyMesh::handleLoginReq(const mesh::Identity& sender, const uint8_t* secr
 
   if (is_flood) {
     client->out_path_len = OUT_PATH_UNKNOWN;  // need to rediscover out_path
+  } else if (client->out_path_len != OUT_PATH_UNKNOWN) {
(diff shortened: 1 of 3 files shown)


fizzyfuzzle:
@liamcottle @ripplebiz can this be reviewed please :sweat_smile:

This will enable the _handleLoginReq_ to respond with a DIRECT message when the path is available, instead of always using a FLOOD.
It can save some unnecessary flood messages on the mesh, where airtime is already limited.

PR #2370, opened 2026-04-22
fix(CommonCLI): add paged output to region command

jirogit:
## Problem
`region` command output is limited to 160 bytes (one packet reply),
causing silent truncation at ~5 entries when region names are long.
Users with many regions cannot see all entries.

## Fix
Add paged output to `handleRegionCmd()`:
- `region` or `region 0` — first page (always includes wildcard `*`)
- `region N` — entries from index N
- Footer `>N` in reply indicates next start index; no footer = last page

## Change
- `src/helpers/CommonCLI.cpp`: replace single `exportTo(reply, 160)` call
  with paged output logic (29 lines)

## Notes
- Mobile app UI may still show only ~5 regions; that is a separate app-side issue
- Compatible with existing `region save`, `region put` etc. subcommands
- No changes to RegionMap or storage format

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -900,8 +900,37 @@ void CommonCLI::handleRegionCmd(char* command, char* reply) {
 
   const char* parts[4];
   int n = mesh::Utils::parseTextParts(command, parts, 4, ' ');
-  if (n == 1) {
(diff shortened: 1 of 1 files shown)


jirogit:
Rebased onto current upstream/dev (no conflicts).

jirogit:
Rebased onto current dev, no conflicts.

Issue #1572, opened 2026-02-02
nRF devices with LTO batteries: PWRMGT_VOLTAGE_BOOTLOCK too high; prevents boot


TheSlowGrowth:
For RAK4631 devices, `PWRMGT_VOLTAGE_BOOTLOCK` is set to 3.3V.
While this is fine for Lithium Ion batteries, it prevents booting without a USB connection when using LTO batteries, where the 0% voltage is around 1.8V.
This affects the popular [mesh base boards by uart.cz](https://uart.cz/2534/solar-mppt-charger-for-meshtastic/) if used with the LTO battery pack from the same website, and any other device using a different type of battery.

Obviously, the voltage can be adjusted using the `PWRMGT_VOLTAGE_BOOTLOCK` definition. But this would require a dedicated firmware for each potential battery chemistry, which sounds like a nightmare from the user perspective and the development perspective.

So ideally, the cutoff voltage would be user configurable in the node configuration.

Happy to help work out a solution and provide a PR. Let's discuss.

tagging @entr0p1 , I see you added the bootlock feature recently.

entr0p1:
Hi,

I'll take a look at this. I'll need to look at the schematics and see how this board powers the Rak. I would've thought something like this would power the board by 5V rail not the battery connector since the MCU becomes unstable below 2.5V (and that is the manufacturer rated minimum if I recall correctly). But I'll confirm.

TheSlowGrowth:
There's a DC/DC (U2) that generates 3.3V from either the USB power or the battery power (switched through U5 and U6).
VBATT is hooked up to AIN0 which is what fires the voltage bootlock, as far as I can see.

I think tailoring this to a particular boad isn't a good solution. If the voltage bootlock is tied to the battery voltage, then there must be a user configuration for selecting the voltage. The power delivery is very much dependent on the particular installation. Some may use a DC/DC boost converter (where battery voltages <2.5V are still totally fine to boot). Others may use a buck converter or a linear converter, where VBAT must be >2.5V.

<img width="1215" height="820" alt="Image" src="https://github.com/user-attachments/assets/f4714867-d569-40e2-ad44-ef8086a68dde" />

entr0p1:
[...]

PR #2377, opened 2026-04-22
Disable NRF52 undervoltage bootlock feature

fschrempf:
The NRF52 undervoltage bootlock feature implemented behind NRF52_POWER_MANAGEMENT was introduced in v1.12.

It configures a fixed voltage level of 3.3V and prevents the software from booting if the battery voltage is below that level.

This is problematic for various reasons. The most important one being that it breaks devices that use cell chemistries other than Li-Ion, such as LFP, LTO or SIB which have much lower minimum voltages.

As a quick fix to prevent devices from failing, disable the feature until it has been decided how to handle this in the future and a proper implementation has been provided.

This solves https://github.com/meshcore-dev/MeshCore/issues/1572.

### DIFF
--- a/variants/gat562_30s_mesh_kit/platformio.ini
+++ b/variants/gat562_30s_mesh_kit/platformio.ini
@@ -7,7 +7,6 @@ build_flags = ${nrf52_base.build_flags}
   -I variants/gat562_30s_mesh_kit
   -D RAK_4631
   -D RAK_BOARD
-  -D NRF52_POWER_MANAGEMENT
   -D PIN_BOARD_SCL=14
   -D PIN_BOARD_SDA=13
   -D PIN_OLED_RESET=-1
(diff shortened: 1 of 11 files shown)


mattzzw:
@ripplebiz, @recrof   - please have a look at this. 
Multiple nrf based boards that suport different cell chemistries like LiFePo or LTO have been locked out since #1413 has been merged and have needed their own specific, patched firmwares ever since.  
This PR provides a simple and clear solution as long as #1413 has not been fixed or reverted.

IoTThinks:
I have been disabled it since MC 1.13.
A bit suprised when it is still added into newer NRF52 boards, so I have to keep disabling it.

1nerdherder:
What is the status of this in V17?  Also - any thoughts of harmonizing this with PR 2088?
Seems you're both indirectly working the solution from a couple different directions.

IoTThinks:
@1nerdherder Currently, NRF52_POWER_MANAGEMENT is disabled in PowerSaving 17.
I would love to use PR 2088 if it is tested and merged. It is still in Draft.

PR #2383, opened 2026-04-24
Added Xiao S3 variant

IoTThinks:
This is for Xiao S3 and a normal Wio SX1262 for Xiao shield.
Tested works fine (repeater and BLE companion) for two PowerSaving realeases.
- Xiao S3: https://www.seeedstudio.com/XIAO-ESP32S3-p-5627.html
- Standard Wio sx1262: https://www.seeedstudio.com/Wio-SX1262-for-XIAO-p-6379.html

image
image

PS: This is not a Xiao S3 Wio combo. The pin mappings are different (especially the mapping to DIO1).
- Xiao S3 Wio: https://www.seeedstudio.com/Wio-SX1262-with-XIAO-ESP32S3-p-5982.html

image

### DIFF
--- a/variants/xiao_s3/XiaoS3Board.h
+++ b/variants/xiao_s3/XiaoS3Board.h
@@ -1,0 +1,13 @@
+#pragma once
+
+#include <Arduino.h>
+#include <helpers/ESP32Board.h>
+
+class XiaoS3Board : public ESP32Board {
+public:
+  XiaoS3Board() { }
+
+  const char* getManufacturerName() const override {
+    return "Xiao S3";
+  }
+};
--- a/variants/xiao_s3/platformio.ini
+++ b/variants/xiao_s3/platformio.ini
@@ -1,0 +1,154 @@
+[Xiao_S3]
+extends = esp32_base
+board = seeed_xiao_esp32s3
+board_check = true
+board_build.mcu = esp32s3
+build_flags = ${esp32_base.build_flags}
+  ${sensor_base.build_flags}
+  -I variants/xiao_s3
+  -UENV_INCLUDE_GPS
+  -D SEEED_XIAO_S3
+  -D PIN_VBAT_READ=1   ; D0
+  -D P_LORA_DIO_1=2    ; D1
+  -D P_LORA_NSS=5      ; D4
+  -D P_LORA_RESET=3    ; D2
+  -D P_LORA_BUSY=4     ; D3
+  -D P_LORA_SCLK=7     ; D8
+  -D P_LORA_MISO=8     ; D0
+  -D P_LORA_MOSI=9     ; D10
+  -D PIN_USER_BTN=-1   ; NC
+  -D PIN_STATUS_LED=21 ; Orange user led, LOW=On
+  -D PIN_BOARD_SDA=D6  ; D6=43
+  -D PIN_BOARD_SCL=D7  ; D7=44
+  -D SX126X_RXEN=6     ; D5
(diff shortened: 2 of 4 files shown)


Sodium-Hydrogen:
My 2¢ input as someone who also is running a Xiao S3 with the WIO shield is; I like this and you beat me to opening a PR.

But instead of adding it as a new variant, why not add it as a sub variant of the already existing Xiao S3? _(see Xiao C6 for example)_

The Xiao that is already supported is a Xiao S3 with a populated B2B connector that doesn't use the D pins. They both still use WIO-SX1262 shields, this proposal just uses the v1.0 version without the B2B connector. I think the naming would be more clear if the new board is named `Xiao_S3_WIO_v1`.

IoTThinks:
So you want this variant as Xiao_S3_WIO_v1 in the current Xiao_S3_Wio?

Sodium-Hydrogen:
That naming just makes more sense to me and combining keeps more code from having to be repeated.

However, this is ultimately your pull request and I am just a random person giving feedback, not a member of the MeshCore development team.

IoTThinks:
> That naming just makes more sense to me and combining keeps more code from having to be repeated.

@Sodium-Hydrogen It is a valid and may be a better idea. Let me see and try your approach in a few days.
I don't mind if you create a PR with your approach too.

ripplebiz:
Hmm, this is going to confuse some people (incl myself).  So, can you drop the Seeed product links for the two sx1262 Wio's ?  So, there's a 'combo' one, and a new/seperate one? (with diff pin mappings)

IoTThinks:
@ripplebiz I have put in more photos to compare in the first post.

This Xiao S3 in this PR is the same way with Xiao C3 and Xiao C6. They will have a esp32s3/c3/c6 connecting to a standard Wio sx1262 via standard pinouts.
That's why I named it Xiao S3.

The Xiao S3 **Wio** in existing Meshcore is a **newer** combo.
The esp32s3 will have tiny "board to board connector" to a "special" Wio sx1262.
The board to board connector provides extra pins between esp32s3 and sx1262.
The pin mappings are different too.

nerdusbirdus:
> @ripplebiz I have put in more photos to compare in the first post.
> 
> This Xiao S3 in this PR is the same way with Xiao C3 and Xiao C6. They will have a esp32s3/c3/c6 connecting to a standard Wio sx1262 via standard pinouts.
> That's why I named it Xiao S3.
> 
> The Xiao S3 **Wio** in existing Meshcore is a **newer** combo.
> The esp32s3 will have tiny "board to board connector" to a "special" Wio sx1262.
> The board to board connector provides extra pins between esp32s3 and sx1262.
> The pin mappings are different too.

I bought some Xiao S3s and Wio SX1262 For Xiao which have the pin mappings you describe for this firmware. I stacked the S3 onto the Wio with the pins and flashed "Xiao S3 Wio" repeater firmware from Meshcore.io. But I don't have any response coming from it, no repeats or anything. Is there something I did wrong with the hardware or with the firmware? There is very little documentation on the Wio SX1262 for Xiao and I really wish I had not tried to save money by buying them. Do they just stack with the pins or do you need to use the SPI port on the Wio SX1262 For Xiao to connect to the S3?

IoTThinks:
@nerdusbirdus Xiao S3 Wio kit (with tiny pins) and Xiao s3 + Sx1262 (standard pins) are NOT compatible.

So you have to use the correct firmware.

This PR is sure working.

I have both versions.

image

Issue #2283, opened 2026-04-09
Bug: SenseCap Solar P1 Pro freezes occasionally on flash write - needs hard reset


winnieXY:
I have occasionally an issue with my SenseCap Solar P1 Pro where the repeater hangs after a successful login.  Here is a short summary of my findings and a possible solution:

---

## Problem Description

**Device**: Seeed SenseCap Solar P1 Pro (nRF52840 + SX1262)  
**Firmware Role**: Repeater
**Firmware Version**: 1.14.1
**Bootloader**: otafix bootloader v0.9.2  

### Hardware Modifications

- USB-C port: 5V Solar Panel attached
- Batteries: Stock 18650 cells replaced with **XTAR 18650 4000mAh 3.6V–3.7V with PCB Protection**
- Battery level: >95%

### Observed Behaviour

After logging into the repeater via a companion node over LoRa:

1. Login succeeds — "Login successful" message is received by the companion
2. The status page request times out
3. All subsequent commands fail
4. The repeater stops emitting advertisements
5. The repeater is **not reachable from any companion node** — including nodes that were not involved in the login
6. No packets are repeated
7. The device is **not** in OTA mode (no BLE OTA device visible)
8. The device does not recover on its own — a manual power cycle is required

---

### Primary Cause: Blocking Flash I/O in Main Loop

Five seconds after a successful **admin** login, the main loop writes the updated client ACL to internal flash. This operation blocks the entire main loop indefinitely on nRF52 hardware.

#### Trigger Sequence

1. Companion sends login request over LoRa
2. Repeater authenticates the companion, sends "Login OK" response
3. Because the login is an admin login, a deferred flash write is scheduled:
   ```cpp
   // MyMesh.cpp line 131
   dirty_contacts_expiry = futureMillis(LAZY_CONTACTS_WRITE_DELAY); // 5000ms
   ```
4. **5 seconds later**, `MyMesh::loop()` executes the save:
   ```cpp
   // MyMesh.cpp lines 1339-1342
   if (dirty_contacts_expiry && millisHasNowPassed(dirty_contacts_expiry)) {
       acl.save(_fs);               // <-- BLOCKS HERE
       dirty_contacts_expiry = 0;
   }
   ```
5. `acl.save()` calls `openWrite()` which performs:
   ```cpp
   // ClientACL.cpp lines 3-8
   static File openWrite(FILESYSTEM* _fs, const char* filename) {
       _fs->remove(filename);                     // flash page erase
[...]

PR #2386, opened 2026-04-24
Use a  tempfile / rename for atomic writes in order to prevent dataloss during a HardFault on a repeater

winnieXY:
This is the first patch to make repeaters more reliable in the field - even if errors occur.

On a repeater the atomic write (tmpfile + rename) should work even on nrf52 Repeaters with internal flash, the amount of data stored is much smaller than the data used on a companion. I estimated the total size needed to be around 40-50kB  with MAX_CLIENTS and MAX_REGIONS = 32.

This patch should ensure that no setting is lost during a failed write operation into the flash on a repeater. The repeater will be still stuck - but after restart all settings are preserved. 

This is most likely the first of several smaller patches to mitigate this bug #2283.

### DIFF
--- a/src/helpers/ClientACL.cpp
+++ b/src/helpers/ClientACL.cpp
@@ -1,9 +1,9 @@
 #include "ClientACL.h"
 
-static File openWrite(FILESYSTEM* _fs, const char* filename) {
+static File openWriteTemp(FILESYSTEM* _fs, const char* filename, const char* tmp_filename) {
   #if defined(NRF52_PLATFORM) || defined(STM32_PLATFORM)
-    _fs->remove(filename);
-    return _fs->open(filename, FILE_O_WRITE);
+    _fs->remove(tmp_filename);
+    return _fs->open(tmp_filename, FILE_O_WRITE);
   #elif defined(RP2040_PLATFORM)
     return _fs->open(filename, "w");
   #else
@@ -54,7 +54,9 @@ void ClientACL::load(FILESYSTEM* fs, const mesh::LocalIdentity& self_id) {
 
 void ClientACL::save(FILESYSTEM* fs, bool (*filter)(ClientInfo*)) {
   _fs = fs;
-  File file = openWrite(_fs, "/s_contacts");
+  const char* tmp_path = "/s_contacts.tmp";
+  const char* real_path = "/s_contacts";
+  File file = openWriteTemp(_fs, real_path, tmp_path);
   if (file) {
     uint8_t unused[2];
     memset(unused, 0, sizeof(unused));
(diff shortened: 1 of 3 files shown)


winnieXY:
@oltaco : Could you review this patch? For repeaters and roomservers the 1MB flash ram is sufficient for this operation. I've tested that on a P1 and it's working.

src/helpers/ClientACL.cpp
```diff
@@ -1,9 +1,9 @@
 #include "ClientACL.h"
 
-static File openWrite(FILESYSTEM* _fs, const char* filename) {
+static File openWriteTemp(FILESYSTEM* _fs, const char* filename, const char* tmp_filename) {
```
ripplebiz: A cleaner way would be to keep method sig the same, but for NRF52 case, concat ".tmp" to the given filename, as use that to open in write mode

src/helpers/ClientACL.cpp
```diff
@@ -74,6 +76,13 @@ void ClientACL::save(FILESYSTEM* fs, bool (*filter)(ClientInfo*)) {
       if (!success) break; // write failed
     }
     file.close();
+
+#if defined(NRF52_PLATFORM) || defined(STM32_PLATFORM)
```
ripplebiz, line 86: @oltaco  Can you verify that _fs->rename() definitely will work in this case?  ie. does the real_path file get overwritten/replaced?

oltaco:
> I estimated the total size needed to be around 40-50kB with MAX_CLIENTS and MAX_REGIONS = 32.

I think the total maximum size of files on disk for repeaters/rooms is much less than that, I think less than 10kB in total? The UserData area that repeaters and room servers use is only 28kb total (actual usable space is less than due to overheads).

src/helpers/ClientACL.cpp
oltaco: Confirmed, the underlying lfs_rename() call will point the old directory entry for file at the blocks referenced by file.tmp and then remove the directory entry for file.tmp

src/helpers/ClientACL.cpp
oltaco: I agree, it would be much cleaner to just let openWrite handle the atomicity for the NRF52 case. All the openWrite callsites could then just stay as they are.

winnieXY: Done.

I added a getTmpPath() function to handle the tmp path name generation and moved back to openWrite().

winnieXY:
@oltaco , @ripplebiz anything else needed to get this merged?

oltaco:
> @oltaco , @ripplebiz anything else needed to get this merged?

We're in the process of getting some other required pieces ready, once that's done I'll go over this properly 👍

winnieXY:
@oltaco, @ripplebiz  - are the other required pieces alrady ready? Two weeks are gone, I really would like to see some more reliablility within the meshcore repeaters to be able without headache to install them on remote inaccessible places.

PR #2388, opened 2026-04-24
Add LilyGo T-ETH Elite SX1262 board support

OhYou-0:
Adds a new variant for the Add LilyGo T-ETH Elite SX1262 board.

This is more of a core board support for now, I feel that adding other accessories and support to it should be separate feature requests, let me know if this is the correct approach. 
I'm unsure what to do for ethernet, so it is not utilized, but still provides board power. In the future it should be trivial to integrate, I've done so in other work but it is more involved than a simple board support PR should be I feel.

validated to build and flash to hardware as repeater and ble companion

### DIFF
--- a/variants/lilygo_teth_elite/TETHEliteBoard.h
+++ b/variants/lilygo_teth_elite/TETHEliteBoard.h
@@ -1,0 +1,10 @@
+#pragma once
+
+#include <helpers/ESP32Board.h>
+
+class TETHEliteBoard : public ESP32Board {
+public:
+  const char* getManufacturerName() const override {
+    return "LilyGO T-ETH Elite";
+  }
+};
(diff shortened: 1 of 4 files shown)


OhYou-0:
oops platformio.ini needs the line
 
  -D MAX_LORA_TX_POWER=22
  
  I had set tx power to default to 8 to support potential 3rd party radio modules but i guess the companion app takes that value as the max unless otherwise specified.

IoTThinks:
This board is interesting.
We can support Ethernet instead of just WiFi.

OhYou-0:
> This board is interesting. We can support Ethernet instead of just WiFi.

I have tried twice to cleanly add ethernet support to some degree into meshcore for this board, but I felt uncomfortable with the changes required. Or if you mean for use in other projects, yeah they are pretty good boards, they have a flaw where the 5v rail is actually 4.5v, and LilyGo decided to pin the thing differently than all existing RasPI lora hats, but past that its alright.

PR #2392, opened 2026-04-24
Introduce DataType Definitions for Meshcore Open

zjs81:
This will be used for sending certain kinds of data on the meshcore open platform.

### DIFF
--- a/docs/companion_protocol.md
+++ b/docs/companion_protocol.md
@@ -318,19 +318,17 @@ Remaining bytes:                Binary payload (variable length)
 
 #### Registered `data_type` values
 
-Declared in `src/helpers/TxtDataHelpers.h`. These values have agreed-upon payload schemas so different client apps on the same channel can interoperate.
+`data_type` is an **application identifier**, not a payload-format identifier. Each registered value identifies an application that owns its own internal payload schemas. The firmware does not inspect payload contents — `data_type` is transported opaquely.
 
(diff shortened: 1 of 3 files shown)


liamcottle:
Hey @zjs81, thanks for the PR.

Please have a read over this doc for updated information on how we are allocating IDs for group data packets.

- https://docs.meshcore.io/number_allocations/
- https://github.com/meshcore-dev/MeshCore/blob/main/docs/number_allocations.md

We will be using the data type as an application identifier, this way applications can maintain their own internal ids/formats for various payloads.

If an application wants 100 different payload types, it doesn't use up 100 slots from the limited pool. It would instead use up 1 slot for the application. Allowing for ~65k unique applications, rather than ~65k unique payload types.

This means we don't need to be responsible for every single unique payload type that an application may want to use. We can instead just approve the application itself.

Please note, the range `0000` - `00FF` is reserved for internal use and `FF00` - `FFFF` is reserved for testing/development without requiring registration here.

I would probably suggest registering MeshCore Open as `0100`, and then you can maintain your own table of IDs for your own payload formats :)

Please update the `number_allocations` table with a contact email, and one link to the application website or listing.

zjs81:
Hey Liam, 

That is really smart. Thank you for the guidance. I will make those changes. 

Regards,

zjs81:
@liamcottle Docs have been updated ready for review

liamcottle:
Looks good, thanks!

PR #2400, opened 2026-04-24
Add support for Grove sensor SHT31

pemensik:
New support for yet another sensor for measuring temperature and humidity. That is another part sold by SeedStudio and provides similar readings to SHT4X.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -125,6 +125,7 @@ build_flags =
   -D ENV_INCLUDE_BME280=1
   -D ENV_INCLUDE_BMP280=1
   -D ENV_INCLUDE_SHTC3=1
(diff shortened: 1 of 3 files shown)


pemensik:
This change make https://www.seeedstudio.com/Grove-Temperature-Humidity-Sensor-SHT31.html working in Sensecap Solar P1. Uses a bit different API to SHT4X.

pemensik:
Oh, but I failed to notice SHT4X has the identical address used. Not sure it is possible to detect their differences. It works to me with SHT31 connected, but it might be caused only by it initialized first. Not sure. Works to me.

pemensik:
There seems duplicate PR #2449, very similar to my updated changes.

PR #2401, opened 2026-04-25
feat: Enable GPS on RAK 1W kit

zevaryx:
Enables GPS on the RAK 1W Kit (RAK3401 + RAK13302) using the RAK12500 GPS module in slot A

The RAK12501 is not compatible at all from testing, and the RAK12500 is only compatible in slot A, not slot D

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,4 +117,4 @@ There are a number of fairly major features in the pipeline, with no particular 
 
 - Report bugs and request features on the [GitHub Issues](https://github.com/ripplebiz/MeshCore/issues) page.
 - Find additional guides and components on [my site](https://buymeacoffee.com/ripplebiz).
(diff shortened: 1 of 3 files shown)


446564 (approved):
LGTM @recrof 

I don't have this hardware combo but the contributor has been running for weeks and confirms working.

recrof:
@liamcottle can be safely merged

liamcottle:
Thanks all, I tried to find my spare RAK12500 gps module to test this on the 1W RAK units I have, but couldn't find it. Will take you word for it :)

recrof:
seems that this PR broke the rak3401 and needs to be reverted. @446564 we should never just merge without testing and just take a word of the author.

oltaco:
I added a comment about what I think the issue might be. https://github.com/meshcore-dev/MeshCore/pull/2536#issuecomment-4431715323

Issue #2320, opened 2026-04-17
[Feature Request] Auto Reboot


holdit:
It seems that from time to time, some repeaters benefit a bit from a reboot. Sometimes things go wrong and in remote nodes, waiting for it to restart would help. Things like that.

So I'd like to suggest a feature where we could set our repeater (chat room too?) to reboot every X days/hours (whatever you see fit, but I guess not too low).

Thank you in advance.

dreirund:
Instead of just "sweeping the issues below the carpet", the code quality should be checked and improved, underlying coding errors found and resolved, maybe a code audit is needed.

Auto reboot because something hickups is fighting symptoms and not causes, and the causes can also lead to more problematic stuff. (Buffer/ memory overflows can be maliciously used, in some circumstances config could get corrupt, etc.)

AND:

If you really need auto reboot, ALSO add debug logging so that the things that go wrong can be captured and analysed to find the root cause.

holdit:
From a user point of view, better code, fixes, better stability and security, etc, sounds great. The problem is, I don't know how to code, can't help with any of that, and can't force anyone to work more on the project.

I can reboot my repeater or companion though, and sometimes do so when something isn't working as smoothly as it should. For example, yesterday before writing this, I was having a few misses between my companion and repeater (Xiao nRF52 kit)... I rebooted it, and it was working better afterwards. Sometimes it's something else and a reboot does nothing... it's hard to say what it is when I'm a noob and the mesh itself isn't that reliable.

Maybe software will be better in this new era of AI/LLMs, but until that happens, the "have you tried turning it off and on" is still a tool to use. Hence the request. If a weekly or even daily reboot makes things run smoother, then why not?

In any case, it's up to the smart people here to decide.

siren386:
> It seems that from time to time, some repeaters benefit a bit from a reboot. Sometimes things go wrong and in remote nodes, waiting for it to restart would help. Things like that.
> 
> So I'd like to suggest a feature where we could set our repeater (chat room too?) to reboot every X days/hours (whatever you see fit, but I guess not too low).
> 
> Thank you in advance.

I fully support this.

towerviewcams:
agree fully support this

winnieXY:
Tbh I opose a auto reboot. If something crashes/ doesn't work as expcepted you can't rely that a configured "reboot" would work. 

Instead a (optional) watchdog would be better. Watchdogs are commonly used in microcontroller to safeguard against possible faults.

[...]

PR #2405, opened 2026-04-25
Add a watchdog for nRF52 based repeaters

winnieXY:
Hey,

I've created a watchdog which can be enabled via cli and defaults to off for nRF52 based repeaters. For all other archs are stubs available which are (currently) doing nothing - so there needs to be some code implemented later for esp32 and co.

This patch would mitigate together with the atomic write patch (#2386) issues with failed writes on the internal flash off the p1 - see #2283 

In order to test it i provoked a HardFault on the repeater - it came up after 10 seconds again:

image

This watchdog would e.g. also fix this bug #2320 without adding unnecessary reboots on the devices. For debugging or testing the watchdog can be disabled by setting the timeout to 0 (which is also the default). 

@oltaco: could you please review the patches?

### DIFF
--- a/examples/simple_repeater/main.cpp
+++ b/examples/simple_repeater/main.cpp
@@ -95,9 +95,7 @@ void setup() {
 
   the_mesh.begin(fs);
 
-#if defined(NRF52_PLATFORM)
   board.initWatchdog(the_mesh.getNodePrefs()->wdt_timeout_secs);
-#endif
 
 #ifdef DISPLAY_CLASS
   ui_task.begin(the_mesh.getNodePrefs(), FIRMWARE_BUILD_DATE, FIRMWARE_VERSION);
--- a/examples/simple_repeater/main.cpp
+++ b/examples/simple_repeater/main.cpp
@@ -95,6 +95,10 @@ void setup() {
 
   the_mesh.begin(fs);
 
+#if defined(NRF52_PLATFORM)
+  board.initWatchdog(the_mesh.getNodePrefs()->wdt_timeout_secs);
+#endif
+
 #ifdef DISPLAY_CLASS
   ui_task.begin(the_mesh.getNodePrefs(), FIRMWARE_BUILD_DATE, FIRMWARE_VERSION);
 #endif
@@ -106,6 +110,7 @@ void setup() {
 }
 
 void loop() {
+  board.feedWatchdog();
   int len = strlen(command);
   while (Serial.available() && len < sizeof(command)-1) {
     char c = Serial.read();
--- a/examples/simple_room_server/main.cpp
+++ b/examples/simple_room_server/main.cpp
@@ -72,9 +72,7 @@ void setup() {
 
   the_mesh.begin(fs);
 
-#if defined(NRF52_PLATFORM)
   board.initWatchdog(the_mesh.getNodePrefs()->wdt_timeout_secs);
-#endif
 
 #ifdef DISPLAY_CLASS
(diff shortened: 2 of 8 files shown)


radiomeano:
I noticed a small compile issue in src/helpers/CommonCLI.cpp.
In the non-NRF52 branch of `set watchdog`, the `sprintf()` line is missing the final semicolon:
sprintf(reply, "OK - Watchdog currently not implemented on this platform")
Suggested fix:
strcpy(reply, "OK - Watchdog currently not implemented on this platform");
or just add the missing semicolon.
One additional thought: since `wdt_timeout_secs` is a new preference byte, it may be safer to default it to 0 when loading older `/com_prefs` files that do not contain it.

winnieXY:
> I noticed a small compile issue in src/helpers/CommonCLI.cpp. In the non-NRF52 branch of `set watchdog`, the `sprintf()` line is missing the final semicolon: sprintf(reply, "OK - Watchdog currently not implemented on this platform") Suggested fix: strcpy(reply, "OK - Watchdog currently not implemented on this platform"); 

Oh - damn. I'm testing it on three different repeaters - but obviously all nRF52 ones - so this line is on all of them dead code and optimized away by gcc - so no compile issues at all :S

winnieXY:
>  One additional thought: since `wdt_timeout_secs` is a new preference byte, it may be safer to default it to 0 when loading older `/com_prefs` files that do not contain it.

I'll check this. Initially I thought that this should be handled by:
```
 memset(&_prefs, 0, sizeof(_prefs)); (see line 861)
```
as a older file is smaller it should defaults to 0 - but I'll double-check that.

radiomeano:
> > I noticed a small compile issue in src/helpers/CommonCLI.cpp. In the non-NRF52 branch of `set watchdog`, the `sprintf()` line is missing the final semicolon: sprintf(reply, "OK - Watchdog currently not implemented on this platform") Suggested fix: strcpy(reply, "OK - Watchdog currently not implemented on this platform");
> 
> Oh - damn. I'm testing it on three different repeaters - but obviously all nRF52 ones - so this line is on all of them dead code and optimized away by gcc - so no compile issues at all :S

No problem Patrick, that makes sense.
Since all your test devices are nRF52, the non-NRF52 branch is never compiled there, so the missing semicolon would only show up when building for another platform.
Thanks for confirming, the watchdog feature itself still looks very useful for remote nRF52 repeaters.

radiomeano:
> > One additional thought: since `wdt_timeout_secs` is a new preference byte, it may be safer to default it to 0 when loading older `/com_prefs` files that do not contain it.
> 
> I'll check this. Initially I thought that this should be handled by:
> 
> ```
>  memset(&_prefs, 0, sizeof(_prefs)); (see line 861)
> ```
> 
> as a older file is smaller it should defaults to 0 - but I'll double-check that.

Thanks Patrick for checking this, and thanks for your work on this feature.
If the prefs struct is already zero-initialized before loading `/com_prefs`, then an older/smaller file should indeed leave `wdt_timeout_secs` at 0, so watchdog disabled.
I only mentioned it as a migration-safety thought. Maybe just worth double-checking that the `memset` clears the actual prefs struct, not only the `_prefs` pointer.
I’m personally very interested in this feature because I have a SenseCAP Solar repeater that I plan to deploy at a high/mountain site, where a manual power cycle would be difficult.

winnieXY:
@oltaco  @ripplebiz - Any chance to get the watchdog feature into meshcore? There are plenty of users waiting to have some kind of watchdog/restart feature.

PR #2315, opened 2026-04-16
feat: add USB Companion Radio env for Heltec Wireless Paper

jirogit:
## Summary

Adds USB Companion Radio firmware environment for the Heltec Wireless Paper (ESP32-S3 + SX1262 + E-Ink).

Closes #2237

## Changes

- Added `[env:Heltec_Wireless_Paper_companion_radio_usb]` to `variants/heltec_wireless_paper/platformio.ini`

## Testing

Tested on hardware (Heltec Wireless Paper):

- Build: SUCCESS
- Flashed and connected via USB (CP2102) using MeshCore Android app over USB OTG
- E-Ink display renders Companion UI correctly

**Note:** Full flash erase required before flashing (`pio run -t erase`) to clear any data left by previously flashed firmware. Without erase, remnants in LittleFS from a previous firmware caused the Companion firmware to not initialize correctly.

**Note:** The Heltec Wireless Paper uses a CP2102 USB-UART bridge (not native USB CDC). Connection via browser Web Serial API (e.g. meshcore.liamcottle.net) may fail due to DTR/RTS reset behavior. The MeshCore Android app connected via USB cable directly to the device worked correctly.

PR #2407, opened 2026-04-26
fix: correct device model name for Heltec Wireless Paper

jirogit:
## Summary

Fixes incorrect device model name displayed in the MeshCore app for Heltec Wireless Paper.

Previously, the app showed **"Heltec V3"** because `target.h` and `target.cpp` used `HeltecV3Board` directly, which returns `"Heltec V3"` from `getManufacturerName()`.

This PR introduces `HeltecWirelessPaperBoard`, a minimal subclass of `HeltecV3Board` that overrides `getManufacturerName()` to return `"Heltec Wireless Paper"`.

## Changes

- `variants/heltec_wireless_paper/target.h`: Added `HeltecWirelessPaperBoard` subclass; updated `extern` declaration
- `variants/heltec_wireless_paper/target.cpp`: Updated board instance to use `HeltecWirelessPaperBoard`

## Testing

Tested on hardware (Heltec Wireless Paper) with both BLE and USB Companion firmware. Device model now correctly shows **"Heltec Wireless Paper"** in the MeshCore Android app settings.

This was discovered while testing the USB Companion firmware added in #2315.

### DIFF
--- a/variants/heltec_wireless_paper/target.cpp
+++ b/variants/heltec_wireless_paper/target.cpp
@@ -1,7 +1,7 @@
 #include "target.h"
 #include <Arduino.h>
 
-HeltecV3Board board;
+HeltecWirelessPaperBoard board;
 
 #if defined(P_LORA_SCLK)
   static SPIClass spi;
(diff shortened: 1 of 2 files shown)


liamcottle:
Looks good, thanks!

PR #2408, opened 2026-04-26
Feat: Add repeater hop limit specifically for advertisement packets

ghost:
This PR introduces a user-defined hop limit specifically for advertisement packets, similar to the existing flood.max implementation, using a new command set/get advert.max.hops.

Currently, our network is seeing a high volume of advertisement traffic travelling excessive distances (40+ hops), which consumes significant airtime without providing a realistic contact or path due to increased probability of collision.

As shown in the attached messaging stats (credit: Mopatops), advertisement packets are accounting for 55% of total Air Time. Many of these packets originate from nodes far beyond a useful range. This creates additional traffic on the mesh and prevents messages from getting through.

image

Instead of lowering the global flood.max (which would impact all traffic), this change targets only advertisement packets, allows users to set a specific threshold for these packets and drops adverts that exceed this hop count while allowing normal messages and DMs to propagate normally.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -429,6 +429,9 @@ void MyMesh::sendFloodReply(mesh::Packet* packet, unsigned long delay_millis, ui
 bool MyMesh::allowPacketForward(const mesh::Packet *packet) {
   if (_prefs.disable_fwd) return false;
   if (packet->isRouteFlood() && packet->getPathHashCount() >= _prefs.flood_max) return false;
+  if (packet->isRouteFlood()
+     && packet->getPayloadType() == PAYLOAD_TYPE_ADVERT
+     && packet->getPathHashCount() >=_prefs.advert_max_hops) return false;
   if (packet->isRouteFlood() && recv_pkt_region == NULL) {
     MESH_DEBUG_PRINTLN("allowPacketForward: unknown transport code, or wildcard not allowed for FLOOD packet");
     return false;
@@ -887,6 +890,7 @@ MyMesh::MyMesh(mesh::MainBoard &board, mesh::Radio &radio, mesh::MillisecondCloc
   _prefs.flood_advert_interval = 12; // 12 hours
   _prefs.flood_max = 64;
   _prefs.interference_threshold = 0; // disabled
+  _prefs.advert_max_hops = 64;
 
   // bridge defaults
   _prefs.bridge_enabled = 1;    // enabled
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -88,7 +88,7 @@ void CommonCLI::loadPrefsInt(FILESYSTEM* fs, const char* filename) {
     file.read((uint8_t *)&_prefs->adc_multiplier, sizeof(_prefs->adc_multiplier));                 // 166
     file.read((uint8_t *)_prefs->owner_info, sizeof(_prefs->owner_info));                          // 170
     file.read((uint8_t *)&_prefs->rx_boosted_gain, sizeof(_prefs->rx_boosted_gain));              // 290
-    // next: 291
+    file.read((uint8_t *)&_prefs->advert_max_hops, sizeof(_prefs->advert_max_hops));              // 291
 
     // sanitise bad pref values
     _prefs->rx_delay_base = constrain(_prefs->rx_delay_base, 0, 20.0f);
@@ -179,7 +179,7 @@ void CommonCLI::savePrefs(FILESYSTEM* fs) {
     file.write((uint8_t *)&_prefs->adc_multiplier, sizeof(_prefs->adc_multiplier));                 // 166
     file.write((uint8_t *)_prefs->owner_info, sizeof(_prefs->owner_info));                          // 170
     file.write((uint8_t *)&_prefs->rx_boosted_gain, sizeof(_prefs->rx_boosted_gain));              // 290
-    // next: 291
+    file.write((uint8_t *)&_prefs->advert_max_hops, sizeof(_prefs->advert_max_hops                 // 291
 
     file.close();
   }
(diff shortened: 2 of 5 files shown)


examples/simple_repeater/MyMesh.cpp
```diff
@@ -887,6 +890,7 @@ MyMesh::MyMesh(mesh::MainBoard &board, mesh::Radio &radio, mesh::MillisecondCloc
   _prefs.flood_advert_interval = 12; // 12 hours
   _prefs.flood_max = 64;
   _prefs.interference_threshold = 0; // disabled
+  _prefs.advert_max_hops = 64;
```
neilalexander: Should agree on a lower sane default here, otherwise there will always be node operators who just don't set it (or don't realise that they should).

ghost: In my own testing I have found lowering this to 10 still generates a substantial list of discovery contacts, but also understand that some regions may not want to use this functionality at all, happy to go with the majority on this one

mattzzw:
This idea is pretty similar to #1338.

neilalexander:
[Had a discussion](https://discord.com/channels/1495203904898728149/1495410606985969765/1501897121320796181) with @liamcottle today on this and we discussed whether tweaking and combining this approach with per-advert TTLs would be a better direction. In effect something like:

- Going forward, adverts get a configurable per-packet TTL, maybe reusing the reserved feature flag bits in the packet
- Adverts that don't have a TTL set (e.g. from old nodes with old firmware) can be caught and dropped by a PR like this one once a configured path limit is reached
- We choose some default values for what repeater TTLs and companion TTLs should be out of the box; realistically repeater ones can default to being much lower

The rationale being that there's a legitimate reason why you might want companion devices to be heard from further away in order to actually make contact with people, but that there's very little reason to know about repeaters more than a couple hops away. (There is still benefit in knowing about local ones, otherwise people think there's no infrastructure around them and give up.)

The problem with a TTL-only approach is that it doesn't deal with nodes that can't or won't be updated. The problem with a hop limit-only approach is that it makes it possible to enforce unfairly low limits on unsuspecting nodes. But combining the two is reasonably fair and solves both angles.

Some more thought would be needed to determine what the default values should be, although my feeling is that it's much better to start with safe/conservative low defaults that can be increased, rather than starting with unsafe/relaxed high defaults that cause problems and need to be decreased for anything to work. We both agreed that defaults were important though.

mattzzw:
I am happy that the issue of advert flood traffic is getting addressed by the devs.

It was one of the reasons to create and maintain a community firmware fork as 50% of the airtime in our mesh was eaten up by flood adverts. By dampening the adverts via #1338 we got the situation under control. But I have also given up on the hope of #1338 getting merged (and I am tired of continuously rebasing a PR...).

ghost:
I'm not sure that this proposal would solve the issue of 'advert spammers', which seems to have started since I submitted this PR and wasnt it's intention. Something would need to be included to mitigate bad actors, otherwise they'll just set whatever the new settings are to the maximum available can carry on flooding the mesh.

ripplebiz:
Can you please change the base branch of the PR to 'dev'.

liamcottle:
Would be good if we can reserve `advert.max.hops` and `prefs.advert_max_hops` for outgoing adverts that originate from the self node/repeater itself.

This would make sense for when/if we implement a TTL/hop limit on the advert packet itself, as described in https://github.com/meshcore-dev/MeshCore/pull/2408#issuecomment-4396836792

An outgoing advert could have the feature flags tag itself with the max hops configured in prefs. Repeaters can then check this value against the current path size. If it's over the limit, it can just drop the packet. Allowing for repeater owners to self limit their own repeater adverts, without the whole network having to impose arbitrary limits for everyone else.

And then refactor the variables/commands in this PR to use something along the lines of:

- `advert.fwd.max.hops`
- and `prefs.advert_fwd_max_hops`.

`fwd` relating to the fact that the config relates to adverts being forwarded/repeated.

The naming with dots is getting a bit long, so maybe something else could be appropriate...?

ripplebiz:
Am willing to merge this, if the merge conflicts can be fixed, and to be consistent with new features probably should be 'flood.max.advert' as CLI config.  There is 'flood.max' and now a new 'flood.max.unscoped'.

ghost:
> Am willing to merge this, if the merge conflicts can be fixed, and to be consistent with new features probably should be 'flood.max.advert' as CLI config. There is 'flood.max' and now a new 'flood.max.unscoped'.

Resolved merge conflicts and updated the CLI configuration to flood.max.advert to match the other flood features. Ready for another look whenever you have time.

liamcottle (approved):

src/helpers/CommonCLI.cpp
```diff
@@ -797,6 +808,8 @@ void CommonCLI::handleGetCmd(uint32_t sender_timestamp, char* command, char* rep
     sprintf(reply, "> %d", (uint32_t)_prefs->flood_max_unscoped);
   } else if (memcmp(config, "flood.max", 9) == 0) {
     sprintf(reply, "> %d", (uint32_t)_prefs->flood_max);
+  } else if (memcmp(config, "flood.max.advert", 16) == 0) {
```
ripplebiz: Need to move this else if to BEFORE the "flood.max" one.

.github/FUNDING.yml
```diff
@@ -0,0 +1 @@
+github: meshcore-dev
```
ripplebiz: Looks like this diff shouldn't be here, and was from the main branch

ghost: not sure what happened there! hopefully fixed now :)

liamcottle: File shouldn't be deleted. It will be best if you rebase your local branch from `dev` instead of `main`. The commits from main branch were pulled in when you changed base branch from `main` to `dev` in github ui. This happens because your local branch was branched off from main, and now github wants to merge those extra main commits missing from dev into it.

PR #2414, opened 2026-04-27
feat(rak3401): AIN1 button for status screen and shutdown

dorfman2:
## AIN1 Button for Status Screen and Shutdown (RAK3401) 🤖🤖

Adds physical button support on AIN1 (pin 31) for the RAK3401 1W repeater, designed for use with a [3D printed portable case](https://www.printables.com/model/1627139-case-for-rak-wisblock-1-watt-starter-kit) that incorporates an Omron-style limit switch as a user button.

> **Note from the Human**: I combined these two feature into to a single PR, but am happy to split this out and discuss further. This is a specific use case where a repeater is being deployed at a festival, and isn't "set & forget." I do think switch functionality could be inherent on this board, as long as we maintain the normally open logic, similar to the OLED screen (always on, but doesn't always physically exist).

### Button Actions
- **0.5s hold** → Status screen (uptime, battery voltage, percentage, gauge bar, charging indicator)
- **5s hold** → Graceful shutdown with "Shutting down..." message, then NRF52 SYSTEMOFF
- **Press while off** → Wake from SYSTEMOFF via GPIO SENSE

### Screenshots

| Home Screen | Status Screen | Final Hardware Build |
|---|---|---|
| IMG_9866 | IMG_9867 | IMG_9869 |

> ### NOTE: Screen does auto-off to save power as in the current code, but uploading photos of the screen black is decidedly not helpful.

### UI Changes
- Redesigned repeater screens with angular theme for small OLEDs
- Inverted header bars, horizontal rule separators, label:value layout
- Full-width battery gauge bar with percentage
- Charging indicator (plug icon + "CHG") when USB power detected via `isExternalPowered()`

### Hardware Changes
- `PIN_USER_BTN_ANA=31` build flag for `RAK_3401_repeater`
- Public `powerOff()` override on `RAK3401Board` delegating to `initiateShutdown(SHUTDOWN_REASON_USER)`
- GPIO SENSE LOW wake configured on AIN1 before SYSTEMOFF
- Waits for button release before entering SYSTEMOFF to prevent immediate wake

### Notes
- Uses `digitalRead` instead of `analogRead` — nRF52 SAADC disconnects `INPUT_PULLUP` during conversion
- Battery percentage: linear mapping 3.0V (0%) → 4.2V (100%) for single-cell 3.7V LiPo
- UI changes are in `examples/simple_repeater/UITask.*` which affects all repeater variants

### Case Design
https://www.printables.com/model/1627139-case-for-rak-wisblock-1-watt-starter-kit

Made as a portable case for the RAK Wisblock 1 Watt starter kit. This case accommodates a very large battery (10000mAh or 15000mAh, 60mm x 110mm x 10mm) and incorporates a standard 128x64 OLED screen, standard micro toggle for power, an Omron style limit switch for user button, and two SMA antenna holes.

### DIFF
--- a/examples/simple_repeater/UITask.cpp
+++ b/examples/simple_repeater/UITask.cpp
@@ -8,6 +8,18 @@
 
 #define AUTO_OFF_MILLIS      20000  // 20 seconds
 #define BOOT_SCREEN_MILLIS   4000   // 4 seconds
+#define STATUS_SCREEN_MILLIS 5000   // 5 seconds
+
+// Electric plug icon 16x7px (side profile, cord left, prongs right)
+static const uint8_t charging_icon [] PROGMEM = {
+  0x01, 0xE0,  // .......####.....
+  0x03, 0xE0,  // ......#####.....
+  0x03, 0xFE,  // ......#########.
+  0xFF, 0xE0,  // ###########.....
+  0x03, 0xFE,  // ......#########.
+  0x03, 0xE0,  // ......#####.....
+  0x01, 0xE0,  // .......####.....
+};
 
 // 'meshcore', 128x13px
 static const uint8_t meshcore_logo [] PROGMEM = {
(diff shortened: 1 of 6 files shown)


ripplebiz:
Nice work.  
Just a small one, can you adjust the alignment of the SF and CR values.  Looks like   '8  CR' is associated, instead of '8' being associated with 'SF':
Screenshot 2026-04-28 at 2 04 41 pm

dorfman2:
> Nice work. Just a small one, can you adjust the alignment of the SF and CR values. Looks like '8 CR' is associated, instead of '8' being associated with 'SF': Screenshot 2026-04-28 at 2 04 41 pm

Yes I can! Should be pushed shortly.

dorfman2:
**Label Alignment complete.** 
IMG_9883

- There were also a minor change to make battery min/max for the display adjustable during build (default 4200mV, no local override). It gives me hives to never reach 100% charge, as my battery tops out at 4.18v 😁 .
- I can pull this out if you want, since this would qualify as feature creep.

**How it works:**

- By default, every board gets 3000mV–4200mV (standard single-cell LiPo range)
- If a board needs different values (different battery chemistry, charging circuit caps at a lower voltage, etc.), they add -D BATT_MAX_MILLIVOLTS=4100 to their platformio.ini build flags
- The -D flag defines the macro at compile time before the #ifndef check, so the default is skipped

**Why it matters:**

- Not all charging circuits charge to 4.2V — some cap at 4.1V or 4.0V for longevity
- Some setups use LiFePO4 cells (3.6V max) or dual-cell packs
- This is the same pattern the companion radio already uses for its battery indicator

For this PR, the defaults stay at 3000/4200 so nothing changes for existing users. Any board variant can override via a single build flag line.

dorfman2:
**Follow-up: dropped the battery threshold build flags**
Force-pushed a new HEAD to remove the #ifndef BATT_MIN_MILLIVOLTS / BATT_MAX_MILLIVOLTS guard block that was in 
UITask.cpp
. The calculation is back to the original 3000/4200 mV literals.

Why: the CLI already exposes runtime tuning for the battery reading, so a compile-time override was duplicating existing functionality.

From the [MeshCore CLI docs](https://docs.meshcore.io/cli_commands/#fine-tune-the-battery-reading):

**Fine-tune the battery reading**
- `get adc.multiplier`
- `set adc.multiplier <value>` (range 0.0–10.0, default 0.0 = board-defined)
- Returns `Error: unsupported by this board` if the hardware doesn't support it.

Net change on the branch:
- `8482a098` (mixed: SF/CR alignment + battery guards) → `71ae7839` (SF/CR alignment only, cleaner message).
- Build still clean on RAK_3401_repeater, flash size unchanged.

The AIN1 button feature itself is untouched.

dorfman2:
@ripplebiz Wanted to check back in on merging this. If you don't think it's helpful, I can close this.

disq:
Just for repeater, why not companion also?

dorfman2:
> Just for repeater, why not companion also?

The companion already has a comprehensive UI, and I didn't want to mess with it. I'm happy to add a page to it if you think there's a desire, but it feels like a duplication of the already existing battery indicator. 

Or are you referring to exposing the additional button pin?

PR #2417, opened 2026-04-27
Tbeam1 w fix

Robowarrior834:


MoppelMat:
maybe a dumb question: but it is in my view these days because I just got one of these Tbeam 1W devices...
What does this change mean exactly?

I see a lot of stuff that seems to have not much to do with the Lilygo T-Beam 1W here and then it is closed, removed and added back again.
image

I am confused and sorry if that is a dumb question!

Robowarrior834:
No dumb questions. I had to close it because I accidently pushed more then just the tbeam to it. and I could not figure out how to undo it.

PR #2418, opened 2026-04-27
Fix: Lillygo Tbeam 1W Fan Control and LNA Sequencing improvements 

Robowarrior834:
Ensures proper sequencing of the LNA and amp for the TBEAM 1W. Also adds fan control to keep the fan on for 10 seconds.


Robowarrior834:
Had to rebuild the repository as I accidently added other local changes I made on my machine to the old pull request

examples/companion_radio/main.cpp
```diff
@@ -229,4 +229,7 @@ void loop() {
   ui_task.loop();
 #endif
   rtc_clock.tick();
```
ripplebiz, line 300: Is probably time to introduce a new virtual MainBoard::loop(), and a no-op as default.
And have the TBeam1WBoard class override, and put the updateFan() logic there

variants/lilygo_tbeam_1w/TBeam1WBoard.cpp
```diff
@@ -3,30 +3,49 @@
...
 
 void TBeam1WBoard::onAfterTransmit() {
-  digitalWrite(LED_PIN, LOW);   // TX LED off
+  digitalWrite(SX126X_RXEN, HIGH); // re-enable LNA immediately after TX
+  digitalWrite(LED_PIN, LOW);
+  // Keep fan running for 10s after TX
+  setFanEnabled(true);
+  _fan_off_millis = millis() + 10000;
+}
+
```
ripplebiz: ie. here  make this the board's loop() method

Robowarrior834: So I did this. Would I also need to add a loop function to every variant?

Robowarrior834: I answered my own question. Added a new virtual loop to the mainboard class so other variants compile and adjusted the tbeam 1w to override.

Robowarrior834: @ripplebiz

r0ck40k:
I basically did the same. Which seems to improve reliability on this board a lot. While changing this i wanted to ask if it makes sense to allow adjusting the PA ramp time dynamically?
I read the Lilygo documentation and for this board they prefer >= 800us for PA ramp time. 
image

Maybe it makes sense to add something like:
1. The build flag (definition) — variants/lilygo_tbeam_1w/platformio.ini
    -D SX126X_PA_RAMP=0x05 

2. Initial application at radio init — variants/lilygo_tbeam_1w/target.cpp
  radio.setTxParams(LORA_TX_POWER, SX126X_PA_RAMP);

3. Re-application after every TX power change — src/helpers/radiolib/CustomSX1262Wrapper
   void setTxPower(int8_t dbm) override {
      ((CustomSX1262 *)_radio)->setOutputPower(dbm);
  #ifdef SX126X_PA_RAMP
      ((CustomSX1262 *)_radio)->setPaRampTime(SX126X_PA_RAMP);
  #endif
  }


This is at least my understanding. Please correct me if i am wrong.

Robowarrior834:
to out of date to try and merge

PR #2421, opened 2026-04-27
add .clangd

446564:


446564:
@recrof easy one

recrof:
@liamcottle can we merge please?

ripplebiz:
As far as I know, the .clangd files are typically committed.  Not an expert tho

446564:
Happy to do that instead, it's just flags to either include or not and some rules for clangtidy.

446564:
@liamcottle if you merge, and want to keep the tidy stuff, you should also follow up with a commit that has run clang tidy on the codebase or PR will have tons of formatting changes. we can also remove the tidy portion for now, until we settle on some style guidlines or rules to apply anyway.

liamcottle:
@446564 I'm happy for the config/rules to apply to newly added code, before it's commit, but would prefer that none of this ever auto updates existing code in irrelevant files. Otherwise PRs become a mess to review.

PRs should only change what is necessary. It's annoying when someone adds a new feature that's a 2 line change, but it comes with a 600 line file reformat.

So I guess, no auto tidy for now then..?

446564:
> @446564 I'm happy for the config/rules to apply to newly added code, before it's commit, but would prefer that none of this ever auto updates existing code in irrelevant files. Otherwise PRs become a mess to review.
> 
> PRs should only change what is necessary. It's annoying when someone adds a new feature that's a 2 line change, but it comes with a 600 line file reformat.
> 
> So I guess, no auto tidy for now then..?

That will only affect files modified in commits, unless someone intentionally runs it against the whole project. And I just remembered this is only diagnostic.. :face_exhaling: so it won't change anything but suggest these changes as hints in editor.

Issue #2132, opened 2026-03-23
Feature Request: Authoritative Time Servers (GPS-Based Time Synchronization)


GavinLanoe:
Dear MeshCore Developers,

I’m proposing a new feature to improve time synchronization across the mesh network by introducing authoritative time servers based on GPS-enabled nodes.

Current Challenge

Accurate and consistent time across nodes is important for logging, coordination, and potential future features. In the current setup:

- Nodes may have unsynchronised, drifting clocks or clocks that are years out due to no RTC being present on node hardware.
- There is no standardized way to distribute accurate time
- GPS-equipped nodes already have access to highly accurate time, but this is not shared across the mesh

This leads to inconsistencies that can affect reliability and coordination between nodes.

Proposed Solution: Authoritative Time Server Nodes

Introduce a feature allowing any node (Client, Room Server, or Router) to be configured as an authoritative time server when equipped with a GPS module.

Server Behavior
- A node with GPS hardware enabled can be configured as a Time Server
- The node responds to time requests from other nodes with its current GPS-derived time
- The node must validate GPS quality before responding:
- Minimum satellite count (configurable threshold)
- Valid GPS fix (e.g., 2D/3D fix)
- Nodes with time server capability should include availability and status flags in existing flood and zero-hop advertisements
This avoids creating additional data transmissions on the network

- Advertisements indicate whether the time server is enabled and GPS fix is valid

If GPS quality is insufficient:
The node responds with a “time unavailable” status instead of inaccurate data to allow a client to move on to another node quicker than waiting for a timeout to expire.

Client Behavior

Client nodes should support configurable time synchronization:

- Ability to enable/disable time synchronization
- Ability to define a priority list of time servers (e.g., up to 5 nodes)
- Ability to auto-populate from discovered nodes advertising time capability
- Clients may automatically build or update their time server list based on these advertisements, removing the need for manual configuration unless desired;
         This minimizes transport of timing messages across the wider network
         Reduces latency in clock synchronization caused by long-distance transmission
         Optimizes overall network efficiency and reduces congestion

Time request logic:

- Attempt to query the closest (fewest hops) time server
- If no response or invalid response → move to next server in the priority list
- Continue until a valid time is received or list is exhausted

Optional enhancements:
- Configurable retry intervals (If the node has previously received a good time update there is probably no need to update the time more than every 12 hours)
- Caching last known good time source
- Timeout thresholds per server

Startup Behavior
To improve initial synchronization and network awareness:

[...]

Issue #1882, opened 2026-02-28
[MeshCore APP] (iOS) uses companion node time for clock sync instead of phone time


solkmaaker:
Companion clock may be wrong, considering that there is no way to set (or sync) companion node time from app.

So if you sync repeater clock from app, it will use that wrong time from companion node, and not phone time.

liamcottle:
Howdy, the app automatically sets the companion time when you connect to it.

When you click "Sync Clock" in the remote management section, it does currently use the companion clock, which should have already been synced when you connected to it.

One issue that can happen here, is if the clock on the companion is running fast, in which case the sync would send a future time from the companion to the repeater. Unfortunately we don't allow for clocks to go backwards without a reboot at this time. Mostly due to other internal logic which drops packets if timestamps are older.

If you need to set a specific time, you can run the CLI command `time 1772317204` where the number is unix timestamp in seconds.

solkmaaker:
This is all true, but there are devices that do not reset clock to specific time in past during reboot.
My LilyGo T-Echo (BLE companion) has time years in future, and there is  no way to get it correct.
So, some sort of solution is needed for such devices to get correct time in them.

When i flashed it as repeater, time was in the past, so i was able to set correct time then.
But as soon as i flash it as companion, time is again in the future, years ahead.
It may be a firmware bug (latest from official flasher), but still, syncing companion time is not best behavior IMHO, since you have device (phone) which has much higher probability for its time to be more correct.

solkmaaker:
For those whose search will end up here:
Soft reboot did not help, reboot from side button did not help.
What helped was: entering DFU mode, and then reboot from side button.
After that, connecting to device with app, time was correct.
So it is NOT firmware bug. Feels more like device quirk or something.

Issue #1332, opened 2026-01-06
sync clock and time commands will fail on repeater if repeaters clock is in future


do6uk:
How ever this could happen ... if repeaters clock gone in future time, you will not be able to correct this without cutting of power-suply including battery and solar-power. The firmware doesn't accept to correct the clock backwards.

I suggest to apply the following change in the firmware - file **CommonCLI.cpp** ...

```
    } else if (memcmp(command, "time ", 5) == 0) {  // set time (to epoch seconds)
      uint32_t secs = _atoi(&command[5]);
      uint32_t curr = getRTCClock()->getCurrentTime();
      if (secs > curr) {
        getRTCClock()->setCurrentTime(secs);
        uint32_t now = getRTCClock()->getCurrentTime();
        DateTime dt = DateTime(now);
        sprintf(reply, "OK - clock set: %02d:%02d - %d/%d/%d UTC", dt.hour(), dt.minute(), dt.day(), dt.month(), dt.yea>
      } else if (secs == 0) {
        getRTCClock()->setCurrentTime(1767225600);
        uint32_t now = getRTCClock()->getCurrentTime();
        DateTime dt = DateTime(now);
        sprintf(reply, "OK - clock RESET: %02d:%02d - %d/%d/%d UTC", dt.hour(), dt.minute(), dt.day(), dt.month(), dt.y>
      } else {
        strcpy(reply, "(ERR: clock cannot go backwards)");
      }
```

after this change admins can reset repeaters clock by sending time 0 as a command to a epoch-time in past.

IoTThinks:
Let me test your code.
I have "clock fsync" to force the timesync.

I'm thinking to let repeater to sync time from companions.

do6uk:
That sounds also nice. I just added this few lines quick and dirty - a special command would be fine.

csrutil:
See https://github.com/meshcore-dev/MeshCore/pull/1326

themelisx:
> Let me test your code. I have "clock fsync" to force the timesync.
> 
> I'm thinking to let repeater to sync time from companions.

This does not work to esp32-s3 cause soft reboot will keep the clock. 
You need to send the command "clkreboot" and then "clock sync".

In any case it's annoying to cannot set time backwards.

PeterAlfonsLoch:
> > Let me test your code. I have "clock fsync" to force the timesync.
> > I'm thinking to let repeater to sync time from companions.
> 
[...]

PR #2423, opened 2026-04-27
Opportunistic clock sync from verified advert timestamps

### LINKED ISSUE #1329: Clock-synchronization feature based on the average time from the last N adverts

pelgraine:
Closes #1329. Related: #2132, #1332, #1882.
 
## Problem
 
GPS-less repeaters lose their clock on every power cycle, reverting to the firmware default (~May 2024). This means their adverts carry stale timestamps until an admin manually syncs the clock via serial, the mobile app, or a remote management command. For inaccessible repeaters (rooftops, hilltops, solar-powered sites that reboot daily), manual sync is impractical.
 
## Solution
 
Add an opt-in mechanism for nodes to sync their local RTC from signature-verified advert timestamps. When a node's clock is clearly wrong (before Jan 2025) and it receives a valid advert with a sane timestamp (Jan 2025 – Jan 2036), it adopts that timestamp.
 
The sync only fires when:
1. The sub-class opts in via `shouldSyncClockFromMesh()` (defaults to `false`)
2. The local clock is before Jan 2025 (clearly wrong / default state)
3. The incoming advert timestamp passes sanity checks
4. The advert signature has already been cryptographically verified
Once the clock is set past Jan 2025, the sync stops firing — it won't overwrite a valid clock with subsequent adverts.
 
## Changes
 
**`src/Mesh.h`** — New virtual method:
```cpp
virtual bool shouldSyncClockFromMesh() const { return false; }
```
 
**`src/Mesh.cpp`** — 14 lines added after advert signature verification (inside the `PAYLOAD_TYPE_ADVERT` handler), before `onAdvertRecv()` call. Checks the opt-in gate, validates both local and incoming timestamps, and calls `setCurrentTime()` if appropriate.
 
**`examples/simple_repeater/MyMesh.h`** — Single-line override:
```cpp
bool shouldSyncClockFromMesh() const override { return true; }
```
 
## What this does NOT do
 
- Does not affect companion nodes, clients, or room servers (they don't override the method)
- Does not overwrite a valid clock — only fires when local time is clearly wrong
- Does not sync backward — a node with a stale clock can't drag others back
- Does not add any new preferences, CLI commands, or protocol changes
- Does not increase packet size or airtime

## Testing
 
1. Flash a repeater without GPS/RTC
2. Confirm clock shows default (~May 2024) on boot
3. Wait for a valid advert from any peer with correct time
4. Confirm serial output shows `clock synced from advert: <epoch>`
5. Confirm subsequent adverts do NOT trigger further syncs
6. Confirm a repeater WITH GPS/RTC is unaffected (clock already valid)

## Tested

- Station G2 repeater (both ESPNow bridge repeater and standard) - example screenshots from test:
<table>
<tr>
<td>IMG_2594</td>
<td>IMG_2595</td>
<td>IMG_2596</td>
</tr>
</table>

- Heltec v3 espnow bridge repeater
- T-Echo Lite repeater (no gps) - example screenshots from test:
<table>
<tr>
<td>IMG_2597</td>
<td>IMG_2598</td>
<td>IMG_2599</td>
</tr>
</table>

### DIFF
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -268,9 +268,7 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
         if (is_ok) {
           MESH_DEBUG_PRINTLN("%s Mesh::onRecvPacket(): valid advertisement received!", getLogDateTime());
 
-          // Opportunistic clock sync: if local clock is clearly wrong and the
-          // verified advert carries a sane timestamp, adopt it.  Only fires
-          // when the sub-class opts in (e.g. repeaters without hardware RTC).
+       // sync clock from verified advert if local clock is clearly wrong (for repeaters without hardware RTC)
           if (shouldSyncClockFromMesh()) {
             uint32_t local_time = _rtc->getCurrentTime();
             if (local_time < 1735689600UL          // local clock before Jan 2025 (clearly wrong)
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -267,6 +267,20 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
         }
         if (is_ok) {
           MESH_DEBUG_PRINTLN("%s Mesh::onRecvPacket(): valid advertisement received!", getLogDateTime());
+
+          // Opportunistic clock sync: if local clock is clearly wrong and the
+          // verified advert carries a sane timestamp, adopt it.  Only fires
+          // when the sub-class opts in (e.g. repeaters without hardware RTC).
+          if (shouldSyncClockFromMesh()) {
+            uint32_t local_time = _rtc->getCurrentTime();
+            if (local_time < 1735689600UL          // local clock before Jan 2025 (clearly wrong)
+                && timestamp > 1735689600UL         // advert after Jan 2025
+                && timestamp < 2082758400UL) {      // advert before Jan 2036
+              _rtc->setCurrentTime(timestamp);
+              MESH_DEBUG_PRINTLN("  clock synced from advert: %lu", (unsigned long)timestamp);
+            }
+          }
+
           onAdvertRecv(pkt, id, timestamp, app_data, app_data_len);
           action = routeRecvPacket(pkt);
         } else {
--- a/examples/simple_repeater/MyMesh.h
+++ b/examples/simple_repeater/MyMesh.h
@@ -157,6 +157,8 @@ protected:
     return _prefs.multi_acks;
   }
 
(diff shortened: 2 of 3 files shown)


4np:
I took a stab at peer-based time synchronization in #1443 , but there were genuine issues with the approach of relying on peer-based timed (such as clock poisoning, future dated clocks on ESP32, etcetera). There is also no way of knowing if some other's advert is on the right time.

My PR used a statistical approach ([Median Absolute Deviation](https://en.wikipedia.org/wiki/Median_absolute_deviation) by sampling many packets, while taking estimated time of flight into account. There are actually quite some genuine concerns by relying on peers to synchronize time, as also voice by reviewers.  Also, with the continuous move towards reducing the number of flood adverts and configuring regions the reliance on flood packets for time keeping has also become less reliable. As such I have closed the PR. 

This PR doesn't have any safeguards, doesn't use statistics, doesn't take time of flight into account and has the same issues (and more) trusting peer time. This PR simply takes any packet that comes around that has a time between Jan 2025 and Jan 2026 and updates their time to match the packet. If I were to send a flood advert using a timestamp set to April 1st, 2030, your repeater would happily adopt it.

src/Mesh.cpp
```diff
@@ -267,6 +267,18 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
         }
         if (is_ok) {
           MESH_DEBUG_PRINTLN("%s Mesh::onRecvPacket(): valid advertisement received!", getLogDateTime());
+
+       // sync clock from verified advert if local clock is clearly wrong (for repeaters without hardware RTC)
+          if (shouldSyncClockFromMesh()) {
+            uint32_t local_time = _rtc->getCurrentTime();
+            if (local_time < 1735689600UL          // local clock before Jan 2025 (clearly wrong)
+                && timestamp > 1735689600UL         // advert after Jan 2025
```
4np, line 275: You cannot assume a packet with a time between Jan 2025 and Jan 2036 has the right time, it can be off by years.

PR #2426, opened 2026-04-28
make g2 more in line with other variants

recrof:
- removed nrf52840_g2_ultra 
- added repeater and room server roles
- moved shared libdeps to Nano_G2_Ultra definition


recrof:
@446564 can you please build and try if those new(and older) envs build and work with your g2? thanks!

446564 (approved):
all targets tested and working

liamcottle:
Thanks all!

PR #2432, opened 2026-04-28
kiss radio: added stm32 support + envs in platformio.ini

recrof:
- add KISS support for stm32 boards
- add kiss_radio envs for most of the boards


tuzzmaniandevil:
OMW THANK YOU!
I was going to buy a heltec to do this haha. Now I can use all the boards with my project!

PR #2438, opened 2026-04-29
Add PERM_ACL_REGION_MGR role for delegated region management

hank:
Introduces a fourth ACL role (value 4) that can manage the region map without full admin privileges. The role is intended for trusted users who curate regions on a repeater but should not have access to general admin commands.

ClientACL:
  - Widen PERM_ACL_ROLE_MASK from 2 to 3 bits so the new value fits. (it's uint8_t in the packet so this is backwards compatible)
  - Add PERM_ACL_REGION_MGR and ClientInfo::isRegionMgr().
  - Exempt region_mgr entries from least-recently-active eviction in putClient(), same as admins.

simple_repeater:
  - Phones may still gate UI on the legacy is_admin byte (reply_data[6]), so report region_mgr as admin there. Without this, the phone CLI falls back to guest view.
  - Allow region_mgr to send TXT_MSG CLI commands. handleCommand() gates non-whitelisted commands with "Err - not permitted". The whitelist covers region.* (read+write) plus a small set of read-only queries (get, ver, board, neighbors, clock, sensor get/list).
  - Pass the ClientInfo* through to handleCommand and drop the redundant sender_timestamp parameter (derived from sender->last_timestamp; NULL means Serial CLI).
  - Use ~PERM_ACL_ROLE_MASK instead of ~0x03 when clearing role bits on login, so the wider mask is honored.

#2437 was tested along with this change. Everything functioned as expected using a modified client with the new permission level for region managers added. Everything should also be backwards compatible with clients that do not support this - using level 3 admin still works the same way it did before.


hank:
I have created a working fork of meshcore-open for testing, you can find it here:

https://github.com/hank/meshcore-open

Issue #2431, opened 2026-04-28
[Documentation request] Document, what `txdelay` and `rxdelay` actually do and what the mechanics behind is.


dreirund:
From [confusion on what `txdelay` and `rxdelay` really do and what the mechanics behind is](https://www.reddit.com/r/meshcore/comments/1sxxrou), I hereby have a documentation request to have this thoroughly documented, and a link to that documentation added to the [repeater CLI documentation](https://docs.meshcore.io/cli_commands/#view-or-change-the-retransmit-delay-factor-for-flood-traffic).

For now, it seems that it is only hidden in code, so one has to be able to read and understand the source code to get to know what those settings actually mean and do.

Regards!

kizniche:
I came up with some descriptions for the docs based on what I could understand from the code and discussions on the Meshcore discord.

dreirund:
> *I came up with some descriptions for the docs based on what I could understand from the code and discussions on the Meshcore discord.*

[Here](https://www.reddit.com/r/meshcore/comments/1sxxrou/comment/oiq2jax/) on reddit there is also some discussion -- and as far as it is now, it seems to be complicated and difficult to understand.

Anyway, someone did _write_ the code, and I think those people should actually come up with the details.

kizniche:
I'm going to keep my comments to the issue, as it makes it confusing to have the same conversation in both the issue and PR.

I was following the style of the documentation. Other Notes are brief, rather than detailed explanations, with equations, implications, etc. If you want a detailed description, equations, and implications, it's probably best to have that described in a separate document rather than in the CLI command doc.

dreirund:
> If you want a detailed description, equations, and implications, it's probably best to have that described in a separate document rather than in the CLI command doc.

That was my intention:

> *to have this thoroughly documented, and a **link to that documentation** added to the [repeater CLI documentation](https://docs.meshcore.io/cli_commands/#view-or-change-the-retransmit-delay-factor-for-flood-traffic).*

kizniche:
Okay. I will come up with a draft and post it in this issue later, and we can review for accuracy. I'll include code references to support any claims.

dreirund:
[...]

PR #2443, opened 2026-04-29
Fix: Enforce upper bounds for rxdelay, txdelay, and direct.txdelay

### LINKED ISSUE #2433: `txdelay` and `rxdelay` can be set higher than documented maximum.
According to the [repeater CLI documentation](https://docs.meshcore.io/cli_commands/#view-or-change-the-retransmit-delay-factor-for-flood-traffic), maximum value of `txdelay` is `2.0` and maximum value of `rxdelay` is `20.0`.

BUT I was able to do `set txdelay 15`, `set rxdelay 22`, and `get txdelay` shows `15` afterwards and `get rxdelay` shows `22` afterwards.

So something is wrong here.

What are the actual correct maximum values? Documentation should be corrected accordingly.

Regards!

kizniche:
Fixes #2433 by enforcing upper bounds for rxdelay, txdelay, and direct.txdelay, based on what is written in the CLI docs.

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -582,21 +582,21 @@ void CommonCLI::handleSetCmd(uint32_t sender_timestamp, char* command, char* rep
     strcpy(reply, "OK");
   } else if (memcmp(config, "rxdelay ", 8) == 0) {
     float db = atof(&config[8]);
-    if (db >= 0) {
+    if (db >= 0 && db <= 20.0f) {
       _prefs->rx_delay_base = db;
       savePrefs();
       strcpy(reply, "OK");
     } else {
-      strcpy(reply, "Error, cannot be negative");
+      strcpy(reply, "Error, must be 0-20");
     }
   } else if (memcmp(config, "txdelay ", 8) == 0) {
     float f = atof(&config[8]);
-    if (f >= 0) {
+    if (f >= 0 && f <= 2.0f) {
       _prefs->tx_delay_factor = f;
       savePrefs();
       strcpy(reply, "OK");
     } else {
-      strcpy(reply, "Error, cannot be negative");
+      strcpy(reply, "Error, must be 0-2");
     }
   } else if (memcmp(config, "flood.max ", 10) == 0) {
     uint8_t m = atoi(&config[10]);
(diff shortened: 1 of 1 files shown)


dreirund:
Maybe increasing the limits can be useful? As There:
https://github.com/user-attachments/files/26126001/Proposal-.MeshCore.factors.in.selecting.rxdelay.txdelay.and.direct.txdelay.pdf (linked at https://github.com/meshcore-dev/MeshCore/discussions/2053#discussioncomment-16214117) `txdelay = 2.5` is also within the range of proposed settings.

_Or_ just the documentation would need correction, whatever suits the intention?

kizniche:
Perhaps changing the defaults is warranted, but something that should be discussed in the issue, not the PR. This PR is simply aligning the code with the expected behavior (the docs).

CullenShane:
Denver already recommends values that are higher than these limits: https://denvermc.com/guides/repeater-setup

dreirund:
> Denver already recommends values that are higher than these limits: https://denvermc.com/guides/repeater-setup

So this PR would break existing use cases -- then I would say find the current maximum values and do a PR that aligns the docs with reality ;-).

446564:
This should not be enforced in this way. There is no reason to have arbitrary limits, the only limit should be that of the underlying datatype. Why would we impose our own idea of a maximum on another user when it is not required for interoperability? There is nothing in the protocol or routing that this will have any effect on so there is no limit, again other than the underlying type which is likely very high (unreasonably so)

The limit should clamp to the max value for the type and the docs should be updated at the same time with maybe some language regarding a sensible value range.

kizniche:
For rx, there's currently a hard 32 sec runtime cap, so the current 20 limit seems arbitrary. If it were to exceed ~62, then the 32 sec cap would come into effect. There's no runtime cap for tx, so high values create proportionately higher delays.

446564:
> For rx, there's currently a hard 32 sec runtime cap, so the current 20 limit seems arbitrary. If it were to exceed ~62, then the 32 sec cap would come into effect. There's no runtime cap for tx, so high values create proportionately higher delays.

i didn't dig too deep, but it seems going with the 32s makes sense for rx and tx i guess if someone wants to set 1000 they can, not that they should.

kizniche:
Sounds reasonable. This would go nicely with additional documentation for tx/rx delay settings (#2431) so users would have some sort of guidance/bearing when changing settings.

oltaco (approved):
Looks good to me 👍

edited to add: Changing the limits can be discussed, but this PR does fix #2433 and aligns CLI feedback with the current limits.

liamcottle:
Since the firmware already enforces these limits at reboot when loading prefs, it makes sense to prevent the user from saving a higher value. Otherwise when they reboot their setting will be wiped, leading to confusion.

https://github.com/meshcore-dev/MeshCore/blob/c940ea5f305df87c4ea215d74e522b64c22c66e9/src/helpers/CommonCLI.cpp#L94

If needed, increasing the limits can be discussed as @oltaco mentioned.

Issue #1086, opened 2025-11-09
Heltec V3.2 wire.cpp i2cRead Error


zerion87:
I get with my Heltec V3.2 the following Error. SW Version 1.9.0

ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x1 (POWERON),boot:0x8 (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fce3808,len:0x4bc
load:0x403c9700,len:0xbd8
load:0x403cc700,len:0x2a0c
entry 0x403c98d0
[  2594][E][Wire.cpp:499] requestFrom(): i2cWriteReadNonStop returned Error -1
[  2603][E][Wire.cpp:513] requestFrom(): i2cRead returned Error -1
[  2610][E][Wire.cpp:513] requestFrom(): i2cRead returned Error -1
[  2617][E][Wire.cpp:513] requestFrom(): i2cRead returned Error -1
[  2624][E][Wire.cpp:513] requestFrom(): i2cRead returned Error -1
[  2630][E][Wire.cpp:513] requestFrom(): i2cRead returned Error -1

Zanop:
Got the same, can only connect to repeater right after flashing, after that USB Configuration cannot connect and console gives this error.

LucLN:
same here .. V1.14.0 with Heltec V3 3.2 and compiled Heltec_v3_sensor .. any solution ?

LucLN:
**Update** :

my Heltec V3 board has 2 tiny pins on the back side without any labeling,
these are GPIO 17 & 18

connected a AHT10 sensor here
data is shown in Telemetry on _Heltec_v3_companion_radio_ble_
compiled _Heltec_v3_sensor_ and data is there too

the log shows that the sensor is found .. but the mentioned error is still there:

```
rst:0x1 (POWERON),boot:0x8 (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fce3808,len:0x4bc
load:0x403c9700,len:0xbd8
load:0x403cc700,len:0x2a0c
entry 0x403c98d0
Sensor ID: B847050C799235A8FB9D45410BB7C43AAEF9D00DD0C652B75F3D5C9379046675
DEBUG: No GPS detected
DEBUG: Found AHT10/AHT20 at address: 38
DEBUG: BME680 was not found at I2C address 76
[...]

PR #2452, opened 2026-04-30
Fixed sensor initialization and automatic selection of I2C addresses for BME/BMP sensors

IoTThinks:
Fixed sensor initialization and automatic selection of I2C addresses for BME/BMP sensors
- Added i2c_probe to probe sensors before initializing to avoid setup unnecessary sensors and to avoid bugs from unused sensors to impact the main code. 
Current issue https://github.com/meshcore-dev/MeshCore/issues/1086. 
One time, I saw Adafruit_VL53L0X crashed Heltec boards and this i2c_probe could mitigate / avoid the crash.
- Added automatically selection of I2C addresses 0x76 and 0x77 for BME280, BMP280 and BME680.
This is to avoid the headache to mess around with I2C addresses of BME280, BMP280 and BME680.
If the I2C address is defined, the automatic selection of I2C addresses is ignored.

Tested and works perfect for 3 releases of my MeshCore variant. 
image

### DIFF
--- a/src/helpers/SensorManager.h
+++ b/src/helpers/SensorManager.h
@@ -2,6 +2,7 @@
 
 #include <CayenneLPP.h>
 #include "sensors/LocationProvider.h"
+#include <Wire.h>
 
 #define TELEM_PERM_BASE         0x01   // 'base' permission includes battery
 #define TELEM_PERM_LOCATION     0x02
@@ -15,6 +16,7 @@ public:
   double node_altitude;       // altitude in meters
 
   SensorManager() { node_lat = 0; node_lon = 0; node_altitude = 0; }
+  virtual bool i2c_probe(TwoWire& wire, uint8_t addr) { return false; }
   virtual bool begin() { return false; }
   virtual bool querySensors(uint8_t requester_permissions, CayenneLPP& telemetry) { return false; }
   virtual void loop() { }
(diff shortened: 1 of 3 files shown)


IoTThinks:
Without this PR, we need to solder most of the BME/BMP sensors to fit the 0x76 address for MeshCore.
And not all boards have the same default I2C addresses.
image
image

446564 (approved):
🙌 tested on xiao nrf with bme/bmp280, no 680 on hand.

nice work

IoTThinks:
Great.
It has been working for a while.

446564:
@recrof

recrof:
there is new sensor detection PR that should be much more robust.. I'm bit conflicted which PR to merge first, because sensor init was not the best for long time. mentioned PR: https://github.com/meshcore-dev/MeshCore/pull/2327

446564:
I like them both actually.

With this one I like the change to have multiple addresses for a given sensor as there are many with such possibilities.

I like #2327 because it scans all addresses.

I _wish_ we moved away from build flags now that we are scanning, I know if saves a bit of space in fw but it would be nice for the end user to just add a sensor and have it work.

IoTThinks:
This is about reliability vs. major feature changes.
More code changes means less reliability.

This PR is a maintenance patch while PR https://github.com/meshcore-dev/MeshCore/pull/2327 is a new feature patch.
This PR has gone through extensive internal and community testing.

If you push PR https://github.com/meshcore-dev/MeshCore/pull/2327, please perform extensive testings too.
I'm fine with both ways.

mikenz (approved):

PR #2457, opened 2026-04-30
Add initial support for M5Stack Cardputer ADV with Cap LoRa-1262

kucharz96:
## Summary

This pull request adds initial support for **M5Stack Cardputer ADV** used with **M5Stack Cap LoRa-1262**.

The goal of this PR is to upstream a minimal, reviewable hardware bring-up path for this device combination, without changing MeshCore protocol behavior, routing, BLE logic, or UI behavior.

## What this PR changes

This PR introduces two focused hardware-related adjustments for the Cardputer ADV LoRa path:

1. **LoRa cap IO expander / RF path initialization**
   - updates the `0x43` IO expander setup so that only the required output is enabled
   - avoids driving the entire expander bank high during startup
   - leaves TX/RX RF switching to the SX1262 DIO2 RF switch control path

2. **SX1262 PA configuration after successful radio initialization**
   - applies PA configuration only after `std_init()` completes successfully
   - aligns the radio bring-up more closely with real hardware behavior observed during Cardputer ADV testing

## Motivation

At the time of opening this PR, the official MeshCore tree does not provide dedicated support for **M5Stack Cardputer ADV**.

During real-device bring-up and testing with **Cardputer ADV + Cap LoRa-1262**, the radio could initialize and operate, but RF behavior was not as consistent as expected compared with hardware-specific builds tested on the same platform.

The most likely cause appeared to be in the low-level hardware bring-up path for the LoRa cap, especially around expander-based initialization and RF control, rather than in MeshCore application logic.

## Scope

This PR is intentionally narrow in scope.

It does **not** include:
- protocol changes
- routing changes
- BLE changes
- UI changes
- experimental AGC tuning
- broader refactoring of other board variants

The intent is to keep the first upstream Cardputer ADV submission small and easy to review.

## Files changed

- `variants/m5stack_cardputer/M5CardputerBoard.h`
- `variants/m5stack_cardputer/target.cpp`

## Hardware used for testing

- **M5Stack Cardputer ADV**
- **M5Stack Cap LoRa-1262**
- SX1262-based LoRa path

## Notes for review

This should be considered **initial support** rather than a final, fully validated implementation.

Feedback would be especially appreciated on:
- whether the IO expander initialization matches expectations for Cap LoRa-1262 hardware
- whether the PA configuration is appropriate for upstream inclusion as-is
- whether Cardputer ADV support should continue under the current `m5stack_cardputer` variant path or be split into more explicit ADV-specific handling

## Closing

The main purpose of this PR is to start upstream support for **M5Stack Cardputer ADV** and open the discussion with maintainers using a small, hardware-focused patch set.

Thanks for reviewing.


kucharz96:
Closes #2458

MultiMote:
Hi. Your code uses wrong io expander init commands.

https://github.com/Stachugit/MeshCore-Cardputer-ADV/pull/17

Mischala:
Just popping in to add my little voice in support of adding Cardputer as a target.
My own hardware is currently in transit, but will build and test this branch asap.

variants/m5stack_cardputer/platformio.ini
```diff
@@ -0,0 +1,105 @@
...
+  -D ARDUINO_USB_CDC_ON_BOOT=1
+  -D ARDUINO_USB_MODE=1
+  -D PIN_USER_BTN=0
+  -D USE_SX1262
+  -D RADIO_CLASS=CustomSX1262
+  -D WRAPPER_CLASS=CardputerRfStabilityWrapper
+  -D HAS_GPS=1
+  -D LORA_TX_POWER=22
+  -D SX126X_DIO2_AS_RF_SWITCH=true
+  -D SX126X_CURRENT_LIMIT=240
```
MultiMote, line 20: RadioLib's setCurrentLimit comment:

```
Sets current protection limit. Can be set in 2.5 mA steps.

Parameters
currentLimit	current protection limit to be set in mA. Allowed values range from 0 to 140.
```

Why it's 240 here?


https://jgromes.github.io/RadioLib/class_s_x126x.html#a8f971dca834be7e0470a9a9f0c01854e

variants/m5stack_cardputer/platformio.ini
```diff
@@ -0,0 +1,105 @@
...
+  -D SX126X_RX_BOOSTED_GAIN=1
+  -D SX126X_DIO3_TCXO_VOLTAGE=3.3f
```
MultiMote, line 22: Official m5's example uses 3.0 TXCO voltate.

Is there an explanation for using 3.3V? 

https://docs.m5stack.com/en/arduino/projects/cap/cap_lora868

Mischala: interesting,
[the default for the 1262 in RadioLib appears to be 1.6V](https://jgromes.github.io/RadioLib/class_s_x1262.html)
according to [This Datasheet](https://cdn.sparkfun.com/assets/6/b/5/1/4/SX1262_datasheet.pdf) the typical value is 1.7V

so both may be a little on the high side?

Reading around, seems the higher voltage just speeds up start-up time of the TCXO, allowing it to warm the crystal faster, getting it to operating temp faster.

variants/m5stack_cardputer/platformio.ini
Mischala: Agreed,
According to [This Datashet](https://cdn.sparkfun.com/assets/6/b/5/1/4/SX1262_datasheet.pdf) 140mA is the default setting, when the device is configured.

variants/m5stack_cardputer/platformio.ini
MultiMote: @Mischala 

[Stamp LoRa-1262 Mini Schematics](https://m5stack-doc.oss-cn-shenzhen.aliyuncs.com/1208/NEW-1262-SCH_A1-Lora_2025_08_27_10_58_52.pdf) also points to 3.0 TXCO 🤷‍♂️

image

PR #2471, opened 2026-05-04
Enhance README with Software Heritage badge

Warmax356:
Added badge for Software Heritage and updated documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,12 @@ For developers;
 
 The Simple Secure Chat example can be interacted with through the Serial Monitor in Visual Studio Code, or with a Serial USB Terminal on Android.
 
+Meshcore is also monitored and calaogued by:
+[![SWH](https://archive.softwareheritage.org/badge/origin/https://github.com/meshcore-dev/meshcore/)](https://archive.softwareheritage.org/browse/origin/?origin_url=https://github.com/meshcore-dev/meshcore)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,6 +48,12 @@ For developers;
 
 The Simple Secure Chat example can be interacted with through the Serial Monitor in Visual Studio Code, or with a Serial USB Terminal on Android.
 
+Meshcore is also monitored and calaogued by:
+[![SWH](https://archive.softwareheritage.org/badge/origin/https://github.com/meshcore-dev/meshcore/)](https://archive.softwareheritage.org/browse/origin/?origin_url=https://github.com/meshcore-dev/meshcore)
+<a href="https://archive.softwareheritage.org/swh:1:dir:a14fa55ca3495c0a4ea8a52db26088b755199055;origin=https://github.com/meshcore-dev/meshcore;visit=swh:1:snp:1542b7c2b2a345f0e81811de644a0d051d60159a;anchor=swh:1:rev:1a7b3614a8439894714a58af55ab7f501e6bb928">
```
446564, line 53: I don't personally thing we need the revision badge as well as the repo.

446564 (commented):
Might also be better higher up, like in the bottom of the first section?

Curious what others think.

README.md
Warmax356: The Software Heritage (SWH) badge is a permanent "save point" for your project. You'll need it because it’s permanent and in the case of GitHub ever goes down or the repo is accidentally deleted, the project's code is safely stored in a universal, non-profit archive forever. (the badges indicate even if there are changes)

It’s citable: It provides a permanent, version-specific link (SWHID) that won't break, which is essential for professional or academic credit.

It shows quality: It signals to users that your project is built for the long haul and follows open-source best practices. Most of all think of it as an insurance policy and a permanent library card for the projects code.

Warmax356: p.s i am an SH ambassador that uses meshcore and catalogued it so that at least there is a redundant save point.

Warmax356: gentle reminder as this is still open

PR #2472, opened 2026-05-04
variants: USB Companion variant for LilyGo T-Echo Lite

emkZero:
This PR adds support for the LilyGo T-Echo Lite to be used as a USB companion.

The LilyGo T-Echo and T-Echo Lite are largely identical boards, but while the T-Echo can be used as a USB Companion, the Lite does not have this variant.

I tested the new variant on my device: LoRa, GPS, contacts, channels and DMs work as expected with the MeshCore web app.

Assuming we didn't miss anything, we (@0xC0FE and me) see no reason not to add a USB Companion variant.

Please let us know if there is anything wrong or missing - we are new to contributing to MeshCore.

### DIFF
--- a/variants/lilygo_techo_lite/platformio.ini
+++ b/variants/lilygo_techo_lite/platformio.ini
@@ -190,3 +190,24 @@ lib_deps =
 extends = LilyGo_T-Echo-Lite
 build_src_filter = ${LilyGo_T-Echo-Lite.build_src_filter}
   +<../examples/kiss_modem/>
+[env:LilyGo_T-Echo-Lite_companion_radio_usb]
+extends = LilyGo_T-Echo-Lite
+board_build.ldscript = boards/nrf52840_s140_v6_extrafs.ld
+board_upload.maximum_size = 712704
+build_flags =
+  ${LilyGo_T-Echo-Lite.build_flags}
+  -I src/helpers/ui
+  -I examples/companion_radio/ui-new
+  -D MAX_CONTACTS=350
+  -D MAX_GROUP_CHANNELS=40
+  -D OFFLINE_QUEUE_SIZE=256
+  -D UI_RECENT_LIST_SIZE=9
+  -D UI_SENSORS_PAGE=1
+  -D AUTO_SHUTDOWN_MILLIVOLTS=3300
+;  -D QSPIFLASH=1
+build_src_filter = ${LilyGo_T-Echo-Lite.build_src_filter}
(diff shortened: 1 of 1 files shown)


variants/lilygo_techo_lite/platformio.ini
```diff
@@ -96,3 +96,25 @@ build_src_filter = ${LilyGo_T-Echo-Lite.build_src_filter}
...
+  ${LilyGo_T-Echo-Lite.build_flags}
+  -I src/helpers/ui
+  -I examples/companion_radio/ui-new
+  -D MAX_CONTACTS=350
+  -D MAX_GROUP_CHANNELS=40
+  -D OFFLINE_QUEUE_SIZE=256
+  -D UI_RECENT_LIST_SIZE=9
+  -D UI_SENSORS_PAGE=1
+  -D AUTO_SHUTDOWN_MILLIVOLTS=3300
+;  -D QSPIFLASH=1
```
oltaco, line 119: Does the T-Echo Lite have an external flash chip? The schematic says it has a ZD25WQ32CEIGR. I have added support for this chip in CustomLFS 0.2.2 but we are currently pointing at 0.2.1 so the chip should still work just in compatibility mode at 1mb instead of the full 4mb.

oltaco:
Thanks for the PR, will merge this as is.

We will revisit enabling QSPIFLASH for the T-Echo Lite in future, because enabling it on devices now will cause contacts to be reset.

oltaco:
Sorry I just realised that this PR was based against main branch, can you resubmit against dev branch please? I tried to change it for you but it's created a big mess.

0xC0FE:
@oltaco We will take a look and let you know

.github/FUNDING.yml
```diff
@@ -0,0 +1 @@
+github: meshcore-dev
```
ripplebiz: Looks like you picked up diffs from main branch, and these have carried over in your PR (after rebasing to dev).
Can you remove these (.github, and /docs changes)

0xC0FE: We will take a look, sorry for any delays, holidays and busy days over here.

emkZero:
Since main and dev are too diverging, a rebase would indeed give an undesirable result. So I cherry-picked the commit onto a "new" branch freshly from dev and then force-pushed. I hope this will resolve the trouble and cleanly merge into dev as requested ^^

0xC0FE:
Hello, this has been open for a while now. Perhaps one would like to check and merge it. Thank you.

PR #2480, opened 2026-05-05
Incorrect time_t TypeDef used in RAK12035_SoilMoisture

KPrivitt:
The time_t typedef was incorrectly used for the variable timeout, it should be uint32_t


KPrivitt (commented):
Please start the approval process.

PR #2482, opened 2026-05-06
Implement WiFi auto-reconnect functionality

Rococo88:
Added auto-reconnect feature for WiFi and event handler for disconnection. Tested on Xiao-S3 board, it reconnects properly after router reboot.

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -195,6 +195,15 @@ void setup() {
 
 #ifdef WIFI_SSID
   board.setInhibitSleep(true);   // prevent sleep when WiFi is active
+  // Enable ESP-IDF auto-reconnect
+  WiFi.setAutoReconnect(true);
+  // Attach an event handler to force a reconnect specifically on disconnect events
+  WiFi.onEvent([](WiFiEvent_t event, WiFiEventInfo_t info){
+      #ifdef WIFI_DEBUG_LOGGING
+      Serial.println("WiFi disconnected. Forcing reconnect...");
+      #endif
+      WiFi.reconnect();
+  }, ARDUINO_EVENT_WIFI_STA_DISCONNECTED);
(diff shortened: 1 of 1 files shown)


examples/companion_radio/main.cpp
```diff
@@ -195,6 +195,15 @@ void setup() {
 
 #ifdef WIFI_SSID
   board.setInhibitSleep(true);   // prevent sleep when WiFi is active
+  // Enable ESP-IDF auto-reconnect
+  WiFi.setAutoReconnect(true);
```
liamcottle, line 199: This shouldn't be needed since auto reconnect defaults to true in the underlying library, however we might as well be explicit about it. So it can stay.

examples/companion_radio/main.cpp
```diff
@@ -195,6 +195,15 @@ void setup() {
...
+  // Attach an event handler to force a reconnect specifically on disconnect events
+  WiFi.onEvent([](WiFiEvent_t event, WiFiEventInfo_t info){
+      #ifdef WIFI_DEBUG_LOGGING
+      Serial.println("WiFi disconnected. Forcing reconnect...");
```
liamcottle, line 203: Please remove the `#ifdef` sections and replace `Serial.println(...)` with `WIFI_DEBUG_PRINTLN(...)`. The latter automatically handles the case where wifi debug logging is disabled.

liamcottle (changes requested):
Hey, thanks for the PR.

I've left a few review comments.

Please also change the PR branch to `dev`. You may need to rebase your fork to prevent the PR pulling in other commits if you change it from `main`.

Rococo88:
Hey Liam, thanks for the review! I've updated the print statements to use WIFI_DEBUG_PRINTLN. I actually couldn't change the base branch to dev in the UI (the option wasn't there for me), so I'm closing this and opening a fresh PR targeting dev right now!

PR #2488, opened 2026-05-06
Gate PIN_GPS_EN

weebl2000:
Do not reference it unconditionally

fixes bug introduced in pr #2401 

[Build firmware here](https://mcimages.weebl.me?commitId=fix-rak-pin-gps-en-undefined)


The-Bober:
In cases such as the default config for PIN_GPS_EN is -1 (such as the .ini file for the rak4631) wouldn't it still reference the pin as it is still defined?

Wouldn't a better way to ensure that the reference only works when a pin is defined be
```cpp
#if defined(PIN_GPS_EN) && (PIN_GPS_EN >= 0)
    gpsResetPin = PIN_GPS_EN;
#endif
```
?

PR #2489, opened 2026-05-06
buildsh refactor menu

mikecarper:
This replaces #2349 after the source branch was renamed from `pr/buildsh-refactor-menu` to `buildsh-refactor`.

This refactors `build.sh` to make the interactive target selection, env parsing, version selection, command dispatch, and artifact collection easier to follow and safer to maintain.

Running `bash build.sh` now opens a 2 tier menu and does decent guessing at the firmware version.

Also fixed various shellcheck issues.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -966,6 +966,47 @@ apply_radio_overrides() {
   fi
 }
 
+print_build_flags() {
+  local env_name=$1
+
+  echo "Build flags for ${env_name}:"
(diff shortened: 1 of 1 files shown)


mikecarper:
@ripplebiz 
image
image
Easy way to build firmware with common settings set via menu. Most things that will need to be updated in the bash script are at the top of the file

PR #2491, opened 2026-05-07
Implement WiFi auto-reconnect functionality

Rococo88:
Re-opening my previous PR against the dev branch. Also updated the print statements to use the WIFI_DEBUG_PRINTLN macro as requested

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -198,6 +198,7 @@ void setup() {
   WiFi.setAutoReconnect(true);
   WiFi.onEvent([](WiFiEvent_t event, WiFiEventInfo_t info){
       WIFI_DEBUG_PRINTLN("WiFi disconnected. Forcing reconnect...");
+      WiFi.reconnect();
   });
   WiFi.begin(WIFI_SSID, WIFI_PWD);
   serial_interface.begin(TCP_PORT);
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -195,6 +195,10 @@ void setup() {
 
 #ifdef WIFI_SSID
   board.setInhibitSleep(true);   // prevent sleep when WiFi is active
+  WiFi.setAutoReconnect(true);
+  WiFi.onEvent([](WiFiEvent_t event, WiFiEventInfo_t info){
+      WIFI_DEBUG_PRINTLN("WiFi disconnected. Forcing reconnect...");
+  });
   WiFi.begin(WIFI_SSID, WIFI_PWD);
   serial_interface.begin(TCP_PORT);
 #elif defined(BLE_PIN_CODE)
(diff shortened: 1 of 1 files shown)


examples/companion_radio/main.cpp
```diff
@@ -195,6 +195,10 @@ void setup() {
 
 #ifdef WIFI_SSID
   board.setInhibitSleep(true);   // prevent sleep when WiFi is active
+  WiFi.setAutoReconnect(true);
+  WiFi.onEvent([](WiFiEvent_t event, WiFiEventInfo_t info){
+      WIFI_DEBUG_PRINTLN("WiFi disconnected. Forcing reconnect...");
```
liamcottle: I think you missed the reconnect in the callback :)

liamcottle (changes requested):

examples/companion_radio/main.cpp
Rococo88: Oh, sorry, fixed :D

Rococo88: Yes, my bad.. i'll try better next time :)

Rococo88:
Hey @liamcottle, I tested adding WiFi.reconnect() inside the callback on my Xiao S3, but it actually causes an infinite crash loop at boot.. Because the callback was catching all WiFi events (or instantly failing if the router is down), calling reconnect() inside the event triggers another disconnect event, which spirals and crashes. I updated the PR to strictly filter for ARDUINO_EVENT_WIFI_STA_DISCONNECTED. Since WiFi.setAutoReconnect(true) natively handles the reconnecting safely in the background, we don't need the manual reconnect() call. The event listener is now just cleanly logging the disconnect state:

��ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0xc (RTC_SW_CPU_RST),boot:0x8 (SPI_FAST_FLASH_BOOT)
Saved PC:0x420b0dc2
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fce3808,len:0x41c
load:0x403c9700,len:0x9a8
load:0x403cc700,len:0x28ac
entry 0x403c98b8
WiFi: WiFi disconnected. OS handling auto-reconnect...
WiFi: Got connection
WiFi: Disconnected
WiFi: WiFi disconnected. OS handling auto-reconnect...
WiFi: Got connection
WiFi: Disconnected
WiFi: Got connection

liamcottle:
Hey, thanks for the info. I don't think this PR will make any difference then.

`WiFi.setAutoReconnect(true)` just sets the internal auto reconnect flag of the Wifi manager to true, but this already defaults to true even if you don't call `setAutoReconnect`, which is the current behaviour in firmware.

There was a previous PR from long ago that tried to merge in the addition of `WiFi.setAutoReconnect(true)`. But it wasn't merged for the same reason.

I assumed that because your earlier PR was explicitly calling `reconnect` that it was tested to resolve the problem.

Rococo88:
Understood, I'll keep it on my branch and see if it will solve my dropout issues as it take some time for APs to reboot at night, will think about other solutions, for now forcing WiFi.reconnect() inside the asynchronous event handler causes a crash loop.. Closing the PR.

PR #2494, opened 2026-05-07
Add missing Wire.h include to sensors

JosiahWI:
This slipped through in PR #2327 and I noticed because the TechoBoard.h for my variant doesn't include the Wire header, so the source file in question does not coincidentally obtain a copy.


liamcottle:
Awesome, thanks for the fix!

PR #2498, opened 2026-05-07
Add sanity check for DS3231 RTC probe

oltaco:
This PR adds a sanity check after probing for DS3231 RTC on the I2C bus.

Some unrelated devices (such as the IMU20948) respond on the same I2C address as DS3231. This can cause AutodiscoverRTCClock to incorrectly identify a non-RTC device as a DS3231, resulting in a non-functioning clock.

The sanity check just makes sure that the data that is coming back is at least plausible.


oltaco:
Closed in favour of #2516 as there was a chance this sanity check could fail when the RTC was uninitialised due to power loss.

Issue #2416, opened 2026-04-27
Feature Proposal: Signed time sync from explicitly trusted advert sources (aka "clock towers")


mangelajo:
This is a refinement on #1329 with a security/authenticity angle, and also relates to #1332, #1426, #2150 and #440. Filing it as a separate thread because the trust-model piece is what's new. Happy to fold it into one of the existing issues if maintainers prefer.

## Why

We run a Spanish mesh on EU868 with a mix of nRF52 / ESP32 nodes. Most of them have no battery-backed RTC, and only a few have GPS (and even those don't always feed the system clock; see #1426). The result is what you'd expect: repeaters drift, some come back from a reset stuck somewhere in 2024, and the spread gets worse over weeks. Multiple existing issues converge on the same root cause.

I added a small `!clocks` query to our companion bot that lists every repeater whose advert timestamp is more than 30 s off from our local clock. The list is uncomfortably long, which got me thinking about how a network like this could correct itself without centralised infrastructure or a dedicated time-sync packet type.

## The idea

Adverts are already signed with the source's ed25519 key, so authenticity comes for free: if a receiver verifies the signature against a key it recognises, the timestamp inside really came from that node. Authority (i.e. "do I believe this node has good time?") is the missing piece, and it should be a per-receiver decision, not a network-wide one.

Concretely:

1. Add a bit of metadata in the advert that says "I have an authoritative clock source" and *what kind*: RTC / GPS / DCF77-style / host-supplied via serial / none. Receivers use this both for filtering and for weighting.
2. Each receiver keeps a local **trust list** of pubkeys it accepts as time references. Trust is explicit: the operator adds the keys out of band. No TOFU, no learned trust.
3. When N adverts from trusted sources arrive within a short window, the receiver takes the **median** timestamp and steps the local clock. The median (rather than mean, as @tonurics noted in #1329) gives robustness against a single bad source: broken RTC, GPS spoofing, whatever.
4. Replay is handled by enforcing strict-monotonic timestamps per pubkey: a new advert from a trusted source must be newer than the last one accepted from that key. The 4-byte epoch field already in the advert is enough. A separate counter is only needed if sub-second guarantees matter.

The signature work is already there. The new pieces are: the source-type bits in the advert, the local trust list, and the median + monotonic check on the receive side.

## How it would look in practice

[...]

PR #2504, opened 2026-05-08
Clock towers + pio testing support

mangelajo:
Implement repeater clock trust update (see #2416 "Clock towers").

Adds an opt-in "trusted clock" list of advert pubkeys whose timestamps
may step the local RTC. Replay-guarded (advert_ts must be strictly newer
 than the last accepted from the same key.

Last few sync events are kept in a small RAM ring queryable via the CLI.

Logic lives in TrustedClockTable (no Arduino deps): table ops, replay
predicate, threshold predicate, sync-event ring, and pure
serialize/deserialize. MyMesh holds it as a member and wraps file I/O
around the helper.

CLI under clock.trust:
```
      clock.trust list
      clock.trust events
      clock.trust add <pubkey> [name]
      clock.trust remove <pubkey-or-name>

```
Also adds a new setting `clock.trust.thresh`.

To make reviewing easy I have structured the PR in two parts:

1) The functionality itself
2) Unit tests for the feature + make test + pio test support, which may help adding tests for other parts of the code as we go.

This is an example of the firmware running on a local T114 repeater I have at home:

```
10:54:32 - 15/5/2024 U RAW: 0A41ED97D250BFB5D6C748C653EA62BE1074E131DD97CA7E
10:54:32 - 15/5/2024 U: RX, len=24 (type=2, route=D, payload_len=20) SNR=13 RSSI=-53 score=1000 time=259 hash=C50C189C44122D59 [50 -> D2]
10:54:32 - 15/5/2024 U: TX, len=22 (type=2, route=D, payload_len=20) [50 -> D2]
DEBUG: RadioLibWrapper: noise_floor = -115
10:54:33 - 15/5/2024 U RAW: 0A4050D29F18FB6E5FCAA4B6333CF3E9ABE1982A4731FCA9AF308C06FD298B2EF8415B6077CB
10:54:33 - 15/5/2024 U: RX, len=38 (type=2, route=D, payload_len=36) SNR=10 RSSI=-94 score=1000 time=320 hash=76C4462508AC411C [D2 -> 50]
```

advert from a trusted clock key received (it's another repeater with GPS clock sync):
```
10:54:35 - 15/5/2024 U RAW: 1140D2590AED87376D577891BDC956026E38412BD4E8E7A972D21A2EF47478F26ECB569FFD69FA8D3475A0C2F8CFA2E9B456E1DC1E329F66F122DDF401E4692C259707750057E9EF5E972B99D770A4EF3078967DE781A43041475FAD3BA901B7520F765DDF0D9261946802AA1AC7FF5075657274612064656C20C3816E67656C20F09FA59D
10:54:35 - 15/5/2024 U: RX, len=133 (type=4, route=F, payload_len=131) SNR=11 RSSI=-93 score=1000 time=812 hash=BF74AF4E5683B06D
DEBUG: 10:54:35 - 15/5/2024 U Dispatcher::checkRecv(), score delay below threshold (0)
DEBUG: 10:54:35 - 15/5/2024 U Mesh::onRecvPacket(): valid advertisement received!
```

clock gets updated: 10:54:35 - 15/5/2024 --> 08:31:18 - 8/5/2026
```
DEBUG: trusted-clock: stepping local clock by +62458603 secs from kiwi [D2590AED87376D57]
DEBUG: RadioLibWrapper: noise_floor = -110
08:31:18 - 8/5/2026 U RAW: 1141CEBAD2590AED87376D577891BDC956026E38412BD4E8E7A972D21A2EF47478F26ECB569FFD69FA8D3475A0C2F8CFA2E9B456E1DC1E329F66F122DDF401E4692C259707750057E9EF5E972B99D770A4EF3078967DE781A43041475FAD3BA901B7520F765DDF0D9261946802AA1AC7FF5075657274612064656C20C3816E67656C20F09FA59D
08:31:18 - 8/5/2026 U: RX, len=135 (type=4, route=F, payload_len=131) SNR=-4 RSSI=-114 score=271 time=832 hash=BF74AF4E5683B06D
DEBUG: 08:31:18 - 8/5/2026 U Dispatcher::checkRecv(), score delay below threshold (0)
08:31:19 - 8/5/2026 U: TX, len=135 (type=4, route=F, payload_len=131)
DEBUG: RadioLibWrapper: noise_floor = -115
DEBUG: RadioLibWrapper: noise_floor = -110
DEBUG: RadioLibWrapper: noise_floor = -110
```

image

### DIFF
--- a/.github/workflows/host-tests.yml
+++ b/.github/workflows/host-tests.yml
@@ -1,0 +1,32 @@
+name: Host Tests
+
+on:
+  pull_request:
+    branches: [main, dev]
+    paths:
+      - 'src/**'
+      - 'examples/**'
+      - 'test/**'
+      - 'platformio.ini'
+      - '.github/workflows/host-tests.yml'
+  push:
+    branches: [main, dev]
+    paths:
+      - 'src/**'
+      - 'examples/**'
+      - 'test/**'
+      - 'platformio.ini'
+      - '.github/workflows/host-tests.yml'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone Repo
+        uses: actions/checkout@v4
+
+      - name: Setup Build Environment
+        uses: ./.github/actions/setup-build-environment
+
+      - name: Run host tests
+        run: pio test -e native
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,5 @@
+# Convenience targets. Firmware builds use PlatformIO directly: `pio run -e <env>`.
+
+.PHONY: test
+test:
+	pio test -e native
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -60,6 +60,129 @@
 
 #define LAZY_CONTACTS_WRITE_DELAY    5000
 
+// On-disk format owned by TrustedClockTable (serialize/deserialize). MyMesh just maps
+// the file open/read/write idioms required for each platform.
+void MyMesh::loadTrustedClocks() {
+  clock_trust.reset();
+  if (!_fs->exists(TRUSTED_CLOCKS_FILE)) return;
+#if defined(RP2040_PLATFORM)
+  File f = _fs->open(TRUSTED_CLOCKS_FILE, "r");
+#else
+  File f = _fs->open(TRUSTED_CLOCKS_FILE);
+#endif
+  if (!f) return;
+  uint8_t buf[TrustedClockTable::HEADER_SIZE
+              + (size_t)MAX_TRUSTED_CLOCK_KEYS * TrustedClockTable::RECORD_SIZE];
+  int n = f.read(buf, sizeof(buf));
+  f.close();
+  if (n > 0) clock_trust.deserialize(buf, (size_t)n);
+}
+
+void MyMesh::saveTrustedClocks() {
+#if defined(NRF52_PLATFORM) || defined(STM32_PLATFORM)
+  _fs->remove(TRUSTED_CLOCKS_FILE);
+  File f = _fs->open(TRUSTED_CLOCKS_FILE, FILE_O_WRITE);
+#elif defined(RP2040_PLATFORM)
+  File f = _fs->open(TRUSTED_CLOCKS_FILE, "w");
+#else
+  File f = _fs->open(TRUSTED_CLOCKS_FILE, "w", true);
+#endif
+  if (!f) return;
+  uint8_t buf[TrustedClockTable::HEADER_SIZE
+              + (size_t)MAX_TRUSTED_CLOCK_KEYS * TrustedClockTable::RECORD_SIZE];
+  size_t n = clock_trust.serialize(buf, sizeof(buf));
+  if (n > 0) f.write(buf, n);
+  f.close();
+}
+
+void MyMesh::formatSyncEventsReply(char* reply, size_t reply_size) {
+  if (reply_size == 0) return;
+  uint8_t n = clock_trust.syncEventCount();
+  if (n == 0) {
+    StrHelper::strncpy(reply, "(none)", reply_size);
+    return;
+  }
+  // Walk newest-first. Per line: <hex16> <name> <±delta>s.
+  reply[0] = 0;
+  size_t used = 0;
+  for (uint8_t k = 0; k < n; k++) {
+    const ClockSyncEvent& ev = clock_trust.syncEventNewestFirst(k);
+    char hex[17];
+    mesh::Utils::toHex(hex, ev.pub_key_prefix, sizeof(ev.pub_key_prefix));
+    hex[16] = 0;
+    int32_t delta = (int32_t)(ev.advert_timestamp - ev.local_before);
+    const char* nm = ev.name[0] ? ev.name : "(no name)";
+    int written = snprintf(reply + used, reply_size - used,
+                           "%s%s %s %+lds", used == 0 ? "" : "\n", hex, nm, (long)delta);
+    if (written < 0 || (size_t)written >= reply_size - used) {
+      reply[reply_size - 1] = 0;
+      break;
+    }
+    used += (size_t)written;
+  }
+}
+
+void MyMesh::formatTrustedClocksReply(char* reply, size_t reply_size) {
+  if (reply_size == 0) return;
+  if (clock_trust.count() == 0) {
+    StrHelper::strncpy(reply, "(none)", reply_size);
+    return;
+  }
+  // Compact listing: <16-hex-prefix> <name> per line. Truncates safely when out of room.
+  reply[0] = 0;
+  size_t used = 0;
+  for (uint8_t i = 0; i < clock_trust.count(); i++) {
+    const TrustedClockSource& s = clock_trust.at(i);
+    char hex[17];
+    mesh::Utils::toHex(hex, s.pub_key, 8);
+    hex[16] = 0;
+    const char* nm = s.name[0] ? s.name : "(no name)";
+    int written = snprintf(reply + used, reply_size - used,
+                           "%s%s %s", used == 0 ? "" : "\n", hex, nm);
+    if (written < 0 || (size_t)written >= reply_size - used) {
+      reply[reply_size - 1] = 0;
+      break;
+    }
+    used += (size_t)written;
+  }
+}
+
+void MyMesh::maybeStepClockFromTrustedAdvert(const mesh::Identity& id, uint32_t advert_timestamp) {
+  TrustedClockSource* src = clock_trust.find(id.pub_key);
+  if (src == NULL) return;
+
+  const char* src_name = src->name[0] ? src->name : "(no name)";
+  char src_hex[17];
+  mesh::Utils::toHex(src_hex, src->pub_key, 8);
+  src_hex[16] = 0;
+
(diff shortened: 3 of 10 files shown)


IoTThinks:
This looks interesting.
I will follow the PR.

Talie5in:
Following and will be testing.

Appreciate the opt-in model and having to define the trusted clock source.

vaizki:
My idea (not implemented yet) was to make a generic trust list that has a set of trust "scopes" for each trusted public key. Example based on your syntax:

```
> trust add kiwi 55483a5d904d3c0a80f3c50fdd4de2a121eca0122f3f6b4ea8b6847df7f66a81 clock
OK 
> trust list
kiwi 55483a5d904d3c0a80f3c50fdd4de2a121eca0122f3f6b4ea8b6847df7f66a81 clock

> trust add kiwi 55483a5d904d3c0a80f3c50fdd4de2a121eca0122f3f6b4ea8b6847df7f66a81 blacklist
OK 
> trust list
kiwi 55483a5d904d3c0a80f3c50fdd4de2a121eca0122f3f6b4ea8b6847df7f66a81 clock,blacklist

> trust add kiwi 55483a5d904d3c0a80f3c50fdd4de2a121eca0122f3f6b4ea8b6847df7f66a81 *
OK 
```

This would contribute a generic and extendable trust mechanism.

The clock-specific commands would be under a different base command:

```
> clock events
admin-reset
admin-sync +431s
peer-sync kiwi +34643634s
```
etc..

mangelajo:
> My idea (not implemented yet) was to make a generic trust list that has a set of trust "scopes" for each trusted public key. Example based on your syntax:
> 
> ```
> > trust add kiwi 55483a5d904d3c0a80f3c50fdd4de2a121eca0122f3f6b4ea8b6847df7f66a81 clock
> OK 
> > trust list
> kiwi 55483a5d904d3c0a80f3c50fdd4de2a121eca0122f3f6b4ea8b6847df7f66a81 clock
> 
> > trust add kiwi 55483a5d904d3c0a80f3c50fdd4de2a121eca0122f3f6b4ea8b6847df7f66a81 blacklist
> OK 
> > trust list
> kiwi 55483a5d904d3c0a80f3c50fdd4de2a121eca0122f3f6b4ea8b6847df7f66a81 clock,blacklist
> 
> > trust add kiwi 55483a5d904d3c0a80f3c50fdd4de2a121eca0122f3f6b4ea8b6847df7f66a81 *
> OK 
> ```
> 
> This would contribute a generic and extendable trust mechanism.
> 
> The clock-specific commands would be under a different base command:
> 
> ```
> > clock events
> admin-reset
> admin-sync +431s
> peer-sync kiwi +34643634s
> ```
> 
> etc..

I am good with that idea if everybody else agrees I can switch to something like that :), would be more reusable as you say :)

Talie5in:
Curious if this couldn't use the same ACL used for repeater admin, but just as a new ACL flag (4) for trust-clock-sync or something?

Just because the ACL system is already there.

Eg, see PR #2438

mangelajo:
> Curious if this couldn't use the same ACL used for repeater admin, but just as a new ACL flag (4) for trust-clock-sync or something?
> 
> Just because the ACL system is already there.
> 
> Eg, see PR #2438



> Curious if this couldn't use the same ACL used for repeater admin, but just as a new ACL flag (4) for trust-clock-sync or something?
> 
> Just because the ACL system is already there.
> 
> Eg, see PR #2438

Thanks for the pointer. I will investigate, because this ACL list is for Clients instead of Repeaters.

It makes sense TBH, but I wonder if the code is structured in a way that we can reuse it for "repeaters" ACL.

I'll report back once I have time to investigate :-)

hank:
> Curious if this couldn't use the same ACL used for repeater admin, but just as a new ACL flag (4) for trust-clock-sync or something?
> 
> Just because the ACL system is already there.
> 
> Eg, see PR #2438

i think this would be a very appropriate use case for ACL 4. In addition to region modification the clock trust is there. Maybe we should come up with a good name like "Trusted Third Party" that isn't as specific as "Region Manager"

ryanbcox:
I'm not sure of the particulars of ACL 4, but I personally would want this separate from repeater admin.  I may trust some random repeater with GPS enough to set the time on my own repeaters, but not enough to be an admin on my repeater.  I don't know if repeaters themselves could do something malicious but (very unlikely) a malicious repeater owner could gain people's trust then take their repeater's private key and use it on a companion device from which they could do something malicious to repeaters that trust that key.

I was actually about to submit a PR very similar to this and I'm glad to see someone else had already been working on it.  (I'll check Github better next time 🙂).  A few things I had done in addition to this were:
- Subtract .5 seconds from the advert time per hop to account for transmission time (seemed like a reasonable estimate)
- I added a max hop count parameter in case you want to ignore adverts beyond a certain distance.  Probably not the most useful parameter.
- Limited to 2 devices, partially because of the next item so it could all fit in one packet:
- Printed out the full key when listing so that a companion UI could more easily present accurate data (no collisions).  Not sure if that's a real concern or not.  Your naming idea might satisfy that.
- Never allowed the time to be set to something older than the date I created the code (maybe firmware build date would be better if that's easily available in a file somewhere)

I like your code better overall and mention these in case you find any of the ideas useful.  Personally I like accounting for time per hop but I'm guessing most people won't use time sources more than a few hops away anyway.

Littleaton:
This would be a great addition that could be connected up with the pymc linux nodes.   they could be the trusted clock and keep all other repeaters clocks on sync.   

@rightup @agessaman

Talie5in:
> This would be a great addition that could be connected up with the pymc linux nodes. they could be the trusted clock and keep all other repeaters clocks on sync.
> 
> @rightup @agessaman

If the repeater trusts it as a source.

We don't want implicit trust.

hank:
> > This would be a great addition that could be connected up with the pymc linux nodes. they could be the trusted clock and keep all other repeaters clocks on sync.
> > @rightup @agessaman
> 
> If the repeater trusts it as a source.
> 
> We don't want implicit trust.

yeah i think any clock source should be in the acl somewhere above guest. that way it's explicit trust only.

mangelajo:
> I'm not sure of the particulars of ACL 4, but I personally would want this separate from repeater admin. I may trust some random repeater with GPS enough to set the time on my own repeaters, but not enough to be an admin on my repeater. I don't know if repeaters themselves could do something malicious but (very unlikely) a malicious repeater owner could gain people's trust then take their repeater's private key and use it on a companion device from which they could do something malicious to repeaters that trust that key.

Also ACLs are designed to be "trust Client" from this Repeater, not for trusting other repeaters.

Although what @Talie5in proposed was using a specific flag to only trust the clock, reusing the same list and extending it for this use case.

> 
> I was actually about to submit a PR very similar to this and I'm glad to see someone else had already been working on it. (I'll check Github better next time 🙂). A few things I had done in addition to this were:
> 
> * Subtract .5 seconds from the advert time per hop to account for transmission time (seemed like a reasonable estimate)

Yeah, that sounds like a reasonable default in line with what I have experienced, we could perhaps make it a configurable parameter.

> * I added a max hop count parameter in case you want to ignore adverts beyond a certain distance.  Probably not the most useful parameter.
+1 Makes sense.

> * Limited to 2 devices, partially because of the next item so it could all fit in one packet:
> * Printed out the full key when listing so that a companion UI could more easily present accurate data (no collisions).  Not sure if that's a real concern or not.  Your naming idea might satisfy that.
> * Never allowed the time to be set to something older than the date I created the code (maybe firmware build date would be better if that's easily available in a file somewhere)
+1
> 
> I like your code better overall and mention these in case you find any of the ideas useful. Personally I like accounting for time per hop but I'm guessing most people won't use time sources more than a few hops away anyway.

I will add this to the patch and test ASAP.

What I have not managed (I believe), is to get any feedback from core developers of meshcore. I'd like to waste a lot of time if this is going into an unwanted direction for them, do you know how to get them have an eye?

ryanbcox:
> What I have not managed (I believe), is to get any feedback from core developers of meshcore. I'd like to waste a lot of time if this is going into an unwanted direction for them, do you know how to get them have an eye?

I'm new to MeshCore so I don't know much about the actual project structure yet or how many developers are involved.

mangelajo:
@liamcottle I'd be grateful if you could have an eye to this PR, and offer some guidance. We are wondering if we should have a more generic trust mechanism, or if the ACL subsystem could be extended for this purpose. I am happy to finish the patch and implement the received feedback, but I'd love to know what is the preference of the project admins.

Best,
Miguel Ángel.

ViezeVingertjes:
Interesting; I’ve implemented a clock sync before that worked by filtering out outliers from incoming packets to get an accurate time. That works until there are malicious actors, of course, but it was fine for some rogue nodes.

I don’t think ACL is the right place for this though, since it’s about companion permissions, and personally I find @vaizki 's idea cleaner in the long run. That said, this approach is also fine for what it’s supposed to do.

On the other hand… it does add quite a bit of code for something (accurate time) that repeaters don’t actually need to function in the first place. 

What’s the real use case here? and does that justify the extra code/complexity?

Avamander:
> What’s the real use case here? and does that justify the extra code/complexity?

If someone operates more than one node, especially if they lack an RTC or GNSS module keeping time in sync becomes a bit annoying at times. This applies to both companions and repeaters. Being able to trust certain operators and their timekeeping would make the management of low-cost nodes easier.

mangelajo:
Ok, it seems like I may need to rebase the patch at this point…

PR #2510, opened 2026-05-08
Implementing on-device ping

hank:
The firmware UI had a feature to send an advert, which is useful, but often when out and about I'd like to ping a repeater and get a signal report, especially when scouting areas. I implemented this by making the last packet RSSI accessible, adding ping functionality to MyMesh and implementing some UI changes to allow selecting rows in the recent adverts screen with a long press, then short presses to move between rows, and a long press to send a ping. I also made the selection mode time out after 10 seconds.

The ping result shows signal in both directions: the rx line is what we hear from the repeater (from the response packet), and the tx line is what the repeater heard from us (parsed out of last_rssi/last_snr in the RepeaterStats response payload, matching what the Android companion app already displays). Extended showAlert to render multi-line popups so both lines fit, and bumped the success alert duration to 5 seconds so there's time to read both numbers.

[Video Example of Use](https://www.youtube.com/shorts/cudaAqsmMWw)

I thought of this because my phone died while hiking and I wasn't able to ping the repeater anymore without it on my companion to see if I could hit it. Now I can!

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -399,6 +399,30 @@ int MyMesh::getRecentlyHeard(AdvertPath dest[], int max_num) {
   return max_num;
 }
 
+int MyMesh::startPing(const ContactInfo& recipient, uint32_t& est_timeout) {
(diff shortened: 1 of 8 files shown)


446564:
I'll test as soon as able, maybe not until next week sadly

PR #2511, opened 2026-05-09
T-Echo Lite SX1262 parameter corrections

entr0p1:
- Set TCXO voltage to 1.8 (not defined previously, falls back to default)
- Set correct variable name for SX1262 RX/TX pins
- Fix SX1262 RX pin number


oltaco:
Looks good, pins match the schematic and the Acsip S62F application note 👍

PR #2512, opened 2026-05-09
Improve battery monitoring algorithm for promicro

todd-herbert:
A few tweaks to get battery monitoring working for the promicro variant:
- modify equation in `Promicroboard::getBattMilliVolts`
- correct `AREF_VOLTAGE`, lower `ADC_MULTIPLIER` accordingly
- raise `ADC_RESOLUTION`

### DIFF
--- a/variants/promicro/PromicroBoard.h
+++ b/variants/promicro/PromicroBoard.h
@@ -23,7 +23,7 @@
 // - use cli to 'set adc.multiplier' (repeaters / rooms)
 
 #ifndef ADC_MULTIPLIER
-#define ADC_MULTIPLIER (1.667f) // Dependent on voltage divider resistors. Default 1MOhm & 1.5MOhm divider
+#define ADC_MULTIPLIER (2.0f) // Dependent on voltage divider resistors. Default 1MOhm resistors
 #endif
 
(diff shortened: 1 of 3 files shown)


mattzzw:
Regading the default adc multiplier value:
Which promicro variant are you referring to?
The FakeTec boards for example use resistors of equal values for the adc voltage devider.

todd-herbert:
> Regading the default adc multiplier value: Which promicro variant are you referring to? The FakeTec boards for example use resistors of equal values for the adc voltage devider.

Oh yes, sorry, well spotted!
I'd previously been using faketec boards with Meshtastic, which for some reason [expects a 1MOhm-1.5MOhm divider](https://github.com/meshtastic/firmware/blob/48274981883b909190963d6dfb459ee4a9a2e39e/variants/nrf52840/diy/nrf52_promicro_diy_tcxo/variant.h#L64). 

It totally slipped my mind that this was abnormal. I'll update to reflect the official faketec setup with the two equal divider resistors.

todd-herbert:
Title updated to reflect the fact that the current implementation does apparently work (by fudging the multiplier value). The changes are still valuable as they tie the ADC_MULTIPLIER value to the ratio of the voltage divider, allowing users to calculate the value, rather than guess and check.

PR #2520, opened 2026-05-11
add heltec-mesh-node-t1

Quency-D:
This PR adds initial Meshcode firmware support for the Heltec Mesh Node T1.

The new board target is based on the nRF52840 MCU and SX1262 LoRa radio.

### DIFF
--- a/boards/heltec_t1.json
+++ b/boards/heltec_t1.json
@@ -1,0 +1,61 @@
+{
+  "build": {
+    "arduino": {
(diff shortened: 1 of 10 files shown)


liamcottle:
Thank you!

grantsavill:
Just received one of these. Is it possible to turn off the buzzer?

jirogit:
> Just received one of these. Is it possible to turn off the buzzer?

Yes — triple-press the user button and it'll disable the buzzer (it persists across reboots).

PR #2517, opened 2026-05-10
Add support for LilyGo T-Echo Card (and tiny-ui!)

oltaco:
This PR adds support for the LilyGo T-Echo Card. It depends on #2516 otherwise the clock doesn't work due to the accelerometer/magnetometer being auto-detected as an RTC.

Added a new tiny-ui for the 72*40 screen, big thanks to @pelgraine for the U8g2 display wrapper 👍 

I decided to try something different with the tiny ui and added a scrolling header with node name, battery voltage, buzzer, GPS and BLE status.
The main screen shows unread msg count, battery voltage and BLE pin when not connected, and there is no recent messages screen as there's not enough space anyway. It might be worth removing the recent adverts screen as well but I've left it in for now.

Triple click to toggle buzzer, double click on C button to toggle "torch". 

Accelerometer/compass is currently disabled but could be enabled by a future update. The speaker could also potentially be enabled for use with alternate notification sounds, who knows :)

PR #2522, opened 2026-05-11
Add support for LilyGo T-Impulse-Plus

liamcottle:
This PR adds initial support for the [LilyGo T-Impulse-Plus](https://github.com/Xinyuan-LilyGO/T-Impulse-Plus) device.

- Schematic: https://github.com/Xinyuan-LilyGO/T-Impulse-Plus/blob/main/project/T-Impulse-Plus_V1.0.pdf
- Pin Out: https://github.com/Xinyuan-LilyGO/T-Impulse-Plus/blob/main/libraries/private_library/pin_config.h

It's very small and could be useful as a pendant/necklace tracker node, or an ultra light portable companion node.

- It has a GPS module
- It has a 200mAh battery
- It has a vibrator module (not implemented at this time)
- The TX/RX signal strength is decent with the built in spring antenna
- It has capacitive touch support on the top face, however currently disabled while screen not implemented
- It has 4MB of external QSPI flash, much improved over other devices with 1MB internal or 2MB external

I've not implemented display support, the screen is very small, and the default SSD1306 library would not init.
This can be done sometime in the future, as a UI redesign would be required anyway. Possibly based on @oltaco's PR https://github.com/meshcore-dev/MeshCore/pull/2517 for the LilyGo T-Echo Card. 

I've tested Companion BLE firmware with the MeshCore App as well as the Repeater firmware.

### DIFF
--- a/boards/lilygo_t_impulse_plus_nrf52840.json
+++ b/boards/lilygo_t_impulse_plus_nrf52840.json
@@ -1,0 +1,64 @@
+{
+  "build": {
+    "arduino":{
+      "ldscript": "nrf52840_s140_v6.ld"
+    },
+    "core": "nRF5",
+    "cpu": "cortex-m4",
+    "extra_flags": "-DARDUINO_NRF52840_T_IMPULSE_PLUS -DNRF52840_XXAA",
+    "f_cpu": "64000000L",
+    "hwids": [
+      [
+        "0x239A",
+        "0x8029"
+      ]
+    ],
+    "usb_product": "T-Impulse-Plus-nRF52840",
+    "mcu": "nrf52840",
+    "variant": "t_impulse_plus_nrf52840",
+    "bsp": {
+      "name": "adafruit"
+    },
+    "softdevice": {
+      "sd_flags": "-DS140",
+      "sd_name": "s140",
+      "sd_version": "6.1.1",
+      "sd_fwid": "0x00B6"
+    },
+    "bootloader": {
+      "settings_addr": "0xFF000"
+    }
+  },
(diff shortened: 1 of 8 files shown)


oltaco (approved):

recrof (approved):

Szetya:
IMG_20260710_051113
The display, touch buttons, and vibration motor are working. 
The UI still needs a little tweaking to make it compatible with both the t-echo card and the t-impulse plus. 
In addition, we need smaller icons for the 64-pixel-high display. @liamcottle , could you make them?

liamcottle:
@Szetya

> The UI still needs a little tweaking to make it compatible with both the t-echo card and the t-impulse plus. In addition, we need smaller icons for the 64-pixel-high display. @liamcottle , could you make them?

Hey there! My T-Impulse-Plus node actually died, so I can't do any more work/testing on this one :/

It's a cute little unit, but seems very fragile.

Szetya:
> @Szetya
> 
> > The UI still needs a little tweaking to make it compatible with both the t-echo card and the t-impulse plus. In addition, we need smaller icons for the 64-pixel-high display. @liamcottle , could you make them?
> 
> Hey there! My T-Impulse-Plus node actually died, so I can't do any more work/testing on this one :/
> 
> It's a cute little unit, but seems very fragile.

@liamcottle 
The battery is discharged below 3V and won't charge?!
It contains an SGM41562s with an ID of 0x09. The original sample code expects 0x04, which is why PMU detection isn't working.
I struggled with it for a long time, but I finally managed to fix it.

PR #2524, opened 2026-05-11
enable gps for all rak 3401

446564:


446564:
this enables for all variant roles as repeater was missed, moves to the top level in ini, if preferred I can move it to each role instead, or omit say KISS modem?

446564:
@recrof

446564:
closing until gps is sorted for this one

PR #2534, opened 2026-05-12
Update CLI command parameters for bridge source

khagele:


StijnBTC:
The change is correct,  the set command already accepts `rx`/`tx` (not `logRx`/`logTx`).

Adding to this, the get/query command still returns the old names:

**`src/helpers/CommonCLI.cpp:832`**
```cpp
sprintf(reply, "> %s", _prefs->bridge_pkt_src ? "logRx" : "logTx");
```

This should be changed to:
```cpp
sprintf(reply, "> %s", _prefs->bridge_pkt_src ? "rx" : "tx");
```

Without this fix, a user running `get bridge.source` will see `logTx` while the docs say `tx`.

Optionally, these comments could also be updated for consistency:
- `src/helpers/CommonCLI.h:48` — `// 0 = logTx, 1 = logRx (default logTx)`
- `examples/simple_repeater/MyMesh.cpp:894` — `// logTx`

But logRx and logTx are used in lots of other parts of the code so maybe it would make more sense to change it to accept both tx/rx and logTx/logRx

PR #2536, opened 2026-05-12
revert: "feat: Enable GPS on RAK 1W kit" (#2401)

recrof:
reverted changes to RAK_BOARD and PIN_GPS_EN. setting `RAK_BOARD` would cause radio to stop working and end with RadioLib error -707

### DIFF
--- a/variants/rak3401/platformio.ini
+++ b/variants/rak3401/platformio.ini
@@ -6,7 +6,6 @@ build_flags = ${nrf52_base.build_flags}
   ${sensor_base.build_flags}
   -I variants/rak3401
   -D RAK_3401
-  -D RAK_BOARD
   -D NRF52_POWER_MANAGEMENT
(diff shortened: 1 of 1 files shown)


oltaco:
Yep, seems like the issue might be when rakGPSinit() is searching for a GPS module it calls gpsIsAwake() against WB_IO2 which is the 5v boost pin on the RAK3401?

Maybe some gating here would fix it? https://github.com/meshcore-dev/MeshCore/blob/12a37a224ad229e5d6507636c2cea5d64933e09a/src/helpers/sensors/EnvironmentSensorManager.cpp#L685

jakymiwm:
Can we revisit this now that #2615 has been merged?

disq:
@jakymiwm I've re-opened the (slightly different) fix at https://github.com/meshcore-dev/MeshCore/pull/2640

PR #2537, opened 2026-05-12
Fix serial bursts on nRF52 with USB CDC

weebl2000:
I was having issues where the usb serial bus would stop responding when running companion USB for a Heltec T096.

[Build this firmware here](https://mcimages.weebl.me?commitId=fix-serial-bursts)


rogovogor (approved):
Approve

PR #2540, opened 2026-05-12
Add bulk region hierarchy command to CLI

agessaman:
Add a new command `region bulk` for defining region hierarchies in a single line. This command allows users to create multiple regions in a single message. Updated the documentation to include usage examples and detailed parameter descriptions.

The primary use-case here is improving the regions configuration process by requiring fewer command messages to load a full set of regions on a repeater. This has the additional benefit that regions can be configured or reconfigured at multiple hops with more reliability—with `region load` or `region put` a dropped message can frustrate or break the configuration process.

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -898,15 +898,15 @@ void CommonCLI::handleGetCmd(uint32_t sender_timestamp, char* command, char* rep
 void CommonCLI::handleRegionCmd(char* command, char* reply) {
   reply[0] = 0;
 
-  // `region bulk ...` — cursor-walk over space-separated tokens (must run before
+  // `region def ...` — cursor-walk over space-separated tokens (must run before
   // parseTextParts, which only keeps 4 segments and mutates the buffer).
   char* cmd = command;
   while (*cmd == ' ') cmd++;
-  if (strncmp(cmd, "region bulk", 11) == 0 && (cmd[11] == ' ' || cmd[11] == '\0')) {
-    char* payload = cmd + 11;
+  if (strncmp(cmd, "region def", 10) == 0 && (cmd[10] == ' ' || cmd[10] == '\0')) {
+    char* payload = cmd + 10;
     while (*payload == ' ') payload++;
     if (*payload == '\0') {
-      snprintf(reply, 160, "Err - empty bulk");
+      snprintf(reply, 160, "Err - empty def");
       return;
     }
     RegionEntry* cursor = &_region_map->getWildcard();
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -898,6 +898,70 @@ void CommonCLI::handleGetCmd(uint32_t sender_timestamp, char* command, char* rep
 void CommonCLI::handleRegionCmd(char* command, char* reply) {
   reply[0] = 0;
 
+  // `region bulk ...` — cursor-walk over space-separated tokens (must run before
+  // parseTextParts, which only keeps 4 segments and mutates the buffer).
+  char* cmd = command;
+  while (*cmd == ' ') cmd++;
+  if (strncmp(cmd, "region bulk", 11) == 0 && (cmd[11] == ' ' || cmd[11] == '\0')) {
+    char* payload = cmd + 11;
+    while (*payload == ' ') payload++;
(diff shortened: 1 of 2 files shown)


AI7NC:
This idea has my full support. This will allow much simpler regional loads for repeater operators by condensing in my region 8 load commands into a single command.

towerviewcams:
8 load commands with many repeaters that I host will create allot of traffic and retry attempts. Anything that can be done to reduce and make this process more efficient should be considered.

ripplebiz:
I quite like this. A nifty shorthand.
I think 'region def ...' would prob read better than 'bulk'. Plus you get an extra byte! :-)

agessaman:
Updated to use `region def ...`. Thanks for the feedback. :)

src/helpers/CommonCLI.cpp
```diff
@@ -898,6 +898,70 @@ void CommonCLI::handleGetCmd(uint32_t sender_timestamp, char* command, char* rep
 void CommonCLI::handleRegionCmd(char* command, char* reply) {
   reply[0] = 0;
 
+  // `region def ...` — cursor-walk over space-separated tokens (must run before
```
thecharge: you can be a bit more efficient with the delimiters and spaces, isn't that the same as
```cpp
// `region def ...` — cursor-walk over space-separated tokens (must run before
  // parseTextParts, which only keeps 4 segments and mutates the buffer).
  char* cmd = command;

  while (*cmd == ' ') {
    cmd++;
  }

  if (strncmp(cmd, "region def", 10) == 0 && (cmd[10] == ' ' || cmd[10] == '\0')) {
    char* payload = cmd + 10;

    while (*payload == ' ') {
      payload++;
    }

    if (*payload == '\0') {
      snprintf(reply, 160, "Err - empty def");
      return;
    }

    RegionEntry* cursor = &_region_map->getWildcard();

    char* p = payload;

    // Trim trailing whitespace in-place; lets us use end-pointer arithmetic once.
    char* pend = p + strlen(p);
    while (pend > p && pend[-1] == ' ') *--pend = '\0';

    // Helper: advance past leading spaces (no allocation, pure pointer arithmetic).
    auto ltrim = [](char* s) -> char* {
      while (*s == ' ') s++;
      return s;
    };

    while (*p) {
      // Skip leading spaces and detect end.
      p = ltrim(p);
      if (*p == '\0') break;

      // Delimit the next space-separated token in-place.
      char* tok = p;
      while (*p && *p != ' ') p++;
      if (*p) *p++ = '\0';

      // Split token on '|' or ',' into name[|jump] — single forward scan.
      char* jump = nullptr;
      for (char* q = tok; *q; q++) {
        if (*q == '|' || *q == ',') {
          *q = '\0';
          jump = ltrim(q + 1);
          // Trim trailing spaces from jump.
          char* je = jump + strlen(jump);
          while (je > jump && je[-1] == ' ') *--je = '\0';
          break;
        }
      }

      char* name = ltrim(tok);

      if (*name == '\0') {
        snprintf(reply, 160, "Err - empty name");
        return;
      }

      if (jump && *jump == '\0') {
        snprintf(reply, 160, "Err - empty jump");
        return;
      }

      auto r = _region_map->putRegion(name, cursor->id);
      if (r == NULL) {
        snprintf(reply, 160, "Err - put failed: %s", name);
        return;
      }

      r->flags = 0;
      cursor = r;

      if (jump) {
        auto j = _region_map->findByNamePrefix(jump);
        if (j == NULL) {
          snprintf(reply, 160, "Err - unknown jump: %s", jump);
          return;
        }
        cursor = j;
      }
    }

    _region_map->exportTo(reply, 160);
    return;
  }
```

agessaman: I like this one too—all paths to the same goal. If @ripplebiz likes the style better, I'm happy to use it instead. I just would like to be able to set up regions with a single command string. :)

n2aws:
I like this approach.  As someone trying to coordinate region implementation locally,  the biggest pushback I'm hearing is that "it's a lot", and they have concerns about doing it correctly, or maintaining it long term.  Everything we can do ti simplify, would be good.

agessaman:
Okay, I gave this a rewrite to use helper functions instead of one big inline process and it significantly improved the readability and reduced the footprint in `CommonCLI::handleRegionCmd()`.

PR #2546, opened 2026-05-13
Fix output power limit on heltec v4 board

TobleMiner:
Fix output power limit on heltec v4 board

The HelTec WiFi LoRa 32 V4 board features a SX1262 connected to a builtin
radio frontend module (FEM) containing both a PA for the TX path and a LNA
for the RX path.
Previously the maxium allowed SX1262 TX power was set to 22 dBm. This is
dangerous since it exceeds the maximum PA input power limit for all
revisions of the board.
V4.2 of the board uses the GC1109 FEM with an absolute maximum PA input
power specfied as +5 dBm. [1]
V4.3 of the board uses the KCT8103L FEM where absolute maximum PA input
power is unclear due to a lack of publicly available datasheeets.
Consequently configuring a TX power of more than +5 dBm on the SX1262 is
dangerous as it can damage the FEM. Furthermore it does not - while not
directly dangerous - make any sense to configure a TX power of more than
+0 dBm on the SX2162 since the GC1109 used in board revision 4.2
specifies a maximum PA output power of +30 dBm and also a typical PA gain
of 30 dB. Similarly the KCT8103L used in board revision 4.3 specifies a
maximum ouput power of +28 dBm [2] and a PA gain of 32 dB [3]. Thus using
an input power of more than +0 dBm and -4 dBm respectively from the SX1262
would overdrive the PA causing distortions and thus degrading TX
performance.

Limit maximum SX1262 TX power to +5 dBm so the FEM will not be damaged and
set default TX power to +0 dBm so the FEM PA is not overdriven.

[1] https://www.geochipinc.com/uploads/23940615_1756826539.pdf

[2] https://kxcomtech.com/en/application/info/1774

[3] https://www.ftelectronic.com/Uploads/Temp/news/20251209/69379d0d25b27.pdf


TobleMiner:
Found more info on the KCT8103L in the meantime and updated the PR accordingly. Based on the claimed gain numbers the situation is even worse in terms of saturation and overdriving.

TobleMiner:
For purpose of validating above findings I obtained a HelTec V4.2 and V4.3 board.

I then characterized the SX1262 TX power vs actual board output power: https://docs.google.com/spreadsheets/d/1lAWqJx1iB4RVzJFnQacfRe2rOg2-eYbKO4bxT7sfeNI/edit?usp=sharing

Surprisingly it shows that the overall TX path gain is a lot lower than expected at just 11 dB (V4.2) and 12 dB (V4.3) respectively. It seems either the FEMs deviate significantly from their specification in terms of RF gain or a lot of power is lost somewhere between the SX1262 and the FEM. In the latter case there is probably no actual risk for the FEMs here.

One aspect the characterization of the TX path shows though is that using SX1262 TX powers beyond ~15 dBm seems to start overdriving the PA in the FEMs. So it would be advisable to stay below that TX power to ensure TX signal quality does not deteriorate.

PR #2554, opened 2026-05-14
[Fix] epaper set color for Heltec E213, E290 and wireless paper

doormatt-dev:
I noticed a massive black bar briefly appearing on the recent page when flipping through the pages on my E290 and when sending an advert. After digging through the code I realised that this is supposed to be the popup notification. It seemed like
the rectangle drawing function did not support changing the drawing. When I tried fixing that I realised that the setColor function was pretty much useless, so I essentially copied the implementation that the monochrome OLEDs and LCDs use. Later I also realised that bitmaps could also only be drawn in a single color and added color change support there too.

The popup on the recent page is a weird choice imo. It only appears when flipping through the pages forwards but not backwards. Some text saying "no recent contracts" when it's empty would probably be best. I might make a future PR for that and
better space utilisation of the big e-paper screen. It just seems like a waste of screen space to only display tiny text and only 4 recents.

The changes in this PR have been tested on a Heltec E290, the E213 and wireless paper should work the same way since they are very similar. The changes to bitmap drawing do not affect anything visually but allow for drawing white bitmaps on black background. The changes to the rectangle drawing functions allow the notification popup to be displayed correctly rather than being a solid black box.

### DIFF
--- a/src/helpers/ui/E213Display.cpp
+++ b/src/helpers/ui/E213Display.cpp
@@ -128,7 +128,9 @@ void E213Display::setTextSize(int sz) {
 
 void E213Display::setColor(Color c) {
   display_crc.update<Color>(c);
-  // implemented in individual display methods
+  _color = c == 0 ? WHITE : BLACK; // for rectangles
+  display.setTextColor(UINT16_MAX * _color); // for text
+  // keep in mind this is inverted on e-ink
 }
 
 void E213Display::setCursor(int x, int y) {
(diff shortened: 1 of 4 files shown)


doormatt-dev:
# Issue: 
The drawing functions in /src/helpers/ui/E213Display.cpp and /src/helpers/ui/E290Display.cpp did not implement any sort of colour changing on rectangles, text and bitmaps. This results in the following black rectangle appearing:

https://github.com/user-attachments/assets/46b02eaf-1d76-4f61-9735-265558ced54e

# How to replicate:
Simply install the official meshcore firmware on any of the Heltec e-paper boards. The black bar will appear on the recent contacts page when cycling through the pages in forward order (not in reverse order but that's a different issue, it shouldn't try to do a popup at all there imo) and when sending an advert via the companion itself (not when triggering an advert through the app).


# Fix:
I have already fixed this in this PR, it is fully functional. Changing text and bitmap colour is also supported now although it is not used so far.

https://github.com/user-attachments/assets/7a9b50cc-391f-47e6-9ae5-c6f6af43b516

PR #2565, opened 2026-05-15
Disable BLE_PIN_CODE for E290 companion USB environment

willumpie82:
## Description
Remove the hardcoded `BLE_PIN_CODE` build flag from the Heltec_E290_companion_usb environment. The presence of `BLE_PIN_CODE` prevents the unit from operating as a USB companion. Commenting it out enables USB companion functionality.

## Changes
- Comment out `-D BLE_PIN_CODE=123456` in `variants/heltec_e290/platformio.ini`

## Impact
This allows the E290 unit to function properly as a USB companion device by removing the hardcoded PIN restriction.


liamcottle:
Thanks! I'm not sure if the hash acts as a comment in the .ini file, and if it is required to be the semi colon. Would be best to just remove the line entirely from the USB variant :)

willumpie82:
you are right, that line does not belong there, so completely removed now 

side note: I flashed my HT E290 earlier today with the line removed and i'm now able to connect over USB (using as Homeassistant proxy)

liamcottle:
Awesome, thanks! Will merge in :)

Issue #1775, opened 2026-02-21
Mesh Reliability Risks Under Scale and Mobility


robekl:
How important are these issues in the current implementation?

1. **1-byte path-hash collision risk in direct routing (semantic addressing ambiguity)**  
Direct routing currently identifies the “next hop” using only a 1-byte hash prefix of a node public key. In practice, that means only 256 possible hop identifiers exist, so collisions are guaranteed as network size and diversity grow. The forwarding logic for direct packets accepts a packet when that single byte matches the local node’s prefix, not when a stronger identity match is proven.  
Why this matters materially: if two repeaters share the same prefix byte, either repeater can consider itself the intended next hop and consume/forward traffic that was meant for the other. This can create route drift, blackholing, mis-forwarding, and increased retransmissions. Under light traffic and small meshes it may appear rare, but in larger or multi-region deployments it becomes increasingly likely and can directly reduce delivery reliability for direct messages.

2. **“First packet wins” route learning with unconditional path replacement (path quality instability)**  
Route learning behavior currently favors whichever valid path-related message arrives first, and then updates stored outbound path state by replacing previous path information without quality gating (no comparison of hop count, recency confidence, delivery success, link quality trend, or stability score). In effect, a single newly received path can overwrite a previously working route immediately.  
Why this matters materially: in RF environments with multipath, congestion, asymmetric links, or mobile repeaters, the first-arriving path is often not the best long-term route. Unconditional replacement can degrade a stable direct route into a brittle one, causing repeated direct failures, timeout-driven retries, and fallback flooding. The user experience is intermittent “works/doesn’t work” behavior for the same peer, and the mesh sees unnecessary airtime usage due to avoidable recovery traffic.

3. **Duplicate suppression robustness is bounded by small cyclic tables (replay/duplicate re-admission window)**  
Duplicate suppression relies on fixed-size cyclic memory of recently seen packet hashes and ACKs. Once these tables wrap, older entries are evicted and can be treated as unseen again. This model is efficient and simple, but it provides only a finite duplicate horizon that shrinks under high traffic.  
[...]

Issue #1489, opened 2026-01-26
Proposal: Improve DM reliability


rvleij:
DM is not working reliable (at least for me in the Netherlands), so I'm proposal some changes to improve reliability, increase the number of working direct DM paths and with that limit the amount of flood traffic.

## What changed:##
Two backward-compatible fixes to improve Direct Message ACK reliability:

### Fix 1: Use Incoming Path for ACK
**Problem:** Direct messages without stored path sent ACK via flood (often lost)
**Solution:** Use the incoming direct path to send ACK back

### Fix 2: Retry Flood ACKs
**Problem:** Flood ACKs sent once only (lost in noisy RF)
**Solution:** Send 3 ACKs at 200ms, 800ms, 2000ms

## Backward Compatibility

✅ **100% backward compatible**
- No protocol changes
- Works with old firmware nodes
- Old repeaters forward ACKs correctly
- Uses existing packet formats

### Reliability
- **+20-40%** ACK delivery rate
- **-30-50%** message retransmissions

### Network Traffic
- **-80%** overhead for first-contact direct messages
- **+20 bytes** per flood ACK (but saves retransmitting full messages)
## Testing

**Quick Test - Fix 1:**
```
1. Send direct message to node without stored path
2. Verify ACK comes back via direct route (not flood)
3. Check logs: "TX ... route=D" not "route=F"
```

**Quick Test - Fix 2:**
```
1. Send flood message
2. Watch for 3 ACK transmissions at 200ms, 800ms, 2s
3. Simulate packet loss - ACK still arrives
```

## Scenarios Covered

| Scenario | Old Behavior | New Behavior | Improvement |
|----------|--------------|--------------|-------------|
| Direct DM, no path | Flood ACK (often lost) | Direct ACK on incoming path | ✅ Major |
| Flood DM | 1× flood ACK | 3× flood ACK | ✅ Significant |
| Direct DM, with path | Direct ACK | Direct ACK (unchanged) | - |

rvleij:
I had Claude Code write the implementation, but haven't had time to actually check if this would work, since I'm not sure if the changes to sendAckTo would actually be valid.

The commit of my current state of work you find here:
https://github.com/rvleij/MeshCore/commit/dc6a0889aa9cc7a9cea36f8b7ef9aba5d7ed439a

timdj:
[...]

PR #2569, opened 2026-05-15
fix(#1775, #1489): path quality gating in onContactPathRecv + 3x flood ACK retry

thecharge:
# Reliability Changes: Path Quality Gating & Flood ACK Retry

Related issues: [#1775](https://github.com/meshcore-dev/MeshCore/issues/1775), [#1489](https://github.com/meshcore-dev/MeshCore/issues/1489)

---

### The Bug

`BaseChatMesh::onContactPathRecv` unconditionally overwrote the stored
`out_path` with any newly-arriving path, regardless of the quality of the
stored path or the quality of the incoming one.

In an RF mesh with multipath propagation, **flood path-return packets can
arrive from multiple routes in quick succession**.  The *first* to arrive wins
— which is not necessarily the shortest route.  More critically, a
longer-hop path arriving shortly *after* an established short-hop path
silently replaced the working route with a worse one.

Consequence (from issue #1775):
- A stable, proven direct route (e.g. 1 hop) could be replaced by a
  suboptimal multipath duplicate (e.g. 3 hops) arriving 50–200 ms later.
- Subsequent direct messages then travel via the longer route, increasing
  airtime, collision probability, and delivery failure rate.
- The user sees intermittent "works / doesn't work" behaviour with the same
  peer even when the mesh topology has not changed.

### The Fix

`onContactPathRecv` now applies a **stickiness window** before accepting a
path replacement:

1. If the stored path is **younger than `PATH_STICKINESS_WINDOW_SECS`**
   (default 600 s / 10 min) **and** the incoming path has **more hops** than
   the stored one, the stored path is kept.
2. Any embedded ACK or response carried inside the path packet is still
   processed regardless (so the sender's ACK timeout is cancelled correctly).
3. The stored path is **always** replaced when it is stale, when it is the
   same or shorter hop count, or when no path was previously known — ensuring
   the node still adapts to topology changes.

A new field `out_path_timestamp` (uint32_t, zero-init) on `ContactInfo`
records the RTC time at which the current `out_path` was last accepted.

### Delivery Success Tracking

A second new field `path_ack_count` (uint8_t, zero-init, saturates at 255)
is incremented in `onAckRecv` whenever an ACK arrives via the stored direct
path (not flood).  This provides a cheap per-contact signal of proven
delivery that can be used in future heuristics (e.g. giving a higher
stickiness weight to a path that has successfully delivered messages).

The counter is reset to zero whenever a new path is accepted, so it always
reflects delivery history on the *current* path.

### Configurable Knobs

| `#define`                     | Default | Description                            |
|-------------------------------|---------|----------------------------------------|
| `PATH_STICKINESS_WINDOW_SECS` | `600`   | Seconds a fresh path is protected from longer-hop replacement |

Override before including `BaseChatMesh.h` or via build flags.

---

### The Bug

When a direct message arrived via flood routing (i.e. the recipient had no
stored direct path to the sender), `sendAckTo` sent **exactly one** flood
ACK packet.

LoRa RF environments are inherently lossy.  A single ACK transmission:
- Can collide with other traffic on the channel.
- Can be lost due to transient interference or fading.
- Has no retry mechanism at the MAC layer.

When the ACK is lost the sender must wait for its full timeout (several
seconds) before attempting to retransmit the message, burning airtime and
battery, and degrading the user experience.  In practice users reported
that doubling or tripling ACK transmissions (already possible via
`getExtraAckTransmitCount()` for direct-path ACKs) dramatically improved
perceived reliability (issue #1489).

### The Fix

`sendAckTo` now sends **three independent flood ACK packets** at staggered
delays when `out_path_len == OUT_PATH_UNKNOWN`:

| Attempt | Delay   |
|---------|---------|
| 1st     | 200 ms  |
| 2nd     | 800 ms  |
| 2000 ms | 2000 ms |

Each copy is a separately-scheduled, independent RF transmission.
If the first copy is lost the second (and third) have a fresh chance of
reaching the sender through a congestion-free window.

### Deduplication on the Receiver Side

Duplicate suppression is handled by the **pre-existing `MeshTables::hasSeen()`
mechanism** at every node (including the destination):

- If a copy arrives and the node has already seen the same packet hash, it
  is discarded immediately — the ACK is not processed twice.
- If the first copy was lost (never received), neither the intermediate
  repeaters nor the destination have seen it, so the second copy propagates
  normally.

No protocol or wire-format changes are required.  The feature is 100%
backward-compatible with older firmware nodes that forward ACK packets
without understanding the retry intent.

### Configurable Knobs

| `#define`              | Default | Description                            |
|------------------------|---------|----------------------------------------|
| `FLOOD_ACK_RETRY_COUNT`| `3`     | Number of independent flood ACK sends  |
| `TXT_ACK_DELAY`        | `200`   | Delay of the first ACK (ms)            |

The 2nd and 3rd delays (800 ms, 2000 ms) are currently fixed in the
`flood_ack_delays` array inside `sendAckTo`.  They can be made configurable
via additional `#define`s if needed.

---

## Files Changed

| File | Nature of change |
|------|-----------------|
| `src/helpers/ContactInfo.h` | Added `out_path_timestamp` and `path_ack_count` fields |
| `src/helpers/BaseChatMesh.cpp` | `onContactPathRecv` — path gating logic; `sendAckTo` — 3× flood retry; `onAckRecv` — delivery tracking |

## Files Added

| File | Purpose |
|------|---------|
| `test/test_reliability/test_path_gating.cpp` | Unit tests for path gating condition |
| `test/test_reliability/test_flood_ack.cpp`   | Unit tests for flood ACK scheduling |

## Running the Unit Tests

```bash
# Run the new reliability tests on native platform
pio test -e native_reliability

# Run the existing utility tests (unchanged)
pio test -e native
```

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -166,3 +166,17 @@ build_src_filter =
   +<../src/Utils.cpp>
 lib_deps =
   google/googletest @ 1.17.0
+
+; Standalone unit tests for reliability changes (issues #1775, #1489).
+; These tests have no Mesh-stack dependencies and compile without the
+; Arduino framework or crypto libraries.
+[env:native_reliability]
+platform = native
+build_flags = -std=c++17
+test_filter =
+  test_path_gating
+  test_flood_ack
+build_src_filter =
+  -<*>
+lib_deps =
+  google/googletest @ 1.17.0
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -457,11 +457,16 @@ Packet* Mesh::createPathReturn(const uint8_t* dest_hash, const uint8_t* secret, 
     if (extra_len > 0) {
       data[data_len++] = extra_type;
       memcpy(&data[data_len], extra, extra_len); data_len += extra_len;
-    } else {
-      // append a timestamp, or random blob (to make packet_hash unique)
-      data[data_len++] = 0xFF;  // dummy payload type
-      getRNG()->random(&data[data_len], 4); data_len += 4;
     }
+    // Always append a random nonce regardless of extra presence.
+    // AES-ECB is deterministic: without a nonce, identical plaintext produces
+    // identical ciphertext → identical calculatePacketHash() → hasSeen() treats
+    // every retransmission of the same PATH+ACK as a duplicate and drops it.
+    // The nonce is ignored by the receiver (it reads only the typed extra fields).
+    // This mirrors the original no-extra handling above (issue: nonce was missing
+    // for the extra_len > 0 branch, breaking flood-ACK retry reliability).
+    data[data_len++] = 0xFF;  // dummy payload type (ignored by receiver)
+    getRNG()->random(&data[data_len], 4); data_len += 4;
 
     len += Utils::encryptThenMAC(secret, &packet->payload[len], data, data_len);
   }
--- a/src/helpers/BaseChatMesh.cpp
+++ b/src/helpers/BaseChatMesh.cpp
@@ -21,6 +21,17 @@
   #define PATH_STICKINESS_WINDOW_SECS  10u
 #endif
 
+// Proven-path lock (seconds): if at least one direct-path ACK has been
+// received for the stored path (path_ack_count > 0), block replacement for
+// much longer.  This prevents a flood-triggered PATH exchange from silently
+// overwriting a working direct path when the remote node temporarily loses
+// its own route back (e.g. after a reboot).  5 minutes is long enough to
+// survive typical disruptions; the app can always call CMD_RESET_PATH if
+// needed.
+#ifndef PATH_PROVEN_LOCK_SECS
+  #define PATH_PROVEN_LOCK_SECS  300u
+#endif
+
 // Number of independent flood-ACK transmissions at increasing delays.
 // Addresses issue #1489 (single flood ACK lost in noisy RF environments).
 #ifndef FLOOD_ACK_RETRY_COUNT
@@ -345,7 +356,14 @@ bool BaseChatMesh::onContactPathRecv(ContactInfo& from, uint8_t* in_path, uint8_
   // The embedded ACK/response is always processed regardless of the lock.
   if (from.out_path_len != OUT_PATH_UNKNOWN && from.out_path_timestamp != 0) {
     uint32_t age_secs = now - from.out_path_timestamp;
-    if (age_secs < PATH_STICKINESS_WINDOW_SECS) {
+    // Proven paths (at least one direct-path ACK received) are locked for
+    // PATH_PROVEN_LOCK_SECS; unproven paths use the shorter blanket window.
+    // This stops a flood-triggered reciprocal PATH exchange from silently
+    // overwriting a working direct path when the remote node loses its route.
+    uint32_t effective_window = (from.path_ack_count > 0)
+        ? PATH_PROVEN_LOCK_SECS
+        : PATH_STICKINESS_WINDOW_SECS;
+    if (age_secs < effective_window) {
       // Path is still within the lock window — keep it, but still process any
       // embedded ACK or response so the sender's retry timer is cancelled.
       onContactPathUpdated(from);
--- a/src/helpers/BaseChatMesh.cpp
+++ b/src/helpers/BaseChatMesh.cpp
@@ -9,14 +9,16 @@
   #define TXT_ACK_DELAY     200
 #endif
 
-// Path-stickiness window (seconds): a stored direct path that is younger than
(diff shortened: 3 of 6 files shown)


CullenShane:
This is really interesting, and there is opportunity for improvement in this area.

Less hops isn't automatically better, RX becomes a problem for high reach nodes so your stable 4 hop path with very high SNR, can degrade to a 2 hop path or 1 hop path where the sender got lucky with that packet. This is especially apparent on meshes that set `rxdelay` above 1, where lower SNR paths are propagated slower than high SNR paths.

If the first path received was locked for 10 seconds rather than 10 minutes and it didn't automatically accept a path with less hops, I think this would work very well.

thecharge:
> This is really interesting, and there is opportunity for improvement in this area.
> 
> Less hops isn't automatically better, RX becomes a problem for high reach nodes so your stable 4 hop path with very high SNR, can degrade to a 2 hop path or 1 hop path where the sender got lucky with that packet. This is especially apparent on meshes that set `rxdelay` above 1, where lower SNR paths are propagated slower than high SNR paths.
> 
> If the first path received was locked for 10 seconds rather than 10 minutes and it didn't automatically accept a path with less hops, I think this would work very well.

@CullenShane 

Good point,  the hop-count preference fights against the rxdelay-based SNR ordering. 
Reworking to a 10s blanket lock on first-arrived path, removing the hop comparison. 
I will keep path_ack_count since it's still useful as a future signal for replacing stale paths after the window (proven-delivery path should be stickier than an unproven one).

And would test two more days on my t-1000-e.

thecharge:
@CullenShane  done - updated the PR

thecharge:
@CullenShane looks ok from my side so far with two nodes talking to each other.

Maybe there are more people that would like to test.

CullenShane:
I've been testing it, and it seems to work, however, sometimes my path will get reset when the other node's path stops working.  This is surprising to me and I'm not sure if the stock firmware has the same behavior.

So a known good path will get reset on my node with no interaction on my part.

thecharge:
> I've been testing it, and it seems to work, however, sometimes my path will get reset when the other node's path stops working. This is surprising to me and I'm not sure if the stock firmware has the same behavior.
> 
> So a known good path will get reset on my node with no interaction on my part.

@CullenShane The remote node losing its own path back caused it to flood, triggering a reciprocal PATH exchange that silently overwrote your working direct path once the 10s stickiness window expired 
- should be fixed by keeping any path that has delivered at least one ACK locked for 5 minutes instead. But I have to test again

CullenShane:
> > I've been testing it, and it seems to work, however, sometimes my path will get reset when the other node's path stops working. This is surprising to me and I'm not sure if the stock firmware has the same behavior.
> > So a known good path will get reset on my node with no interaction on my part.
> 
> @CullenShane The remote node losing its own path back caused it to flood, triggering a reciprocal PATH exchange that silently overwrote your working direct path once the 10s stickiness window expired
> 
>     * should be fixed by keeping any path that has delivered at least one ACK locked for 5 minutes instead. But I have to test again

Locking the path for 5 minutes would mean that mobile operations would stop working where you could have a pretty quick direct message to flood message situation.  I understand the protocol is not designed with mobile operation in mind, even the radios are de-sensed from each other when one of them is moving, I assume from Doppler shift.  

In any case, I think this isn't a problem and doesn't need to be solved for.

If this change was coupled with `rxdelay` becoming a default setting again, I think it could be a great addition to reliable pathing.

thecharge:
> > > I've been testing it, and it seems to work, however, sometimes my path will get reset when the other node's path stops working. This is surprising to me and I'm not sure if the stock firmware has the same behavior.
> > > So a known good path will get reset on my node with no interaction on my part.
> > 
> > 
> > @CullenShane The remote node losing its own path back caused it to flood, triggering a reciprocal PATH exchange that silently overwrote your working direct path once the 10s stickiness window expired
> > ```
> > * should be fixed by keeping any path that has delivered at least one ACK locked for 5 minutes instead. But I have to test again
> > ```
> 
> Locking the path for 5 minutes would mean that mobile operations would stop working where you could have a pretty quick direct message to flood message situation. I understand the protocol is not designed with mobile operation in mind, even the radios are de-sensed from each other when one of them is moving, I assume from Doppler shift.
> 
> In any case, I think this isn't a problem and doesn't need to be solved for.
> 
> If this change was coupled with `rxdelay` becoming a default setting again, I think it could be a great addition to reliable pathing.

@CullenShane not really `rxdelay` related but if you want i can revert it as change.
Anyhow soon or later the protocol I think (but may be wrong) should become also mobile operation relevant.

PR #2571, opened 2026-05-16
Add packet filtering support to Meshcore

donutsoft:
After some discussions on the Meschore development channel, I saw two issues faced by our friends in Europe:
1. Repeaters are clogging the airwaves with unnecessary adverts.
2. Public channel spamming being caused repeatedly by a user running bad firmware: https://github.com/AkitaEngineering/Akita-Zmodem-MeshCore/issues/5

This PR aims to build the infrastructure for a mesh firewall that gives repeater owners the ability to filter specific packets. Two types are added to this PR as a starting point:
1. Advert rate limiter - limits the number of adverts a repeater/room/companion can issue in a specified time period (e.g. repeaters cannot advertise more than once every 24 hours). A bloom filter is used to keep memory use light.
2. A channel message filter - Allows public channel messages containing strings to be filtered. Either node names or messages can be filtered.

Configuring the packet filters is done over the CLI through the use of the block command. Rules are persisted between repeater restarts.

Assuming this is approved:
1. I'd like to move the loop detection and duplicate packet detection to this infrastructure, simplifying code elsewhere.
2. I want to build a system where repeater owners can allow limited access to Mesh administrators to add rules. Efficacy is limited unless most repeaters are configured the same.

Notes:
1. There's a rule that dynamic object allocation isn't allowed. PacketFilterRule.h does dynamically allocate the packet filter classifiers (AdvertRateLimitClassifier/ChannelMessageClassifier) as needed. There is a hard limit of 20 rules, but without this dynamic allocation, all rule instances would have to be statically allocated which makes things significantly less flexible and wasteful.
2. I did use AI to assist me in writing this. I also spent the best part of 2 days reviewing and rewriting code in order to make sure it works well.

Any feedback is much appreciated.


titusn:
Although I like the ability to add a firewall, please consider the principle of net neutrality. A central repeater or group of repeaters should probably not decide what messages are allowed on the mesh. Temporarily blocking spammers may be a good idea though.

PR #2586, opened 2026-05-18
Set return path for repeaters

mikecarper:
Adds repeater CLI support for viewing and controlling the current remote client's reply path.

  - Adds `get outpath`
  - Adds `set outpath <hop1_hex,hop2_hex,...>` 1,2,3 byte supported
  - Adds `set outpath clear` forgets the current direct path and allows normal path discovery to repopulate it
  - Adds `set outpath flood` forces replies to use flood packets until the client logs in again. Prevents discovered paths from overwriting flood mode while it is active

#### Why it's needed 
image

This is the ideal path there, which I set  
```7773, 7E76,``` _11E1_
But the path back is bad. It should be this  
_11E1_, ```7E76, 7773```  

I would like to be able to set the return path via the CLI 
```set outpath 7E76, 7773``` 
 In this case I get the login but nothing was given back to my client and so I get the timeout message.  
image

### CODE REVIEW

**Comment by towerviewcams** [comment]
We need this so very much. I cant tell you how hard it is to get stats. Log in works fine but then its going to war with retries to get the info. @liamcottle hope you agree ? this would take away allot of traffic with tons of retries.

@IoTThinks can we try this in your next test firm?

**Comment by IoTThinks** [comment]
I always assume the out and return path (in reverse) are the same.
It should.

Hi @mikecarper, 
Do we need to set this to all repeaters on the way back? Or just the repeater we want to remote manage to?

**Comment by mikecarper** [comment]
@IoTThinks just the one you're logged into needs the path set. Admin Repeater packets act like a DM where either side can set the path to the other end.

**Comment by IoTThinks** [comment]
@mikecarper Nice.
I will test with @towerviewcams 

No idea if we can automatically set this out path when we select the path to Remote manage.
May be in next version?

I have this issue too.
But I thought due to big size of repeater stats.

**Comment by IoTThinks** [comment]
@towerviewcams This is PowerSaving 15.0.3 (based on Meshcore 1.15 main line) and this PR2586.
The code looks solid.
Please help to test.

https://github.com/IoTThinks/EasySkyMesh/tree/main/firmware/Testing/WirelessRocks/PR2586

**Comment by towerviewcams** [comment]
> @towerviewcams This is PowerSaving 15.0.3 (based on Meshcore 1.15 main line) and this PR2586. The code looks solid. Please help to test.
> 
> https://github.com/IoTThinks/EasySkyMesh/tree/main/firmware/Testing/WirelessRocks/PR2586

I can confirm that this works and its sooooo nice! I have set 3 repeaters hex (3byte) and it replies with my set route for my companion.  If i test with a second companion (different from first) then the return path is normal.  

**Hope that next step on this can be to have a option for the "your logged in as Admin" green bar reply can auto add to reply the telemetry stats. As we all know that is always reliable as it follows the incoming path from the companion. But this is a GREAT step in the right direction and works!

very nice !!!!!!!  @AI7NC @liamcottle @ripplebiz @weebl2000 

{2AE87A44-B824-4FDE-A297-D2210B999261}

**Comment by IoTThinks** [comment]
> Hope that next step on this can be to have a option for the "your logged in as Admin" green bar reply can auto add to reply the telemetry stats. As we all know that is always reliable as it follows the incoming path from the companion. But this is a GREAT step in the right direction and works!

This item needs support / code change in MeshCore app, I guess.
I have merged this into PS 15. 👯

**Comment by towerviewcams** [comment]
@ripplebiz @liamcottle  Please consider looking at this for FW16.  We have this merged into PS 15.1 now and it works fantastic. Amazing for large scall systems where we cant get telemetry after log in. @IoTThinks did a great job putting this in at my request and do very well indeed!!

Thank you

**Comment by mikecarper** [comment]
added a small code change to make this work
`get/set outpath direct` 
for zero hop

**Comment by IoTThinks** [comment]
@mikecarper Hi Mike,
What is 63? 
Is it the mask 0b111111?
Thanks alot.

```
  if ((path_len & 63) == 0) {
    snprintf(out, out_len, "> direct");
    return;
  }
```

@towerviewcams I have merged to PS15 for you to test.
It looks great.

towerviewcams:
Ok sounds great! I'll test when I wake up. Got a repeater going in tomorrow and couple updates also

mikecarper:
Yeah bitwise math. Easy to do here

towerviewcams:
Testing this for little over 24 hours. works great. My repeater is 4 hops away from my home and solid return path now for telemetry stats.

IoTThinks:
> Testing this for little over 24 hours. works great. My repeater is 4 hops away from my home and solid return path now for telemetry stats.

@towerviewcams Merged to current PS15.  We will include this in PS16.
I have this issue too. 
Remote manage is instant for a few hops but getting the status is very hard. Need to retry a few times with luck.

mikecarper:
@ripplebiz this now targeting dev and is ready for you to review

towerviewcams:
We are using this already in PS16 and its working great. no more multiple retires. set your path for the companion that is logging in and its solid after that.

towerviewcams:
@ripplebiz @liamcottle We have been using this for over a month. Please look at for firm 17

IoTThinks:
@mikecarper This PR has a conflict now.
Could you please rebase / resolve the conflict?

IoTThinks:
@towerviewcams Are you actively use this PR? E.g the outpath.

towerviewcams:
> @towerviewcams Are you actively use this PR? E.g the outpath.

Yes sir!

m0urs:
I try to understad what "set outpath" really do_

I have set "set outpath=a5f2" and a get "outpath" gets back "hs=2 hops=1 hex=A5F2". So far it seems to be ok, as I am normally administer this repeater via one hop (via repeater "a5f2...."). However, it seems that the repeater does forget this setting from time to time and it either set it to "flood" or "direct". In both cases I am still able to administer my server via repeater "a5f2...".

1. Is it normal, that my hop setting is forgotten from time to time?
2. What does the repeater do, if it is set to "direct" or "flood" and I have no direct connection? I think even if set to "flood" it does not correctly set that to the repeater via I do my admin tasks. I always have to do that manually.

I think I did not really understand the mechanism behind "set outpath". Could someone plesa eexplain that to me in more detail? Also compared to the original 1.17.1 firmware.

In addition: "set outpath clear" returns "unknown". Is this works s designed or does that command just do nothing?

mikecarper:
The top message explains the issue; the path back to your node in a large mesh can be wrong; so you can successfully login but then getting anything back from the node doesn't work. This command allows you to set the return path that the reply will use. If everything still works then you probably don't need this. The act of logging in will reset this, you can continue the session to keep your out path

AI7NC:
As a sidenote, I've been told that the login ack always goes via flood.

73,

Nathan Campbell, AI7NC

mikecarper:
> As a sidenote, I've been told that the login ack always goes via flood. 73, Nathan Campbell, AI7NC

Yes go back to the screenshot in at the top https://github.com/meshcore-dev/MeshCore/pull/2586#issue-4473335895
It says "You are logged in as Admin" but then all attempts at getting the repeater stats fails and all commands seem to fail as well. This is a way to set the path back

m0urs:
> The top message explains the issue; the path back to your node in a large mesh can be wrong; so you can successfully login but then getting anything back from the node doesn't work. This command allows you to set the return path that the reply will use. If everything still works then you probably don't need this. The act of logging in will reset this, you can continue the session to keep your out path

Thanks, I guess " The act of logging in will reset this, you can continue the session to keep your out path" is what I was missing. But still my other question " "set outpath clear" returns "unknown". Is this works s designed or does that command just do nothing?"

mikecarper:
Returns > unknown - this means “route cleared/no route known,” not “unknown command.”

I should make this more clear. Thanks for the heads up

m0urs:
> Returns > unknown - this means “route cleared/no route known,” not “unknown command.”
> 
> I should make this more clear. Thanks for the heads up

Ah, thanks for clarification. Yes, "unknown" could be misunderstood  here ...

PR #2590, opened 2026-05-20
Fix typos in comments and docs

jocasmark:
Fixes spelling typos in comments and documentation across `src/` and `variants/`.

No functional changes. Typos in third-party code under `lib/` left untouched.


oltaco:
Thanks, looks good but maybe revert the typos in the nordic libs 😉

liamcottle:
Yeah, please revert all changes to files in `lib`:

- `lib/ed25519/add_scalar.c‎`
- `‎lib/nrf52/include/nrf52/nrf_mbr.h`
- `‎lib/nrf52/include/ble_l2cap.h`

Since these are external library files, ideally they should only be updated if pulled in from original source. Otherwise they will diverge from upstream.

jocasmark:
Fair enough. Didn't think of that. Reverted ✅

PR #2593, opened 2026-05-20
Add configurable SOS message support for T-Echo

pedrocaldeira66:
This PR adds a configurable SOS/Help message feature for T-Echo devices.

The feature allows the device to send a predefined emergency/help message from the device UI/button flow.

Current behaviour:
- Uses a configurable SOS message with `{gps}` placeholder support.
- Replaces `{gps}` with the current GPS coordinates when available.
- Appends a Google Maps link when GPS coordinates are available.
- Falls back to `GPS sem fix` when no GPS fix is available.
- Sends to the configured SOS channel slot.
- Falls back to Public / slot 0 if the configured channel is not available.
- Includes optional debug logging controlled by `SOS_DEBUG`.

This is intended as a lightweight field/off-grid helper for outdoor MeshCore use.

Future possible improvements:
- Optional acknowledgment/reply handling.
- Configurable retry interval.
- ATAK / TAK emergency marker integration.


pedrocaldeira66:
Additional testing update:

Tested successfully on:
- LilyGo T-Echo
- Heltec Wireless Tracker
- Heltec V3

On all three devices:
- SOS page appears correctly in the UI.
- Long press from the SOS page sends the message.
- Triple-press shortcut also sends the SOS message.
- Messages are received across bridged 868 MHz / 433 MHz MeshCore networks.

PR #2600, opened 2026-05-21
Added Power Saving for room servers

IoTThinks:
Hi all,
This PR is to add Power Saving to room servers for ESP32 and NRF52 room servers.
ESP32 room servers can achieve 10+ mA (instead of 100mA).
NRF52 room servers can achieve 5.8+ mA (instead of 12mA).

This features have been running fine for a few months on my variant.
Have a nice day.

PS: This is all related [pending PRs for Power Saving](https://github.com/meshcore-dev/MeshCore/issues?q=(state%3Aopen)%20is%3Apr%20author%3A%40IoTThinks)

### DIFF
--- a/examples/simple_room_server/MyMesh.cpp
+++ b/examples/simple_room_server/MyMesh.cpp
@@ -1022,3 +1022,11 @@ void MyMesh::loop() {
   uptime_millis += now - last_millis;
   last_millis = now;
 }
+
+// To check if there is pending work
+bool MyMesh::hasPendingWork() const {
+#if defined(WITH_BRIDGE)
+  if (bridge.isRunning()) return true; // bridge needs WiFi radio, can't sleep
+#endif
+  return _mgr->getOutboundTotal() > 0;
+}
--- a/examples/simple_room_server/MyMesh.h
+++ b/examples/simple_room_server/MyMesh.h
@@ -222,4 +222,7 @@ public:
   void clearStats() override;
   void handleCommand(uint32_t sender_timestamp, char* command, char* reply);
   void loop();
+
+  // To check if there is pending work
+  bool hasPendingWork() const;
 };
--- a/examples/simple_room_server/main.cpp
+++ b/examples/simple_room_server/main.cpp
@@ -18,6 +18,9 @@ void halt() {
 
 static char command[MAX_POST_TEXT_LEN+1];
 
+// For power saving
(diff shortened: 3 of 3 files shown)


oltaco (approved):

ripplebiz:
The MyMesh::hasPendingWork() needs more work.

There needs to be a check that there aren't outstanding post pushes still queued. Would have to do comparisions like:
               p->post_timestamp > client->extra.room.sync_since 

ie. find the LAST post (p), ie. with highest post_timestamp, then go through the Client ACL, and do that compare above. If true, then that client still needs that post pushed, so hasPendingWork() should return true.

IoTThinks:
> The MyMesh::hasPendingWork() needs more work.
> 
> There needs to be a check that there aren't outstanding post pushes still queued. Would have to do comparisions like: p->post_timestamp > client->extra.room.sync_since
> 
> ie. find the LAST post (p), ie. with highest post_timestamp, then go through the Client ACL, and do that compare above. If true, then that client still needs that post pushed, so hasPendingWork() should return true.

@ripplebiz What is the test use case / scenario for the above rule?
If the room server wakes up every 30s or even 5s, will it serve the above use case?

IoTThinks:
> Would have to do comparisions like:
> p->post_timestamp > client->extra.room.sync_since

Let me explore and test this suggestion.

IoTThinks:
New code changes from main.
Will rebase.
```
// To check if there is pending work
bool MyMesh::hasPendingWork() const {
  return _mgr->getOutboundTotal() > 0 || dirty_contacts_expiry != 0;
}
```

ViezeVingertjes:
This can get flaky really quickly as room server in particular has really weird edge-cases. You would also need to handle if any client has outstanding pending retries, if there are pending acks, and its very time sensitive, so if it would sleep in the incorrect place, it would invalidate those when waking up as its all expired, introduce weird and hard to debug behavior etc.

Unless youre going to dogfood this yourself very carefully... i would pass on this one for your own sanity, but thats just my 2 cents. 💯 

In real-life; a room server will always have pending work, undelivered messages, pending acks, 24/7, so i wonder if it would ever trigger once all checks are in.

IoTThinks:
@ViezeVingertjes Sure, I will relook at this PR when there is some spare time.
So I make this as Draft now.
Have a nice day.

IoTThinks:
I will use PowerSaving approach for BLE companions to apply to room servers instead.
Any I may safely bypass / ignore hasPendingWork().

PR #2604, opened 2026-05-22
Companion repeat: update to more usable default EU frequency

recrof:
869.495 is in 869.4 to 869.65 range which does have 10% duty cycle and 500mW ERP.

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -987,7 +987,7 @@ static FreqRange repeat_freq_ranges[] = {
   ALLOWED_REPEAT_FREQ_RANGE
   #else
   { 433000, 433000 },
-  { 869000, 869000 },
+  { 869495, 869495 },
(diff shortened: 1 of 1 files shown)


j0uni:
This should as well limit the duty cycle to 10% to be compliant with EU regulations.

recrof:
> This should as well limit the duty cycle to 10% to be compliant with EU regulations.

that's out of scope for this PR

j0uni:
> > This should as well limit the duty cycle to 10% to be compliant with EU regulations.
> 
> that's out of scope for this PR

"to be compliant with EU regulations" 

It is not.

@ripplebiz

recrof:
> "to be compliant with EU regulations"
> 
> It is not.
> 
> @ripplebiz

duty cycle is out of scope for this PR, it's different problem that needs to be adressed elsewhere, open new issue. companions don't have option to adjust it and it needs more work. this PR will still make companion repeat lot more usable for people in EU/UK.

PR #2605, opened 2026-05-22
Thinknode M6 Repeater Enable LEDs and Function Button

AtlavoxDev:
## What this fixes

On the original M6 repeater build:

- **No visual LED feedback for power-on, power-off, or reset**: the device gave the user no indication that a button press, shutdown command, or boot was recognized. Gave the impression of a dead device.
- **No way to turn off the device**: `poweroff` didn't actually sleep the chip. The original code called a SoftDevice API (`sd_power_system_off()`) that requires SoftDevice to be running — but SoftDevice isn't enabled in the repeater build, so the call silently failed and the chip stayed awake. No button mapped to poweroff.
- **Function Button was disabled.**: It was defined in `variant.h` but never read or armed as a wake source or poweroff.
- **USB cable triggered power on**: Charging the device would trigger the radio to boot up
## What this adds

- **Full LED choreography for every power event.**
  - **Boot:** dual-LED full bright → 1 s gap → red solid + blue "loading-activity" flicker _(To prevent user from attempting to press buttons during boot phase)_ → blue flash on boot complete.
  - **Hold ~2 s:** power off, with red blink ack at 0 s and 1 s + full-bright commit cue + dual-LED goodbye flash.
  - **Shutdown via `poweroff` CLI:** same shutdown cue.
- **Robust SYSTEMOFF.** New `enterDeepSleep()` helper with SoftDevice/direct-register fallback. Arms Function Button as `SENSE-LOW` wake source.
- **Reset-cause-aware boot.** `RESETREAS` captured in a priority-101 constructor (before Nordic's errata-136 workaround scrubs `RESETPIN`).
- **Resilient deployment.** `GPREGRET2` user-intent flag distinguishes deliberate shutdown from transient resets. Survives `SYSTEMOFF`, cleared by full power loss. Result: device stays off when _you_ asked it to, but self-heals from brownouts and auto-boots when a dead battery recovers.

## Implementation notes

- All M6-only and `#if defined(THINKNODE_M6)`-guarded.
- Other M6 envs (room server, companion radio, kiss modem) inherit the `SYSTEMOFF` fix but not the repeater UX.
- `analogWrite` PWM-binds pins on the Adafruit nRF52 core — boot path use `digitalWrite` to keep blue compatible with the existing LoRa TX LED indicator.

## Test plan

Verified on hardware:

- [x] Hold ~2 s → power off, function button wake works
- [x] `poweroff` over serial → real `SYSTEMOFF`
- [x] Reset button → clean reboot
- [x] USB plug-in to off device → stays off

Not field-verified:

- [ ] Dead-battery + solar recharge auto-boot (logic-tested only)

## Out of scope

- Charging-status LED (needs `EXT_PWR_DETECT` / `EXT_CHRG_DETECT` and a wake-on-charge mode) — planned follow-up.


AtlavoxDev:
Hey MeshCore team, this is my first ever attempt to contribute to a repo. Let me know if there are any questions or if there's something I should do differently next time. I only targeted the Repeater firmware this time, but will explore the other M6 firmware too for feature parity. Note I used Claude to assist and manually reviewed all code. Novice developer, wanted to take a stab at adding these critical features myself and contributing to the project. Tested over the past couple days.

AtlavoxDev:
I moved as much as a could out of main.cpp. Only a couple small blocks inside main.cpp are minimal dispatch calls (board.bootComplete() and board.pollButton()) to enable boot monitoring and button functionality.

AtlavoxDev:
I need to research the affects this has on other firmware builds for the boot LED indicator. I think this causes a regression with companion firmware builds, which would cause an indefinite LED flicker.

Issue #1741, opened 2026-02-19
Proposal: Coarse GPS locations


fsedarkalex:
I would like to propose adding coarse GPS locations.

To make it least effort, I'd suggest the following:

In GPS settings add an option "coarse" and rename "on" to "exact".

Whenever the device is using the GPS location, it will calculate and round the lon/lat coordinate to a precision which will effectively cause a worldwide "grid" of approximately 5x5km cells (at equator), up to 5x8km cells (near poles).

I used ChatGPT to create such a calculation code:
```
#include <math.h>

#define GPS_COARSE_GRID_DEG 0.05
#define GPS_COARSE_MIN_FACTOR 0.7

static inline double lat_factor(double lat)
{
    double f = 1.0 - 0.2 * (fabs(lat) / 90.0);  // linear approximation
    return (f < GPS_COARSE_MIN_FACTOR ) ? GPS_COARSE_MIN_FACTOR : f;
}

double snap_lat(double lat)
{
    return round(lat / GPS_COARSE_GRID_DEG ) * GPS_COARSE_GRID_DEG ;
}

double snap_lon(double lon, double lat)
{
    double step = GPS_COARSE_GRID_DEG / lat_factor(lat);
    return round(lon / step) * step;
}
```

I made this simple on purpose. Of course, there is more and nicer options but I think we should stick to the basics and keep it simple.

//Edit: Maybe we should also add a hysteresis. The Location should only update, when the device has actually moved 500m (on the exact location) or more. This prevents jumping if it is physically at a cell border.

mrlmuk:
I have proposed a similar idea before, using the decimal places to increase or decrease the precision and changing the default from 0.000001 (11cm) to 0.001 (111meters).

Here are a few more related posts which would help battery life, allow better tracking if devices like T1000/RakTag and provide an initial NTP/RTC time clock update to repeaters.

[Feature Request] Position - add the ability to vary the precision https://github.com/meshcore-dev/MeshCore/issues/870
Add optional GPS obfuscation / reduced precision to mitigate unintended tracking #1265

[Feature Request] GPS polling at 1 Hz is wasteful — interval & motion-based activation needed [#873](https://github.com/meshcore-dev/MeshCore/issues/873)
moko smart/WisMesh Tag needs ability to always send position https://github.com/meshcore-dev/MeshCore/issues/866

fsedarkalex:
I see, precision is the word I should have searched for 😆

tedder:
[...]

Issue #870, opened 2025-10-01
[Feature Request] Position - add the ability to vary the precision.


mrlmuk:
Currently the app show positions as a lat lon of 6 decimal places, an accuracy of 11cm at 0,0. While this has its uses if you are tracking a device, most people don't want or need this level of accuracy. People with repeaters are reluctant to upload them to a map or add a location for fear of people stealing them or otherwise interfering.

Perhaps if the precision was made variable and the default lowered to 3 decimal places (111 meters) this would helpt offset these fears and also increase privacy for those who don't realise they are advertising their position.

Although it may need some config storing the firmware would just see lots of zeros at the end. The app could be fully aware of this and show a larger circle of possibility on the map rather than pinpointing the wrong spot.

Meshtastic has a similar feature already.

techyporcupine:
Agreed. I would like to share my node GPS, but don't want it as accurate as the GPS can possibly go.

Danie10:
I have to manually fudge my repeater's exact location, but this is a bigger issue if I travel with my mobile node. At home, I'd like to toggle the precision to be obscured a bit, but if out travelling or doing rescue work I need full precision. In this day and age, one does need less precision especially at a home location, or for where the repeater is actually located.

kiwiant:
> I have to manually fudge my repeater's exact location, but this is a bigger issue if I travel with my mobile node. At home, I'd like to toggle the precision to be obscured a bit, but if out travelling or doing rescue work I need full precision. In this day and age, one does need less precision especially at a home location, or for where the repeater is actually located.

Although not for real time tracking, Garmin has the idea of randomised blackspots for home/work/school so others can't see the start and end location. These are jagged edge shapes to make it harder to discover the real location over time but obviously a house in the middle of a large farm will be more difficult.

Danie10:
@kiwiant yes I've seen the same on the Strava exercise app. I think that may make things more complicated though and these devices and networks want to stay super lightweight. I think the precision adjuster may be simpler to implement. But I'd be happy with either.

kiwiant:
> [@kiwiant](https://github.com/kiwiant) yes I've seen the same on the Strava exercise app. I think that may make things more complicated though and these devices and networks want to stay super lightweight. I think the precision adjuster may be simpler to implement. But I'd be happy with either.

Agreed, but I didn't want to start suggesting implementations :-)

jschrempp:
[...]

PR #2611, opened 2026-05-23
feat: add location privacy / GPS blurring

tedder:
addresses #870 and #1741 by chopping the floating point digits at `(6-blur_digits)`. Tested and running on my Station G2, but should work on all.

Using integer math for the truncation.

Further, we're doing deterministic **GPS fuzz offsets**.  When in "blur mode", using the device pubkey, offset the device from the truncated GPS coordinate. This prevents multiple devices from being at the same point.
    
By using modulo we may not be evenly distributed, but they should have decent spread.
    
Here are some examples from one device:
```
gps
  -> on, active, fix, 7 sats
gps blur
  -> > 3, offset lat+0.000071 lon-0.000310


gps blur 4
  -> ok, offset lat+0.000021 lon+0.000040
gps blur 3
  -> ok, offset lat+0.000071 lon-0.000310
gps blur 2
  -> ok, offset lat-0.003429 lon-0.004810
```

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -905,6 +905,27 @@ region save
 
 ---
 
+#### View or change GPS location blur (privacy)
+**Usage:**
+- `gps blur`
+- `gps blur <digits>`
+
+**Parameters:**
+- `digits`: `0`–`6` decimal places to publish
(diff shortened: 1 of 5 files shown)


skyepn:
This would be very welcome

PR #2622, opened 2026-05-25
Add MainBoard::pollButton() hook + correct SenseCAP Solar button naming

AtlavoxDev:
## Summary

Two related changes in two commits:

**1. `MainBoard::pollButton()` framework hook** — adds a no-op virtual hook mirroring the existing `onBootComplete()` pattern (#2621). Every example's `loop()` calls `board.pollButton()` unconditionally. Boards with buttons override; boards without inherit the no-op default. Removes the existing one-off device-specific conditional `#if defined(_SEEED_SENSECAP_SOLAR_H_)` blocks from `simple_repeater/main.cpp` and moves the equivalent hold-1.5s-to-power-off logic into `SenseCapSolarBoard::pollButton()` as an override. Behavior unchanged.

**2. SenseCAP Solar: correct button naming** —  The firmware aliased `PIN_USER_BTN` to pin 13, but pin 13 is the PWR button (SYSTEMOFF wake source + hold-to-power-off target). The actual USR button (pin 20) was completely unused. Renames so code matches silkscreen: `PIN_PWR_BTN` → pin 13, `PIN_USER_BTN` → pin 20. RST is hardware-only (not GPIO). Behavior unchanged.

## Rationale

This PR follows the same pattern as the recently-merged `onBootComplete()` framework hook (#2621). It exists as a precondition for two in-flight device-specific PRs (ThinkNode M6 power-button/LED work, SenseCAP Solar boot LED/Button improvements) so that those PRs don't have to carry both the framework decision *and* the device implementation in a single review.

## Follow-up

Two device-specific PRs are queued to land after this one:
- ThinkNode M6: overrides `pollButton()` for a 2 s hold-to-power-off sequence with red-flash UX
- SenseCAP Solar: adds boot LED sequence and adds hold-feedback flashes to the existing `pollButton()` override

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -243,6 +243,8 @@ void setup() {
 }
 
 void loop() {
+  board.pollButton();
+
   the_mesh.loop();
   sensors.loop();
 #ifdef DISPLAY_CLASS
--- a/examples/kiss_modem/main.cpp
+++ b/examples/kiss_modem/main.cpp
@@ -124,6 +124,8 @@ void setup() {
 }
 
 void loop() {
+  board.pollButton();
+
   modem->loop();
 
   if (!modem->isActuallyTransmitting()) {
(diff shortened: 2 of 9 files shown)


AtlavoxDev:
Continuing to lay groundwork for improving LED and button functionality for devices. Added a hook that devices can use for buttons, and removed the device-specific code for the P1-Pro into its own codebase instead of the main.cpp.

Issue #1031, opened 2025-10-30
MeshCore Telemetry Scaling Problem


awolden:
# MeshCore Telemetry Scaling Problem

## Problem Statement
MeshCore moved from Meshtastic’s flood-push model to a PULL model (DM request when a path is known; flood request as fallback). While this reduces constant background chatter at small scale, it introduces risks as networks grow:

- **REQ/RES doubling:** Every telemetry read requires a request and a response; on lossy links this compounds with retries.
- **Inversion of control:** Harvesters can’t tell *node offline* vs *link flakey* vs *route stale*. That opacity drives retries and flood fallbacks that create bursts.
- **Reliance on user good-faith:** Meshtastic did **not** use polling, its failure mode was automatic *flood* telemetry saturating airtime. But MT still depended on users to configure responsibly (power, intervals, channels, topology). MeshCore risks a similar dynamic if it relies on external harvesters behaving; a single misconfigured harvester can overwhelm shared airtime. If left as is, users will inevitably abuse the network in a way that impacts it for everyone.
- **Multi-request patterns for repeaters:** Fetching “details” and “telemetry” separately costs airtime when it could be a single operation.
- **Zombie harvesters** - When telemetry harvesters are set-up there is a high likelihood they run indefinitely reaching out for telemetry sensors that may not even exist anymore.

These combine into airtime saturation, unfairness (some nodes starve), and instability during config changes or partial outages.

## Pain Observations
- **Traffic inflation from PULL:** REQ + RES per read; retries multiply airtime.
- **Flailing under uncertainty:** Harvesters over-poll when responses are intermittent, often escalating to flood requests.
- **Good-faith failure mode:** Even one motivated user monitoring many telemetry sensors can unintentionally bring down a mesh, despite local rate-limits.
- **Operational churn:** Config changes correlate with reconnect storms and restarts; lifecycle gaps can leave background pollers running across reloads.
- **Redundant queries for repeaters:** Separate requests for details and telemetry increase contention and latency.

## Example
**PugetMesh Health.** A single well-intended telemetry harvester monitored multiple deployed repeaters. When responses were intermittent, the harvester retried and escalated (including flood fallbacks), which led to:
- **One origin saturating airtime:** one harvester inadvertently degraded the mesh while “watching” many repeaters.
- **Retry/flood amplification:** intermittent replies triggered aggressive retries and flood requests, increasing contention for everyone.
[...]

PR #2629, opened 2026-05-26
[mmc] Add commands to enforce direct path and power settings in repeater telemetry replies 

xyzw-io:
### Motivation

Tangential to [#1031](https://github.com/meshcore-dev/MeshCore/issues/1031): local telemetry meshes do not always need flood-routed replies. For same-site repeater polling, flood responses add avoidable public mesh traffic.

### Changes

Adds opt-in repeater controls for admin telemetry replies:

```text
set telem.reply.zerohop on
get telem.reply.zerohop

set telem.reply.power 10
get telem.reply.power

set telem.reply.power off
```

When `telem.reply.zerohop` is enabled, the repeater sends matching replies as direct zero-hop packets instead of using flood fallback. Zero-hop means the packet is transmitted once by the repeater and is not forwarded by other nodes.

`telem.reply.power` optionally overrides TX power for that reply only, then restores the normal node TX power afterward. `off` uses the normal configured `tx` value.

Defaults stay stock: zero-hop is off and reply power override is off. This PR is a workaround for local telemetry meshes, not a general fix for telemetry scaling.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -41,6 +41,10 @@
   #define TXT_ACK_DELAY 200
 #endif
 
+#ifndef TELEM_REPLY_ZEROHOP_DEFAULT
+  #define TELEM_REPLY_ZEROHOP_DEFAULT 0
+#endif
+
 #define FIRMWARE_VER_LEVEL       2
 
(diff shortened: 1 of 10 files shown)


xyzw-io:
I can also scale back to only adding zero hop command for now. This will make PR a bit smaller.

Issue #2628, opened 2026-05-26
Add `MESH_CONSOLE_SERIAL` macro to redirect repeater console/debug to `Serial1` (RAK3401)


disq:
On RAK19007 + RAK3401 the WisBlock base exposes TX1/RX1 on the header. With the RAK13302 1W add-on we’d like to power the node from an third-party 5V supply (not the Pi’s USB) and talk to a Raspberry Pi over GPIO UART. Today that isn’t possible without code edits, because all console/debug output in `simple_repeater` is hardcoded to `Serial` (USB CDC).

## Proposal

Add a single macro that selects which serial the repeater console + debug logging uses, defaulting to `Serial` so behaviour is unchanged.

In src/MeshCore.h:
```
#ifndef MESH_CONSOLE_SERIAL
  #define MESH_CONSOLE_SERIAL Serial
#endif
```

Use `MESH_CONSOLE_SERIAL` for:
`MESH_DEBUG_PRINT` / `MESH_DEBUG_PRINTLN`
`BRIDGE_DEBUG_PRINTLN`
`examples/simple_repeater/main.cpp` console I/O (boot ID, command loop, power-off msg)
`examples/simple_repeater/MyMesh.cpp` (logRxRaw, dumpLogFile, get acl)

Do:

```
MESH_CONSOLE_SERIAL.begin(115200);
```

Enable per-variant with:

```
-D MESH_CONSOLE_SERIAL=Serial1
```

## Why not just use the bridge / USB

USB CDC requires powering the node from the Pi, which fights the 1W add-on's preferred third-party 5V supply path.
`WITH_RS232_BRIDGE` is a binary bridge protocol, not a debug/console.
Compile-time `-D Serial=Serial1` collides with libraries that reference `Serial`.

## Scope

Default behavior unchanged (`MESH_CONSOLE_SERIAL` falls back to `Serial`).
Only `simple_repeater` example touched in the first pass; other examples could follow the same pattern.
Happy to open a PR with the change.

disq:
Looks like a dupe of https://github.com/meshcore-dev/MeshCore/issues/2513, will close

Issue #2513, opened 2026-05-09
Feature Request: Ability to switch between USB-Serial and UART1-serial


K5KTF:
Currently on a RAK4630/31 + 19007 all console/functions goes only to the USB serial. Connecting (wired) to UART1 TX/RX does nothing.  

Proposal: Add a setting--checkbox/radio button in the gui/setting in config via cmdline--that selects whether all functions normally done via USB serial reroutes to the UART1 pins on the node/19007. 

Use case: Using a Serial-to-Ethernet adapter and PoE->5VDC USB-C splitter (for power via CAT6) at the repeater node, the ability to control the node over ethernet (with virtual COM port software @ local PC) versus USB would be extremely helpful when up a 50' tower and climbing ability is limited. Plus if the node needs a power cycle, easier to be done by unplug/replug at the PoE switch on the ground than climbing 50' up a tower to mess with battery/solar. 
Node control/console/connectivity would then be available via the LAN (or WAN if desired and router port-forwarding setup separately). 

I currently have this setup on bench, connected to 8dBi antenna up 20' over 1/2" heliax, all works great (sees 6 neighbors via USB Serial com port) and Ethernet COM connectivity all the way to the TX/RX wires at the 19007, but then no response in serial terminal. 

Would be glad to be guinea pig/test. 
I tried looking at the code to do myself, but Im a hardware/RF guy, not software, so my brains leaked out when looking through /git/MeshCore/src ...
I thought that for anyone who knows the code this might be an easy thing?  ( s/serial/serial1/g ?)

Devices in use with RAK4631/19007: 
PoE Splitter: https://www.amazon.com/dp/B09GM8FB3X?ref=ppx_yo2ov_dt_b_fed_asin_title&th=1I&content_source=fb&fb_content_id=Q9-wBQGjyctRSjHjO34UIzAbPvBNdngECCqrILh4Vxv1NQrjcwzluzz4y-qYnYGJaA&channel_type=fb&fbclid=IwY2xjawRsOnpleHRuA2FlbQIxMABicmlkETFUNmJUN1JRZFhnd2tjRlVrc3J0YwZhcHBfaWQQMjIyMDM5MTc4ODIwMDg5MgABHn7JVT6f2GqmBzSwK3QsI0If_l81fsmQJtbp-dQh6h4zLZ8e3T-UrOvuG8HE_aem_C7NhY6ZUwPmQ3VxMpoIJog

[...]

PR #2630, opened 2026-05-26
Add UART console configurability to repeater 🤖🤖

disq:
Implements https://github.com/meshcore-dev/MeshCore/issues/2513 (as it was summarized in https://github.com/meshcore-dev/MeshCore/issues/2628)

I had the code already (tested and seems to work) so why not submit it.

## Configurable UART console

Adds a `MESH_CONSOLE_SERIAL` indirection so console/log output can be routed off USB (`Serial`) onto a spare UART (`Serial1`). On RAK3401 this lets USB stay power-only while serial dumps flow out the extra UART.

### How it works
- `MESH_CONSOLE_SERIAL` defaults to `Serial`; override via build flag (e.g. `-D MESH_CONSOLE_SERIAL=Serial1`).
- `MESH_DEBUG`/`BRIDGE_DEBUG` macros and all repeater/room-server console I/O print through it.
- Opt-in `USE_SERIAL1_CONSOLE` wires up the Serial1 pins on nRF52.

### Changes
- `src/MeshCore.h` — define `MESH_CONSOLE_SERIAL`; route the debug macros through it.
- `src/Dispatcher.cpp` — redirect the `MESH_PACKET_LOGGING` RX/TX packet dumps.
- `examples/simple_repeater/` — console output + Serial1 setup.
- `examples/simple_room_server/` — same redirection for parity.
- `variants/rak3401/platformio.ini` — commented `-DMESH_CONSOLE_SERIAL=Serial1` and `-UENV_INCLUDE_GPS` flags (the latter to avoid GPS UART clash)

### Out of scope
- `WIFI_DEBUG`/`ESPNOW_DEBUG` (esp32) still hardcode `Serial` — left untouched.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -461,10 +461,10 @@ const char *MyMesh::getLogDateTime() {
 
 void MyMesh::logRxRaw(float snr, float rssi, const uint8_t raw[], int len) {
 #if MESH_PACKET_LOGGING
-  Serial.print(getLogDateTime());
-  Serial.print(" RAW: ");
-  mesh::Utils::printHex(Serial, raw, len);
-  Serial.println();
+  MESH_CONSOLE_SERIAL.print(getLogDateTime());
+  MESH_CONSOLE_SERIAL.print(" RAW: ");
+  mesh::Utils::printHex(MESH_CONSOLE_SERIAL, raw, len);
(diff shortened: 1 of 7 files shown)


disq:
~~er, this currently won't help with serial packet logging (observers/mqtt) as the affected code sits in an upper level (Dispatcher). Checking if there's a better way to do it (with RS232/custom env/etc)~~

Update: following commits fix this

PR #2640, opened 2026-05-28
feat: Reliable GPS on all RAK WisBlock GPS boards (RAK12500 ublox + RAK12501 Quectel)

disq:
Re-application of https://github.com/meshcore-dev/MeshCore/pull/2401

| Base | Core | GPS @ slot | Status |
|------|------|-----|--------|
| 19003 | RAK4631 | RAK12500 (I²C) @ C | ✅ tested |
| 19003 | RAK4631 | RAK12501 (serial) @ C | ✅ tested |
| 19007 | RAK3401+13302 | RAK12500 @ A | ✅ tested |
| 19007 | RAK3401+13302 | RAK12501 @ A | ✅ tested |
| 19007 | RAK4631 | RAK12501 @ A | ✅ tested |
| 19007 | RAK4631 | RAK12500 @ A | ✅ tested |

All 19003 tests had RTC in slot D and all 19007 tests had RTC in slot C.
All RAK4621 tests had `-UPIN_USER_BTN` to prevent P0.09 conflict with RTC on 19003. (including 19007 tests)

---

## Changes beyond the original #2401 re-application

**RAK3401 GPS enablement (b6e6730f)**
- Restore `-D RAK_BOARD` on `[rak3401]` (removed in the #2401 revert due to "RadioLib error -707"). The real cause was `rakGPSInit()`'s WB_IO4 fallback probe pulsing pin 4 (= SX126X_RESET on RAK3401) LOW for 500 ms, hard-resetting the radio after `radio_init()`. The fixes below make `RAK_BOARD` safe to re-enable.
- `rakGPSInit()` WB_IO2 probe: WB_IO2 also controls the 3V3_S switched peripheral rail; a failed probe left it as INPUT, dropping the rail (and the RTC/display/GPS with it). Restore WB_IO2 HIGH before falling through to the WB_IO4/WB_IO5 probes.
- Add `-D FORCE_GPS_ALIVE` on the RAK3401 base env so `stop_gps()` can't later pull `gpsResetPin` (which may resolve to the radio reset / rail pin) LOW after init.

**RAK4631 GPS enablement — `FORCE_GPS_ALIVE` parity (47d3753c)**
- Add `-D FORCE_GPS_ALIVE` to the `[rak4631]` base, matching RAK3401. Without it, `rakGPSInit()` ends by calling `stop_gps()` to park the receiver. On RAK4631 (`PIN_GPS_EN=-1`, GPS powered from the always-on 3V3_S rail) that "stop" is a *software* GNSS-stop the receiver did not reliably resume from over I²C, so the GPS booted dark (SIV stuck at 0). The flag compiles out that single boot-time stop; the runtime `start_gps()`/`stop_gps()` paths are unchanged.
- Move the GPS UART pin defines (`PIN_GPS_TX`/`PIN_GPS_RX`) from `platformio.ini` into `variant.h` (value-identical), alongside the other GPS constants — again matching RAK3401.

**No dedicated GPS enable pin: `PIN_GPS_EN = -1` (14f9d08a)**
- Make `-1` a real "no GPS enable pin" sentinel. Default `gpsResetPin` changed from `0` (a real GPIO — P0.00 / LFXO on nRF52) to `-1` (`0xFFFFFFFF`, an out-of-range index that `pinMode()`/`digitalWrite()` treat as a no-op), and `gpsIsAwake()`'s serial branch now tests `PIN_GPS_EN >= 0` instead of a truthy `PIN_GPS_EN` (which wrongly accepted `-1`). Adds `-D PIN_GPS_EN=-1` to RAK3401 so it matches RAK4631. Aligns the WisBlock path with `MicroNMEALocationProvider`, where `GPS_EN` already defaults to `-1`.

**GPS stop/start decoupled from the shared 3V3_S rail (ff82e66a)**
- `start_gps()`/`stop_gps()` now branch on whether a dedicated enable pin exists. With `PIN_GPS_EN >= 0`, behaviour is unchanged (drive the pin). With **no** enable pin (the common WisBlock case, where `gpsResetPin` resolves to **WB_IO2 = the shared 3V3_S rail**), they no longer toggle that pin — they idle/resume just the GNSS engine in software and leave the rail HIGH:
  - **I²C u-blox (RAK12500):** `UBX-CFG-RST` *controlled GNSS stop* (resetMode `0x08`) / *start* (`0x09`) over I²C — stops RF/tracking but keeps the receiver's host interface alive, so it restarts reliably. (UBX-RXM-PMREQ backup was rejected: on this hardware it only wakes on EXTINT/UART, **not** I²C — confirmed dead.)
  - **Serial MTK/Quectel:** `$PMTK161,0` standby / single `0xFF` wake byte.
- Previously both paths unconditionally drove `gpsResetPin`; in the no-enable-pin case that meant `stop_gps()` pulled **WB_IO2 LOW**, browning out everything on the 3V3_S rail (RTC, I²C peripherals, boost) and relying on it coming back. GPS-off is now non-destructive, and the receiver warm/hot-starts.

**Stale PVT data fix (b6e6730f)**
- `RAK12500LocationProvider` passed maxWait of 2/8 ms to the ublox getters — below the I²C response window, so polls timed out and the SparkFun library returned stale/garbage PVT (observed: a current latitude paired with a longitude from a previous fix). Bumped to a 250 ms ceiling.
- Switched `getAltitude()` (height above WGS84 ellipsoid) to `getAltitudeMSL()` for more intuitive altitude.

**Cold-start false fix (f2ad72e6)**
- Require both `gnssFixOK` **and** `fixType == 3` before reporting coordinates. Ublox reports `gnssFixOK == true` with bogus almanac-derived coords during cold start; requiring a 3D fix filters those out.

**Main-loop / button responsiveness fix (a13098a5)**
- The 250 ms maxWait above, combined with `RAK12500LocationProvider::loop()` running on *every* main-loop iteration, meant the loop spent most of its time blocked on redundant I²C polls (the GNSS solution only updates at the 1 Hz nav rate). This made the AIN1 button sluggish and dropped double-clicks ("prev").
- Throttle the poll to ~1 Hz inside the provider, and gate all field reads on a single `getPVT(250)` (returning early on a timeout so no getter fires its own ~1100 ms default poll on a stale cache). The throttle lives in the provider rather than the shared outer loop because the serial/NMEA providers need `loop()` every iteration to avoid dropping bytes.

**RAK12501 (Quectel) serial GPS on RAK19007 Slot A (9e8cbf6b)**
- Correct the GPS UART RX/TX orientation in the RAK3401 variant (host TX → GPS RX, host RX → GPS TX). The previous straight-through mapping (written for a u-blox module) left no bytes flowing in either direction, so the RAK12501 (Quectel, serial) in Slot A was never detected. Pure pin-assignment fix — baud (9600), power rail, reset/enable handling, and the `LocationProvider` path are untouched.

**UI: real GPS fix type on the GPS screen (8f867bfd)**
- Add `getFixType()` to the `LocationProvider` interface (default derived from `isValid()`; `RAK12500LocationProvider` overrides it with the actual u-blox fix type). The ui-new GPS screen now shows **`off` / `no fix` / `2D` / `3D`** instead of a binary `fix`/`no fix`, and the satellite count updates **during acquisition** (SIV is refreshed every tick rather than only on a 3D lock). Display/diagnostics only — `isValid()` and when a position is reported/logged are unchanged (still strict 3D).

**Debug-print cleanup (c9dfd1bd)**
- Dropped a duplicate lat/lon debug print and throttled the remaining one (~every 10th update tick).

**Docs: `PIN_USER_BTN=9` / Slot D note (13d34967)**
- Documentation-only. Annotate the three `-D PIN_USER_BTN=9` RAK4631 envs noting P0.09 = WisBlock IO5 (Slot D): a Slot D module that drives IO5 (e.g. the RAK12002 RTC's RV-3028 CLKOUT) pins it LOW, which reads as a permanently-held user button → boots into CLI rescue → the BLE/USB companion link appears dead. No binary change.

**GPS screen: accurate fix-state display (d2e47487)**
- Add `LocationProvider::getFixType()`; the companion GPS page now shows **off / no GPS / no fix / 2D / 3D** with a live satellite count, instead of a binary fix/no-fix.
- Render `---` for position until there's a real fix, and reject what would otherwise flash as bogus coords: the ublox cold-start `fixType==3`-with-junk frames (gated on `SIV>0` + in-range coords) and MicroNMEA's `999°` no-fix sentinel (both normalised to `0`).
- Position renders **eagerly** on-screen as soon as there's a 2D/3D solution; telemetry stays strict — `isValid()`/`node_lat` only report an accuracy-OK (`gnssFixOk`) 3D fix.
- `getLocationProvider()` returns `NULL` when no GPS is detected → the screen reads **"No GPS detected"** instead of a phantom unfixed GPS.

---

## Scope / affected boards

The `RAK12500LocationProvider`, `rakGPSInit()`, and `start_gps()`/`stop_gps()` changes compile under `RAK_WISBLOCK_GPS` (`ENV_INCLUDE_GPS && RAK_BOARD && !RAK_WISMESH_TAG`), so they affect **every RAK I2C-ublox GPS board, not just the RAK 1W kit**: `rak3401`, `rak4631` (all envs except the two RS232 bridges, which `-UENV_INCLUDE_GPS`), and `gat562_30s_mesh_kit`. (`gat562_mesh_watch13` has GPS disabled; `rak_wismesh_tag` is excluded and keeps its real `PIN_GPS_EN=34`.)

Implications for the already-shipping RAK4631 / GAT562 boards:
- **1 Hz poll throttle** — strictly beneficial; the GPS poll previously ran on every main-loop iteration. (The sluggish-button symptom was only *visible* on RAK3401 because of its AIN1 analog button, but the redundant blocking affected all of these boards.)
- **`fixType == 3` gating** — coordinates are now withheld until a 3D fix (filters cold-start false fixes).
- **`getAltitudeMSL()`** — altitude is now height above mean sea level instead of the WGS84 ellipsoid.
- **`rakGPSInit()` WB_IO2 rail-restore** — safety fix for a failed first probe dropping the 3V3_S peripheral rail.
- **Rail decouple + software GNSS stop/start** — on boards with no dedicated GPS enable pin, toggling GPS no longer browns out the shared 3V3_S rail (RTC / I²C peripherals stay powered), and GPS warm/hot-starts. Boards with a real `PIN_GPS_EN >= 0` are unchanged.
- **`PIN_GPS_EN = -1` sentinel** — hardens the no-enable-pin path; no change for boards with a real enable pin.
- **GPS screen fix-type / live sat count** — ui-new display change only; no effect on reported/logged position.


src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -168,16 +168,21 @@ class RAK12500LocationProvider : public LocationProvider {
...
+    // routinely timed out and returned stale/garbage PVT data (observed: random
+    // wrong-hemisphere longitudes despite getGnssFixOk() returning true).
+    // 250 ms matches the SparkFun library's typical response time.
+    if (ublox_GNSS.getGnssFixOk(250)) {
       _fix = true;
-      _lat = ublox_GNSS.getLatitude(2) / 10;
-      _lng = ublox_GNSS.getLongitude(2) / 10;
-      _alt = ublox_GNSS.getAltitude(2);
-      _sats = ublox_GNSS.getSIV(2);
+      _lat = ublox_GNSS.getLatitude(250) / 10;
```
jakymiwm: Remove the "/ 10" and add it below?

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -168,16 +168,21 @@ class RAK12500LocationProvider : public LocationProvider {
...
+      _lng = ublox_GNSS.getLongitude(250) / 10;
```
jakymiwm: Remove the "/ 10" and add it below?

src/helpers/sensors/EnvironmentSensorManager.cpp
jakymiwm, line 988: Was 1000000 Now 10000000
This seems to be more of the standard.

jakymiwm, line 989: Was 1000000 Now 10000000
This seems to be more of the standard.

jakymiwm, line 998: Was 1000000 Now 10000000
This seems to be more of the standard.

jakymiwm, line 999: Was 1000000 Now 10000000
This seems to be more of the standard.

src/helpers/sensors/EnvironmentSensorManager.cpp
disq: Nope, this will diverge the code paths between other branches: It normalises ublox's native 1e-7 degree integer down to the implicit `LocationProvider` interface contract of microdegrees (1e-6).
Touching these has no real benefit other than complicating things, even if we changed it everywhere else. It won't be any faster or any better.

unex:
Running this on RAK10724 kit with RAK12501, and it fixed my GPS, thanks!

IoTThinks:
RAK4631 and RAK125001 work for me with the current code.
I will pick some changes to make RAK125000 work for RAK3401.

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -787,16 +832,25 @@ void EnvironmentSensorManager::rakGPSInit(){
...
-    return;
+  else {
+    // WB_IO2 controls the 3V3_S switched peripheral rail on these RAK boards.
+    // gpsIsAwake() leaves the pin as INPUT on failure, which drops the rail and
+    // takes I²C peripherals (RTC, display, the GPS itself) down with it.
+    // Restore it before probing other sockets.
+    pinMode(WB_IO2, OUTPUT);
+    digitalWrite(WB_IO2, HIGH);
+
+    if(gpsIsAwake(WB_IO4)){
```
IoTThinks, line 843: Why we need to check at WB_IO4 and IO5?
WB_IO2 is the only 3v3_S pin.

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -787,16 +832,25 @@ void EnvironmentSensorManager::rakGPSInit(){
...
-    gps_detected = false;
-    Serial1.end();
-    return;
+  else {
+    // WB_IO2 controls the 3V3_S switched peripheral rail on these RAK boards.
+    // gpsIsAwake() leaves the pin as INPUT on failure, which drops the rail and
+    // takes I²C peripherals (RTC, display, the GPS itself) down with it.
+    // Restore it before probing other sockets.
+    pinMode(WB_IO2, OUTPUT);
+    digitalWrite(WB_IO2, HIGH);
```
IoTThinks, line 841: In my test, we need a delay(500) after set HIGH.
Else RAK12501 (L76K) can not start on time.

variants/rak4631/platformio.ini
```diff
@@ -12,8 +12,7 @@ build_flags = ${nrf52_base.build_flags}
   -D NRF52_POWER_MANAGEMENT
   -D PIN_BOARD_SCL=14
   -D PIN_BOARD_SDA=13
-  -D PIN_GPS_TX=PIN_SERIAL1_RX
-  -D PIN_GPS_RX=PIN_SERIAL1_TX
+  -D FORCE_GPS_ALIVE
```
IoTThinks, line 15: Why we need to force gps alive for RAK4631?
We can safely turn it of if there is not sensor using sensor slots.

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -856,8 +910,23 @@ bool EnvironmentSensorManager::gpsIsAwake(uint8_t ioPin){
 void EnvironmentSensorManager::start_gps() {
   gps_active = true;
   #ifdef RAK_WISBLOCK_GPS
-    pinMode(gpsResetPin, OUTPUT);
-    digitalWrite(gpsResetPin, HIGH);
+    #if defined(PIN_GPS_EN) && (PIN_GPS_EN >= 0)
```
IoTThinks, line 913: PIN_GPS_EN is -1 for RAK3401 and RAK4601, so this always returns false.
I guess this should be an "if" instead of #if.
For RAK4631, PIN_GPS_EN is initialized as -1 and later gpsResetPin is set to WB_IO2.

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -872,8 +941,25 @@ void EnvironmentSensorManager::start_gps() {
 void EnvironmentSensorManager::stop_gps() {
   gps_active = false;
   #ifdef RAK_WISBLOCK_GPS
-    pinMode(gpsResetPin, OUTPUT);
-    digitalWrite(gpsResetPin, LOW);
+    #if defined(PIN_GPS_EN) && (PIN_GPS_EN >= 0)
```
IoTThinks, line 944: PIN_GPS_EN is -1 for RAK3401 and RAK4601, so this always returns false.
I guess this should be an "if" instead of #if.
For RAK4631, PIN_GPS_EN is initialized as -1 and later gpsResetPin is set to WB_IO2.

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -872,8 +941,25 @@ void EnvironmentSensorManager::start_gps() {
...
+      if (i2cGPSFlag) {
+        // UBX-CFG-RST "controlled GNSS stop": stops RF/tracking but keeps the host
+        // interface alive, so it restarts reliably over I2C. (PMREQ backup only wakes
+        // on EXTINT/UART, not I2C -- confirmed dead on this hardware.)
+        static const uint8_t ubxGnssStop[] = {0xB5,0x62,0x06,0x04,0x04,0x00,0x00,0x00,0x08,0x00,0x16,0x74};
+        Wire.beginTransmission((uint8_t)TELEM_RAK12500_ADDRESS);
+        Wire.write(ubxGnssStop, sizeof(ubxGnssStop));
+        Wire.endTransmission();
+      } else if (serialGPSFlag) {
+        Serial1.println("$PMTK161,0*28");     // Quectel/MTK standby
```
IoTThinks, line 960: I have tested this command. It has no effect on power consumption.

In Quectel forum, they said "The L76K does not support the PMTK command."
https://forums.quectel.com/t/l76k-cold-and-hot-start-for-low-power-periodic-location-acquisition/49439/3

IoTThinks (commented):
I guess the only hope to have GPS powersaving for RAK3401 is with I2C RAK12500.
Have you tested the power consumption of stop_gps() on RAK3401 with I2C RAK12500?

For RAK4631, if there is no sensor, we can just turn off RAK12501 using WB_IO2.
With RAK12500, we can hope your I2C command can lower power consumption as same as with RAK3401.

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -872,8 +941,25 @@ void EnvironmentSensorManager::start_gps() {
...
+      pinMode(gpsResetPin, OUTPUT);
+      digitalWrite(gpsResetPin, LOW);   // dedicated GPS enable pin
+    #else
+      // No dedicated enable pin: gpsResetPin is the shared 3V3_S rail (WB_IO2),
+      // which also powers the RTC/boost and must stay HIGH. Don't drop it; put
+      // the receiver into software low-power instead (rail/RTC stay up, warm-start).
+      if (i2cGPSFlag) {
+        // UBX-CFG-RST "controlled GNSS stop": stops RF/tracking but keeps the host
+        // interface alive, so it restarts reliably over I2C. (PMREQ backup only wakes
+        // on EXTINT/UART, not I2C -- confirmed dead on this hardware.)
```
IoTThinks, line 954: Could you please help to check if stop_gps() can reduce power consumption for RAK3401 and RAK 12500 I2C?

Issue #2645, opened 2026-05-29
[Feature Request] Mirror CLI to secondary UART to enable remote console access (e.g. WiFi bridge)


gagorski:
## Problem

When a repeater is deployed in the field (no USB access), there is no way to interact with
the serial CLI without physically retrieving the device. This makes it impossible to run
diagnostic commands (`log`, `stats-radio`, `neighbors`, etc.), change settings, or synchronize
the clock remotely.

## Proposed Solution

Add a `DualConsole` wrapper class that forwards CLI I/O to both the primary USB serial port
and a secondary hardware UART simultaneously. When a second UART is available, any device
connected to it (e.g. a WiFi bridge, BLE module, or another MCU) can relay CLI commands
over the network.

The key idea is that the firmware's CLI accepts a `Stream*` - passing a `DualConsole` instead
of `&Serial` is a non-breaking change that requires no modification to `CommonCLI` or any
other core code.

### Implementation sketch

```cpp
class DualConsole : public Stream {
    Stream &_a, &_b;
public:
    DualConsole(Stream &a, Stream &b) : _a(a), _b(b) {}

    // Input: read from whichever port has data
    int available() override { return _a.available() || _b.available(); }
    int read() override {
        if (_a.available()) return _a.read();
        if (_b.available()) return _b.read();
        return -1;
    }
    int peek() override {
        if (_a.available()) return _a.peek();
        if (_b.available()) return _b.peek();
        return -1;
    }

    // Output: write to both
    size_t write(uint8_t b) override { _a.write(b); _b.write(b); return 1; }
    void flush() override { _a.flush(); _b.flush(); }
};

```
In main.cpp, conditionally replace Serial with DualConsole:
```
#if defined(SERIAL_TX) && defined(SERIAL_RX)
    Serial1.begin(115200, SERIAL_8N1, SERIAL_RX, SERIAL_TX);
    static DualConsole dual_console(Serial, Serial1);
    // pass dual_console to CLI instead of Serial
#endif
The #if guard ensures the change only compiles on boards that define secondary UART pins.
```

Use Case
With a small ESP32 running MicroPython acting as a TCP-to-UART bridge:

```
 Remote operator → TCP socket → ESP32 bridge → UART (Serial1) → XIAO S3 CLI
```

This enables fully remote CLI access with no hardware changes to the repeater itself -
just adding a low-cost WiFi module wired to the secondary UART pins.

[...]

PR #2646, opened 2026-05-30
Fix: ESP32 KISS modem becomes permanently unresponsive under TX backpressure

agessaman:
### Problem
On ESP32-S3 boards (Heltec V4, Station G2), the KISS modem would stop responding after a while and stay dead—`SetHardware` PING (`0x17`) timed out and the condition **survived a host service restart**, forcing the pyMC_repeater client into no-radio mode. nRF52 (RAK4361) on identical firmware was unaffected.

### Root cause
The ESP32 modem ran over Arduino **TinyUSB `USBCDC`** (`ARDUINO_USB_MODE=0`), which wedges under TX backpressure: `USBCDC::write()` busy-spins indefinitely while "connected" (its `tx_timeout_ms` only guards the lock acquire, not the send loop), and RX events post to a 5-deep queue with `portMAX_DELAY`. Since `loop()` is single-threaded and the ESP32-S3 has no DTR/RTS→EN reset, nothing recovers it. nRF52 is immune (hardware UART, non-blocking FIFO writes).

Confirmed by a deterministic repro (flood the modem while never reading replies → permanent wedge) and a heartbeat LED that kept blinking while the data path was frozen, isolating the hang to the TinyUSB layer, below app code.

### Fix
Switch the ESP32 KISS envs to the ESP32-S3 **USB-Serial-JTAG peripheral (`HWCDC`, `ARDUINO_USB_MODE=1`)**, whose `write()` is bounded (`tries`/timeout countdown, not an infinite spin) and which posts RX from ISR.

- `variants/{heltec_v4,station_g2}/platformio.ini`: `build_unflags` the board default `=0`, set `=1` (a bare `-U` is reordered after the `-D` by SCons).
- `examples/kiss_modem/`: transport-neutral hardening—non-blocking writes that drop a frame instead of stalling `loop()` when the TX buffer is full, plus `setTxTimeoutMs()`. A dropped reply is harmless; the host retries.

### Validation
- Builds clean for `heltec_v4_kiss_modem`, `Station_G2_kiss_modem`, `RAK_4631_kiss_modem`.
- `nm`: `HWCDC` linked, zero TinyUSB symbols.
- On hardware (Heltec V4): the flood that permanently wedged TinyUSB now recovers in a retry or two, which the host's continuous-read + ping-retry handles.

### A Wrinkle
`ARDUINO_USB_MODE=1` changes the USB descriptor, so the `/dev/serial/by-id/` path becomes `usb-Espressif_USB_JTAG_serial_debug_unit_<MAC>-if00`.

### DIFF
--- a/examples/kiss_modem/KissModem.cpp
+++ b/examples/kiss_modem/KissModem.cpp
@@ -38,15 +38,15 @@ void KissModem::beginFrameWrite() {
 }
 
 void KissModem::rawWrite(uint8_t b) {
-  /* Once a frame has been aborted, swallow the rest of it so we never emit a
-     truncated KISS frame to the host. */
+  /* A frame is sent all-or-nothing; once we start dropping, swallow the rest so
+     we never emit a truncated KISS frame. */
   if (_tx_write_aborted) return;
 
-  /* Strictly non-blocking: if there is no room in the TX buffer, drop the rest
-     of this frame rather than wait. A stalled/absent USB-CDC host must never
-     stall loop() (any wait here starves the RX read loop and lets the TinyUSB
-     RX path overflow). Only write() when there is space, so write() can never
-     enter its internal busy-wait either. */
+  /* Non-blocking: if the TX buffer is full, drop this frame instead of waiting.
+     loop() is single-threaded and also services RX and the radio, so it must not
+     stall on a host that has stopped reading. Writing only when space is free
+     keeps the underlying write() from blocking; a dropped reply is harmless as
+     the host retries. */
   if (_serial.availableForWrite() <= 0) {
     _tx_write_aborted = true;
     return;
--- a/examples/kiss_modem/KissModem.cpp
+++ b/examples/kiss_modem/KissModem.cpp
@@ -22,7 +22,6 @@ KissModem::KissModem(Stream& serial, mesh::LocalIdentity& identity, mesh::RNG& r
   _getStatsCallback = nullptr;
   _config = {0, 0, 0, 0, 0};
   _signal_report_enabled = true;
-  _tx_write_aborted = false;
 }
 
 void KissModem::begin() {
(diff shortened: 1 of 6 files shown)


examples/kiss_modem/KissModem.cpp
```diff
@@ -32,35 +33,58 @@ void KissModem::begin() {
...
-    _serial.write(KISS_TFESC);
+    rawWrite(KISS_FESC);
+    rawWrite(KISS_TFESC);
   } else {
-    _serial.write(b);
+    rawWrite(b);
   }
 }
 
 void KissModem::writeFrame(uint8_t type, const uint8_t* data, uint16_t len) {
```
ripplebiz, line 81: Just wondering if a better approach is to pre-calc the total frame length, then introduce a new method like:
   bool  canWriteFrame(size_t len);

And for logic to be:
writeFrame( ... ) {
  size_t total_len = ... ;
  if (!canWriteFrame(total_len)) return;  // bail, all or nothing
  _serial.write( ... ); ...
}

ViezeVingertjes: I was thinking about the same, but looks like those buffers are quite small (like 64 bytes in some cases). Need some sleep first, but its an interesting one to look into.

disq:
Duplicate of #2645 (identical branch/commit, created twice due to a tooling hiccup). Closing in favor of #2645.

cwichura:
@disq The #2645 mentioned appears totally different.

examples/kiss_modem/KissModem.cpp
agessaman: @ViezeVingertjes any additional thoughts on this? Happy to make revisions following Scott's suggestion, but I wanted your feedback because the KISS firmware is your baby.

I can confirm that this change, plus a couple changes on the pyMC side have completely eliminated the lockup and unresponsive frame issues that we were facing (which is super exciting).

ViezeVingertjes: > [@ViezeVingertjes](https://github.com/ViezeVingertjes) any additional thoughts on this? Happy to make revisions following Scott's suggestion, but I wanted your feedback because the KISS firmware is your baby.
> 

I think to remember `availableForWrite()` is like only 64 or 128 bytes max depending on HWCDC/USBCDC, so now i wonder why it works haha! Does it drop all frames above it now making it look like it does because it does send smaller frames through?

Sorry have been busy, will prioritize looking at it further today. 💯 Most important is that we dont drop/lose frames without the proper feedback to the client, otherwise it wouldnt know to retry etc. (Although no response is also an indicator i guess)

ViezeVingertjes:
Quick check in between; did `ARDUINO_USB_MODE=1` by itself solve it and is the rest improving the handling? Because then it might be a quick merge with that change isolated, then seperate PR to improve it further.

agessaman:
It's a good question. `ARDUINO_USB_MODE=1` gets us pretty close. It is currently only applied to the ESP32-S3 variants for the Heltec V4 and the Station G2, because they displayed the wedging behavior when the client failed to read and the buffer filled.

The buffer size change exists because HWCDC's ring defaults to 256 bytes, and would drop full-size frames. The increase to 1024 bytes is a required part of the first change for normal operation.

With the first two changes, we'll avoid a permanent lockup to but we risk occasional bounded stalls should a host fail to read in a timely manner. The all-or-nothing writes are just hardening against backpressure in that case. If you want to defer this, we can. I can split into two PRs.

Update on the soak test: The radio has been running flawlessly with the combo of changes for several days.

ripplebiz:
I'm happy to merge this. Anything else pending for this one?

ViezeVingertjes:
> Update on the soak test: The radio has been running flawlessly with the combo of changes for several days.

Sounds good then! we can always revisit and improve it further at a later time (famous last words).

agessaman:
Nothing else pending from me on this one.

PR #2649, opened 2026-05-31
fix(rak12035): add missing returns, remove unused locals, and cleanup

LeoCatsune:
This commit seeks to address a number of issues in [`src/helpers/sensors/RAK12035_SoilMoisture.cpp`](https://github.com/meshcore-dev/MeshCore/blob/d2c2a6ef0b6b6e7323867c45cecbdae66179759a/src/helpers/sensors/RAK12035_SoilMoisture.cpp), namely:
- Undefined Behaviour: Missing return in `RAK12035_SoilMoisture::sensor_on()` (return type `bool`)
- Numerous unused local variables
- Redundant or unnecessary function calls in non-debug cases
- Miscellaneous spelling, formatting, indentation, and whitespace issues

The missing return caused downstream projects depending on MeshCore to fail compilation under `-Werror=return-type`, which subsequently led to discovery of the other issues addressed here.

One notable change affects initialisation in `EnvironmentSensorManager.cpp`. `RAK12035_SoilMoisture::begin()` was observed to ignore its `addr` parameter and instead rely on the address supplied via the constructor. To make this behaviour explicit and consistent with existing patterns, the initialisation flow was updated to match how `INA226` is handled in the same file. No change in effective addressing behaviour is intended.

Please note that I do not have access to the relevant hardware and therefore am unable to fully validate runtime behaviour. I would appreciate any input from those who depend on this module.

CC: @KPrivitt


LeoCatsune:
Just double-checked, wio build fails on `dev` before this, no change here.

sefinek:
> wio build fails

Fixed in #2675.

PR #2650, opened 2026-05-31
fix: update wio-e5-mini pio config to address SubGhz dependency issue

LeoCatsune:
Title is self-explanatory, should fix an issue with the `wio-e5-mini` variants failing to build due to a failure in resolving the `SubGhz` library of `framework-arduinostm32` - builds no longer fail locally, though may require testing on hardware to confirm functionality.


LeoCatsune:
This should also resolve the workflow failures observed in #2649.

PR #2651, opened 2026-05-31
Early receives: allow nodes to process most incoming direct packets ASAP

redxdev:
Right now a node can only process a direct packet if it is receiving the final hop (i.e. the entire path has been consumed), but this seems like it is unnecessary for payload types that are easy to check destination hashes on.

For example, consider a packet with path A,B,C with a destination of D. If node D can hear A directly, there's no reason for it to wait for C to repeat the packet before replying.

There was already an exception for ACKs that worked this way, so this change extends that to a few other payload types that all have easy access to the destination hash.

### DIFF
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -73,16 +73,6 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
   }
 
   if (pkt->isRouteDirect() && pkt->getPathHashCount() > 0) {
-    // check for 'early received' ACK
-    if (pkt->getPayloadType() == PAYLOAD_TYPE_ACK) {
-      int i = 0;
-      uint32_t ack_crc;
-      memcpy(&ack_crc, &pkt->payload[i], 4); i += 4;
-      if (i <= pkt->payload_len) {
-        onAckRecv(pkt, ack_crc);
-      }
-    }
-
     if (self_id.isHashMatch(pkt->path, pkt->getPathHashSize()) && allowPacketForward(pkt)) {
       if (pkt->getPayloadType() == PAYLOAD_TYPE_MULTIPART) {
         return forwardMultipartDirect(pkt);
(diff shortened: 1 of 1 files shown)


src/Mesh.cpp
```diff
@@ -73,16 +73,6 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
...
-    // check for 'early received' ACK
-    if (pkt->getPayloadType() == PAYLOAD_TYPE_ACK) {
-      int i = 0;
-      uint32_t ack_crc;
-      memcpy(&ack_crc, &pkt->payload[i], 4); i += 4;
-      if (i <= pkt->payload_len) {
-        onAckRecv(pkt, ack_crc);
-      }
-    }
-
```
redxdev, line 85: Just noting that this is a slight behavior change, and opens up one weird edge case: if a packet has the local node as part of the path *and* as the destination (i.e. the full path is something like A,B,C with a destination of B) then the packet won't be received early as retransmit logic takes priority here. I could restructure the code to avoid this situation, but it's a bit nonsensical in the first place and I'm not sure it's worth fixing.

redxdev:
Also noting a potential downside of this change: this will result in a node processing more packets in total, since it will now process any direct packet with a destination that has the same 1-byte hash as itself. Previously this would be somewhat protected against since it would only process a packet with a colliding destination hash if it could hear the final hop in the path.

That said, I don't think this is a big deal - flood messages end up having to be processed the same way already where a node can't tell if it is intended as the destination until it attempts to decrypt the payload, so this seems like it should be fine.

shortwavesurfer2009:
Mesh is moving to two and three byte paths, theoretically would reduce this even further, correct?

redxdev:
@shortwavesurfer2009 correct, larger hash sizes will be even less likely to be processed when you aren't the intended recipient.

I still don't think it'll be an issue on 1-byte, though. It shouldn't happen that often, and when it does it is at worst wasting a bit of CPU time failing to decrypt a packet.

PR #2652, opened 2026-05-31
RAK3401: reliable AIN1 button — interrupt latch, debounce, digital read, instant screen-on 🤖🤖

disq:
## Problem

The RAK3401 companion build has a single user button on **AIN1** (P0.31), driven by `analogRead()` polled once per main-loop iteration — there is no interrupt serving it. Symptoms:

- Presses **dropped** when the loop briefly stalls (LoRa RX/TX, BLE, flash writes).
- Single clicks feel laggy: the screen only wakes after the multi-click window expires.
- The button can't wake the device from **hibernate** (SYSTEMOFF) — and a naive GPIO SENSE wake makes it wake instantly and never stay off.

This PR makes the AIN1 button story self-contained: reliable input **and** a working hibernate wake. It folds in #2642's RAK3401 board pieces (so that PR can be closed in favor of this one).

## Changes

**`MomentaryButton` (generic)**
- Configurable multi-click window on the analog constructor (default unchanged).
- `reset()` — abandon the in-flight gesture.
- `enableInterrupt()` — permanent FALLING-edge GPIOTE interrupt that **latches presses** so they survive loop stalls. Debounced: one count per press, re-armed only after `check()` sees a clean release (rejects press/release contact bounce). Skipped for analog/SAADC-threshold buttons (GPIOTE vs `analogRead` conflict on the same pin).

**companion_radio `UITask`**
- Wake the display **instantly on the press edge** when it's off, instead of after the multi-click window.
- Arm the button interrupt at startup for `PIN_USER_BTN_ANA` boards.

**`variants/rak3401` — input**
- Read AIN1 **digitally** (active-low, internal pull-up) instead of via SAADC — `digitalRead` coexists with the GPIOTE edge interrupt where `analogRead` does not (analogRead otherwise reads stuck "always pressed" once the interrupt is attached). Multi-click retained: single=NEXT / double=PREV / triple=SELECT / long=ENTER.

**Hibernate wake via LPCOMP** (2nd commit; supersedes #2642's GPIO-SENSE approach)
- `NRF52Board::configureVoltageWake()` gains a `detect_down` param (default `false` = UP crossing for voltage recovery; `true` = DOWN crossing for a button press), selecting `ANADETECT` Down/Up and `INTENSET` DOWN/UP accordingly.
- `RAK3401Board.h`: `powerOff()` override routing through `initiateShutdown(USER)`.
- `RAK3401Board::initiateShutdown()`: when `PIN_USER_BTN_ANA` is defined, arm **LPCOMP** on AIN7 at REFSEL 3 (~½ VDD) for a DOWN crossing as the SYSTEMOFF wake source. Waits for a confirmed release (analogRead above threshold) before arming, bounded by a 5 s timeout. Channel/threshold overridable via `PWRMGT_BTN_LPCOMP_AIN` / `PWRMGT_BTN_LPCOMP_REFSEL`.

### Why LPCOMP, not GPIO SENSE

AIN1 is wired as an analog button (`pressed == analogRead() < threshold`). Its **digital** input buffer reads the *released* idle level as LOW even though `analogRead` reports ~VDD, so arming `SENSE_Low` latches DETECT the instant we enter SYSTEMOFF — `sd_power_system_off()` returns immediately and the fall-through reset reboots ("instant wake"). `LATCH`/`EVENTS_PORT` clearing and release debounce don't help, because the released level itself reads LOW digitally. LPCOMP operates in the analog domain and sees the idle level correctly, so a DOWN-crossing arm works: released idles above threshold, a press pulls below → wake.

## Relationship to #2642

This **supersedes** #2642 — it ports #2642's `powerOff()` override and AIN SYSTEMOFF wake, but replaces the GPIO-SENSE wake (which instant-wakes on this pin) with the LPCOMP approach. #2642 can be closed once this lands.

## Testing

- Builds clean: `RAK_3401_companion_radio_ble`, `RAK_3401_companion_radio_usb`.
- Button input bench-tested on RAK3401 `companion_radio_ble` hardware: digital read fixes the SAADC+GPIOTE "always pressed" conflict; presses reliable through loop stalls; single/double/triple/long all work, no dropped or double-counted clicks.
- Hibernate: GPIO-SENSE instant-wakes on this pin (confirmed on hardware); the LPCOMP DOWN-crossing approach is the fix. Please re-flash/re-test the hibernate path before merge.

## Known edge case

If the loop stalls through the middle of a *fast double-click* so `check()` never re-arms between the two presses, the second press is dropped (fails toward "one click", never a phantom extra). Normal clicking is unaffected.

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -564,6 +564,9 @@ void UITask::begin(DisplayDriver* display, SensorManager* sensors, NodePrefs* no
 #endif
 #if defined(PIN_USER_BTN_ANA)
   analog_btn.begin();
+  // Arm a permanent edge interrupt so presses are never lost when the main loop stalls
+  // (radio/BLE/flash) and the poll misses the edge. This also wakes the MCU from light sleep.
+  analog_btn.enableInterrupt();
 #endif
 
   _node_prefs = node_prefs;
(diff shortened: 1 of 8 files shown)


disq:
Probably not needed.

PR #2658, opened 2026-06-01
xiao_s3_wio: Allow connection of the L76 GPS module using its RX,TX and power pins.

m0rjc:
Add in support for GPS on this board.
Use the !RESET/ENABLE pin (D2) as the ENABLE pin in code. This is not wired into my circuit. (I will run a wire from D2 to the ENABLE pin on my module. They are side by side on veroboard in my buils so easy enough.)

### DIFF
--- a/variants/xiao_s3_wio/platformio.ini
+++ b/variants/xiao_s3_wio/platformio.ini
@@ -6,7 +6,10 @@ board_build.mcu = esp32s3
 build_flags = ${esp32_base.build_flags}
   ${sensor_base.build_flags}
   -I variants/xiao_s3_wio
-  -UENV_INCLUDE_GPS
+  -D ENV_INCLUDE_GPS=1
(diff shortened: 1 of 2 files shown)


m0rjc:
This is very close to the pioggyback setup of these modules, with the L76K plugged into the back of the ESP32 board. My LoRa board is plugged into the piggyback socket on top. The only problem is the ENABLE pin, which should be D0 if the boards are plugged this way. This apparently conflicts.

Suggestion I found online was to stack the boards having cut the EN and !RESET pins. I've laid them out on veroboard wihch is easier and less damaging to the modules.

m0rjc:
I'd also like to get my nRF52 board working. At the moment that's my Companion Node so fine without GPS. Whether to add it here, or separate, or drop this pull request and keep my fork for my own boards?

PR #2660, opened 2026-06-01
fix(#2659): Repeater neighbours not populated correctly

### LINKED ISSUE #2659: Repeater neighbours not populated for directly received flood adverts where source path.hash.mode > 0
**Describe the bug**

A repeater that hears another repeater's **flood** advertisement directly (zero hops) does not add it to the neighbours list when that source repeater's path.hash.mode is set to 1 or 2. Setting path.hash.mode back to 0 restores correct behaviour. 

Originally reported by **Dan VK2MR**: _something very strange going on with repeaters repeating adverts after hearing them direct, but not adding the repeater to the neighbours list_

**To reproduce**

1. Configure two repeaters with set path.hash.mode 1 (or 2).
2. Place them within direct radio range.
3. Initiate a flood advert on either repeater.
4. Query neighbours on the other repeater; the source repeater does not appear.
5. Set path.hash.mode 0 on the source repeater and repeat the advert; neighbour is populated correctly.

**Root cause**

path_len encodes both the hash size (upper 2 bits) and the hop count (lower 6 bits) via setPathHashSizeAndCount. When sendFlood emits an advert with zero accumulated hops it calls:
```
packet->setPathHashSizeAndCount(path_hash_size, 0);
// mode 0 → path_len = 0x00
// mode 1 → path_len = 0x40
// mode 2 → path_len = 0x80
```

The neighbour check compares the raw field against zero:
```
if (packet->path_len == 0 && !isShare(packet)) {   // only true for mode 0
    putNeighbour(id, timestamp, packet->getSNR());
}
```

With mode 1 or 2, path_len is 0x40 or 0x80 on a zero-hop packet, so the condition is never satisfied and putNeighbour is never called.

**Fix**

Replace the raw path_len comparison with getPathHashCount(), which masks off the hash-size bits and returns only the hop count:
```
if (packet->getPathHashCount() == 0 && !isShare(packet)) {
    putNeighbour(id, timestamp, packet->getSNR());
}
```

Affected file: examples/simple_repeater/MyMesh.cpp:638

Irate-Architect:
Fixes #2659

Tested on RAK 4630s.

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -635,7 +635,7 @@ void MyMesh::onAdvertRecv(mesh::Packet *packet, const mesh::Identity &id, uint32
   mesh::Mesh::onAdvertRecv(packet, id, timestamp, app_data, app_data_len); // chain to super impl
 
   // if this a zero hop advert (and not via 'Share'), add it to neighbours
-  if (packet->path_len == 0 && !isShare(packet)) {
+  if (packet->getPathHashCount() == 0 && !isShare(packet)) {
     AdvertDataParser parser(app_data, app_data_len);
     if (parser.isValid() && parser.getType() == ADV_TYPE_REPEATER) { // just keep neigbouring Repeaters
       putNeighbour(id, timestamp, packet->getSNR());


Irate-Architect:
Fixed in dev branch. Closing PR.

Issue #2555, opened 2026-05-14
[Feature Requests] Region UX Improvements: add metadata to zero hop adverts, allow different flood.max for unscoped messages


mrzarquon:
Regions, when implemented, work pretty well for reducing the congestion on the Mesh. However the adoption of regions is hampered by the fear that new users will somehow turn on their radios to a silent mesh and assume nothing is there and not continue any further.

To address these concerns, and to make it easier in general for new users to onboard, the following features could be added:

1) Zero-Hop Adverts from repeaters include a region (and byte size). This could either be the configured Home, Default, or Other region (either tbd or let users pick: `set zerohop.broadcast.region Home, Default, existing region`). This information can then be used in multiple ways:
    - Companions & Apps can include a "detect regions" section that will show what regions are available to the user, based on the heard zerohop adverts
    - Repeater Discovery requests will return the region information when users request it
    - Repeater Admins can view the regions of their neighbors - even if they aren't supporting flood on * (or that neighbors default region)

2) Along with `set flood.max`, there should be a new setting for `set flood.max.unscoped` that will allow Repeater administrators to set a different hop limit on for messages that are unscoped. New repeater admins can get their messages heard and repeated across their local mesh, but not carried further than that. Allowing a new user who has setup a new repeater to have their messages relayed by neighboring established repeaters. In observation of rolling out repeaters with regions here on the Island of Ireland, this is actually rare that a new user can _only_ user a repeater to access the mesh - maybe reliably, but they usually get some signal on their companion first or as part of determining where to setup a repeater. If they aren't immediately within ZeroHop of an established MeshCore repeater, they are 1-2 hops away from one, so `set flood.max.unscoped 2` would work well here. Others may need `set flood.max.unscoped 5` etc.

Instead of trying to prescript region names as the solution, these improvements would make using the existing method of letting users discover and coordinate what regions are in place organicly as they are engaging with the mesh. There's not presumed "correct" region name for an area other than what the people who are building the mesh in the region decide it to be. The RF waves don't conform to UNLOC region codes, human social groups definitely don't, so focusing on improving the UX of learning what regions are being scoped around the user should be the focus and would likely improve the traffic congestion situation as it is less daunting or "drastic" to disable flood on * since now:
1) New users can find out the regions that are in use around them through their companion app and their own repeaters neighbors discovery list. App improvements can make this easier and part of a new user setup of just pinging them when the first time a region is detected when they open the app (not notifying outside the app)
[...]

PR #2661, opened 2026-06-01
Implement flood.max.unscoped to allow selective repeating of unscoped messages

mrzarquon:
As mentioned in https://github.com/meshcore-dev/MeshCore/issues/2555 this, having flood.max.unscoped as an alternative to `region denyf *` would address most peoples primary fear of blocking all unscoped messages - mostly that nearby users new to the mesh won't be able to contact or communicate with _anyone_ until they have the correct regions set.

This fix implements `flood.max.unscoped` as an additional setting, with the following design considerations:
- If unset, default value is 0xFF, and the check falls back to the flood.max value (and reports the flood.max.unscoped value as the same as the flood.max value - because it is handled the same way).
- Once set, packets with ROUTE\_TYPE\_FLOOD will be checked against this value, while all other packets will be evaluated against the max.flood value. There is currently no way to _unset_ this value, so users will have to change flood.max.unscoped independently of flood.max in the future.

Preliminary testing has been done with repeater firmware, and have verified that both unscoped of <= flood.max.unscoped have passed through, while > flood.max.unscoped were dropped. Scoped messages continued to pass without issue at the same time.

Compiled firmware, based off dev snapshot from the afternoon of 2026-06-01, is available here: https://static.loraproject.org/unscoped/

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -619,6 +619,21 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 
 ---
 
+#### Limit the number of hops for an unscoped flood message
+**Usage:**
+- `get flood.max.unscoped`
+- `set flood.max.unscoped <value>`
+
+**Parameters:**
+- `value`: Maximum flood hop count (0-64) for a packet without a scope (no region set)
+
+**Default:** `0xFF` - indicates it hasn't been set, will track flood.max until it is.
+
+**Note:** An alternative to `region denyf *`, setting `flood.max.unscoped` to a lower value such as `3` would allow for local unscoped messages to propagate, while preventing noisy neighbors from flooding a local region.
+
+
+---
+
 ### ACL
 
 #### Add, update or remove permissions for a companion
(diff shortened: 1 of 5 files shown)


ripplebiz:
Thanks for this. Will merge, but will do a few simplifications.

formtapez:
It would be good if this feature could be limited to group messages and adverts only.
So pathless DMs can still go through.

That should be easy to implement, like here:
https://github.com/meshcore-dev/MeshCore/pull/1810/changes

formtapez:
The danger of the actual implementation of this PR is:
People are turning this on, because it sounds good - But with that, they break DM-functionality on their repeaters.
Direct messages won't find a path to its destination anymore...

@ripplebiz

mrzarquon:
@formtapez counter is this is an alternative to people blanket deploying `region denyf *` so with this change as is, people using devices that haven't set a default region can still send DMs within the radius of the flood.max.unscoped number of hops, which is better than not being able to send any at all.

PR #2663, opened 2026-06-02
Disable auto-shutdown when externally powered, add auto-shutdown warning for OLED displays

oltaco:
This PR skips low battery auto-shutdown if isExternalPowered() returns true, and also introduces a 3 second warning for OLED displays before auto-shutdown.

Currently only works for NRF52 boards, ESP32 would probably need board-level overrides I think.


ripplebiz:
Just wondering if, instead of HAS_EINK_DISPLAY, add a new DisplayDriver virtual bool isEInk() ??

oltaco:
@ripplebiz good call, much tidier and only new e-ink display drivers need to remember to implement isEink()

I've made the changes, tested on T-Echo Card and ThinkNode M1, should be good to go.

liamcottle (approved):

Issue #2668, opened 2026-06-02
[Fix] Heltec t096 RTC option


dl1rf:
Hi dev team,

I have added the possible RTC PCF8563 option to a Heltec T096 board.
The RTC is not detected by the actual repeater firmware.
I changed the I2C bus 0 pin definitions from:

**#define PIN_WIRE_SDA (0 + 7)**
**#define PIN_WIRE_SCL (0 + 8)**

to

**#define PIN_WIRE_SDA (32 + 4) // P1.04 RTC_SDA**
**#define PIN_WIRE_SCL (32 + 2) // P1.02 RTC_SCL**

Now RTC is detected and work perfect with one additional small change to AutoDiscoverRTCClock.cpp.

May I offer a pull request to you?

With Best Regards

Martin, DL1RF

PR #2669, opened 2026-06-02
Issue #2668 [Fix] Heltec t096 RTC option

dl1rf:
Fixes to support RTC PCF8563 option with Heltec T096 board.
Only 1 I2C bus enabled to reduce unused resources.

### DIFF
--- a/src/helpers/AutoDiscoverRTCClock.cpp
+++ b/src/helpers/AutoDiscoverRTCClock.cpp
@@ -43,6 +43,13 @@ void AutoDiscoverRTCClock::begin(TwoWire& wire) {
 
   if (i2c_probe(wire, PCF8563_ADDRESS)) {
     rtc_8563_success = rtc_8563.begin(&wire);
+    if (rtc_8563_success) {
+      if (rtc_8563.lostPower()) {
+        // only set time if RTC lost power, otherwise leave it alone
+        // (e.g. if battery is dead but still keeping time, or if user has set to a specific time)
+        rtc_8563.adjust(DateTime(2024, 5, 15, 20, 50, 0)); // set date and time as with CMD clkreboot: (15 May 2024, 8:50pm)
+      }
+    }
   }
 
   if (i2c_probe(wire, RX8130CE_ADDRESS)) {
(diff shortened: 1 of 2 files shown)


Aliasmk:
Wouldn't this break systems that use I2C devices on pin 7 & 8?

dl1rf:
The fix only target Heltec T096 device. No other devices.
The Heltec T096 has no I2C devices connected to pin 7 & 8. (Reference: schematics of Heltec T096).
External user connected I2C devices have to switch to RTC_SDA (header P2 pin 18) and RTC_SCL (header P2 pin 17) lines.
These lines are documented in the Heltec T096 documentation (schematics).

dl1rf:
Benefit of the provided fix is that the already fully supported RTC PCF8563 can be placed to the yellow marked space at the T096 board.
IMG_4775

IoTThinks:
@dl1rf The yellow space is for GPIO 17 and 18?

dl1rf:
> @dl1rf The yellow space is for GPIO 17 and 18?

I don't understand the question. What is it aiming at?

IoTThinks:
I have a T096 too.
I would like to know what is the yellow area about.

Thanks a lot.

dl1rf:
As described with the foto it is the place to solder an optional Real Time Clock chip with the crystal, tuning capacitors, a supercap capacitor and an isolation diode.
I solder the RTC option to some T096 boards used as repeaters.
And after changing the I2C lines as offered here this work perfect.

Regards

Martin

Aliasmk:
Understood now - the i2c pins you propose are both connected on the header AND internally on the RTC pads. So you are proposing the code changes to use the latter. Users that want to have external sensors can still attach them to the header, and if anyone wants to solder the RTC directly to the board, it would work right away.

My only issue is that this will break currently deployed projects that use pins 7 & 8. For instance, I have a solar node with an INA219 and an AHT20, soldered to pin 7 and 8. If I were to update Meshcore, those now no longer work.

I'm not familiar enough with the project yet to comment on how things like this should be resolved, I'm just pointing out a potential breaking issue.

dl1rf:
I see. Yes this will be a problem and shall be avoided.
How is it to change the I2C pins of the second enabled I2C bus 1 ?
Is this bus used by actual firmware and hardware with T096 board?
I did not see any use yet.
I think it will be possible to change the RTC code with conditional compile so that the RTC code use second I2C bus 1 instead of I2C bus 0 with T096 board. And only with this board.
This may overcome the actual problem and I2C bus 0 pins will stay at 7 & 8.

Any concerns or suggestions?

PR #2673, opened 2026-06-03
fix: clear hasPendingWork queue before shutdown on multiple node types to prevent corruption

NickDunklee:
This PR is a cleanup and is in ref to my comments in https://github.com/meshcore-dev/MeshCore/pull/2627

**Update:** This clears pending work before shutdown to try and prevent corruption during power off. Which was different from another PR mentioned in comments.

There have been several updates around messaging, shutdown, etc. the past couple of months, so refactoring this to be current and cleaning it up where I can.

@ripplebiz had some recommendations for existing code flows, so I refactored to leverage that. It compiles against a few sensor/repeater/companion node types.

### New cases covered for saveContacts/hasPendingWork in this PR:

* COMPANION: button-triggered deferred shutdown
* COMPANION: serial CLI reboot
* REPEATER: button-triggered on display-equipped boards

#### Cases already covered before this PR:

* COMPANION: nrf52 idle sleep
* COMPANION: CMD_REBOOT via BLE/USB app command
* COMPANION: factory reset - NOT covered since FS wipe
* REPEATER: SenseCAP solar button
* REPEATER: idle sleep


disq:
PR title doesn't fully reflect the changes. Also some of it is similar to https://github.com/meshcore-dev/MeshCore/pull/2663

NickDunklee:
Updating here for read consistency - that other PR disables auto-shutdown on powered nodes. This PR makes sure contacts are synced before the node powers off to prevent corruption.

Thanks @disq ! 🤜 🤛

NickDunklee:
Updated to take into account recent dev changes, removed some code that would never be executed, and cleaned up messaging.

NickDunklee:
Update: #2663 was about disabling power off when on mains power. This PR is about a different failure mode: making sure a manual/physical shutdown doesn't cut power while a contact-list write is still pending in the 5s lazy-write window (`dirty_contacts_expiry`), which can corrupt the contact list.

NickDunklee:
Updated and took into account work done by others since this PR was first opened to simplify/clean up.

Issue #2686, opened 2026-06-05
Add room server system post helper for local events


Che177:
I would like to propose a small room server enhancement: provide a clean helper method that allows the room server to create a room post as the room server itself.

The proposed helper is:

```cpp
MyMesh::addSystemPost(const char* postData)
```

This gives `simple_room_server` a clean local/server-side posting path without pretending the message came from a connected client and without duplicating the existing room post storage/push logic.

The helper keeps system/local posts on the existing room post path:

* posts use the room server’s own identity
* post storage is handled through the existing `storePost()` logic
* post indexing, timestamping, and client sync behavior remain consistent with normal room posts

As part of this change, I also exposed the helper through a console command:

```text
room.post <message>
```

This makes the feature easy to test manually from the serial console.

I also created a small disabled-by-default GPIO contact demo to show the specific use case I was working on:

* equipment box opened
* dry contact closes
* room server posts a message into the chat

For example:

```text
Equipment box opened
```

This could also be useful for similar simple local status events:

* tamper switch
* gate/contact switch
* equipment enclosure opened

Known limitation: this follows the current room server timestamp behavior. The room server clock needs to be synchronized before local/system posts are created. Otherwise, the post may be timestamped in the past and may not appear as expected in the client room view.

I published my current implementation here for review: https://github.com/Che177/MeshCore/tree/feature/room-server-system-posts

In that branch, the GPIO contact example and debug flags are disabled by default / commented out in the PlatformIO configuration.

The commented PlatformIO example is currently included in the LilyGo T-LoRa V2.1-1.6 room server environment:

variants/lilygo_tlora_v2_1/platformio.ini

[env:LilyGo_TLora_V2_1_1_6_room_server]

The GPIO part is only intended as a small demo/use-case implementation, not as a general sensor framework, protocol change, or app change.

Che177:
I opened a draft PR with the current proposed implementation here: https://github.com/meshcore-dev/MeshCore/pull/2688

Che177:
Rebased on current dev and cleaned the branch so the PR now only contains the room_server-related changes.

komacke:
This would be a helpful feature to share announcements related to the room server.

Che177:
Update: PR #2688 has been rebased onto the latest `dev` branch and narrowed to the core `room.post` functionality only.

The GPIO demo, debug flags, and board-specific configuration have all been removed from the PR. It now only:

- adds `MyMesh::addSystemPost()`
[...]

PR #2688, opened 2026-06-05
room_server: add room.post command for server-originated posts

Che177:
Based on #2686.

## Summary

This PR adds support for room-server-originated posts to `simple_room_server`.

It introduces:

```cpp
MyMesh::addSystemPost(const char* postData)
```

and exposes it through the serial console command:

```text
room.post <message>
```

## Motivation

The room server currently stores posts received from clients, but it has no simple API for creating a post using its own identity.

This change provides a reusable path for local room-server code to publish messages without duplicating the existing post storage and notification logic.

## Implementation

This PR:

- adds `MyMesh::addSystemPost(const char* postData)`
- stores system posts using the room server's own identity
- refactors the existing post logic into a shared `storePost()` method
- preserves the existing timestamp, cyclic queue, push scheduling, and synchronization behavior
- adds the `room.post <message>` serial console command
- rejects empty messages with `ERR empty message`
- introduces no protocol changes or new dependencies

## Usage

From the room server's serial console:

```text
room.post Equipment box opened
```

Successful commands return:

```text
OK
```

## Testing

Built successfully with:

```text
pio run -e LilyGo_TLora_V2_1_1_6_room_server
```

## Scope

This PR only adds the system-post helper and the `room.post` command. It does not include GPIO handling, sensor examples, or board-specific configuration.

### DIFF
--- a/examples/simple_room_server/MyMesh.cpp
+++ b/examples/simple_room_server/MyMesh.cpp
@@ -45,9 +45,7 @@ void MyMesh::addPost(ClientInfo *client, const char *postData) {
 void MyMesh::addSystemPost(const char *postData) {
   if (!postData || postData[0] == 0) return;
 
-#if defined(ENABLE_ROOM_POST_DEBUG) && ENABLE_ROOM_POST_DEBUG == 1
-  Serial.print("room.post: addSystemPost: "); Serial.println(postData);
-#endif
+  MESH_DEBUG_PRINTLN("room.post: addSystemPost: %s", postData);
 
   storePost(self_id, postData);
 }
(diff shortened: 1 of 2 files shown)


examples/simple_room_server/MyMesh.cpp
```diff
@@ -39,18 +39,43 @@ struct ServerStats {
 };
 
 void MyMesh::addPost(ClientInfo *client, const char *postData) {
+  storePost(client->id, postData);
+}
+
+void MyMesh::addSystemPost(const char *postData) {
+  if (!postData || postData[0] == 0) return;
+
+#if defined(ENABLE_ROOM_POST_DEBUG) && ENABLE_ROOM_POST_DEBUG == 1
```
liamcottle: Howdy! These should probably just use the existing `MESH_DEBUG_PRINTLN` instead of adding a new `ENABLE_ROOM_POST_DEBUG` and using `Serial.print`.

examples/simple_room_server/MyMesh.cpp
```diff
@@ -39,18 +39,43 @@ struct ServerStats {
...
+  int idx = next_post_idx;
   // TODO: suggested postData format: <title>/<descrption>
-  posts[next_post_idx].author = client->id; // add to cyclic queue
-  StrHelper::strncpy(posts[next_post_idx].text, postData, MAX_POST_TEXT_LEN);
+  posts[idx].author = author; // add to cyclic queue
+  StrHelper::strncpy(posts[idx].text, postData, MAX_POST_TEXT_LEN);
 
-  posts[next_post_idx].post_timestamp = getRTCClock()->getCurrentTimeUnique();
+  posts[idx].post_timestamp = getRTCClock()->getCurrentTimeUnique();
+#if defined(ENABLE_ROOM_POST_DEBUG) && ENABLE_ROOM_POST_DEBUG == 1
```
liamcottle: Same with these

liamcottle (changes requested):

Che177:
Updated in 5a04e060. I removed the dedicated ENABLE_ROOM_POST_DEBUG guards and replaced all room.post Serial logging with the existing MESH_DEBUG_PRINTLN macro. I also verified the Tbeam_SX1262_room_server build both normally and with MESH_DEBUG enabled. Thanks for the review!

PR #2689, opened 2026-06-05
HeltecV4: Fix battery ADC reading accuracy

TPRoberts:
Correct `ADC_MULTIPLIER` from `5.42` to `4.9` to match the `R27/R28` resistor divider on the schematic `(390kΩ + 100kΩ) / 100kΩ = 4.9`.

Also switch from `analogRead()` with manual `3.3V/1024` conversion to `analogReadMilliVolts()` at 12-bit resolution, which uses the ESP32's built-in ADC calibration for improved accuracy.

Schematic reference: [https://resource.heltec.cn/download/WiFi_LoRa_32_V4](https://resource.heltec.cn/download/WiFi_LoRa_32_V4)

### DIFF
--- a/variants/heltec_v4/HeltecV4Board.cpp
+++ b/variants/heltec_v4/HeltecV4Board.cpp
@@ -62,18 +62,18 @@ void HeltecV4Board::begin() {
   }
 
   uint16_t HeltecV4Board::getBattMilliVolts()  {
-    analogReadResolution(10);
+    analogReadResolution(12);
     digitalWrite(PIN_ADC_CTRL, HIGH);
     delay(10);
-    uint32_t raw = 0;
+    uint32_t raw_mv = 0;
     for (int i = 0; i < 8; i++) {
-      raw += analogRead(PIN_VBAT_READ);
+      raw_mv += analogReadMilliVolts(PIN_VBAT_READ);
     }
-    raw = raw / 8;
+    raw_mv /= 8;
 
     digitalWrite(PIN_ADC_CTRL, LOW);
 
-    return (adc_mult * (3.3 / 1024.0) * raw) * 1000;
+    return adc_mult * raw_mv;
   }
 
   const char* HeltecV4Board::getManufacturerName() const {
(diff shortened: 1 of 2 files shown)


fizzyfuzzle:
Thanks for this PR, I also coincidental changed this in my local branch, so I can confirm it works and is more accurate then the current code :+1: 

One question: do we also need to change the Attenuation to a smaller range to have an even more accurate reading? Because I think the default is: ADC_ATTEN_DB_11 [0 mV ~ 3100 mV], but because of the voltage divider we will never measure above the 1000mv.
https://docs.espressif.com/projects/esp-idf/en/v4.4/esp32s3/api-reference/peripherals/adc.html

e.g.:
`analogSetAttenuation(ADC_2_5db);`

TPRoberts:
Hi @fizzyfuzzle

Thanks for testing it. I'm actually still testing mine, was going to wait for full drain and see how it performs before I move it from draft.

You raise a good point though. I think the default attenuator is 11dB but it could go as low as 0dB as the standard Lithium Polymer should only go from 615mV to 857mV at the ADC pin with that divider, so well within the 0dB range. Leaving it at the 11dB default would waste ADC counts and introduce more nonlinearity. I think 0dB gives the best linearity and resolution the ESP32-S3 ADC can offer, what do you think?

TPRoberts:
Actually, you're suggestion if probably better @fizzyfuzzle , 0 dB might not give enough room on the top range for chip to chip variations, 2.5 dB would probably give better compatibility between users.

Change is made, I  moved it from draft for now but I am still going through a full drain.

TPRoberts:
Mine reports 93%/ 4.12V when fully charged. I didn't take into account the charging IC will cut off the charging. 

I personally feel this PR is still improves the ADC reading for the Heltec v4 and should anyone need finer adjustments they should redefine `ADC_MULTIPLIER` for their requirements. It could be possible to adjust it in `meshcore-cli` or the app should they add support for setting `adc.multiplier` in the future.

ErikBrown2:
I hope that this PR will be approved. Currently the battery voltage monitoring is done on the Heltec v4 in steps of about 17 millivolt which is rather course.

Issue #2690, opened 2026-06-05
Allow for security vulnerability reporting


NoodlesNZ:
In order to keep security vulnerabilities from being reported via issues and PRs, it's advisable to enable vulnerability reporting through the Github repository settings.
https://docs.github.com/en/code-security/how-tos/report-and-fix-vulnerabilities/configure-vulnerability-reporting/configuring-private-vulnerability-reporting-for-a-repository

This will allow submissions of vulnerabilities and for the maintainers to issue advisories for vulnerabilities. For example: https://github.com/containerd/containerd/security

A security policy is also required and I'll create a PR to add something I think fits this project

IoTThinks:
Great suggestion.

liamcottle:
Hey, thanks! I've merged in #2691 and also enabled the feature for users to privately submit vulnerabilities on the repo.

PR #2691, opened 2026-06-05
Add Github Security policy

NoodlesNZ:
As referenced in #2690 this is what I feel is a sane security policy for MeshCore. This outlines vulnerability scopes and versions where bugs are accepted. It also tries to set expectations for responses to submissions.


liamcottle:
Looks good, thanks! We'll make adjustments as needed :)

PR #2694, opened 2026-06-05
feat: rak ble companion led behavior change

NickDunklee:
RAK nodes use default board behavior for BLE handling.

That behavior keeps the BLE LED turned on solid all the time, and blue LEDs tend to have a shorter life than most. Most MeshCore node behavior is minimalist to save power.

This is a simple change for RAK boards to do basic BLE LED handling.

**Feature:**

  - Companion BLE LED blinks during initial pairing
  - Companion BLE LED turns off once connected to a device
  - Companion BLE LED blinks when disconnected from a device

Tested on RAK19003, RAK19001


mmoll:
Tested on RAK WisMesh Board ONE Pocket, works fine 👍🏿

PR #2695, opened 2026-06-05
feat: rak12047 sgp40 voc sensor support

NickDunklee:
This short feature adds the RAK12407 SGP40 sensor support for RAK nodes.

Follows the current sensor implementation.

Does a serial number check to bootstrap and the sensiron lib for VOC calculations.

Query reports as addGenericSensor.

I've had this one running for a few months now, so yanked the code out to put in a PR in case others want to be able to use this.

Tested on RAK19001, compiled against several other board types and compile didn't break.


NickDunklee:
voc-rak12047
Oh, forgot to attach a graph of this VOC sensor outputting data, last 15 days.

PR #2699, opened 2026-06-05
remove dependencies on direnv and nix

notmike-5:
remove dependencies on direnv and nix


liamcottle:
Howdy. Please provide explanation why this needs to be removed. It was added in https://github.com/meshcore-dev/MeshCore/pull/205 Also please ensure pull requests are opened against the `dev` branch.

notmike-5:
From what I can tell `direnv` and `nix` are not dependencies of MeshCore. The file `.envrc` is usually left to users to setup their own local environments as suits their needs. The file `.envrc` in the repo currently enforces the development environment of some `nix`/`direnv` users onto *all* users, overriding whatever particular environments users may have themselves. https://github.com/meshcore-dev/MeshCore/pull/205 adds `.direnv/` to the `.gitignore` in recognition that this environment is not needed by most users. It is advisable to similarly remove `.envrc` from version control since the environment may need look different for every different clone of the repo.

So, my claim is that https://github.com/meshcore-dev/MeshCore/pull/205 should really have added `.envrc` and `.direnv/` to `gitignore.txt` and left it at that. The other files are particular to the authors' development environment.

hickey:
#3025 may resolve this issue. In #3025 I have added the condition to check if `nix-shell` is available to trigger setting up the Nix environment. This alleviates the problem that `direnv` fails without a Nix environment (my problem) and allows existing environments that use `direnv` to continue functioning.

PR #2702, opened 2026-06-06
new CLI config: flood.max.advert

ripplebiz:
* new CLI config: flood.max.advert  (default 16)

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -431,6 +431,7 @@ bool MyMesh::allowPacketForward(const mesh::Packet *packet) {
   if (packet->isRouteFlood()) {
     if (packet->getPathHashCount() >= _prefs.flood_max) return false;
     if (packet->getRouteType() == ROUTE_TYPE_FLOOD && packet->getPathHashCount() >= _prefs.flood_max_unscoped) return false;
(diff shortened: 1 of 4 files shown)


liamcottle:
Below are my comments from Discord:
https://discord.com/channels/1495203904898728149/1495410606985969765/1512657279869259867

> This is similar to `flood.max` but it limits flood adverts only. The idea here, is instead of adverts defaulting to a max of 64 hops, it's set to 16 hops, enforced at the forwarding repeater. With the ability to increase if needed and decrease even more if desired.
>
> There was a PR opened for this by someone else, however it looks like the owner of the PR deleted their GitHub account, so the PR is no longer mergable. https://github.com/meshcore-dev/MeshCore/pull/2408
>
> The idea is that this will drop max advert flooding from 64 hops to 16. I personally wanted to have the TTL/max hops on the advert packet, but unfortunately using the feature flags (in my proposal) would increase advert size by 2-bytes and truncate the name max length by another 2-bytes...
>
> The main reason I'm asking for feedback here, is because instead of defaulting to 64, we would default to 16, and it won't really be possible for adverts to travel any further than this, without network repeater owners increasing the limit.
>
> I'm personally in favour of the PR, as it should help to decrease the affect of flood adverts on the network. Generally an advert going 64 hops is not needed. So finding a balance on how far they really should travel (by default) is the hard part. We settled on 16 as a start.

IoTThinks:
It should be fine.
In the future, the repeater should automatically set this parameter based on how busy it is.

PR #2703, opened 2026-06-06
Change default power saving setting to 'off'

IoTThinks:
The default powersaving value is off.
It is only on when entering "powersaving on".

Minimal change but hope to clear possible confussion.


IoTThinks:
So fast the merge.
Thanks a lot. 👯‍♂️

PR #2707, opened 2026-06-06
feat: add path.hash.mode.min CLI command for repeaters

StijnBTC:
## Description
Adds a new CLI command `path.hash.mode.min` to allow repeaters to enforce a minimum path hash size for forwarded flood packets.

## Changes
- Added `min_path_hash_mode` field to `NodePrefs` struct for persistence
- Added CLI handlers for `get` and `set path.hash.mode.min` commands
- Implemented in `simple_repeater` `allowPacketForward()` method
- Updated documentation with usage examples

## Behavior
The repeater will drop flood packets with insufficient hash size:
- `0`: No minimum enforced (default, forward all packets)
- `1`: Require hash_size >= 2 bytes (drop mode 0 packets)
- `2`: Require hash_size >= 3 bytes (drop mode 0-1 packets)

## Testing
- Verified build succeeds for (RAK_4631_repeater)
- Tested CLI commands on RAK 4631
- Confirmed logic works correctly by testing it with my companion and repeater

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -442,7 +442,24 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 
 **Note:** the 'path.hash.mode' sets the low-level ID/hash encoding size used when the repeater adverts. This setting has no impact on what packet ID/hash size this repeater forwards, all sizes should be forwarded on firmware >= 1.14. This feature was added in firmware 1.14
 
-**Temporary Note:** adverts with ID/hash sizes of 2 or 3 bytes may have limited flood propagation in your network while this feature is new as v1.13.0 firmware and older will drop packets with multibyte path ID/hashes as only 1-byte hashes are supported. Consider your install base of firmware >=1.14 has reached a criticality for effective network flooding before implementing higher ID/hash sizes. 
+**Temporary Note:** adverts with ID/hash sizes of 2 or 3 bytes may have limited flood propagation in your network while this feature is new as v1.13.0 firmware and older will drop packets with multibyte path ID/hashes as only 1-byte hashes are supported. Consider your install base of firmware >=1.14 has reached a criticality for effective network flooding before implementing higher ID/hash sizes.
+
+---
+
+#### View or change minimum path hash size for forwarded packets (repeaters only)
+**Usage:**
+- `get path.hash.mode.min`
+- `set path.hash.mode.min <value>`
+
+**Parameters:**
+- `value`: Minimum path hash mode (0-2)
+  - `0`: No minimum (forward all packets)
+  - `1`: Require hash size >= 2 bytes (drop 1-byte hashes)
+  - `2`: Require hash size >= 3 bytes (drop 1-2 byte hashes)
+
+**Default:** `0`
+
+**Note:** This setting only applies to repeaters and only affects flood packet forwarding. When set to a non-zero value, the repeater will drop flood packets whose embedded path hash size is smaller than the minimum.
 
 ---
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -299,7 +299,12 @@ However, with 1 byte, there are only 254 unique IDs (exclude 00 and FF which are
 Firmware version 1.14 and newer introduces the ability for repeaters to advert with 1-, 2-, or 3-byte adverts. Companions can also send out channel and direct messages with 1-, 2-, or 3-byte path. Adverts and messages sent in 1-byte path is compatible with repeater firmware older or newer than 1.14. They will travel up to 64 hops. 2-byte adverts and messages will travel up to 32 hops. 3-byte adverts and messages will travel up to 21 hops.
 
 ### 3.9.1. Q: **What path hash sizes will my repeater forward?**
-Repeaters running firmware 1.14+ repeat packets sent with 1-, 2-, or 3-byte path hash. Repeaters on firmware older than 1.14 only repeat 1-byte path hash packets and silently drop 2- and 3-byte packets.
+Repeaters running firmware 1.14+ repeat packets sent with 1-, 2-, or 3-byte path hash by default. Repeaters on firmware older than 1.14 only repeat 1-byte path hash packets and silently drop 2- and 3-byte packets.
+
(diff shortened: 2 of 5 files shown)


ViezeVingertjes (approved):

niekvenema:
👍

ripplebiz:
I'm actually not sure what the rationale behind this one is? 
The path.has.mode is the _sender's_ preference. If they don't want extra diagnostics, OR would prefer a bit more privacy, then that is their choice. Not sure why a repeater admin should have a say in that choice?

Avamander:
> Not sure why a repeater admin should have a say in that choice?

Because airtime is scarce, old misconfigured repeaters that make routing and monitoring difficult with prefix collisions are making things worse for everyone. People reasonably want to restrict that.

agessaman:
I have to voice my disagreement with merging this PR. It is a breaking change at a mesh level, because it is fundamentally coercive in a way that flood.max or flood.max.advert are not. It discriminates against an established user choice, not a neutral factor like number of hops, and provides users zero way to understand the failure mode. Additionally, the amount of airtime gained back by an occasional DIRECT packet that goes a hop down the wrong path is negligible, and if this is an effort to manage FLOOD packets there are other tools. I have to agree with Scott, this seems contrary to existing design decisions.

Avamander:
A lot of repeater settings are in the end "coercive" and against "user choice". The "failure mode" is nothing more than not being heard, there are numerous cases where by-design MC does not reply with anything.

At the same time repeater owners and their needs are just as important, be it proper observability or just slight airtime optimization,

agessaman:
"Proper observability" within the design framework of MeshCore is the amount of observability the packet permits. This is a change to that contract. (I say this as a guy who has done a lot of work to support observability and who likes mesh analytics.)

Now, if you RX a repeater passing messages, there's basically 100% chance that with proper LoS and power it will TX your messages onward. With this change, that's no longer the case. That's a meaningful behavior change.

Avamander:
> Now, if you RX a repeater passing messages, there's basically 100% chance that with proper LoS and power it will TX your messages onward. 

Not if it's an old unmaintained repeater that only supports 1-byte or if it only allows certain scopes.

agessaman:
An old repeater will pass on your messages one-byte path hashes just fine—even if they are scoped—unless the next repeater down the line has turned on this PR's analytics purity filter. In which case, you're stranded until you deploy new hardware.

mikecarper:
Not a fan as well.

onnozweers:
Perhaps I could elaborate a bit on why we need this option.

The Netherlands is a densely populated country with roughly 3,000 repeaters. Radio time is scarce, and we constantly need to optimize to keep the Mesh working. We've had to implement regions so that users don't accidentally flood the whole country — and even parts of Germany and Belgium — with their messages. Bottom line: the Dutch Mesh is congested.

On top of that, for the past few months, a spammer has been active. He (or she) regularly overloads the Dutch Mesh with garbage messages. When this happens, the Mesh comes to a grinding halt. At first these messages were sent without a region, but once we started enforcing region use, they switched to using region `nl` to still have maximum possible impact. Now at least Germany and Belgium are spared from this spam.

People in the community are trying to locate this spammer, but it's difficult, because the spammer spoofs the first hops of the path. The spammer uses 1-byte hashes, so even when you know which hops are real, you still can't be sure where those hops actually are. With 3,000 repeaters in a single small country, each single-byte hash occurs more than 10 times. We're trying to triangulate, but with so many uncertain hops, there's a lot of guesswork and complex algorithms involved.

It would help tremendously if repeater admins could block single-byte hashes. We would block the spammer instantly. We expect them to adapt and switch to 2-byte hashes — and that's fine, because it would give us a much better idea of where they're located. We could then try to persuade them to stop. I'm optimistic that simply proving they can be located will be enough to stop this behavior, and it would also discourage other would-be spammers from trying the same thing.

I hope this clarifies why we need this option.

xeros:
In Poland, we had similar spam case, but not intentional.
There was one bot, which hit a bug, while his owner was on vacations.
It was sending empty messages few times a second, flooding all repeaters.
Most bots are still sending messages using 1-byte paths, while all users use 3-byte paths.
If there would be possibility to eliminate 1-byte traffic at that time, we could easily disable 1-byte traffic on nearby repeaters to heal the network.
Without this, the air was flooded and people couldn't communicate for huge range.

ViezeVingertjes:
It's indeed meant to prevent spam; we have a spammer for months already that floods the network 2 times a day for several hours, rendering the mesh unusable for most area's. Now there are forks that drop 1-byte and also OpenHop repeater is capable of dropping most of it, but would be nice if there is some way to defend the mesh from these malicious actors in atleast some way or another in the official MC so things dont diverge etc.

Privacy shouldnt be the argument against it imho, as we can locate even 1-byte transmitters with ease these days and legitimate users are easier to find than these malicious actors, due to them not faking paths etc.

If this approach isnt the one, it would be very very welcome to atleast come up with something that would make NL use MeshCore again, as it doesnt seem to resolve itself and its already ongoing for some months now. (~3 or so?)

recrof:
I'm also against merging this PR. @ripplebiz is right - it's a choice of user to use 1b or 2b. we should not give repeater operator way to restrict that.

xeros:
You might have probably only good, polite users in your mesh network, no unattended bots, etc.
If you'd have problematic ones, you'd change your mind and you'd probably even abandon all this and shutdown your hardware totally.

onnozweers:
Guys, you simply can't take away _all_ responsibility from the repeater admins and give it to the users. Because some users out there can't be trusted. We're desperate here in the Netherlands. People here are even [creating portals to indicate when a spam attack is ongoing](https://mc-spamdetector.nl/). Some people say they are quitting Meshcore out of frustration.

How else would you suggest we handle this problem?

onnozweers:
I think we forgot to mention, that in the Dutch Meshcore community, we have already agreed to use at least 2 byte hashes. https://settings.dutchmeshcore.nl/en/

xeros:
The same in Poland. 3 byte hashes is general standard with some smaller cities in 2 byte.
1 byte hashes are only for spam now and very annoying.

PR #2710, opened 2026-06-06
Feature/rtc cli status

polipl:


ViezeVingertjes (changes requested):
Few findings:
- You're branch is pointing to `main`
- Debug flags are uncommented (`MESH_PACKET_LOGGING` & `MESH_DEBUG`)

Also; i did see RTC being reworked somewhere, sadly do not know the exact PR, but might be worth checking out if you can find it to see if it overlaps or could be combined.

Last but not least: - any - even a very short or concise description would always be helpfull. 👍🏼

PR #2712, opened 2026-06-06
boards: Add Ebyte Eora-HUB-900TB

VynDragon:
Ebyte LR1121 EoRa board

Note this does not include the LR1121 helper files and the companion code doesn't know about P_LORA_DIO_9 so P_LORA_DIO_1 is provided.

### DIFF
--- a/boards/ebyte_eora-hub.json
+++ b/boards/ebyte_eora-hub.json
@@ -1,0 +1,45 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "esp32s3_out.ld",
+      "partitions": "default.csv",
+      "memory_type": "qio_qspi"
+    },
+    "core": "esp32",
+    "extra_flags": [
+      "-DBOARD_HAS_PSRAM",
+      "-DARDUINO_USB_CDC_ON_BOOT=1",
+      "-DARDUINO_RUNNING_CORE=1",
+      "-DARDUINO_EVENT_RUNNING_CORE=1",
+      "-DARDUINO_USB_MODE=1"
+    ],
+    "f_cpu": "240000000L",
+    "f_flash": "80000000L",
+    "flash_mode": "qio",
+    "mcu": "esp32s3",
+    "variant": "esp32s3"
+  },
+  "connectivity": [
+    "wifi",
+    "bluetooth"
+  ],
+  "debug": {
+    "openocd_target": "esp32s3.cfg"
+  },
+  "frameworks": [
+    "arduino",
+    "espidf"
+  ],
+  "name": "Ebyte EoRa-HUB-XXXTB Radio",
+  "upload": {
+    "flash_size": "4MB",
+    "maximum_ram_size": 327680,
+    "maximum_size": 4194304,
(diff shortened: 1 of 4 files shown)


mihvoi:
+1 to support Ebyte Eora-HUB
I tested the EoRa-hub port and it worked fine for several days.
I only needed to force the flash size to 4MB as my board had, as the firmware defaulted to 8MB and was not booting.

The _merged firmware needed to be flashed at 0x0 via flash_download_tool_3.9.10, web upgrade might not work.
https://github.com/meshcore-dev/MeshCore/commit/3977c0d49d89ef5307600d083057c28872e4961a

mrrdmr:
Does the rep mode work as well?
I couldn't get it to work.

dkmr27:
Definitely an issue with the serial running repeater. I changed ARDUINO_USB_MODE and ARDUINO_USB_CDC_ON_BOOT to zero in the board config, and could connect with a terminal program, but not through the console on the flasher site.

VynDragon:
The USB does not work because it has the esp32 flash thing which touches reset depending on RTS and CTS, it only worked for flashing for me.

mrrdmr:
Something also very wrong with tx power in repeater mode. Guess because of incomplete LR1121 wrapper.

mrrdmr:
And I think we don't need LR1121 wrapper at all. It works as LR1110. In this case DIO9 is not needed.

dkmr27:
Not all the wrappers attached to PR's here are the same. There are considerations for 2.4G compatibility that an LR1110 will not address, nor do most of the LR1121 wrappers. I don't think P_LORA_DIO_9 is needed either, only the confusion from the naming of P_LORA_DIO_1 which should probably really be called P_LORA_INT or similar elsewhere in code, but that's a bigger issue. Seems battery measurement needs work too.

USB companion works, as does repeater with a serial terminal, but not the flasher console, if you change ARDUINO_USB_CDC_ON_BOOT=0 in the board config and get rid of ARDUINO_USB_MODE

Which band is the power low? I'll test mine.

mrrdmr:
Meshcore doesn't use 2.4GHz atm. 868 62.5/8/8 works within few meters only in repeater mode. It works well in meshtastic. Which wrapper do you use?

dkmr27:
I started with one of them and added in changes, it now talks to SX128x modules on 2.4G just fine. Power output I tested tonight 20.5dBm on 915M, I don't have anything close to cal for 2.4G but it's within a couple of dB of spec. I suspect the switching is working OK, seems to get a few Km from inside the house, will test it with the roof antenna tomorrow.

mrrdmr:
[My wrapper. ](https://github.com/mrrdmr/MeshCore-EoRa-HUB/tree/main/src/helpers/radiolib)
I've tested again and it works. Can connect to another repeater 1.7km but signal strength very weak -7...-5db (0...+2db back). Heltec with SX1262 gives +8...+10db from same position. Overall my HUB works fine on short distances (+15db) but almost useless on longer. 
Maybe hw issue but didn't see any problem with meshtastic.

PR #2340, opened 2026-04-20
Fix FEM/LNA enbaled by default for Heltec T096, Heltec Wireless Tracker v2

weebl2000:
Should only really be disabled when it causes issues. It's enabled by default on all similar Heltec boards.

PR #2713, opened 2026-06-07
Enable LNA by default t096, tracker v2, heltec v4

weebl2000:
#2340 
#2439 

Not sure why it was disabled again.


palachzzz:
Why not to merge PR with cli control #2140 instead if doing that strange ping-pong?

PR #2716, opened 2026-06-07
Added several Flags to fix some issues with RAK3401

martenumberto:
Added Flags for RAK3401 to match up with the RAK3461. This also uses the Bosch Driver for BME Sensors like in https://github.com/meshcore-dev/MeshCore/pull/2327 and https://github.com/meshcore-dev/MeshCore/pull/2634


martenumberto:
Closed because wrong target branch

Issue #2599, opened 2026-05-21
Feature Request.  Mobile Repeater that won't confuse the mesh.


pnhbxpmpqw-del:
It would be nice to have a mobile/car repeater option that won't ruin the mesh.

aclaghorn:
There are a few open issues discussing this already

https://github.com/meshcore-dev/MeshCore/issues/2396

https://github.com/meshcore-dev/MeshCore/issues/2261

IoTThinks:
How about set it as repeater and set max advert flood = 1?

The mesh should be dynamic and handle the mobility on its on.

Issue #2396, opened 2026-04-24
[Feature request] Companion Node Relay.


mcneely:
Prior Understandings:
- Moving a repeater around, destabilizes the Network
- Flood repeating creates too much noise, reduces hop count, etc..
- small devices like credit card sized have limited range and may not reach a repeater.
- there is a desire for base nodes in cars that rebroadcast short range devices  to a longer range to repeaters.
- meshcore supports companion repeater for adhoc situations in remote areas but restricts you to a specific frequency. Doesn't allow repeating to the general network.

Proposal:  Companion Relay.
-  Involves setting a limited list of companion nodes that a designated companion will listen for and rebroadcast
-  mechanism can be put in place to disallow companions rebroadcasting each other if needed.
- Allows for car nodes that won't destabilize the network as a mobile repeater would.

Nice to have:
- Ability to manage relay node from the clients on the list.

zevaryx:
Someone was experimenting with this, here's their fork: https://github.com/ZaosDaOne/MeshCore-relay/tree/relay

wesselah:
Feels like going back to meshtastic . The reason repeating on a companion to use fixed freq. has a reason. People would forget to disable it and the mesh is already very crowded this would make it worse. And the mesh would collapse . And when you need better range buy a decent companion

mcneely:
> Feels like going back to meshtastic . The reason repeating on a companion to use fixed freq. has a reason. People would forget to disable it and the mesh is already very crowded this would make it worse. And the mesh would collapse . And when you need better range buy a decent companion

ooh, you invoked the word "Mashtastic" maybe you should have also mentioned "Andy Kirby" in your sentence as well. I presented real world issue and your solution is "buy better hardware"?   Neither Meshcore or Meshtastic get it fully right, The evil Voldemort Meshtastic does a better job at address smaller and more portable nodes, such as ones that you keep in your pocket or wear around your neck.  The solution I suggested addresses this need with massive noise reduction by limiting the nodes that get repeated. 1 node repeating 1 node is way different from every node repeating every other node. It addresses issues a lot of us have particularly in place where we don't live on a island or in a large city  where there is mass coverage.

If you have a better solution than "buy better hardware", by all means suggest it. Otherwise please keep your useless toxic commentary off my thread.

wesselah:
[...]

Issue #2261, opened 2026-04-05
Car Repeater build - Feature Request


jeremeguenther:
I have a feature request for a firmware build or new setting that allows me to configure a repeater for use in a mobile vehicle.

A common challenge we have is the desire to mount a repeater in a vehicle to act as a mobile “range extender” for one or more companion nodes. Today this requires running a full repeater which can pick up and rebroadcast traffic from the entire mesh, creating network instability as the available routes shift and change.

I propose adding a lightweight “car-repeater” (or “companion-repeater”) mode with two simple behaviors:

- Only repeat packets originating from a configurable whitelist of companion node IDs.
- When a packet is received from another repeater, automatically force its hop count to 0 before any further processing or flooding. This guarantees the packet will not be repeated again by any other repeater, significantly reducing the chance of it being included in memorized paths, while still allowing the boosted packet to reach nearby companions.

aclaghorn:
I think this is a great idea. 

Having a vehicle roof repeater would be a huge benefit to being able to use low power trackers nearby your vehicle while mobile.

mepholic:
You'd want the "carpeater" to force hops to max count for whatever path hash size the packet is using, not to 0. Resetting the hop count on a flood will make it flood further, not stop flooding, but this also has the risk of messing with the direct path discovery, so it's probably not the ideal solution here.

I've thought about this particular problem a bit, and I think the best solution would be a "car-repeater" mode that shoves the repeater on a unique private region/transport scope and effectively enables `denyf *`. The client would then be configured to use the private region/transport scope that is configured on the car repeater. As far as I can tell, the only "atypical" thing that the car repeater would need to do in order to make this work would be to rewrite traffic sent to the target scope to the private scope (target scope being either entirely unscoped `*` traffic, or traffic to any other operator configurable target scope), and then rewrite any messages heard on the private scope to target scope.

This could be further extended to allow pods of "traveling repeaters", say for like a convoy, or friends going on a multi-car road trip. But further thought would be needed regarding the specific implementation.

pete1450:
Got pointed to this issue when discussing the same idea. Here's builds from mine if you want to try it. https://github.com/pete1450/MeshCore-EdgeRelay/actions/runs/35294007377

[...]

PR #2717, opened 2026-06-07
Add Companion Selective Repeat Ability (CSRA)

copelaje:
## Background

There are plenty of situations where you want a well-positioned companion (always powered, good antenna) to extend the reach of another device stuck in poor coverage, but only for your own traffic. A roof-mounted companion covering a pocket device that ducks into a store is one example.

The existing options don't fit. The built-in `client_repeat` mode repeats everything for everybody, and critically, it requires moving off your normal companion frequency onto a dedicated repeater band. The common community workaround, a "car-peater" or mobile repeater, also introduces trouble in that it isn't selective and thus can create known paths that are transient and unstable for many users.

## Proposed Solution

This PR adds a third option alongside the two that exist today. `off` repeats nothing and `all` repeats everything on a dedicated repeater frequency, both unchanged. The new `selective` mode **repeats only traffic on your allowlist** and stays on your normal companion frequency. I'm calling the capability CSRA, Companion Selective Repeat Ability, a nod to the [CSRA mesh region](https://meshcore.csramsh.org) (Central Savannah River Area) I'm building out near Augusta GA, USA. I'm definitely open to alternate methods of indicating allowlist should the community decide that is a good path to not overload use of favorites.

As implemented, the allowlist reuses the existing `favorite` flag on contacts, so there's no new setting to manage and the capability works without companion app updates. You mark the contacts you want selectively repeated, and it also repeats the specific group channels you already carry.

## Limiting impact on the larger mesh

This was the central design consideration, so I want to address it head on. A repeater that isn't selective is a burden on the mesh. Every flood it hears it rebroadcasts, and that retransmission consumes airtime every nearby node has to wait through. CSRA is built specifically to avoid that. It rebroadcasts a message only when the source contact is on your allowlist, or the message is on a channel you carry. Everything else is dropped with no retransmission at all. The practical result is that a CSRA node adds essentially no flood traffic to the mesh beyond the specific conversations its owner has opted into. That is a real improvement over an `all`-mode car-peater sitting in the same location and rebroadcasting indiscriminately.

The same selectivity keeps the node from quietly turning into a general next-hop for unrelated routed traffic. It will only relay along paths it created for allowlisted conversations, and it forwards nothing else.

## What gets repeated

Traffic moves in two directions. Outbound is a message leaving your device toward the wider mesh. Inbound is the reply coming back.

On the outbound side, a direct message involving a favorited contact gets repeated, and so does a group message on a channel you carry on the CSRA enabled companion. Anything else, meaning a message from a contact you haven't favorited or on a channel you don't carry, is ignored and not repeated.

On the inbound side, the reply or acknowledgement heading back to you is forwarded even though the far party is not on your allowlist. The return arrives in one of two ways. Once the node has built a path, the reply comes back along that path, and dropping it would break the route the node created. Before any path exists, which is the usual case while your device is shielded behind the node, the reply arrives as a flood addressed to your device. **The node recognizes your device as the destination and repeats it**, and that repeated flood is what lets the return path be found so the acknowledgement gets home. Traffic that is not part of one of your conversations is ignored here as well.

Three implementation points for reviewers. First, when the node repeats one of your messages it inserts itself into the message path. That is what gives the reply a way back. Without it the message goes out but the acknowledgement has nowhere to return, and the DM looks like it silently failed. Second, the allowlist check looks at both ends of a message, the sender and the recipient, rather than the sender alone. A favorited device is amplified whether it is sending a flood out or receiving one back, and that second case is what carries the first acknowledgement home before any path exists. Third, the node only steps in when a message is actually flooding. If your device already knows a direct route to a contact, that message does not need help and the node leaves it untouched. CSRA engages precisely when coverage is poor and the device has fallen back to flooding.

## Testing

I verified all cases DM/flood worked (and didn't work) as expected by flashing KISS on a device and watching all the raw OTA packets traversing through the various cases in a "bench test" scenario. I and other have also field tested the firmware and found the capability works as intended (see "test yourself! section below).

## Companion tooling

There is a companion-side piece, a new command to set the repeat mode, that I have ready in `meshcore_py` and `meshcore-cli`. I am holding those until there is intent to merge this firmware change, and they will need to be coordinated with a client app update as well.

## Test yourself!

Given there is no current client app support, I've gone ahead and built CSRA auto-enabled companion firmware for all supported devices available here: https://github.com/copelaje/MeshCore/releases  If you flash these, these your companion will behave in the manner described above for personal testing purposes. Long term my intent is that this mode be selectable within client apps (see "Companion tooling" above.  I'm happy to provide MRs and coordinate in this regard.

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -520,17 +520,24 @@ bool MyMesh::allowPacketForward(const mesh::Packet* packet) {
 bool MyMesh::shouldSelectivelyRelay(const mesh::Packet* packet) {
   uint8_t pt = packet->getPayloadType();
 
-  // Chat-type flood packets carry src_hash at payload[1].
+  // Chat-type flood layout: payload[0]=dest_hash, payload[1]=src_hash.
+  // Match a favourite at either endpoint so both directions of its conversation relay:
+  // outbound floods it sends (src), and floods headed back to it (dest). When the
+  // favourite is shielded behind this relayer, its peer has no path home yet, so the
+  // returns come back as floods rather than direct routes. Without the dest match the
+  // return leg is dropped and the message is never ACKed.
   if (pt == PAYLOAD_TYPE_TXT_MSG || pt == PAYLOAD_TYPE_REQ ||
       pt == PAYLOAD_TYPE_RESPONSE || pt == PAYLOAD_TYPE_PATH) {
     if (packet->payload_len < 2) return false;
+    uint8_t dst_hash = packet->payload[0];
     uint8_t src_hash = packet->payload[1];
     int n = getNumContacts();
     for (int i = 0; i < n; i++) {
       ContactInfo ci;
       if (!getContactByIdx(i, ci)) continue;
-      // 1-byte hash collisions (~favourite_count/256) are intrinsic to the on-wire src_hash.
-      if (ci.id.isHashMatch(&src_hash, 1) && (ci.flags & FLAG_FAVOURITE)) return true;
+      // 1-byte hash collisions (~favourite_count/256) are intrinsic to the on-wire hash.
+      if ((ci.id.isHashMatch(&src_hash, 1) || ci.id.isHashMatch(&dst_hash, 1))
+           && (ci.flags & FLAG_FAVOURITE)) return true;
     }
     return false;
   }
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -62,6 +62,14 @@
 #define CMD_SET_DEFAULT_FLOOD_SCOPE   63
 #define CMD_GET_DEFAULT_FLOOD_SCOPE   64
 #define CMD_SEND_RAW_PACKET           65
+#define CMD_SET_CLIENT_REPEAT_MODE    66  // non-radio-params variant; SELECTIVE bypasses isValidClientRepeatFreq
+
+#define CLIENT_REPEAT_OFF        0
+#define CLIENT_REPEAT_ALL        1
+#define CLIENT_REPEAT_SELECTIVE  2
+
+// Existing ContactInfo.flags bit; doubles as the SELECTIVE allowlist gate.
+#define FLAG_FAVOURITE           0x01
 
 // Stats sub-types for CMD_GET_STATS
 #define STATS_TYPE_CORE               0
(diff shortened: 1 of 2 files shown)


copelaje:
Over the past week or so I've attempted to engage devs in the #development channel on meshcore.io discord but haven't received any response. I recall someone attempting something like this in a thread on the old discord, but don't think that topic started again post-split.  I may add one for discussion on this if needed. There are already several "issues" regarding a capability like this #2261 #2396 #2599

Lets see if we can get a workable solution!

copelaje:
Testing has shown that #1707 applies to CSRA paths as well. Everything functions right, but companion contributions to paths will always be unknown due to reasonable constraints mentioned by @liamcottle in that issue.

ifeth0:
Awesome work so far! I've had success in relaying packets from my low power t1000 to heltec v4 running your firmware. The only question I have is regarding the auto functionality of repeats. I'm currently sitting in a dead spot with my low power companion and the custom firmware seems to not trigger the repeat functionality immediately. I'll have to take a deeper dive but just wanted to give you some additional feedback. I hope you're able to get acknowledged by the dev team and get this implemented. It's greatly needed in my area that isn't MC heavy.

copelaje:
@ifeth0 , thanks for testing and providing the feedback! I took a look at things again and I don't see any reason why repeat off a CSRA enabled companion would be excessively delayed. Is the message for flood or DM message?  Could it be that your t1000 has to re-learn a new path through the heltek v4 and the delay is the retry logic up until it floods again and learns the path? Is is also slow on 2nd message after path learned through companion?  What kind of delay we talking? If you've got more details I can take a look closer.

ifeth0:
I guess I'm trying to understand the expected behavior while being out of range within the MC app. When I'm near areas that have repeaters I'll see 2+ repeats with the one repeat being unknown (the radio running your firmware). When I'm in an area without coverage I never see the app show a repeat. I still haven't had a chance yet to dive further into troubleshooting yet sorry!

copelaje:
@ifeth0 , I got a chance to test today in total isolation and it worked as I'd expect. The setup was a normal 1.16 companion (lets call A), and a CSRA 1.16 companion B, with no other repeaters present. I assume you're talking about channel messages since you're talking about # repeaters observed. CSRA companion will only repeat channel messages for those channels to which it is subscribed. So in my setup, if B is not configured for a channel, it won't be repeated.

When A sends a channel message to a group that B IS subscribed to, A indicates 1 repeat (the key from CSRA node B). When I send channel message to group that B does NOT have subscribed, no repeat is observed on A. If you find otherwise I need more info to understand.

NOTE: the fact that CSRA repeats based on what groups it is subscribed to the *easy* capability I could use to get selective repeat behaviors without requiring companion app feature updates.  I'm open to having specific flags for channels that should repeat rather than just configured channels if that's what the devs feel best.  Still waiting feedback though...

copelaje:
I see a merge conflict has developed as dev has moved on. Since no dev engagement thus far I'll just let it ride until next release. If that changes let me know and I'll make the necessary updates for merge.

vanous:
I do **similar** thing for selective local coverage by using regions. It is similar but not the same as this PR. It uses the current standard MC building blocks and doesn't require another node type - further complexity for the users.

IoTThinks:
> situations where you want a well-positioned companion (always powered, good antenna) to extend the reach of another device stuck in poor coverage,

Convert it to become a repeater.
Then done.

copelaje:
> I do **similar** thing for selective local coverage by using regions. It is similar but not the same as this PR. It uses the current standard MC building blocks and doesn't require another node type - further complexity for the users.

This is not another node type, it's just a checkbox in companion options (at least in final form). Is there some MR where your approach can be examined?


> > situations where you want a well-positioned companion (always powered, good antenna) to extend the reach of another device stuck in poor coverage,
> 
> Convert it to become a repeater. Then done.

The whole point of this is to mitigate the ill effects of mobile repeaters (literally the 2nd paragraph) by being selective in what is repeated. Are you suggesting that mobile repeaters wouldn't be a problem?

PR #2719, opened 2026-06-08
fix(techo-lite): remove duplicate EXTERNAL_FLASH_DEVICES definition

jirogit:
## Problem

`variants/lilygo_techo_lite/variant.h` defines `EXTERNAL_FLASH_DEVICES` twice:

```c
// Line 73 (correct)
#define EXTERNAL_FLASH_DEVICES ZD25WQ32CEIGR  // 32Mbit = 4MB, actual hardware

// Line 98 (incorrect)
#define EXTERNAL_FLASH_DEVICES MX25R1635F     // 16Mbit = 2MB, used on T-Echo
                                               // (non-Lite) and Heltec T114
```

Both definitions were introduced simultaneously in the initial commit
adding T-Echo Lite support (b11f0842). Per LILYGO's official schematic,
`ZD25WQ32CEIGR` is the only flash chip used on T-Echo Lite hardware.

## Fix

Remove the erroneous second definition (lines 98–99) to avoid confusion
for anyone reading the variant file.

## Testing

Built `LilyGo_T-Echo-Lite_non_shell_companion_radio_ble` — Success.
Also fixes missing newline at end of file.

### DIFF
--- a/variants/lilygo_techo_lite/variant.h
+++ b/variants/lilygo_techo_lite/variant.h
@@ -95,9 +95,6 @@
 #define PIN_BUTTON2             _PINNUM(0, 18)
 #define BUTTON_PIN2             PIN_BUTTON2
 
-#define EXTERNAL_FLASH_DEVICES MX25R1635F
-#define EXTERNAL_FLASH_USE_QSPI
-
 ////////////////////////////////////////////////////////////////////////////////
 // Lora
 
--- a/variants/lilygo_techo_lite/variant.h
+++ b/variants/lilygo_techo_lite/variant.h
@@ -70,9 +70,6 @@
 #define PIN_QSPI_IO2            _PINNUM(1, 9)
 #define PIN_QSPI_IO3            _PINNUM(0, 26)
 
-#define EXTERNAL_FLASH_DEVICES ZD25WQ32CEIGR
-#define EXTERNAL_FLASH_USE_QSPI
-
 ////////////////////////////////////////////////////////////////////////////////
 // Builtin LEDs
 


liamcottle:
Hey, thanks for the PR.

As far as I'm aware, `EXTERNAL_FLASH_DEVICES` is not actually used by the MeshCore firmware, and [CustomLFS](https://github.com/oltaco/CustomLFS) retrieves the JEDC ID to determine which flash chip is available.

https://github.com/oltaco/CustomLFS/blob/9a6a514c2ba0d64dad21463fc153630891fa7975/src/CustomLFS_QSPIFlash.cpp#L647

So I can't see how changing this will make any difference to LittleFS. However, I'll tag @oltaco as we probably should remove the define that's incorrect.

jirogit:
Thank you for the clarification! I wasn't aware that CustomLFS uses
JEDEC ID detection at runtime rather than relying on EXTERNAL_FLASH_DEVICES.
I'll update the PR description to remove the incorrect claim about LittleFS
behavior.

The fix still seems worthwhile to avoid confusion for anyone reading
variant.h, even if there's no functional impact on the current firmware.
Happy to close this if you'd prefer to handle it differently.

entr0p1:
Liam is correct, we don't have to define it at all; its all handled in CustomLFS. The T-Echo Lite does have QSPI flash as you found, but it isn't currently active for meshcore. Wouldn't hurt to just strip out the EXTERNAL_FLASH_DEVICES definition altogether since we don't need it even when QSPI flash support is added for the board.

jirogit:
Thanks both for the input — I've updated this PR to remove EXTERNAL_FLASH_DEVICES/EXTERNAL_FLASH_USE_QSPI entirely from lilygo_techo_lite/variant.h rather than just fixing the duplicate. I checked and EXTERNAL_FLASH_USE_QSPI isn't referenced anywhere outside variant.h files either, so it looked equally safe to drop. Verified build success on LilyGo_T-Echo-Lite_non_shell_companion_radio_ble.

While looking into this, I noticed EXTERNAL_FLASH_DEVICES is still defined in 22 other variant.h files across the repo (e.g. rak4631, heltec_t114, wio-tracker-l1, lilygo_techo). If it's genuinely unused firmware-wide per CustomLFS's JEDEC ID detection, would it be worth opening a separate issue to track cleanup across those too? Happy to open one if useful, just didn't want to scope-creep this PR.

PR #2731, opened 2026-06-09
FindMy Beaconing support

alfs:
This patch adds integrated "airtag" functionality to nodes. When enabled, a node advertises an Apple FindMy / OpenHaystack "offline finding" beacon alongside its normal mesh duties. Nearby iPhones anonymously relay its encrypted location to Apple's servers, letting you locate a deployed node through the global Find My network. No extra
infrastructure, GPS, or cellular needed.

This is particularly useful for repeater nodes. To maximise coverage they are often deployed in
exposed, unattended, hard-to-reach spots — rooftops, towers, hilltops, remote solar sites — which
makes them especially prone to going missing, whether through theft, weather, a failed mount, or being forgotten. A FindMy beacon gives you a way to locate (or recover) such a node, or at least get a last-known position, without having to physically visit the site.

It can also be used to geofence a repeater, to send notifications when it's moving from the installed location,
which if great for repeater nodes that don't have a built in GPS (such as the popular wismesh repeater mini solar).

The beacon genrerating code is adapted from https://github.com/pix/heystack-nrf5x

### DIFF
--- a/docs/findmy.md
+++ b/docs/findmy.md
@@ -21,11 +21,10 @@ requires the matching **private** key, which you keep off-device on your own mac
 
 ## Supported hardware
 
-nRF52840 boards only (Adafruit Bluefruit BLE stack),
-but the implementation is easily ported to ESP32 and other BLE capable chips.
-The challenge is with concurrent BLE usage, ie. companion nodes. **TODO/Future work** check if
-ble adverisements can be sent when the companion is disconnected from the phone,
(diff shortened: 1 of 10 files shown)


cra0:
This is a great idea, at the moment I put an individual tag in my repeater builds to keep track of their locations and to prevent them from being stolen. This feature would make it a lot easier for asset management as well.

PR #2733, opened 2026-06-09
Add public/channel content filtering to simple_repeater (low priority)

disq:
Lets a repeater decrypt channels it holds the key for (incl. the built-in public channel) and drop messages matching blocked keywords or sender names before forwarding. Configured via new "filter" CLI commands and persisted to /channel_filter.

Tested to be working with `Public` and hashtag channels, with or without scope set.


m0urs:
@disq Should this work also for senders which have a blank in their name, e.g. "XYXY|ABC C"? I tried but it does not seem  to work?

disq:
@m0urs What is your config, ie. what are you trying to block in this `XYXY|ABC C` scenario? Could be a trailing space issue?

m0urs:
> @m0urs What is your config, ie. what are you trying to block in this `XYXY|ABC C` scenario? Could be a trailing space issue?

"XYXY|ABC C" is the name of a companion in a channel and I would like to block that sender name. I am right, that you do NOT need to use quotation marks if the name contains a blank?

disq:
> > @m0urs What is your config, ie. what are you trying to block in this `XYXY|ABC C` scenario? Could be a trailing space issue?
> 
> "XYXY|ABC C" is the name of a companion in a channel and I would like to block that sender name. I am right, that you do NOT need to use quotation marks if the name contains a blank?

Correct, `filter sender XYXY|ABC C` would take everything after sender as the name, spaces included...  Also, it's a case-insensitive substring match, so blocking just `XYXY` would already catch that sender.

Issue #2678, opened 2026-06-03
[Feature request] LilyGo T-Watch S3 + T3S3 (LR1121 / SX1280) board support + a 2.4 GHz LoRa frequency plan


GrayHatGuy:
### Use-case / what problem this solves
MeshCore today has no support for the LilyGo T-Watch S3 family, and no support for **2.4 GHz LoRa** (the SX128x driver is currently excluded in-repo). I'd like to contribute both:

- **2.4 GHz LoRa** (SX1280 + LR1121 in 2.4 GHz high-band) — opens MeshCore to the worldwide, license-free 2.4 GHz ISM band (no regional sub-GHz duty-cycle/region constraints).
- **LilyGo T-Watch S3 family** (S3 + S3 Plus) as a wearable companion/repeater — SX1262 (sub-GHz) **or** SX1280 (2.4 GHz), AXP2101 PMU, on-device ST7789 TFT + FT6336 touch, and BMA423 wrist-lift screen-wake.
- **LilyGo T3S3** with **LR1121** (multi-band: 2.4 GHz + sub-GHz) and with **SX1280** (2.4 GHz) — both with SSD1306 OLED, all roles (companion BLE/USB, repeater, room server).

### Already implemented + hardware-verified
This is built and running — I've published a working release on my fork so it can be tried/reviewed before any PR:

➡️ **Release v0.3 — 20 flashable bins:** https://github.com/GrayHatGuy/MeshCore/releases/tag/v0.3
Branch: `feat/lilygo-twatch-s3-and-2.4ghz-lora` — https://github.com/GrayHatGuy/MeshCore

Verified on hardware:
- **2.4 GHz messaging** on a 3-node mesh — public + cross-chip SX1280<->LR1121, private channels @ 2.4 GHz, companion client-repeat @ 2450.
- **Sub-GHz** (LR1121 US 915) messaging + BLE pairing + OLED.
- **T-Watch S3 Plus**: ST7789 display, FT6336 touch, dynamic BLE PIN, BMA423 wrist-lift screen-wake.

### Approach / API notes
[...]

Issue #2441, opened 2026-04-29
[Suggestion] [repater] Support multiple radios, to build backhauls or bridges.


dreirund:
Ahoj,

I have a suggestion:

Implement repeater firmware for devices with more than one radio, which can drive both radios with different parameters and intelligently repeat packets received on one radio on the other.

Use case would be e.g.

* Backhauls (like [this idea](https://stuff.cisien.com/meshcore-backhaul.html))/ directional links, to reduce load on the local network,
* bridges between areas that use different radio parameters (like Czech Republic and surrounding countries),
* &hellip;

Since different scenarios would need different repeating behaviours, the following needs to be configurable.

* If a flood packet is received on radio 1, repeat it on radio 1, radio 2, &hellip;, or all.
    - Make this configurable on a per-scope-basis.
* If a flood packet is received on radio 2, repeat it on radio 2, radio 1, &hellip;, or all.
    - Make this configurable on a per-scope-basis.
* etc. for all radios.

(If more than 2 radios makes too much complexity, just restrict to two radios.)

Furthermore, Meshtastic has grown and has started to diversify their repeat options. They also have a setting where only specific flood packets (from specific senders) are repeated. In the case for repeaters connecting a local region to a backhaul, it could be useful to repeat some packets only on the local antenna, others on both the local and the backhaul antenna -- e.g. by scope.

Regards!

Talie5in:
Is there currently any devices with multiple radios in the wild (they may be, just haven't seen any)

dreirund:
> Is there currently any devices with multiple radios in the wild (they
> may be, just haven't seen any)

There exists, but of course not in the wild with MeshCore since there
is no support.

Talie5in:
Any link to said hardware?

dreirund:
On Thu, 30 Apr 2026 01:22:29 -0700, @Talie5in wrote:

> Any link to said hardware?

As far as I stumbled upon:

* "LilyGo T-Lora Dual": https://lilygo.cc/en-us/products/t-lora-dual
* "LilyGo T-Lora Pager (LR1121 variant)":
  https://lilygo.cc/en-us/products/t-lora-pager?variant=51587138683061

(OK, that's only one LoRa chip, but dual band with two antennae, which
could serve as well as "dual radio Repeater", although maybe not both
active at the same time.)

I did not research further because no MeshCore support yet.

Talie5in:
T Lora Dual is a great example thanks

[...]

PR #2198, opened 2026-03-30
Add MuziWorks Base Duo board support (nRF52840 + LR1121) 🤖🤖

zaycev:
## Summary                                                                                                                                                                        
                                                                                                                                                                                     
  - Add board support for MuziWorks Base Duo (Elecrow nRFLR1121 module: nRF52840 + Semtech LR1121, sub-GHz LoRa)
  - New `CustomLR1121` / `CustomLR1121Wrapper` radio helpers (RadioLib's LR1121 class extends LR1120, not LR1110 — reusing CustomLR1110 fails at chip identification)                
  - 5 build environments: companion_radio_ble, companion_radio_usb, repeater, room_server, terminal_chat                                                            
  - Board-specific config: 2-DIO RF switch table (DIO5+DIO6), TCXO 3.0V, 8MB QSPI flash (W25Q32JV), SoftDevice S140 v6.1.1, UF2 bootloader                                          
                                                                                                                                                                                     
  ## Hardware tested                                                                                                                                                                
                                                                                                                                                                                     
  - Radio init (LR1121 sub-GHz, 22 dBm TX power)                                                                                                                                    
  - BLE pairing via MeshCore companion app                                                                                                                                          
  - LoRa TX/RX with good RSSI and SNR    
  - Battery ADC reading                                                                                                                                                              
  - QSPI flash persistence across reboots                                                                                                                                            
  - TX LED indicator                    
                                                                                                                                                                                     
  ## Note                                                        
                                                                                                                                                                                     
  This PR was drafted by Claude Code, then manually reviewed, fixed, and tested on physical hardware.
                                                                                                                                                                                     
  ## Test plan
                                                                                                                                                                                     
  - [x] All 5 build environments compile                          
  - [x] Flashed and verified on physical MuziWorks Base Duo board                                                                                                                    
  - [x] BLE companion app connects and communicates              
  - [x] LoRa packets sent and received on sub-GHz band                                                                                                                              
  - [x] Settings persist across reboots (QSPI flash)

PR #2739, opened 2026-06-10
feat: Add NiceRF LoRa2021 (LR2021 Gen 4) variant support

c03rad0r:
## Summary

Adds MeshCore support for the **NiceRF LoRa2021** module (Semtech LR2021 Gen 4) on ESP32-C3_Mini_V1 development board.

## Hardware

| Component | Details |
|-----------|---------|
| **MCU** | ESP32-C3_Mini_V1 (Maker go, 22.52x18mm, USB-C, 4MB flash) |
| **Radio** | NiceRF LoRa2021 (Semtech LR2021 Gen 4, 19.72x15mm, 18-pin) |
| **Antenna** | Wire dipole (~8.2cm for 868 MHz Sub-GHz) |
| **Cost** | ~$8 total (MCU + radio module) |
| **Frequency** | Sub-GHz: 150-960 MHz + 2.4 GHz (dual-band) |

## RadioLib Class

Uses the `LR2021` class (extends `LRxxxx` → `PhysicalLayer`). This is a **separate** RadioLib class from `LR1121` (which extends `LR11x0`). RadioLib v7.6.0 (already in MeshCore deps) includes LR2021 support.

**Key differences from SX1262:**
- IRQ on DIO9 (not DIO1) — set via `irqDioNum = 9` + `setDioFunction(9)`
- `setRxBoostedGainMode(uint8_t level)` takes 0-7, not `bool`
- No DIO2-as-RF-switch (NiceRF handles internally)
- Crystal oscillator (XTAL), not TCXO — `tcxoVoltage = 0`
- `getPacketLength()` override handles header errors (same as LR1110 pattern)

## Known Issue: Arduino SPI

The Arduino `SPIClass` on ESP32-C3_Mini_V1 returns all-zero SPI responses, causing `RADIOLIB_ERR_SPI_CMD_FAILED` (-707). The same hardware works with ESP-IDF's `spi_bus_initialize()`.

This variant includes `EspIdfHal.h` — a custom RadioLib HAL that bypasses Arduino `SPIClass` and uses ESP-IDF's `spi_bus_initialize()` / `spi_device_polling_transmit()` directly. This is the same approach used in our ESP-IDF tracker firmware.

**Root cause hypothesis**: The ESP32-C3 has one SPI peripheral (FSPI/SPI2_HOST). The Arduino SPIClass may not properly reconfigure FSPI when the SPI pins differ from the default boot-time configuration (our board uses GPIO2/6/7 instead of default SPI pins).

If anyone has insight into why Arduino SPI fails on this board, I would love to fix this properly and remove the custom HAL.

## Build Targets

All verified building against `dev` branch:

| Target | Flash | RAM | Status |
|--------|-------|-----|--------|
| `LR2021_companion_radio_usb` | 542 KB (41%) | 140 KB (43%) | ✅ |
| `LR2021_repeater` | — | — | ✅ |
| `LR2021_room_server` | — | — | ✅ |
| `LR2021_secure_chat` | — | — | ✅ |
| `LR2021_kiss_modem` | — | — | ✅ |

All fit within ESP32-C3 4MB flash. Reference build (`Heltec_v3_companion_radio_ble`) passes — no regressions.

### Radio Helpers (in `src/helpers/radiolib/`)
- `CustomLR2021.h` — Extends `LR2021`, adds `std_init()`, `isReceiving()`, `getPacketLength()` override
- `CustomLR2021Wrapper.h` — Extends `RadioLibWrapper`, implements `setParams()`, `doResetAGC()`, `getSpreadingFactor()`

### Board Variant (in `variants/nicerf_lr2021/`)
- `platformio.ini` — 5 build environments with LR2021 pin mapping
- `target.h` / `target.cpp` — Radio init via EspIdfHal
- `NiceRFLR2021Board.h` — Board class (deep sleep, battery ADC)
- `EspIdfHal.h` — ESP-IDF SPI HAL (documented workaround)

### Board Definition (shared)
- `boards/esp32c3_supermini.json` — ESP32-C3 SuperMini V1 board definition
- `variants/esp32c3_supermini/pins_arduino.h` — Pin mapping (D0-D10 = GPIO0-GPIO10)

## Testing

- [x] All build targets compile on `dev` branch
- [x] No regressions on reference builds
- [x] Flash to ESP32-C3_Mini_V1 + NiceRF LoRa2021 hardware
- [x] Radio init succeeds
- [x] Noise floor calibration: -102 dBm stable
- [x] Companion connects via meshcore-cli v1.5.7
- [x] Advert TX verified
- [ ] Two-device encrypted chat (second board in progress)
- [x] Community node reception test (Berlin, no nearby nodes found indoors)

## Related

- Related to #2678 — LR2021 also supports 2.4 GHz LoRa (future work)
- Related to #2441 — LR2021 is dual-band (Sub-GHz + 2.4 GHz), relevant to multi-radio
- Note: LR2021 is a different RadioLib class from LR1121 (PRs #2696, #2198) — not conflicting

## Pin Mapping

```
ESP32-C3_Mini_V1  →  NiceRF LoRa2021
3V3               →  Pin 1 (VCC)
GND               →  Pin 2,8,11,12,18
D2 (GPIO2)        →  Pin 3 (MISO)
D7 (GPIO7)        →  Pin 4 (MOSI)
D6 (GPIO6)        →  Pin 5 (SCK)
D10 (GPIO10)      →  Pin 6 (NSS/CS)
D4 (GPIO4)        →  Pin 7 (BUSY)
D3 (GPIO3)        →  Pin 14 (RST)
D5 (GPIO5)        →  Pin 15 (DIO9/IRQ)
```


c03rad0r:
**Community mesh test (Berlin)**: Prior to opening this PR, I tested the LR2021 + ESP32-C3 against the live Berlin MeshCore community mesh:

- Discovered community relays and contacts via the MeshCore Android companion app
- Sent and received messages on the public group channel with users multiple hops away
- Confirmed multi-hop flood routing works correctly through the LR2021
- Validated interoperability with the standard EU MeshCore PHY (869.618 MHz, SF8, BW62.5, CR4/5)

ripplebiz:
Looks good.  Just a small suggestion:  I'd probably move the IDF Hal to /src/helpers/radiolib, as it's related only to RadioLib and could potentially be re-used by other variants

variants/nicerf_lr2021/target.cpp
```diff
@@ -0,0 +1,34 @@
+#include <Arduino.h>
+#include "target.h"
+#include "EspIdfHal.h"
+
+NiceRFLR2021Board board;
+
+static EspIdfHal hal(P_LORA_SCLK, P_LORA_MISO, P_LORA_MOSI);
```
oltaco, line 12: Did you try using ``static SPIClass spi(0)`` or one of the other SPI peripherals like FSPI or HSPI? Some of the ESP32 boards like C6 for example don't have the same SPI peripherals and it can cause SPI to fail silently and return 0s.

src/helpers/radiolib/CustomLR2021.h
```diff
@@ -0,0 +1,55 @@
...
+  bool _rx_boosted = false;
+
+  public:
+    CustomLR2021(Module *mod) : LR2021(mod) {
+      irqDioNum = 9;
+    }
+
+    size_t getPacketLength(bool update) override {
+      size_t len = LR2021::getPacketLength(update);
+      if (len == 0 && getIrqStatus() & RADIOLIB_LR11X0_IRQ_HEADER_ERR) {
```
oltaco: Likely not required unless it's confirmed that this workaround is necessary on the LR2021?

src/helpers/radiolib/CustomLR2021.h
```diff
@@ -0,0 +1,55 @@
...
+    int16_t setRxBoostedGainMode(uint8_t level) {
+      _rx_boosted = (level > 0);
+      return LR2021::setRxBoostedGainMode(level);
+    }
+
+    bool getRxBoostedGainMode() const { return _rx_boosted; }
+
+    bool isReceiving() {
+      uint32_t irq = getIrqStatus();
+      bool detected = (irq & RADIOLIB_LR11X0_IRQ_SYNC_WORD_HEADER_VALID) ||
```
oltaco, line 32: Should these be RADIOLIB_LR2021_IRQ_PREAMBLE_DETECTED and RADIOLIB_LR2021_IRQ_LORA_HEADER_VALID?

oltaco (changes requested):

src/helpers/radiolib/CustomLR2021.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include <RadioLib.h>
+#include "MeshCore.h"
+
+class CustomLR2021 : public LR2021 {
+  bool _rx_boosted = false;
+
+  public:
+    CustomLR2021(Module *mod) : LR2021(mod) {
+      irqDioNum = 9;
```
oltaco: irqDioNum will need to be configured via platformio define instead of hardcoded, since the LR2021 allows connecting the interrupt do any of the DIO pins.

Probably ``-D LR2021_IRQ_DIO`` and change this to ``irqDioNum = LR2021_IRQ_DIO`` or something?

oltaco (changes requested):

c03rad0r:
Thanks for the detailed review @oltaco! I'm working through all 4 items:

1. **SPI**: Investigating `SPIClass(0)` / FSPI approach — will update with results
2. **getPacketLength override**: Reviewing whether LR2021 actually needs it — will remove if not confirmed necessary
3. **IRQ constants**: Will switch to `RADIOLIB_LR2021_IRQ_*` chip-specific constants
4. **irqDioNum**: Will make configurable via `-D LR2021_IRQ_DIO` platformio build flag

Will push updates to this branch soon.

c03rad0r:
Thanks again @oltaco and @ripplebiz! All review items are now addressed and tested on real hardware. Latest commit: 05ad173a.

**Review items resolved:**

1. **SPI — `SPIClass(0)` (oltaco):** Switched to `static SPIClass spi(0)` (FSPI) as default. EspIdfHal is now opt-in via `-D USE_ESPIDF_HAL`. (e5a0e5b)

2. **`getPacketLength` override (oltaco):** Removed entirely — not confirmed necessary on LR2021. (627909a)

3. **LR2021 IRQ constants (oltaco):** Now uses `RADIOLIB_LR2021_IRQ_PREAMBLE_DETECTED` / `RADIOLIB_LR2021_IRQ_LORA_HEADER_VALID` with `#ifdef` fallback to `LR11X0_IRQ_*`. (627909a)

4. **Configurable IRQ DIO (oltaco):** `-D LR2021_IRQ_DIO=9` in platformio.ini, `irqDioNum = LR2021_IRQ_DIO` instead of hardcoded. (627909a, 684e9d9)

5. **EspIdfHal location (ripplebiz):** Moved to `src/helpers/radiolib/EspIdfHal.h` — reusable by other variants. (05ad173a)

**Build verification:** All 7 LR2021 targets + 2 reference targets (Xiao_C3, Heltec_v3) compile clean. No regressions.

**2-device hardware test:** Two ESP32-C3 + LR2021 boards, ~10cm apart, EU MeshCore PHY (869.618 MHz, SF8, BW62.5):
- Board #2 (repeater): boots, generates Ed25519 identity (53A53B5D...), starts broadcasting adverts
- Board #1 (companion): boots, receives Board #2's advert packet at 19s — cross-board flood routing confirmed
- Previously validated against live Berlin community mesh (multi-hop, encrypted chat)

Ready for another pass when you have a moment.

c03rad0r:
Full hardware test evidence (annotated serial logs, build matrix, wiring diagram, test timeline) published on Nostr:

https://npub1l06qpv6kr4l4sfjqm30jcd0jdtqlrel5nhhdtxfprjdu6dfyns0qnghahp.nsite.lol/

Includes 2-device cross-board flood routing confirmation, all 9 build targets, and the EspIdfHal explanation.

variants/nicerf_lr2021/target.cpp
c03rad0r: Yes — switched to `static SPIClass spi(0)` (FSPI) as the default in e5a0e5b. EspIdfHal is now opt-in via `-D USE_ESPIDF_HAL` only for boards where Arduino SPI returns zeros. Both paths tested on hardware (2-device smoke test, cross-board advert exchange confirmed). See commit e5a0e5bf1656.

src/helpers/radiolib/CustomLR2021.h
c03rad0r: Removed entirely in commit 627909ad. Confirmed not necessary on LR2021 — the header-error edge case it worked around does not manifest on this chip.

src/helpers/radiolib/CustomLR2021.h
c03rad0r: Done — switched to RADIOLIB_LR2021_IRQ_PREAMBLE_DETECTED and RADIOLIB_LR2021_IRQ_LORA_HEADER_VALID in 627909ad, with an #ifdef fallback to LR11X0_IRQ_* for older RadioLib builds that do not define the LR2021-specific macros.

src/helpers/radiolib/CustomLR2021.h
c03rad0r: Done — now configurable via -D LR2021_IRQ_DIO=9 in platformio.ini. The constructor reads irqDioNum = LR2021_IRQ_DIO instead of hardcoding 9. Commits 627909ad + 684e9d9.

dkmr27:
Tried your wrapper today with a promicro type setup. All working well but my module is different, Ebyte E80-900M2212S and it has a TCXO. It would be nice if you can include provision in the wrapper to set a TCXO voltage if it's defined by the variant.

Also tested 2.4Ghz which is working but has the same SX128x compatibility issues (LDRO) as does the LR1121.

fkallay1:
We are running what looks like the same module on a different MCU - a Seeed XIAO
nRF52840 with the NiceRF LoRa2021F33-2G4 (18-pin, dual-band) - so a couple of things we
measured here may be useful, and there is some overlap worth flagging early.

**The front-end DIOs do need programming.** The description says "No DIO2-as-RF-switch
(NiceRF handles internally)". They are internal in the sense that they are not brought
out to the MCU, but they are LR2021 DIOs wired to the front-end on the module PCB, and
the chip has to be told to drive them - `setRfSwitchTable()` writes that mode mask into
the chip, it does not touch MCU GPIOs. The module datasheet says so directly, in the note
under the sensitivity table:

> Note: For the 2.4GHz LNA (DIO5), it should normally be set to high level (Bypass OFF).
> If set to low level, sensitivity will decrease by 12dB (Bypass ON).

On the sub-GHz path the one that matters is DIO6, the PA enable, active during TX_LF.
Without it the external PA never turns on and you transmit with the bare chip output.
That one is not spelled out in the datasheet - it comes from NiceRF's own demo code
(LoRa/Core/Src/lr2021.c), which programs DIO6 for sub-GHz and DIO5/DIO7/DIO8 for the
2.4 GHz side, and their support confirmed the mapping for us.

One trap if you add it: `LR2021::setRfSwitchTable()` indexes the accumulated config by
the table column instead of the DIO number, so the pin array has to stay dense and
ordered from DIO5 - listing DIO6 on its own writes a zero config and the PA stays off
anyway.

**The TCXO is very likely the cause of your -707.** The description lists "Crystal
oscillator (XTAL), not TCXO - tcxoVoltage = 0", but the module datasheet has this as a
headline feature on page one:

> Industrial-grade TCXO Crystal Oscillator 0.5PPM

We configure 3.3 V (matching `LR20XX_SYSTEM_TCXO_CTRL_3_3V` in NiceRF's demo), and
`begin()` succeeds with it - we print the accepted value at boot and it reads 3.30 V, so
no fallback was taken. And -706/-707 is exactly what a wrong TCXO configuration returns,
which is why the LR2021 support already in `dev` retries `begin()` with 0.0 V on those
two codes. Since you get -707 with `tcxoVoltage = 0`, that is worth ruling out before
attributing it to the Arduino SPI driver - it is a one-line test. It may of course be
both.

Separately, and unrelated to the zeros you see: there is a race in the LR11x0/LR2021
read path where the reply to a get command can be read before the chip has prepared it,
and the chip then returns its status stream instead. It shows up as a bogus received
length. We reported it upstream to RadioLib with a reproduction. Mentioning it only
because it lives in the same code path you had to work around.

**A note on the power dial.** With RadioLib's stock PA table the lowest step already
drives this module to roughly +19 dBm at the antenna, so `LORA_TX_POWER=22` means about
+30 dBm once the PA is enabled - which is the datasheet maximum for 868/915 (1 W,
< 800 mA at 5 V; the 2 W headline figure is the 433/470 band, max 33 dBm - hence the F33
in the part number). That is also well above the limit of most 868 sub-bands. We ended up
remapping the dial so the number means dBm at the module output, which keeps `set tx`
consistent with MeshCore's other high-power boards.

**What we are preparing.** Two pieces, and the first one is deliberately
board-independent so it can be shared: a `NiceRF_LoRa2021F33` support header under
`src/helpers/radiolib/` holding the module specifics only - RF switch table, the remapped
PA drive table, the PRAM patch loader and a boot report - plus a variant for the XIAO
nRF52840. The header takes no MCU assumptions, so your ESP32-C3 board should be able to
include it unchanged and get the PA enabled without duplicating any of it. Happy to
coordinate so we are not solving the same module twice.

Last practical thing: `src/helpers/radiolib/CustomLR2021.h` and `CustomLR2021Wrapper.h`
have landed in `dev` in the meantime, so a rebase will drop those two files from this PR
and leave the variant, the board json and the HAL.

c03rad0r:
Rebased onto current dev and slimmed down now that shared LR2021 support has landed via #3115 / #3112 / #3146 / #3218. What remains is only the NiceRF-specific pieces (7 files, +553):

- variants/nicerf_lr2021/ — board definition for the plain NiceRF LoRa2021 module (160 mW, chip output only, no PA front-end) on ESP32-C3
- boards/esp32c3_supermini.json + variants/esp32c3_supermini/pins_arduino.h — C3 SuperMini host mapping (SPI on non-default pins)
- src/helpers/radiolib/EspIdfHal.h — opt-in ESP-IDF SPI path (-D USE_ESPIDF_HAL) for boards where Arduino SPIClass returns zeros; default remains static SPIClass(0)/FSPI
- IRQ pin configurable via -D LR2021_IRQ_DIO (defaults to 9)

The rebase drops our CustomLR2021 copies in favor of the shared dev ones, as @fkallay1 anticipated. On his TCXO note: this variant targets the plain crystal-only module, so tcxoVoltage=0 is correct here; our -707 was the Arduino SPI driver (the ESP-IDF path fixed RX with TCXO settings unchanged) — his F33 work and @dkmr27's ProMicro success are great additions to the ecosystem.

On the earlier review points: FSPI default / opt-in workaround split kept; unnecessary CustomLR2021 workaround dropped with the shared file taking over; LR2021-specific IRQ macros now live in the shared file. All 7 PlatformIO environments compile green on this head.

@oltaco — could you take another look when you get a chance? Happy to adjust anything that's still off.

PR #2741, opened 2026-06-10
factory default address for INA260 sensor

matkam:
According to the [INA260 spec sheet](https://www.ti.com/lit/ds/symlink/ina260.pdf) section 8.5.3.1 (page 18), the factory default of A0 and A1 are tied to GND results in an address of 0x40 (1000000). The current firmware configuration assumes a less usual configuration, where A0 is tied to VCC and A1 to GND, resulting in an address of 0x41.

Tested on a Station G2. The sensor did not show up before the change, but now shows up as in channel 3 telemetry. For what it's worth, the sensor worked in Meshtastic as well with no changes.


variants/station_g2/platformio.ini
```diff
@@ -228,7 +228,7 @@ build_flags =
   -D WIFI_DEBUG_LOGGING=1
   -D WIFI_SSID='"myssid"'
   -D WIFI_PWD='"mypwd"'
-  -D OFFLINE_QUEUE_SIZE=256  
+  -D OFFLINE_QUEUE_SIZE=256
```
matkam, line 231: unrelated auto formatting

PR #2742, opened 2026-06-10
feat(techo-lite): add non-shell repeater and room server environments

jirogit:
## Summary

Adds missing build environments for LilyGo T-Echo Lite Without Shell (H747-01/H747-02/H747-04), and fixes a missing TCXO voltage flag in the existing non-shell USB companion environment.

### New environments
- `LilyGo_T-Echo-Lite_non_shell_repeater`
- `LilyGo_T-Echo-Lite_non_shell_room_server`

The standard `LilyGo_T-Echo-Lite_*` envs work correctly on non-shell hardware (confirmed on H747-02), but dedicated `non_shell_*` envs make the intent explicit and allow the web flasher to surface these roles as a distinct device entry. ADVERT_NAME uses `-NS` suffix to distinguish non-shell nodes on the network from display-equipped units.

### Bug fix
- `LilyGo_T-Echo-Lite_non_shell_companion_radio_usb`: added missing `-D SX126X_DIO3_TCXO_VOLTAGE=1.8`. All other non-shell envs had this flag; the USB variant was the only one missing it, causing the TCXO to fall back to 1.6V and potentially preventing radio init.

### Housekeeping
- Moved `LilyGo_T-Echo-Lite_kiss_modem` entry before the non-shell block for readability.

## Hardware tested
- H747-02 (T-Echo Lite Without Shell, 915MHz)
- `non_shell_repeater`: build + hardware verified ✅
- `non_shell_room_server`: build + hardware verified ✅
- `non_shell_companion_radio_usb`: build + hardware verified ✅ (after TCXO fix)

### DIFF
--- a/variants/lilygo_techo_lite/platformio.ini
+++ b/variants/lilygo_techo_lite/platformio.ini
@@ -98,6 +98,38 @@ lib_deps =
   ${LilyGo_T-Echo-Lite.lib_deps}
   densaugeo/base64 @ ~1.4.0
 
+[env:LilyGo_T-Echo-Lite_kiss_modem]
+extends = LilyGo_T-Echo-Lite
+build_src_filter = ${LilyGo_T-Echo-Lite.build_src_filter}
+  +<../examples/kiss_modem/>
+
+[env:LilyGo_T-Echo-Lite_non_shell_repeater]
+extends = LilyGo_T-Echo-Lite
+build_src_filter = ${LilyGo_T-Echo-Lite.build_src_filter}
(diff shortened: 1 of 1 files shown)


jirogit:
Rebased onto upstream/dev (TCXO voltage conflict resolved — adopted upstream's 3.0V per LILYGO's updated pinout, PR #2742's own 1.8V addition was for a previously-unset env block, now superseded). Build+flash tested on T-Echo Lite Non-Shell H747-02: non_shell_repeater and non_shell_room_server both confirmed working.

Issue #2754, opened 2026-06-12
[Feature request] Built-in debug bot for the companion radio firmware (!ping, !path, ...)


zfouts:
## Use-case / problem

  A lot of people on the mesh are using things like `pathbot`, where you send
  `!path` in a #bot channel and it tells you how your message got there. This is
  really valuable for knowing how good the pathing is in your local mesh.

  I did not want to rely on running a separate radio for this, hooked into pyMC
  or similar. I thought, why not build it directly into the firmware and skip
  needing yet more hardware and software? The radios can handle this simple
  task, they have access to all of the same data natively.

  ## What I'm proposing

  An opt-in firmware variant (`<board>_companion_radio_ble_bot`) that is just
  the stock BLE companion plus a small responder. When the node gets a direct
  message starting with `!` it replies over the mesh with diagnostics, and the
  message still gets forwarded to the phone app like normal.

  Commands so far: `!ping` (the requester's SNR/RSSI/hop count into the node),
  `!rf`, `!path` (hop hashes resolved to contact names), `!status`, `!stats`,
  `!neighbors`, `!uptime`, `!time`, `!ver`.

  It can also answer in group channels, but only ones on an allow-list, which
  is empty by default. So no spamming Public.

  Config is local only, through the existing custom vars frames, so nobody on
  the mesh can mess with your bot. With meshcore-cli:

```
      set bot 0|1                      # enable/disable
      set bot.channels "#test,#bots"   # channel allow-list
      get custom
```

  Settings persist to flash.

  ## Keeping it polite on the airwaves

  Replies are rate limited three ways: a global minimum gap between replies, a
  per-sender cooldown, and a hard cap per minute. A flood of `!` messages won't
  turn the node into an airtime hog.

  Replies never start with `!` so bots can't trigger each other, and unknown
  commands are ignored silently, so typos and other bots' prefixes generate no
  traffic at all.

  ## Implementation notes

  I have this working already, with zero changes to existing files. Everything
  lives in `examples/companion_radio_bot/` plus one env file, using the stock
  companion's public API and a couple of link-time aliases. Builds clean for
  8 boards across ESP32 and nRF52 (Xiao S3 WIO, Xiao nRF52 WIO, Heltec V3/V4,
  RAK4631, RAK WisMesh 1W Booster, T1000-E, Heltec T114), and adding another
  board is about 12 lines of platformio.ini.

[...]

PR #2755, opened 2026-06-12
Add companion radio debug bot firmware variant (multi-board)

zfouts:
Implements the built-in debug bot discussed in #2754 

  ## What this adds

  A new opt-in firmware type: the stock BLE companion with a small mesh debug
  bot built in, for range testing and checking pathing in your local mesh.
  When the node gets a direct message starting with `!` it replies over the
  mesh with diagnostics, and the message still gets forwarded to the phone app
  like normal.

  Commands: `!ping` (the requester's SNR/RSSI/hop count into the node), `!rf`,
  `!path` (hop hashes resolved to contact names), `!status`, `!stats`,
  `!neighbors`, `!uptime`, `!time`, `!ver`.

  It can also answer in group channels, but only ones on an allow-list, which
  is empty by default. So no spamming Public.

  ## Keeping it polite on the airwaves

  Replies are rate limited three ways: a global minimum gap between replies, a
  per-sender cooldown, and a hard cap per minute. A flood of `!` messages won't
  turn the node into an airtime hog. Replies never start with `!` so bots can't
  trigger each other, and unknown commands are ignored silently.

  ## Runtime configuration

  No mesh-side config, so nobody on the mesh can mess with your bot. It's
  controlled through the existing custom vars frames (CMD_GET_CUSTOM_VARS /
  CMD_SET_CUSTOM_VAR), e.g. with meshcore-cli:
```
      set bot 0|1                      # enable/disable replies
      set bot.channels "#test,#bots"   # channel allow-list
      get custom
```
  Settings persist to flash and survive reboots.

  ## Design: zero changes to existing files

  Everything lives in `examples/companion_radio_bot/` plus one env file
  (`variants/bot_envs/platformio.ini`). No upstream file is modified, so this
  can't break any existing build:

  - `main.cpp` includes the stock companion `main.cpp` with `MyMesh` renamed
    to a `BotMesh` subclass that overrides `onMessageRecv` /
    `onChannelMessageRecv`. The env links the original symbol back with
    `-Wl,--defsym,the_mesh=the_bot_mesh`.
  - `BotTarget.cpp` compiles the board's stock `target.cpp` with the global
    `sensors` object swapped for a thin subclass that adds the two custom
    vars. Works unmodified across boards, including the T1000-E's custom
    `T1000SensorManager`.
  - The bot only uses `MyMesh`'s public API. Stock builds never compile any
    of these files.

  ## Boards

  `Xiao_S3_WIO`, `Xiao_nrf52`, `Heltec_v3`, `heltec_v4`, `RAK_4631`,
  `RAK_3401` (WisMesh 1W Booster), `t1000e`, `Heltec_t114`, each as
  `<board>_companion_radio_ble_bot`. Adding another board is about 12 lines
  of platformio.ini that extend its stock BLE companion env. Full docs in
  `examples/companion_radio_bot/BOT_README.md`.

  ## Testing

  - All 8 envs build clean (ESP32-S3 and nRF52840 toolchains).
  - Tested on real hardware: Xiao S3 WIO, Xiao nRF52 WIO, Heltec V4, and the
    RAK WisMesh 1W Booster (RAK3401 + RAK13302).
  - Verified in the linked ELFs that `sensors` resolves to the wrapper and
    `the_mesh` to the bot subclass on both architectures.

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -1981,7 +1981,7 @@ void MyMesh::handleCmdFrame(size_t len) {
         sendPacket(pkt, priority, 0);
         writeOKFrame();
       } else {
-        _mgr->free(pkt);
+        releasePacket(pkt);
         writeErrFrame(ERR_CODE_ILLEGAL_ARG);
       }
     } else {
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -1981,6 +1981,7 @@ void MyMesh::handleCmdFrame(size_t len) {
         sendPacket(pkt, priority, 0);
         writeOKFrame();
       } else {
+        _mgr->free(pkt);
         writeErrFrame(ERR_CODE_ILLEGAL_ARG);
       }
     } else {
--- a/examples/companion_radio_bot/BOT_README.md
+++ b/examples/companion_radio_bot/BOT_README.md
@@ -1,0 +1,163 @@
+# Companion Radio + Built-in Debug Bot
+
+This is the standard MeshCore **BLE Companion** firmware with a built-in
+"debug bot". When the node receives a **direct text message** over the
+LoRa mesh whose text begins with `!`, it auto-replies over the mesh with useful
+debugging info — while still behaving as a normal companion (the incoming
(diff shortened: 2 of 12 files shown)


zfouts:
Closing as this should not be a core feature.

PR #2763, opened 2026-06-14
* fix for anon contacts when full

ripplebiz:
BaseChatMesh::contacts array now initialised with first 8 (MAX_ANON_CONTACTS) reserved 
ContactsIterator now starts at idx = MAX_ANON_CONTACTS, skipping the reserved slots
allocateContactSlot(true) now only considers first 8 (MAX_ANON_CONTACTS) slots
allocateContactSlot(false) now skips first 8  (MAX_ANON_CONTACTS) slots


oltaco:
👌

PR #2793, opened 2026-06-17
Ikoka Handheld: Fix LED state polarity

tuxmobil:
LEDs are lit with LED_STATE_ON == 0 as with all Xiao nRF52 boards.
This caused inverted LED behavior when code used the default PIN_LED + LED_STATE_ON defines, such as the Bluefruit Library.


entr0p1 (approved):
Looks good, matches Xiao nRF52840 which this board uses.

PR #2794, opened 2026-06-17
Add variant for ikoka handheld to use Ebyte E22P module

tuxmobil:
The original ikoka handheld was build for the Ebyte E22 module.
This commit includes the slight changes needed to adapt to the improved E22P module.
To allow both variants, a new build variant for the E22P was created.


pizkesh:
I sucessfully built and flashed my E22P based ikoka handheld with this commit, thank you for making it available!

PR #2795, opened 2026-06-17
[Feature request] Disable blue LED during Bluetooth connection

tuxmobil:
To save power, do not indicate an active Bluetooth connection with a steadily lit blue LED.


rolandwhoffmannrh-alt:
same for the annoying flashing LED all 4 seconds on nRF52 based boards.

mmoll:
> same for the annoying flashing LED all 4 seconds on nRF52 based boards.

@rolandwhoffmannrh-alt Check out #2694

entr0p1:
In my opinion this type of thing comes down to preference of the user and really should be configurable from the app. That'll require Liam's input though.

PR #2797, opened 2026-06-18
Allow repeaters to restrict Flooded REQ, ANON_REQ, and RESPONSE packets

netprince17:
In our meshcore region, we are receiving a large number of flooded request, flooded anonymous request, and flooded response packets.  We think this is probably from people using automation which is not quite configured properly.  There are more flooded anonymous request and flooded response packets than flooded advertisement packets.  I have monitored the packets using the 'Rx Log' tool, as well as using the pyMC:Console.

image

This patch allows the repeater owner to configure a maximum number of hops for flooded request, flooded anonymous request, and flooded response packets using the CLI.  

set flood.max.response 8
set flood.max.request 8
set flood.max.anon.request 8

**All options default to 64**, so no changes to the default repeater behavior.

This patch was modeled after the recent flood.max.advert addition to meshcore found here:
https://github.com/meshcore-dev/MeshCore/commit/74adda6316f16d4c8ae605538e3198bb20fdfa28

I have been testing this patch for several days on two repeaters.  Using the test repeaters, I can still log-in to further away repeaters, less than the configured 8 hops away.

Dropping these packets will likely prevent logging-in to repeaters/rooms that are further away than the configured hops.  There may be other consequences of preventing flooded request, anon request, and response packets further away than the configured hops, but I have not found any other downside yet.


IoTThinks:
Should we use the same flood max setting for these requests too?

musznik:
What’s wrong with requests? I don’t write anything on the channels myself - I only try to get into repeaters or check the telemetry. I don’t like writing because it’s not always reliable; not everyone replies.

I’ve noticed a strong push to limit the number of hops. This affects people who live in mountain valleys, like me (or dense mesh). Sometimes, to get a signal through to a bigger town only about 20 km away, we need 8-10 hops.

netprince17:
> What’s wrong with requests? I don’t write anything on the channels myself - I only try to get into repeaters or check the telemetry. I don’t like writing because it’s not always reliable; not everyone replies.
> 
> I’ve noticed a strong push to limit the number of hops. This affects people who live in mountain valleys, like me (or dense mesh). Sometimes, to get a signal through to a bigger town only about 20 km away, we need 8-10 hops.

Thanks for the comment... there is not anything inherently wrong with flooded requests, just like there is not anything inherently wrong with flooded advertisements.  In this case, a few users have setup automations which flood the network with flooded requests and flooded replies every minute or two.  

If it were up to me, I would set the defaults max-hops on flooded requests and flooded responses to 8 to match the new default on flooded adverts, but considering there may be unforeseen consequences, leaving the default set to 64 keeps the current behavior by default.

SaschaKt:
> What’s wrong with requests? I don’t write anything on the channels myself - I only try to get into repeaters or check the telemetry. I don’t like writing because it’s not always reliable; not everyone replies.
> 
> I’ve noticed a strong push to limit the number of hops. This affects people who live in mountain valleys, like me (or dense mesh). Sometimes, to get a signal through to a bigger town only about 20 km away, we need 8-10 hops.

Every login, also with direct path, results in an flood response. And this response is mostly Unscoped. So you get 28, 34 hops away responses and so on. I don't think that it would find the destination after 28 hops. Normally you are logging max 2-3 hops away to repeaters, more is not comfortable

Alain2019:
A suggestion : adding  flood.max.advert.repeater

Adverts from a repeater are not needed for message transport and don't need to go as far as other adverts.
Adverts from repeaters do take a lot of airtime, sometimes up to 50%.

lbibass:
This is definitely necessary for large networks to remain usable. This is a well placed repeater. 50ft AGL on the highest ground in the county. It's linked into a network from RDU to Nashville to Atlanta.


request, anon request, and response are 33.9% of the airtime on this repeater.
image


request, anon request, and response are 42.7% of the packets on this repeater. 
image

Most of this traffic is coming from upwards of 10-15 hops away. 
If just a few high elevation repeaters enabled scoping like this, it would massively cut down on the amount of traffic flooding large networks, and make them significantly more reliable.

PR #2799, opened 2026-06-19
Add LilyGo T-Watch S3 variant

djkazic:
ESP32-S3 + SX1262 companion-radio (BLE) support for the LilyGo T-Watch S3:
- AXP2101 PMU brought up via XPowersLib (enables the LoRa, display and sensor power rails)
- ST7789 240x240 display and FT5x06/FT6x36 capacitive touch via LovyanGFX
- on-device UI driven by touch, as the watch has no navigation buttons

Shared changes, both guarded/no-op for existing boards:
- DisplayDriver: add virtual getTouch() (default returns false; LGFXDisplay already implements it)
- companion ui-new: map touch taps to nav keys when UI_HAS_TOUCH is defined

Tested this with the official MeshCore app and was able to confirm that LoRA and BLE both work.

### DIFF
--- a/boards/lilygo_twatch_s3.json
+++ b/boards/lilygo_twatch_s3.json
@@ -1,0 +1,41 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "esp32s3_out.ld",
+      "partitions": "default_16MB.csv",
+      "memory_type": "qio_opi"
+    },
+    "core": "esp32",
+    "extra_flags": [
+      "-DARDUINO_USB_MODE=1",
+      "-DARDUINO_RUNNING_CORE=1",
+      "-DARDUINO_EVENT_RUNNING_CORE=1"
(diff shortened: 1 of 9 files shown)


djkazic:
I also have some power-saving tweaks locally, but was unsure if that was in scope for adding variant support. Let me know if I should push it.

More or less, it just defines a `LOOP_IDLE_YIELD_MS` var, and yields when idle so that the core can drop into `waiti` between events (rather than busy-spinning).

niks200:
I'm also really looking forward to the firmware release for the Lilygo T-Watch-S3. I'm currently using v11.0 for the T-Watch-S3 Plus, but there is no full compatibility and some functions do not work. 
If you need help testing it, let me do it too. I'm really looking forward to finally adding the firmware for my T-Watch-S3 to the official flasher.

Issue #2786, opened 2026-06-17
nRF - broken manual and automatic GPS time sync on repeaters


LennertVA:
T114 with active GPS module does not sync time via GPS:

- Automatically: uptime of 11+ hours, clock showing 2024 despite gps on, fix, 15 satellites
- Manual: `gps sync` returns `OK` but does not sync the clock

Finally `clock sync` sends it nicely back to 2026.

Currently running on 1.15. I'll test with 1.16.

Same behaviour reported specifically for RAK: https://github.com/meshcore-dev/MeshCore/issues/2359

LennertVA:
Confirmed on T114 with 1.16. No automatic GPS clock sync after 40 minutes; `gps sync` returns OK but does not sync the clock.

ckoehler:
I'm not entirely sure, just getting started in the codebase, but it could be that no clock is passed into the GPS setup [here](https://github.com/meshcore-dev/MeshCore/blob/main/variants/heltec_t114/target.cpp#L25).

If so, adding it to the constructor may be all that's required:

```c
MicroNMEALocationProvider nmea = MicroNMEALocationProvider(Serial1, &rtc_clock);
```

I think the RAK fix is a bit more involved.

GOTO-GOSUB:
This is interesting, I suspect that I have just run into a similar issue on my Heltec V4 repeater. It had been running for a couple of months and then ceased updating before disappearing completely from the maps. When I checked the clock it was reported as being ahead and attempting to re-sync the clock either timed out or informed me that the clock cannot go back (although simply issuing the "clock" command via the CLI returned a time that looked right to me). It took a combination of disabling the GPS module from the CLI, power cycling (not reboot), re-enabling the GPS module and then re-syncing the clock to get the repeater to seem happy with the time again.

From Discord:

<img width="1011" height="420" alt="Image" src="https://github.com/user-attachments/assets/c029b41a-784b-4b13-b621-986f75e7fc1d" />

IoTThinks:
gps sync is not working for my T114+L76K too.
It saids ok but clock showed 1/3/2026 UTC.

gps shows on, active, fix, 9 sats.

IoTThinks:
> I'm not entirely sure, just getting started in the codebase, but it could be that no clock is passed into the GPS setup [here](https://github.com/meshcore-dev/MeshCore/blob/main/variants/heltec_t114/target.cpp#L25).
> 
> If so, adding it to the constructor may be all that's required:
[...]

PR #2810, opened 2026-06-22
Fixed GPS time sync for Heltec T114

IoTThinks:
Hello all,
This PR is to fix the issue in https://github.com/meshcore-dev/MeshCore/issues/2786
The GPS time sync is not working (not syncing up time) for Heltec T114.

Tested working for Heltec T114.
The time is autosynced fine after rebooting and locking GPS.
gps sync works fine too.

Credits to Ckoehler.


IoTThinks:
So fast.
Thanks a lot.

PR #2815, opened 2026-06-23
Fix blank OLED on T-Beam Supreme V3 (I2C bus init + 0x3D address)

### LINKED ISSUE #2609: T-Beam Supreme V3 OLED Display is not showing information
# Problem description
After installing the companion firmware v1.15.0 on a Lilygo T-Beam Supreme ESP32-S3 V3.0 the display stays dark.

The device came originally with Meshtastic preinstalled. It showed the Meshtastic information on its OLED display. Installing MeshCore Companion Bluetooth from https://flasher.meshcore.io/lilygo-t-beam-supreme-sx1262/ seemed to work, but the OLED display stays dark after rebooting.

# Analysis
Lilygo's documentation at https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/docs/en/t_beam_supreme/t_beam_supreme_hw.md states that the OLED has a another I2C address (0x3D instead of 0x3C), if the device has a MAG sensor QMC6310N at 0x3C.

# Proposed solution 
Extend the code to include 0x3D as I2C address.

1. change `bool SH1106Display::begin()` similar to the proposal in #2591, but extending `bool SH1106Display::i2c_probe(TwoWire &wire, uint8_t addr)` to probe for a SH1106 at the probed address instead.
2. remove `DISPLAY_ADDRESS` completely. This will also affect the code of wio-tracker-l1.
3. change `void TBeamBoard::scanDevices(TwoWire *w)` to add the additional address for debugging.

The best solution seems to check if at 0x3C is a SH1106 and try 0x3D, if it is not.

Please advise me, if I should just add another variant (e.g. lilygo_tbeam_supreme_SX1262_v3).

UltimateCodeWarrior:
## Summary

Fixes the blank OLED on the **LilyGo T-Beam Supreme V3** (ESP32-S3 + SX1262 + SH1106). Closes #2609 (see also #2591).

There are two independent bugs on this board, and both are needed for the display to work:

1. **The OLED I2C bus (`Wire`) is never initialized.** On the Supreme the PMU/RTC are on `Wire1` (GPIO 42/41, brought up by `XPowersLib`), while the OLED + BME280 + magnetometer are on the primary `Wire` bus (GPIO 17/18). Nothing ever calls `Wire.begin()` for it, so `Adafruit_SH110X` ends up on the wrong (default) pins.
2. **The OLED is at I2C address `0x3D`, not `0x3C`.** On this board `0x3C` is occupied by the QMC6310 magnetometer, so the panel was moved to `0x3D` (per LilyGo's [hardware docs](https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/docs/en/t_beam_supreme/t_beam_supreme_hw.md)). `SH1106Display::begin()` only ever tried `DISPLAY_ADDRESS` (`0x3C`).

## Changes

- `src/helpers/esp32/TBeamBoard.cpp` — initialize `Wire` on `PIN_BOARD_SDA`/`PIN_BOARD_SCL` for `TBEAM_SUPREME_SX1262` so the OLED bus is actually up before `display.begin()`.
- `src/helpers/ui/SH1106Display.cpp` — probe `0x3D` first (only the OLED uses it) and fall back to `0x3C`. This avoids the false-positive where `0x3C` ACKs because of the magnetometer.

Both changes are backwards-compatible: boards with the OLED at the standard `0x3C` and nothing at `0x3D` still resolve to `0x3C`.

## Evidence

Boot-time I2C scan on the affected board (Wire, GPIO 17/18):

```
I2C scan -> 0x3C 0x3D 0x77
```

- `0x3C` = QMC6310 magnetometer
- `0x3D` = SH1106 OLED
- `0x77` = BME280

After this change the display initializes at `0x3D` and renders normally.

## Test plan

- [x] T-Beam Supreme V3 (`T_Beam_S3_Supreme_SX1262_repeater`): OLED renders boot logo + home screen.
- [ ] Regression check on a board with the OLED at `0x3C` (e.g. standard T-Beam) to confirm the fallback path still initializes the display.


hansimgamr (approved):
This is working on my t-beam supreme currently. Thanks for this.

hansimgamr (approved):
Tested on my T-Beam Supreme V3 and the OLED now works correctly. LGTM, good to merge.

hansimgamr:
Re: the outstanding regression-check item — I only had a T-Beam Supreme on hand (the 0x3D board, already covered by the checked test-plan item), not a 0x3C board, so I did a code-level review of the fallback path instead of a hardware run:

```cpp
uint8_t addr = 0;
if (i2c_probe(Wire, 0x3D)) {
  addr = 0x3D;
} else if (i2c_probe(Wire, DISPLAY_ADDRESS)) {
  addr = DISPLAY_ADDRESS;
}
return addr && display.begin(addr, true);
```

On a standard board with the OLED at 0x3C and nothing at 0x3D:
1. `i2c_probe(Wire, 0x3D)` NACKs (no device there) → `endTransmission()` returns nonzero → probe returns `false`, `addr` stays 0.
2. Falls through to `i2c_probe(Wire, DISPLAY_ADDRESS)` (0x3C) → OLED ACKs → `addr = 0x3C`.
3. `display.begin(0x3C, true)` is called — identical to pre-PR behavior.

Probing an absent address is a standard non-destructive I2C operation, and `Wire` init is unchanged for non-Supreme boards (the new `Wire.begin()` call in `TBeamBoard.cpp` is gated to `TBEAM_SUPREME_SX1262` only), so the precondition for `SH1106Display::begin()` is the same as before this PR.

Logic checks out for the 0x3C fallback case. I don't currently have a plain 0x3C board (standard T-Beam / T-Beam 1W / Station G2 / etc.) to physically confirm, so this is a code review, not a hardware-verified test run — flagging that distinction for whoever merges.

IoTThinks:
No idea if it is better to define "DISPLAY_ADDRESS" in platformio.ini for T-Beam Supreme V3.
Then in the code ifdef DISPLAY_ADDRESS then use the defined address.

UltimateCodeWarrior:
Good suggestion. I pushed a follow-up commit (`0de2b4fc`) that moves the T-Beam Supreme V3 OLED address into the variant config with `DISPLAY_ADDRESS=0x3D`, and keeps `SH1106Display` generic by using the configured `DISPLAY_ADDRESS` instead of hard-coding a Supreme-specific 0x3D probe in the shared display driver.

IoTThinks:
This works perfect.
Thanks a lot.

IoTThinks:
@UltimateCodeWarrior Reported from my friends with T-Beam Supreme, this PR fixed the OLED for some and blacked out the OLED for others.

I guess due to this
Your two "identical" Supremes are almost certainly different sub‑variants (likely V3.0 vs V3.1, or VHF vs UHF V3.1) —
the OLED daughter‑boards have different solder‑bridge configurations. The MeshCore firmware happens to work on the
**0x3C unit** and fails on the **0x3D unit**.
https://github.com/meshcore-dev/MeshCore/pull/2591#issue-4484221337

So we need to fix this PR to include both 0x3C and 0X3D

UltimateCodeWarrior:
Thanks, that makes sense with the mixed Supreme sub-variants. I pushed another follow-up (`29b4d53a`) so the Supreme variant now declares both addresses:

- `DISPLAY_ADDRESS=0x3D`
- `DISPLAY_ADDRESS_ALT=0x3C`

`SH1106Display` stays generic: it tries the configured primary address first, then tries `DISPLAY_ADDRESS_ALT` when a variant provides one. That should cover both Supreme OLED solder-bridge configurations without making 0x3D a global SH1106 behavior.

IoTThinks:
Yah, true.
I will be helpful to all TBeam Supreme.
Thanks a lot for your effort.

hansimgamr:
Yeah we talked about this a little but it seems this was the best solution. Thanks for the feedback. I'm glad it's working for both

IoTThinks:
It can be like this.
```
#ifndef DISPLAY_ADDRESS_2
    return i2c_probe(Wire, DISPLAY_ADDRESS) && display.begin(DISPLAY_ADDRESS, true);
#else
    return (i2c_probe(Wire, DISPLAY_ADDRESS) && display.begin(DISPLAY_ADDRESS, true)) ||
           (i2c_probe(Wire, DISPLAY_ADDRESS_2) && display.begin(DISPLAY_ADDRESS_2, true));
#endif
```

hansimgamr:
> It can be like this.
> 
> ```
> #ifndef DISPLAY_ADDRESS_2
>     return i2c_probe(Wire, DISPLAY_ADDRESS) && display.begin(DISPLAY_ADDRESS, true);
> #else
>     return (i2c_probe(Wire, DISPLAY_ADDRESS) && display.begin(DISPLAY_ADDRESS, true)) ||
>            (i2c_probe(Wire, DISPLAY_ADDRESS_2) && display.begin(DISPLAY_ADDRESS_2, true));
> #endif
> ```

it's been merged already I believe.

IoTThinks:
@hansimgamr I know. This PR was merged with no-oled bug for old T-Beam Supreme.
I may fix for my variant and create a new PR to make it work for all T-Beam Supreme.

hansimgamr:
> Thanks, that makes sense with the mixed Supreme sub-variants. I pushed another follow-up (`29b4d53a`) so the Supreme variant now declares both addresses:
> 
> * `DISPLAY_ADDRESS=0x3D`
> * `DISPLAY_ADDRESS_ALT=0x3C`
> 
> `SH1106Display` stays generic: it tries the configured primary address first, then tries `DISPLAY_ADDRESS_ALT` when a variant provides one. That should cover both Supreme OLED solder-bridge configurations without making 0x3D a global SH1106 behavior.

does this address the issue @IoTThinks ?

IoTThinks:
@hansimgamr Let me merge this fix to mine.
https://github.com/UltimateCodeWarrior/MeshCore/commit/29b4d53af944a368e2524d3b2f08dc68e39eadce

Since the commit was done after this PR had been merged.
The commit will not be in dev unless there is a new PR for it (not sure if there is).

PR #2817, opened 2026-06-23
Add identity sign CLI command

vaizki:
Adds a command to sign arbitrary data with the repeater / room server private key. The use is to authenticate repeater ownership on analysis platforms etc. 

- data is limited to 64 bytes due to CLI input / output buffer limitations, didn't want to change them
- verification command `identity verify <message> <signature>` was planned but scrapped due to the same limitations
- not sure if "identity sign" is the best command name but might be an ok command namespace for future additions
- `mesh::Utils::fromHex()` returns 0 for errors like non-hexchars which is a bit ambiguous so a new helper `parseHexBytes()` was added, this could be inlined of course

Example: I want to view my node's detailed statistics so the platorm asks me to sign a random string (in this case "MeshCore"):

```
identity sign
  -> Usage: identity sign <data_hex>

identity sign 4d657368436f7265
  -> > E861F72FC8BBF2DBB7C4997257289CC0727D48D9A011DF6AF66F264802A1F4BF51FC864A6FF967A16AF6475A4898A23B4F8CF7F0B99A7735C3289F5C6C793802
```

After pasting the resulting signature to the platform it can be certain that I have access to the node and its private key, allowing me to link the repeater to my account.

Note: This function already exists in the KISS firmware as the SetHardware Extensions (0x06) -> SignData (0x04) TNC command and includes the verify command as well. So this brings the same feature to the repeater CLI.

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -27,6 +27,21 @@ static bool isValidName(const char *n) {
   return true;
 }
 
+static bool parseHexBytes(uint8_t* dest, size_t max_len, const char* hex, size_t* len_out) {
+  size_t hex_len = strlen(hex);
+  if (hex_len == 0 || (hex_len & 1)) return false;
+  if (hex_len / 2 > max_len) return false;
+
+  for (size_t i = 0; i < hex_len; i++) {
+    if (!mesh::Utils::isHexChar(hex[i])) return false;
+  }
+
+  size_t len = hex_len / 2;
+  if (!mesh::Utils::fromHex(dest, len, hex)) return false;
+  *len_out = len;
+  return true;
+}
+
 void CommonCLI::loadPrefs(FILESYSTEM* fs) {
   if (fs->exists("/com_prefs")) {
     loadPrefsInt(fs, "/com_prefs");   // new filename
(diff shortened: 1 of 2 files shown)


ViezeVingertjes:
> Note: This function already exists in the KISS firmware as the SetHardware Extensions (0x06) -> SignData (0x04) TNC command and includes the verify command as well. So this brings the same feature to the repeater CLI.

I almost threw in a shameless plug to promote the KISS firmware untill i read that last line, but that would only work in your situation (i think) if you would use PyMC Repeater already, as otherwise it wouldnt make sense to run KISS next to a regular Repeater.

LGTM.

src/helpers/CommonCLI.cpp
```diff
@@ -27,6 +27,21 @@ static bool isValidName(const char *n) {
...
+  size_t hex_len = strlen(hex);
+  if (hex_len == 0 || (hex_len & 1)) return false;
+  if (hex_len / 2 > max_len) return false;
+
+  for (size_t i = 0; i < hex_len; i++) {
+    if (!mesh::Utils::isHexChar(hex[i])) return false;
+  }
+
+  size_t len = hex_len / 2;
+  if (!mesh::Utils::fromHex(dest, len, hex)) return false;
```
ViezeVingertjes, line 40: In this case `fromHex` will always true, so can be removed. On the other hand; if someone ever decides to move code around, it might be a nice safegaurd... its up to you, can resolve this comment if you wish. 👍🏼

ViezeVingertjes (approved):
One comment on a check that _could_ be removed; but not a blocker for me personally as its a preference.

src/helpers/CommonCLI.cpp
vaizki: do you mean just removing the `if ( ... ) return false;` as it (currently) is an unreachable branch?

ViezeVingertjes: Yes, it should never reach `return false` in this case but its a preference thing, could also leave it in as a safeguard. Probably easier. 💯

PR #2819, opened 2026-06-23
fix: prevent ESP32 KISS modem stalls under USB backpressure

ViezeVingertjes:
Switch ESP32-S3 KISS modem environments to HWCDC and move outbound KISS writes to a non-blocking queued frame path so loop() and TX completion keep progressing when the host reads slowly. Add native backpressure regression tests and correct the native_kiss_modem test filter so this suite runs directly with pio test.

Also should resolve #2646 

| Scenario | Old behaviour | New behaviour |
|---|---|---|
| Flood commands, stop reading | `write()` blocks → `loop()` permanently wedged | queue drops overflow frames, `loop()` keeps running |
| Resume reading after flood | dead, PING never responds | recovers in ~1.3s as backlog drains |

### DIFF
--- a/.github/workflows/run-unit-tests.yml
+++ b/.github/workflows/run-unit-tests.yml
@@ -20,7 +20,7 @@ jobs:
         uses: ./.github/actions/setup-build-environment
 
       - name: Run Unit Tests
-        run: pio test -e native -vv
+        run: pio test -e native -e native_kiss_modem -vv
 
       - name: Upload Test Results
         # Upload test results even if the test step failed.
--- a/docs/kiss_modem_protocol.md
+++ b/docs/kiss_modem_protocol.md
@@ -41,7 +41,7 @@ Maximum unescaped frame size: 512 bytes.
 
 | Command     | Value  | Data               | Description                                                 |
 |-------------|--------|--------------------|-------------------------------------------------------------|
-| Data        | `0x00` | Raw packet         | Queue packet for transmission                               |
+| Data        | `0x00` | Raw packet         | Queue packet for transmission (one pending at a time)       |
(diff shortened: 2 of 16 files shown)


ViezeVingertjes:
What do you think about this solution, @agessaman?

I didn't want to butcher your branch/PR, so I made it a separate one. If we end up going with this approach, all the credit goes to you since you brought up the idea in the first place. Happy to add you as a co-author as well.

It's a bigger change in terms of LOC, but with KISS gaining popularity, I think it makes sense to make it as resilient as possible. Most of the changes in this PR are test related anyway. If you filter on the three files under `examples/kiss_modem`, the actual implementation changes aren't that extensive.

I've tested it both in code and on hardware, but if you get a chance to do the same and give it a try, it can get merged a bit quicker.

ViezeVingertjes:
In case this one gets merged, it might need a change on your side too or not? @rightup
(Due to changing DARDUINO_USB_MODE)

rightup:
Thanks, @agessaman

agessaman:
Sorry I didn't see this earlier. I'm vacationing at the moment and haven't been giving my email much time. I gave it a skim and I don't have any additional thoughts. Users will definitely appreciate being able to use those devices.

ViezeVingertjes:
Would be great if this one could go through @liamcottle @ripplebiz 💯 

It doesnt affect many devices; but the devices it does (Heltec v4 for example) are probably amongst the most popular ones for this purpose.

ViezeVingertjes:
Been running this for a while now; as stable as it gets. 💯

liamcottle:
> Been running this for a while now; as stable as it gets. 💯

Epic! Thanks for your work on this :)

PR #2838, opened 2026-06-24
GPS: poll up to 5s for cold-start detection instead of a single 1s check

### LINKED ISSUE #2258: GPS detection in initBasicGPS() too short — most repeater variants affected
## Problem

The GPS detection in `EnvironmentSensorManager::initBasicGPS()` uses a hardcoded 1-second delay before checking `Serial1.available()`:

```cpp
_location->begin();
_location->reset();    // 10ms reset pulse
delay(1000);           // only 1 second

#ifdef ENV_SKIP_GPS_DETECT
  gps_detected = true;
#else
  gps_detected = (Serial1.available() > 0);
#endif
```

Cold-starting GPS modules often need 2-5 seconds before they begin outputting NMEA sentences. When detection fails, `gps_detected` is set to `false`, which hides all GPS CLI commands and prevents GPS time sync entirely.

## Current workaround

The `ENV_SKIP_GPS_DETECT=1` and `PERSISTANT_GPS=1` build flags bypass detection and force GPS active. But only **3 out of ~20+ GPS-equipped variants** have these flags in their repeater builds:

- LilyGo TBeam 1W
- ThinkNode M5
- ThinkNode M1 (PR #2256)

Plus Station G2 (PR #2255). All other GPS-equipped repeater variants (Heltec V4, Heltec Tracker/V2, T096, T114, TBeam SX1262, TBeam Supreme, TDeck, T-Echo, SenseCap Solar, ProMicro, Nano G2 Ultra, ThinkNode M3/M6, GAT562 variants, etc.) are susceptible to GPS detection failure on cold start.

## Suggested fix

Instead of adding `ENV_SKIP_GPS_DETECT` to every variant, fix the detection itself. For example, a short retry loop:

```cpp
_location->begin();
_location->reset();

#ifdef ENV_SKIP_GPS_DETECT
  gps_detected = true;
#else
  gps_detected = false;
  for (int i = 0; i < 5 && !gps_detected; i++) {
    delay(1000);
    gps_detected = (Serial1.available() > 0);
  }
#endif
```

This gives GPS modules up to 5 seconds to start sending NMEA data, returning early as soon as data is detected (no wasted boot time for fast modules). The reset pulse (currently 10ms) could also be extended to 50-100ms for broader module compatibility.

An even more robust approach would be to not permanently disable GPS on detection failure — allow users to manually enable it via settings even if auto-detection missed it.

## Related

- #1426 — Update RTC from GPS
- #2255 — Station G2 repeater GPS fix (per-variant flag workaround)
- #2256 — ThinkNode M1 repeater GPS fix (per-variant flag workaround)
- #1503 — GPS not working on SenseCAP repeater firmware
- Reddit discussion: https://www.reddit.com/r/meshcore/comments/1qyoxh2/can_you_use_gps_on_a_repeater/

benskigomez:
## Problem
`EnvironmentSensorManager::initBasicGPS()` waits a fixed **1 second**, then decides the GPS is present only if NMEA data has already arrived. Cold-start GPS modules often need **2–5 seconds** before emitting their first sentence, so on a cold boot detection frequently fails.

A failed detection sets `gps_detected = false`, which **hides all GPS CLI settings and disables GPS time sync until the next reboot** — so the GPS appears permanently dead even though the hardware is fine. This affects the ~20 GPS-equipped variants that don't set the `ENV_SKIP_GPS_DETECT`/`PERSISTANT_GPS` bypass flags.

## Fix
Poll for up to ~5 seconds (in 250 ms steps) and return **as soon as any data appears**. Fast modules are detected almost immediately and add no boot delay; slow cold-start modules now get enough time. The `ENV_SKIP_GPS_DETECT` bypass is unchanged.

## Trade-off
Boards that have `ENV_INCLUDE_GPS` but no physical GPS will now wait the full ~5 s at boot (vs. 1 s before), since detection never succeeds. This seemed the right call given the alternative is GPS silently not working, but happy to shorten the cap if preferred.

## Testing
Builds clean (verified on a ThinkNode M6 companion build). The change is a drop-in replacement for the single check.

Fixes #2258.

### DIFF
--- a/src/helpers/sensors/EnvironmentSensorManager.cpp
+++ b/src/helpers/sensors/EnvironmentSensorManager.cpp
@@ -748,14 +748,18 @@ void EnvironmentSensorManager::initBasicGPS() {
     MESH_DEBUG_PRINTLN("No GPS wake/reset pin found for this board. Continuing on...");
   #endif
 
-  // Give GPS a moment to power up and send data
-  delay(1000);
-
-  // We'll consider GPS detected if we see any data on Serial1
+  // Give the GPS time to power up and emit its first NMEA sentence. Cold-start
+  // modules can take several seconds, so poll for up to ~5s instead of checking
+  // once after a fixed 1s delay -- returning as soon as any data appears, so
+  // fast modules add no boot delay (issue #2258).
 #ifdef ENV_SKIP_GPS_DETECT
   gps_detected = true;
 #else
-  gps_detected = (Serial1.available() > 0);
+  gps_detected = false;
+  for (int i = 0; i < 20 && !gps_detected; i++) {
+    delay(250);
+    gps_detected = (Serial1.available() > 0);
+  }
 #endif
 
   if (gps_detected) {


ViezeVingertjes (changes requested):
PR targets `main` instead of `dev`, approve when fixed. 👍🏼

ViezeVingertjes (approved):

PR #2844, opened 2026-06-25
Fix for Rx boosted gain on LR1110

oltaco:
This PR fixes Rx boosted gain not being applied on LR1110 and improves the error handling when setting Rx boosted gain.

Previously LR1110 was never getting Rx boosted gain set due to an overly narrow guard condition in MyMesh. Nobody noticed that the override wasn't getting called because the call was void and the default was no-op so there was no error to report.

I have changed setRxBoostedGain() and setRxBoostedGainMode() to return bool and the default is to return false. Now errors can propagate back to CommonCLI so that we know when radio.rxgain is unsupported or the radio returned an error.

Tested on LR1110 (T1000e) and (RAK 4631).

Confirmed building on SX1268 (Meshadventurer) and ESPNOW (Heltec V3), untested but should be fine.

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -573,13 +573,13 @@ void CommonCLI::handleSetCmd(uint32_t sender_timestamp, char* command, char* rep
     strcpy(reply, _prefs->disable_fwd ? "OK - repeat is now OFF" : "OK - repeat is now ON");
   } else if (memcmp(config, "radio.rxgain ", 13) == 0) {
     bool enabled = memcmp(&config[13], "on", 2) == 0;
+    _prefs->rx_boosted_gain = enabled;
+    savePrefs();
     if (_callbacks->setRxBoostedGain(enabled)) {
-      _prefs->rx_boosted_gain = enabled;
       strcpy(reply, "OK");
-      savePrefs();
     } else {
-      strcpy(reply, "Error: unsupported or rejected");
-    }    
+      strcpy(reply, "Error: unsupported");
+    }
   } else if (memcmp(config, "radio.fem.rxgain ", 17) == 0) {
     if (!_board->canControlLoRaFemLna()) {
       strcpy(reply, "Error: unsupported");
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -571,13 +571,15 @@ void CommonCLI::handleSetCmd(uint32_t sender_timestamp, char* command, char* rep
     _prefs->disable_fwd = memcmp(&config[7], "off", 3) == 0;
     savePrefs();
     strcpy(reply, _prefs->disable_fwd ? "OK - repeat is now OFF" : "OK - repeat is now ON");
-#if defined(USE_SX1262) || defined(USE_SX1268) || defined(USE_LR1110)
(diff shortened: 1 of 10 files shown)


variants/t1000-e/platformio.ini
```diff
@@ -44,8 +44,8 @@ build_flags = ${t1000-e.build_flags}
   -D ADVERT_LON=0.0
   -D ADMIN_PASSWORD='"password"'
   -D MAX_NEIGHBOURS=50
-;  -D MESH_PACKET_LOGGING=1
-;  -D MESH_DEBUG=1
+ -D MESH_PACKET_LOGGING=1
```
recrof: you probably forgot to comment those lines again...

platformio.ini
```diff
@@ -25,6 +25,7 @@ lib_deps =
   melopero/Melopero RV3028 @ ^1.1.0
   electroniccats/CayenneLPP @ 1.6.1
 build_flags = -w -DNDEBUG -DRADIOLIB_STATIC_ONLY=1 -DRADIOLIB_GODMODE=1
+  -D RADIOLIB_DEBUG_BASIC
```
recrof: same

oltaco:
Thanks @recrof didn't mean to include those 👍

src/helpers/CommonCLI.cpp
```diff
@@ -835,10 +837,8 @@ void CommonCLI::handleGetCmd(uint32_t sender_timestamp, char* command, char* rep
     sprintf(reply, "> %s", StrHelper::ftoa(_prefs->node_lat));
   } else if (memcmp(config, "lon", 3) == 0) {
     sprintf(reply, "> %s", StrHelper::ftoa(_prefs->node_lon));
-#if defined(USE_SX1262) || defined(USE_SX1268) || defined(USE_LR1110)
```
liamcottle, line 838: Only issue I can see, is that boards/radios that don't support RX Boosted Gain, might report this as `on` since the pref defaults to `1`?

oltaco: That's a fair point. At the moment there aren't any radios that don't support it, the main reason I added the error message is so we avoid the trap where turning it on and off reports OK but nothing actually happens, and then it seemed natural to also extend the error to the case where the call to the radio was rejected.

One approach would be to implement a register check in ``getRxBoostedGainMode()`` (the SX1262 wrapper already does this) and in the case of error check the result of ``getRxBoostedGainMode()`` and set the prefs to the result. Would also need to implement the same check in ``MyMesh.cpp`` where it gets setup at startup (in case it fails for some reason).

The problem with this is that if for some reason the radio fails to set boosted gain mode and returns an error, then your prefs are switched to disabled forever and you would never know that it had happened.

Maybe it's better to change the debug print that happens at startup in ``MyMesh.cpp`` so that instead of just reporting the preference for rx gain it reports the result of setting the preference?

liamcottle: > At the moment there aren't any radios that don't support it

Yeah, not a big deal. Was just a nitpick ;) No need to worry then.

ViezeVingertjes (approved):
LGTM; did open a PR to yours with a few suggestions but those are more style/cleanup, so not mandatory. 👍🏼 

_(Still having nightmares of the LR1110 deafness problems when i read that type haha)_

PR #2854, opened 2026-06-27
Add new variant LilyGo TLora V1

nica-f:
This PR add a new variant for the Lilygo TLora32 V1, the predecessor of the TLora32 V2. These devices are still around and useful.
Tested on real hardware, works as expected including the OLED display.


Black6spdZ:
even some v1.1/1.2 floating around that have TXCOs
TTGO Lora32 versions
t-fox jpg

Issue #1864, opened 2026-02-27
thinknode m3 doesnt have gps


mikerocklewitz:
Gps doesnt work on thinknode m3.  Not even detected.

IoTThinks:
@mikerocklewitz As instructed by the same bug for M6. https://github.com/meshcore-dev/MeshCore/pull/2863
You may try this and feedback if it fixes your GPS on M3.
https://github.com/IoTThinks/EasySkyMesh/releases/tag/PowerSaving16.1

<img width="291" height="103" alt="Image" src="https://github.com/user-attachments/assets/ae8229be-5213-4989-a5fd-a4d839d1e993" />

If it works, please help to put in your screenshot so we will push a PR to MeshCore.

PR #2863, opened 2026-06-29
thinknode_m6: don't drive GPS REINIT pin so the L76K is detected

### LINKED ISSUE #2862: ThinkNode M6 GPS not supported — CASIC L76K (AT6558) needs $PCAS config; UART TX hangs with SoftDevice active
## Summary
On the Elecrow ThinkNode M6, the MeshCore companion app reports the device has **no GPS / GPS not supported**, even though the M6 has a GPS module and it **works correctly on Meshtastic** (confirmed real position fixes with 8 satellites on the same unit). This is a firmware gap, not a hardware fault — and it almost certainly affects the **ThinkNode M1 and M5** as well (same GPS family).

## Root cause
The M6's "L76K" GPS is an **Airoha AG3335**. It does **not** auto-stream NMEA on power-up — it stays completely silent until it receives a **UART wake command** (`$PAIR382,1*2E`), and it must be **baud-probed** (9600 / 115200 / 38400) and then configured for NMEA output.

Meshtastic handles all of this in `src/gps/GPS.cpp` (the `GNSS_AIROHA` path: `wakeAirohaForActiveProbe()` sends `$PAIR382,1`, followed by `$PAIR021` identify and the NMEA-output config commands).

MeshCore's `MicroNMEALocationProvider` only powers the module (EN pin) and **reads** — it assumes the GPS free-runs NMEA. On an Airoha chip nothing ever arrives, so `EnvironmentSensorManager::initBasicGPS()` sees no data, sets `gps_detected = false`, and the GPS feature is hidden from the app/CLI.

## Evidence (on-hardware)
- With the GPS powered and `PIN_GPS_EN` / `PIN_PWR_EN` / `PIN_GPS_STANDBY` / `PIN_GPS_RESET` driven in **every** combination, there is **zero serial activity on any GPIO at any baud** — confirmed by both UART reads (15 s) and a baud-agnostic GPIO edge-count scan across all pins.
- Sending the Airoha wake (`$PAIR382,1`) at 9600/115200/38400 from a standalone sketch did not elicit a response either, so the full Meshtastic probe/config sequence (not just the wake) appears to be required.
- Meshtastic, which performs that sequence, gets fixes on the same board and pins (GPS EN = pin 6, RX = pin 3 default).

## Implementation note / caveat
A naive port that switches the GPS UART baud (`Serial1.end()/begin()`) **inside the companion BLE firmware hangs the nRF52** — re-initializing the UART while the SoftDevice is active locks up. So the Airoha bring-up needs **SoftDevice-safe** baud handling, not a quick patch.

## Affected boards
ThinkNode M6 (confirmed), and very likely M1 / M5 (all L76K / AG3335).

## Suggested fix
Add Airoha / AG3335 support to MeshCore's GPS bring-up: send the wake (`$PAIR382,1`), probe baud, and issue the NMEA-output config commands — portable from Meshtastic's `GPS.cpp`, with care around UART re-init under the SoftDevice. Happy to test on an M6.

benskigomez:
## Problem

On the ThinkNode M6, the L76K GPS was never detected — the companion app reported "no GPS / not supported", and on-device the GPS stayed at `sat 0`. The same sealed unit's GPS works fine under Meshtastic, so the hardware is good.

## Root cause

Pin 29 (`PIN_GPS_RESET`, which is the module's **REINIT** line on this board) was being driven **HIGH**, which holds the L76K silent — it never emits any NMEA, so detection (`Serial1.available()`) sees zero bytes and the GPS setting is hidden.

It was driven HIGH in two places:
- `variants/thinknode_m6/variant.cpp` `initVariant()` drove it HIGH at boot.
- Because `PIN_GPS_RESET` was defined, `MicroNMEALocationProvider::begin()` also drove it HIGH (`GPS_RESET` = 29, `GPS_RESET_FORCE` = LOW, so `begin()` writes `!LOW` = HIGH).

The Meshtastic M6 variant leaves this exact pin as a floating input and never drives it — which is why GPS works there.

## Verification (no logic analyzer; M6 is a sealed unit)

A standalone sketch powered the GPS and passively captured NMEA at 9600 baud while changing **one variable at a time**:

| pin 29 state | result |
| --- | --- |
| driven HIGH (current firmware) | **0 bytes** |
| left floating (INPUT) | **full NMEA stream** (`$GNGGA/$GNRMC/$GNGSA/$GNGSV`) |

An EN power-cycle was tried as well and made no difference — floating pin 29 alone is sufficient.

The GPS streams standard NMEA at 9600 by default and needs no `$PCAS`/config commands.

## Fix

- `variant.h`: define `GPS_RESET (-1)` so the location provider never touches pin 29.
- `variant.cpp`: stop driving pin 29; leave it floating.

This matches the Meshtastic M6 variant. Confirmed on hardware: GPS is now detected and the app shows live coordinates, altitude, and map position.

Fixes #2862.

## Note for maintainers

The same pattern may affect other Elecrow ThinkNode boards whose reset pin is labeled "REINIT" — notably the **M3** (`variant.h` has `PIN_GPS_RESET (25) // REINIT`), which has a matching open report (#1864). I don't have M3 hardware to verify, so it's intentionally left out of this PR.

### DIFF
--- a/variants/thinknode_m6/variant.h
+++ b/variants/thinknode_m6/variant.h
@@ -102,7 +102,12 @@
 #define PIN_GPS_RX              (2)
 #define PIN_GPS_TX              (3)
 #define PIN_GPS_EN              (6)    // EN
-#define PIN_GPS_RESET           (29)
+#define PIN_GPS_RESET           (29)   // REINIT - must FLOAT; driving it (esp. HIGH) silences the L76K
+// The M6's L76K streams NMEA on its own and must not have its REINIT pin driven.
+// Tell the location provider there is no reset pin so it never touches pin 29
+// (driving it HIGH holds the module silent). Matches Meshtastic, which leaves
+// this pin as an input. See variant.cpp (pin 29 is intentionally not configured).
+#define GPS_RESET               (-1)
 #define PIN_GPS_STANDBY         (30)   // STANDBY
 #define PIN_GPS_PPS             (31)
 #define GPS_BAUD_RATE           9600
--- a/variants/thinknode_m6/variant.cpp
+++ b/variants/thinknode_m6/variant.cpp
@@ -30,6 +30,8 @@ void initVariant() {
   digitalWrite(PIN_GPS_STANDBY, HIGH);
   pinMode(PIN_GPS_EN, OUTPUT);
   digitalWrite(PIN_GPS_EN, HIGH);
-  pinMode(PIN_GPS_RESET, OUTPUT);
-  digitalWrite(PIN_GPS_RESET, HIGH);
+  // PIN_GPS_RESET (pin 29 / REINIT) is intentionally left floating (input).
+  // Driving it HIGH holds the L76K silent, so it never streams NMEA and the
+  // firmware reports "no GPS". Letting it float lets the module run, matching
+  // the Meshtastic M6 variant. See GPS_RESET (-1) in variant.h.
 }


benskigomez:
**Likely the same bug on the ThinkNode M3 (#1864).** For a maintainer with M3 hardware:

`variants/thinknode_m3/variant.h` has `#define PIN_GPS_RESET (25)  // REINIT` (note the same "REINIT" labelling as the M6), and `variant.cpp` sets it as an OUTPUT. As on the M6, `MicroNMEALocationProvider::begin()` then drives that pin HIGH, which on the M6 is exactly what held the L76K silent. The open issue #1864 ("thinknode m3 doesnt have gps") matches the M6 symptom.

If that reproduces on an M3, the same one-line fix should apply: stop driving pin 25 / set `GPS_RESET (-1)` so the provider leaves it floating. I don't have an M3 to verify, so I've left it out of this PR.

variants/thinknode_m6/variant.h
```diff
@@ -102,7 +102,12 @@
 #define PIN_GPS_RX              (2)
 #define PIN_GPS_TX              (3)
 #define PIN_GPS_EN              (6)    // EN
-#define PIN_GPS_RESET           (29)
+#define PIN_GPS_RESET           (29)   // REINIT - must FLOAT; driving it (esp. HIGH) silences the L76K
+// The M6's L76K streams NMEA on its own and must not have its REINIT pin driven.
+// Tell the location provider there is no reset pin so it never touches pin 29
+// (driving it HIGH holds the module silent). Matches Meshtastic, which leaves
```
IoTThinks, line 108: I suggest to remove the comment "Matches Meshtastic, which leaves
// this pin as an input. See variant.cpp (pin 29 is intentionally not configured)." to make the code "neutral".

Let me test the PR.
Seems good.

IoTThinks:
Yeah, this PR is confirmed working.
It brings back GPS to M6.

fdlamotte:
I'm quite sure gps was working when I did the port

Unfortunately I don't have my unit here to do the test (stayed in France with my brother, I didn't want to carry items with batteries in my luggage)

So everyone confirm gps was not working and now works ... would be ok for me I think

IoTThinks:
@fdlamotte Likely due to variants with NO PULL vs. PULLUP resistor.

F4FPR:
ThinkNode M6 / repeater 1.17.0

The #2863 REINIT fix is present and GPS has successfully obtained
real fixes on this same device after upgrading to 1.17.0.

However, the GPS can later enter a persistent state:

gps
on, active, no fix, 0 sats

Observed for hours.

The following did NOT recover it:
- gps off / gps on
- software reboot
- reboot on battery
- reboot while externally powered by USB
- USB power maintained with active charging
- reflashing the exact same repeater 1.17.0 image via OTAFIX

The GPS antenna and hardware configuration were unchanged from the
previous successful fixes.

This suggests an intermittent GPS/UART/module state remains possible
after the #2863 REINIT fix.

PR #2864, opened 2026-06-29
OTA-over-LoRa: signed full/delta firmware updates (.mota) for ESP32 & nRF52 🤖🤖

vk496:
## TL;DR

Eventually upgradable. Low priority.

firmware OTA over LoRa. DHT/bittorrent propagation. Support ESP32 and nRF52. nRF52 requires [special bootloader to apply OTA.](https://github.com/vk496/Adafruit_nRF52_Bootloader_OTAFIX/releases/tag/dev-latest). Serve folder with multiple mOTA through WiFI/serial (motatool)

Tested with Heltec V3, RAK4631, T114

Test firmware: https://github.com/vk496/MeshCore/releases/tag/dev-latest
User Quick Start:  [`docs/ota_user_guide.md`](https://github.com/vk496/MeshCore/blob/feature%2Fota-lora/docs%2Fota_user_guide.md).
Dev specs: [`docs/ota_protocol.md`](https://github.com/vk496/MeshCore/blob/feature%2Fota-lora/docs/ota_protocol.md)

## What this adds

Over-the-air firmware updates that travel over the **existing LoRa mesh** — no
internet, BLE, or USB needed at the target node. A node discovers update sources
among its neighbours, fetches a signed image block-by-block, verifies it, and
applies it. Images can be **full** or **delta** against a known base (deltas are
far smaller — essential on LoRa's tiny bandwidth).

Motivation: repeaters/sensors are often deployed somewhere physically awkward to
reach. Today updating them means going to the device. This lets you push a signed
update through the mesh itself.

> Discussion/issue: <link the feature issue here>

## How it works

- **`.mota` container** — a signed update package: manifest + payload + a merkle
  tree over fixed-size blocks, plus a 56-byte `EndF` self-identity trailer baked
  into every firmware (target id, version, hw id, image hash). Spec in
  [`docs/ota_protocol.md`](docs/ota_protocol.md).
- **Discovery** (two-tier, anti-storm) — a tiny periodic `OTA_ADV` beacon
  (seeder + count + set-digest); interested peers send `OTA_QUERY` and build a
  catalog from `OTA_HAVE`. Jitter + overhear-suppression keep it quiet.
- **Transfer** — 1 KB logical blocks split into ~160 B `DATA` fragments + a merkle
  `PROOF`, reassembled and verified per block; resumable across reboots.
- **Apply** — ESP32 stages into the inactive A/B slot (stock partition scheme,
  **no bootloader change**); nRF52 applies the delta **in place** via a companion
  bootloader (see *Dependencies*). Refuses on hw-id mismatch (brick-safety).
- **Trust** — Ed25519 signatures + a per-node signer allowlist; `autofetch` and
  `autoinstall` policies are opt-in and conservative by default.
- **CLI** — `ota status | ls | get <#> | install | cancel | announce | self | …`.

## What's in the PR (layered for review)

The 13 commits build up in dependency order — container format → vendored decoder
→ transfer protocol → platform apply → folder relay → CLI/node integration →
build enablement + tooling → host packager → tests → docs → WiFi serve →
discovery tuning → nRF52 variant enablement. Each is independently reviewable.

Also included:
- **`motatool`** — a portable C++17 host tool to build/verify/inspect/serve/keygen
  `.mota` (cross-checked byte-for-byte against the Python reference).
- **Folder relay** — a node (or `motatool`) can serve a folder of `.mota` to peers
  over USB serial **or** WiFi/TCP (ESP32 companion: dedicated seeder port 5001 +
  an OTA text console on 5002, both coexisting with the phone-app port).
- **Tests** — native unit tests (`pio test -e native`), `motatool` ctest, and a
  Python reference suite; generated tables (`OtaTargets.h`, `mota_vectors.h`) are
  committed.
- **Docs** — protocol spec + a user guide.

## Platform support

- **ESP32** (A/B OTA): works with the stock partition scheme, no bootloader change.
- **nRF52** (in-place delta): enabled on the **14 existing variants** the companion
  bootloader supports (RAK4631, Heltec T114, LilyGo T-Echo family, ThinkNode,
  Wio-Tracker, Xiao nRF52, ProMicro, T1000-E, …). No new variants are introduced.

## Dependencies

nRF52 in-place apply requires a companion bootloader change
(`Adafruit_nRF52_Bootloader_OTAFIX`, separate repo/PR — link TBD). ESP32 needs
nothing extra. Full/verify/serve paths need no bootloader on any platform.

## Testing

- ✅ Native unit tests, `motatool` ctest, Python reference suite all pass.
- ✅ Builds: ESP32 (Heltec V3) and nRF52 (RAK4631, ThinkNode M1) with EndF injection.
- ✅ **HW-validated end-to-end** on RAK4631 (nRF52840), Heltec T114 (nRF52840), and
  Heltec V3 (ESP32-S3): mesh discovery, full + delta fetch, signature/hash verify,
  and apply (ESP32 A/B + nRF52 in-place).

## Notes / scope

- `detools` 0.53.0 decoder is **vendored** (isolated, decoder-only, 3rd-party) — the
  delta codec is never reimplemented.
- No public-API breakage; additive `NodePrefs` fields are versioned/back-compatible.

## Checklist
- [x] Examples updated (companion/repeater hooks)
- [x] Code commented; protocol + user docs added
- [x] Builds + native tests pass; HW-validated
- [ ] Linked feature issue / maintainer 👍 (per CONTRIBUTING for larger features)


dreirund:
Nice :-).

I see -- not knowing programming or the protocol in depth -- some issues:

`OTA_ADV` beacon interval should be configurable.

Does it really go multi-hop, or only zero-hop? (If multi-hop, it should really be thought how to get network burden down _and_ to stop that just any remote evil person can flush the whole network with evil updates. I opt for zero-hop.)

I would further opt to advertise only manuall by default.

I tried to follow your links to `docs/ota_user_guide.md` and `docs/ota_protocol.md`, but they lead me only to some technical GitHub page saying "*There isn’t anything to compare.*" I do not understand.

vk496:
> Nice :-).
> 
> I see -- not knowing programming or the protocol in depth -- some issues:
> 
> `OTA_ADV` beacon interval should be configurable.
> 
> Does it really go multi-hop, or only zero-hop? (If multi-hop, it should really be thought how to get network burden down _and_ to stop that just any remote evil person can flush the whole network with evil updates. I opt for zero-hop.)
> 
> I would further opt to advertise only manuall by default.
> 
> I tried to follow your links to `docs/ota_user_guide.md` and `docs/ota_protocol.md`, but they lead me only to some technical GitHub page saying "*There isn’t anything to compare.*" I do not understand.

Thanks, links fixed.

By default I set max 3 hops, but I agree to actually change that dynamically (will merge in the following commits)

Advertising manually would prevent propagation of the firmware. I don't know what config should be by default, but beacons are intended to be small and cheap. Advertising one beacon (20 bytes IIRC) every 3-10min randomly should not have big impact in the mesh

dreirund:
> Advertising manually would prevent propagation of the firmware. I
> don't know what config should be by default, but beacons are intended
> to be small and cheap. Advertising one beacon (20 bytes IIRC) every
> 3-10min randomly should not have big impact in the mesh

Ehm, many meshes did cut down on forwarding adverts because flood
routing of them did overload. That was the reason the EVO firmware was
born.

Every 3..10 min some flood message -- thats way too much.

Weekly I think is a good option. We do not have so frequent firmware
updates to deploy.


So, you want to make an auto-update function by default, not a manual
update over LoRa, as other OTAs are also manual?

vk496:
I don't mind to follow what community agrees. You can always change the default behaviour. The only downside about every week is that you need to wait a week to know what firmwares are available around surrounding (`ota ls`)

Yes, the default behaviour is supported. Your target node must discover the firmware from your relay node and start pulling it.

The problem is that you need days (or even weeks in heavy traffic scenarios) to pull it. Just 1 device is fine, but if you have 7 nodes, that's a headache. That's why the idea of DHT/P2P firmware sharing. All node can share its own full firmware installed in the flash + mOTA (if available). Special nodes like ESP32 with WiFi can rely a folder from a host full of mOTA, so they don't need to store them in order to share them.

The goal is: if you have N same nodes, push once and let the firmware spread

akiraysh:
Hey vk496, noticed you built the .mota OTA-over-LoRa protocol with merkle tree verification and delta updates across ESP32 and nRF52, that's rlly rlly cool, especially the DHT style propagation for nodes that are hard to physically reach.

We're building an open core OTA platform with a similar goal, pushing updates to fleets you can't easily access, though our approach goes through WiFi/cellular rather than mesh.

Would you be down for a quick call to talk through how you implemented the delta updates and merkle verification? Always good to compare notes with someone solving the same class of problem from a different angle.

vk496:
> Hey vk496, noticed you built the .mota OTA-over-LoRa protocol with merkle tree verification and delta updates across ESP32 and nRF52, that's rlly rlly cool, especially the DHT style propagation for nodes that are hard to physically reach.
> 
> We're building an open core OTA platform with a similar goal, pushing updates to fleets you can't easily access, though our approach goes through WiFi/cellular rather than mesh.
> 
> Would you be down for a quick call to talk through how you implemented the delta updates and merkle verification? Always good to compare notes with someone solving the same class of problem from a different angle.

Hi. If you want quicker interaction, feel free to join the Discord thread:

https://discord.com/channels/1495203904898728149/1518163443750797332

akiraysh:
> > Hey vk496, noticed you built the .mota OTA-over-LoRa protocol with merkle tree verification and delta updates across ESP32 and nRF52, that's rlly rlly cool, especially the DHT style propagation for nodes that are hard to physically reach.
> > We're building an open core OTA platform with a similar goal, pushing updates to fleets you can't easily access, though our approach goes through WiFi/cellular rather than mesh.
> > Would you be down for a quick call to talk through how you implemented the delta updates and merkle verification? Always good to compare notes with someone solving the same class of problem from a different angle.
> 
> Hi. If you want quicker interaction, feel free to join the Discord thread:
> 
> https://discord.com/channels/1495203904898728149/1518163443750797332

Hey, thanks for sharing, the link doesn't seem to be working for me, getting an invalid invite error. Mind sending a fresh one?

vk496:
> > > Hey vk496, noticed you built the .mota OTA-over-LoRa protocol with merkle tree verification and delta updates across ESP32 and nRF52, that's rlly rlly cool, especially the DHT style propagation for nodes that are hard to physically reach.
> > > We're building an open core OTA platform with a similar goal, pushing updates to fleets you can't easily access, though our approach goes through WiFi/cellular rather than mesh.
> > > Would you be down for a quick call to talk through how you implemented the delta updates and merkle verification? Always good to compare notes with someone solving the same class of problem from a different angle.
> > 
> > 
> > Hi. If you want quicker interaction, feel free to join the Discord thread:
> > https://discord.com/channels/1495203904898728149/1518163443750797332
> 
> Hey, thanks for sharing, the link doesn't seem to be working for me, getting an invalid invite error. Mind sending a fresh one?

https://discord.gg/9sRhx5wvJ (OTA over LoRa in the `development`)

mikecarper:
Can this be paired with the temp radio command? Faster tx speeds are possible then

vk496:
> Can this be paired with the temp radio command? Faster tx speeds are possible then 

What do you mean? Any link or reference?

mikecarper:
https://docs.meshcore.io/cli_commands/#change-the-radio-parameters-for-a-set-duration  
Change the radio parameters for a set duration  
Usage: - tempradio <freq>,<bw>,<sf>,<cr>,<timeout_mins>  
Parameters: - freq: Frequency in MHz (300-2500) - bw: Bandwidth in kHz (7.8-500) - sf: Spreading factor (5-12) - cr: Coding rate (5-8) - timeout_mins: Duration in minutes (must be > 0)  
Note: This is not saved to preferences and will clear on reboot



So you move the radio to a different freq and then can go nuts with sending lora packets.  
If in the EU pick the bottom of the freq; otherwise in other places go 1mhz down and SF7 with BW500

vk496:
> https://docs.meshcore.io/cli_commands/#change-the-radio-parameters-for-a-set-duration  
> Change the radio parameters for a set duration  
> Usage: - tempradio <freq>,<bw>,<sf>,<cr>,<timeout_mins>  
> Parameters: - freq: Frequency in MHz (300-2500) - bw: Bandwidth in kHz (7.8-500) - sf: Spreading factor (5-12) - cr: Coding rate (5-8) - timeout_mins: Duration in minutes (must be > 0)  
> Note: This is not saved to preferences and will clear on reboot
> 
> 
> 
> So you move the radio to a different freq and then can go nuts with sending lora packets.  
> If in the EU pick the bottom of the freq; otherwise in other places go 1mhz down and SF7 with BW500

I don't think this blocks somehow mOTA. The protocol doesn't understand radio. Just if nodes are available or not. Is up to the user IMHO

mikecarper:
Giving the admin a hint to switch to a different frequency would be a good reminder.

vk496:
> Giving the admin a hint to switch to a different frequency would be a good reminder. 

Yes, but why? Being alone is not the intention of this OTA. Changing radio to have more speed is out of scope (you can go to node and flash it through BLE/Wifi/USB if speed matters)

It's not like I don't want doing that. I just don't understand your usecase

mikecarper:
In a congested mesh, sending the firmware on the same frequency is a no go. Use case is a tree repeater at the top of the mountain in the snow.

vk496:
I see. When do you suggest that reminder? I feel that it's just a note/idea that can be in the docs instead of the firmware

mikecarper:
First question is if there is a way to push the update and not use the torrent scheme. If there's a command like `start lora ota` that'd be where I'd put the reminder to switch the frequency and go with faster settings (like SF5 BW250/500). I would want the the get the firmware from the mesh be a switch; in my use case I'd want it off.

Use Case:
Tree node up in the mountains and it's snowed in.  
`start lora ota`  
`tempradio 910.100,500,5,5,120` // Gives me 2 hours to do the update  
Wave magic wand and start sending data to the tree node  
Update is done, node reboots moving radio back to the normal radio settings

vk496:
The idea behind DHT is to not interrupt normal operations. Some nodes (like one in the mountain ) could be the only node capable of doing hops. You assume that switching to a different radio is better because you will be alone. But if everyone else did this, you would have the same issue. And assuming that you need to do it to just 1 node and not 7.

The way I solve this is by making OTA a non priority traffic. And ideally, upgradable without user intervention (you flash your home node and just let it propagate to your  others nodes).

Anyway, with the current PR, you can do what you want. You can switch your remote and local node to a different radio and start polling the mOTA from the cli. Worst case, is that it will not finish on time and some blocks remains to pull. Switch again, finish poll and apply update

IoTThinks:
This will be very useful to upgrade repeaters more frequently.

IoTThinks:
@mikecarper We can change to 500kHz then upgrade.

LoRaWAN uses FSK and 500kHz for FOTA for faster speed.

mikecarper:
Yep I want to make sure this auto update can be turned off and only turn it on when I switch the settings to be fast Lora

vk496:
There are 3 things:

- DHT. You ask (and forward) OTA traffic, torrent like style (not exactly the same)
- Auto fetch. You start pulling a compatible new version from neighbors
- Auto update. You automatically apply the update (only possible with signatures)

You insist on doing it faster and alone. I remember you that the speed is out of scope (you can always go to the node and do the OTA over BLE/USB).

This OTA proposal solves when:

- No easy physical access at all. Requires summer season or drone.
- No direct connection. You need to go far for it, and cant rest there for days
- Unstable link. slow, retries, reboots, ....

towerviewcams:
I have 36 repeaters in the pacific north west and I could not agree more with @mikecarper that changing freq, bandwidth and uploading firmware is the key here. I will take mine off main channel for all upgrades. just to much at risk and much faster/easier to not be on main bet. Allot of tower sites are so noisy that 500 would just toast the RX where I am, but, i have tested 125khz at towers that have many cellular/broadcast and that is reliable many miles away.

I can also see the need to do this on main system. Having the option is key for folks.  Low usage areas i can see this working well. supper busy like the PNW that has the largest mesh in North America would be very tough. we cant get a telemetry read from a repeater with rxdelay 2 or 3 without several retries. we also are using programmable return paths now for telemetry and that has helped.

@vk496  sir, can we get bootloader for V4 (4.2/4.3) and T096  These are in use allot and can complete lots of testing at a good scale.  Tested 4631 and worked great

towerviewcams:
> First question is if there is a way to push the update and not use the torrent scheme. If there's a command like `start lora ota` that'd be where I'd put the reminder to switch the frequency and go with faster settings (like SF5 BW250/500). I would want the the get the firmware from the mesh be a switch; in my use case I'd want it off.
> 
> Use Case: Tree node up in the mountains and it's snowed in. `start lora ota` `tempradio 910.100,500,5,5,120` // Gives me 2 hours to do the update Wave magic wand and start sending data to the tree node Update is done, node reboots moving radio back to the normal radio settings

We need both options. But THIS ^^^^^^ is what I would be doing 98% of the time.

vk496:
I don't see how the usecase of being alone in different frequency is incompatible with this PR. If only 2 nodes exists, the DHT don't hurt you.

The bootloader and firmware are deployed as release in my forks

mikecarper:
What command do I use to not sniff for firmware upgrade packets? I want this off until I move the freq

vk496:
@mikecarper you cant avoid snifiing for those packets. Other nodes flood announce them

@towerviewcams I just updated my repo to add V4/t096 builds for testing. Feel free to play with it and report back :)

https://github.com/vk496/MeshCore/releases/tag/dev-latest

towerviewcams:
> @mikecarper you cant avoid snifiing for those packets. Other nodes flood announce them
> 
> @towerviewcams I just updated my repo to add V4/t096 builds for testing. Feel free to play with it and report back :)
> 
> https://github.com/vk496/MeshCore/releases/tag/dev-latest

We cant use T096 as we dont have a bootloader, right?

vk496:
Without the correct bootloader on nRF52, when you apply the OTA, it will just reboot the device instead of performing the upgrade

PR #2876, opened 2026-07-01
Fix incorrect deduplication of flood packets

neilalexander:
We were, in many cases, adding packets into the deduplication table before checking whether the packet was even eligible for forwarding. This could mean that we drop packets that should have been allowed.

For example, the flood limit is set to 10. The first copy of a packet arrives with a path length of 12. The second copy of the same packet arrives via a different route with a path length of 8. The first packet is not forwarded because it exceeds the path length but it is still entered into the deduplication table, which results in the second packet also being dropped when it should have been forwarded.

### DIFF
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -120,11 +120,13 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
       memcpy(&ack_crc, &pkt->payload[i], 4); i += 4;
       if (i > pkt->payload_len) {
         MESH_DEBUG_PRINTLN("%s Mesh::onRecvPacket(): incomplete ACK packet", getLogDateTime());
-      } else if (!_tables->wasSeen(pkt)) {
+        break;
+      }
+      if (!_tables->wasSeen(pkt)) {
         _tables->markSeen(pkt);
         onAckRecv(pkt, ack_crc);
-        action = routeRecvPacket(pkt);
       }
+      action = routeRecvPacket(pkt);
       break;
     }
     case PAYLOAD_TYPE_PATH:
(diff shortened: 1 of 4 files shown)


src/Mesh.cpp
```diff
@@ -106,7 +106,29 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
     return ACTION_RELEASE;   // this node is NOT the next hop (OR this packet has already been forwarded), so discard.
   }
 
-  if (pkt->isRouteFlood() && filterRecvFloodPacket(pkt)) return ACTION_RELEASE;
+  // Only consult the dupe table later on if eligible for forwarding.
+  bool flood_forward_eligible = true;
+  auto hasSeenForDispatch = [&]() {
+    return flood_forward_eligible ? _tables->hasSeen(pkt) : false;
```
ckoehler: I think this would bypass the dedup altogether if `flood_forward_eligible` were false? If it's false, `_tables->hasSeen` will never be called, so no hash will ever be computed, so any checks in the dedup table will be false ("we've never seen this before").

neilalexander: The initial value is true, not false. It's only set to false if the initial forwarding checks fail.

ckoehler: Right, so wouldn't companion radios (leaf nodes) get multiple messages if they hear them from multiple repeaters?
`allowPacketForward` returns false, so `flood_forward_eligible` is set to false on 129, so any subsequent call to `hasSeenForDispatch` will return false, never calling `_tables->hasSeen`, and thus never deduping any packets?

neilalexander: I see what you mean. I didn't also account for the client repeat behaviour. I'll think on this a bit more.

neilalexander:
I've changed the approach here somewhat so that the tables track dispatches and forwards separately, otherwise it's extremely hard to solve this problem properly. That should stop us from suppressing dispatches for packets that aren't eligible for forwarding, and still allow us to defer setting the forwarded bit until after the packet forwarding checks have been done.

ripplebiz:
I think separating hasSeen() into two methods like this: https://github.com/meshcore-dev/MeshCore/pull/2880
Could be simpler?

neilalexander:
That maybe simplifies things, but I don't think it fixes the problem. The issue is that a repeater must only mark the packet as a duplicate for forwarding if we actually intend to forward it.

I've rebased and tweaked things, curious on your thoughts.

ckoehler:
Not sure I'm much help here, the behavior isn't super clear to me. I did write a test that I think shows the behavior you want, which currently fails. Once that passes, the behavior would be correct.

https://github.com/ckoehler/MeshCore/commit/cf40b761b44f4266eefbfc63c8f84e67c6e989b4

Maybe it'd be a starting point to add some tests to your PR/verify it's working as intended?

PR #2880, opened 2026-07-02
refactor: split MeshTables::hasSeen into pure query + markSeen

ckoehler:
`hasSeen` checked seen but also marked seen status. Some places in the code just called it to mark seen, discarding return value. This PR just splits the old `hasSeen` into a pure check and a new `markSeen` function that marks seen, and updates call sites of the old `hasSeen` (the check) to also call `markSeen`.

Also added some unit tests for `SimpleMeshTables`.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -165,5 +165,6 @@ test_build_src = yes
 build_src_filter =
   -<*>
   +<../src/Utils.cpp>
+  +<../src/Packet.cpp>
 lib_deps =
   google/googletest @ 1.17.0
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -55,7 +55,8 @@ DispatcherAction Mesh::onRecvPacket(Packet* pkt) {
       uint16_t offset = (uint16_t)pkt->path_len << path_sz;
       if (offset >= len) {   // TRACE has reached end of given path
         onTraceRecv(pkt, trace_tag, auth_code, flags, pkt->path, &pkt->payload[i], len);
-      } else if (self_id.isHashMatch(&pkt->payload[i + offset], 1 << path_sz) && allowPacketForward(pkt) && !_tables->hasSeen(pkt)) {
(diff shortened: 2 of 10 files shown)


liamcottle (approved):
Looks good! And thanks for adding the tests! :)

ripplebiz:
Looks good. Thanks.

One thing I prefer to do, when changing the basic meaning/use of a method, is to _rename_ it.  1) compiler will redline every call site that needs re-thinking.  2) all downstream code will know they need to fix their uses too.

Not sure what to suggest for hasSeen() because that's already the optimal name :-(  Maybe  wasSeen()  ?

ckoehler:
> One thing I prefer to do, when changing the basic meaning/use of a method, is to _rename_ it. 1) compiler will redline every call site that needs re-thinking. 2) all downstream code will know they need to fix their uses too.
> 
> Not sure what to suggest for hasSeen() because that's already the optimal name :-( Maybe wasSeen() ?

@ripplebiz Oof, yeah the name is pretty perfect. `wasSeen` would be okay. This PR changes all the occurrences of the function; do you mean other code bases? If it's just this codebase, I got all the places I think. But if there are other users of this code, I'd be happy to rename it to `wasSeen`. Just say the word.

ripplebiz:
> > One thing I prefer to do, when changing the basic meaning/use of a method, is to _rename_ it. 1) compiler will redline every call site that needs re-thinking. 2) all downstream code will know they need to fix their uses too.
> > Not sure what to suggest for hasSeen() because that's already the optimal name :-( Maybe wasSeen() ?
> 
> @ripplebiz Oof, yeah the name is pretty perfect. `wasSeen` would be okay. This PR changes all the occurrences of the function; do you mean other code bases? If it's just this codebase, I got all the places I think. But if there are other users of this code, I'd be happy to rename it to `wasSeen`. Just say the word.

Yeah, there are nearly always more references than you _know_ about.  Safest to rename.

liamcottle:
Maybe go with `hasSeenPacket` to keep the `hasSeen` ;)

ckoehler:
I think I prefer `wasSeen`. Change made.

PR #2881, opened 2026-07-02
refactor: common allocForType function

ckoehler:
The create* methods all do the same thing at the start. This extracts that out into a single function.


ripplebiz:
Thanks for the suggestion, but this doesn't reduce complexity, and some debug logging info is actually lost if using this.

PR #2883, opened 2026-07-03
repeater: poweroff on long press

fdlamotte:
changed the button login in repeater to use MomentaryButton as in companion radio

implemented poweroff on long press

tested on TEcho, Thinknode M5 and heltec_v3
 
should support all repeaters with display

might need some review of all variants to see if it breaks, could support all repeaters with buttons by adding a null display


.vscode/extensions.json
```diff
@@ -1,4 +1,6 @@
 {
+    // See http://go.microsoft.com/fwlink/?LinkId=827846
```
fdlamotte: I'll need to remove this ...
was added by pio ... should probably add this file to .gitignore ...

fdlamotte:
seems not to work after sending some commands from LORA ...

edit: this only happened to me on TEcho for the moment, maybe some issue specific to this board ... noticed some incoherencies in btn init (pullups or not ...), might also have to do with eink screen refresh will have to check the loop. This does not prevent to poweroff the node as one just need to reboot it before powering off

fdlamotte:
> seems not to work after sending some commands from LORA ...

This was a screen refresh issue on techo, it has been solved

Now I'll check result of compilation

examples/simple_repeater/UITask.cpp
```diff
@@ -73,8 +80,26 @@ void UITask::renderCurrScreen() {
...
+  } else if (_powering_off_at > 0) {
+    // meshcore logo
+    _display->setColor(DisplayDriver::BLUE);
+    int logoWidth = 128;
+    _display->drawXbm((_display->width() - logoWidth) / 2, 3, meshcore_logo, logoWidth, 13);
+
+    // meshcore website
+    const char* website = "https://meshcore.io";
+    _display->setColor(DisplayDriver::LIGHT);
+    _display->setTextSize(1);
```
ripplebiz, line 87: We have drawTextCentered() as shorthand for this

fdlamotte: right, and I should know that ;)

I'll try to move the poweroff logic out of the ui (the 3s timeout), to enable a poweroff command that would actually show on the screen that the repeater is powering off 

the ui would only initiate the sequence ...

IoTThinks:
_board.powerOff() will not turn off everything including power-hungry GPS module.
_board.powerOff() also does not power off the NRF52 on repeaters at all because repeaters are not soft-devices.

This PR will help to make your PR works.
https://github.com/meshcore-dev/MeshCore/pull/2895

fdlamotte:
Well in the end poweroff is already implemented in the CLI (it only did not work on my techo before #2895)

So I won't bother changing current mechanism in this PR, it will stay as it is

Further work should focus on implementing minimal UI support for boards that don't have screen but have buttons that can be used to poweroff.

Also repeater UI could be better, some screen to show status, neighbors, control adverts, show gps status ... could help

IoTThinks:
UI for repeater similar to companion is more friendlier to sleep.

Press to wakeup and allow to surf the UI for 30s...

examples/simple_repeater/UITask.cpp
```diff
@@ -124,4 +140,15 @@ void UITask::loop() {
       _display->turnOff();
     }
   }
+
+  if (_powering_off_at > 0) { // power off timer armed
+#ifdef LED_PIN
+    digitalWrite(LED_PIN, LED_STATE_ON); // switch on the led until poweroff
```
IoTThinks, line 146: This leaves LED on forever after pressing powering off T096 repeater.
It consumes 1.7mA instead of 0.4mA.

Not all boards have defined LED_STATE_ON and powered off LED during power off.

I will attemp to fix it.

IoTThinks: This is the PR to fix the above issue.
https://github.com/meshcore-dev/MeshCore/pull/3449

Issue #2879, opened 2026-07-01
ThinkNode M3 blue led stays enabled after transmission


sleeply4cat:
blue (TX) led stays enabled for a long time and doesn't turn off as intended. When transmitting, led is blinking dark.

The problem seems to be related to an inverse led control signal. I generated a patch that fixed this issue for me.

[thinknode_m3_tx_led_fix.patch](https://github.com/user-attachments/files/29574992/thinknode_m3_tx_led_fix.patch)

recrof:
Please open Pull request with the fix.

hermes-gadget:
test - please ignore

PR #2887, opened 2026-07-03
Fix GPS and LEDs on ThinkNode M3

spinda:
GPS support was implemented incorrectly for the ThinkNode M3. The reset pin was being driven when it should be left floating for this unit. The enable pin also wasn't being picked up by `MicroNMEALocationProvider` because of a mismatch in constant naming conventions. I did a general cleanup of the GPS and ThinkNode M3 bring-up code so that constant names line up and "*_ACTIVE" constants are used consistently vs hardcoding `HIGH`/`LOW`. After making these changes, serial data immediately starts streaming in from the NMEA on boot and GPS detection just works.

LED handling was also not quite right for the ThinkNode M3. The LoRa TX LED was being driven high to turn it on when it should actually be driven low. I changed the code to use the `LED_STATE_ON` constant and also added a little code to the shutdown path to properly make sure that all LEDs are turned off.

Tested and confirmed working on real hardware.

Resolves both #1864 and #2879.

### DIFF
--- a/src/helpers/sensors/MicroNMEALocationProvider.h
+++ b/src/helpers/sensors/MicroNMEALocationProvider.h
@@ -13,8 +13,12 @@
     #endif
 #endif
 
-#ifndef PIN_GPS_EN_ACTIVE
-    #define PIN_GPS_EN_ACTIVE HIGH
+#ifndef GPS_EN_ACTIVE
+    #ifdef PIN_GPS_EN_ACTIVE
+        #define GPS_EN_ACTIVE PIN_GPS_EN_ACTIVE
+    #else
+        #define GPS_EN_ACTIVE HIGH
+    #endif
 #endif
 
 #ifndef GPS_RESET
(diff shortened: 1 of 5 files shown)


fdlamotte:
Oh that's great ...

I did the initial port quick and then I forgot the unit in France (and it is still there). I remember I had issues with gps, it's great you could sort this out ...

I see there is still a lot of mess in symbols for the GPSes on different boards, could you verify how your changes affect other variants ? 

We usually prefer keeping the PRs focused and as small as possible ... I think here it would be better to focus on making the M3's peripheral work and in another PR do some refactoring on symbol names (there is really a need for that part, and I agree for the bad choice of GPS_RESET_FORCE). You can mention me in the other PR so I review and merge it. Thanks.

fdlamotte:
GPS_RESET_ACTIVE makes more sense than GPS_FORCE. So I'll merge it even if it makes the PR a little big

IoTThinks:
@fdlamotte I'm a fan of few lines to fix the stuff.
This one works for M6 with 2-3 lines of code.

A similar fix should work for M3 too.
I'm a bit scared to touch the fragile GPS code. Esp: GPS_EN, GPS_RESET
https://github.com/meshcore-dev/MeshCore/pull/2863/changes

These definitions look too similar.
```
#ifndef GPS_EN_ACTIVE
    #ifdef PIN_GPS_EN_ACTIVE
```

PR #2888, opened 2026-07-04
Node-to-companion frames fragmentation (reworked)

HDDen:
Hello! This is rework of [PR#2885](https://github.com/meshcore-dev/MeshCore/pull/2885)

This PR adds fragmentation of packets transmitted from the node to the application.

## Problem

In the MeshCore app, you can enter up to 144 bytes of text (and up to 155 in alternative apps), but with such a long message, the app doesn’t receive a retranslation report, even though the message was actually successfully transmitted across the network and received at the other end of town. Problem is presented on screenshots.

<details>
  <summary>Screenshots</summary>

Original app: 1, 2 - messages that are sendable, but due to this issue, their retranslations will not be displayed. 3 - message with factical 133-bytes limit and displayed retranslations counter.

3 - official app  

Meshcore_open app: 1 - factical limit-fine message with confirmation; 2 - plus 1 byte is breaking retranslation confirmation; 3, 4 - patched firmware+app, overlimitting is now doesnt breaking retranslations counter and counts correctly.

2 - comparsion between original and patched
  
</details>

I tried to get to the bottom of this issue (first manually via debugging using VSCode and the ‘meshcore_open’ client, and then analysed my hypothesis using Codex), and came to the conclusion that the node hears the retranslation of my long message, attempts to pass it to the app, but a certain ‘node-to-companion’ packet limit (MAX_FRAME_SIZE=176), and a corrupted data packet is received by the app, which cannot be decrypted and, consequently, cannot be recognised as a repeat of my message.

In other words, MeshCore radio packets with length up to maximum factical limit (not app-side limit) text can be valid and successfully propagated through the network while the corresponding companion frame-packet sent from the node to app is too large for the existing `MAX_FRAME_SIZE = 176` node-to-app frame limit.

I encountered this error with messages sent to channels. Let me repeat: message can be sent over the air, successfully retranslated by repeaters, and received by other nodes. However, when the sending node hears it's retranslation and tries to report it to the connected app, the companion frame may exceed the single-frame limit. Client app then does not receive a complete parseable receive frame and cannot recognize the heard retranslation, so the retranslation counter is not updated.

In my case, with a 3-byte node name, 134-byte channel messages still produced retranslation confirmations, while 135-byte messages did not. The 134 -> 135 byte boundary is likely visible because AES block padding can increase the encrypted payload by 16 bytes at once, pushing the companion response over the frame limit.

## Solution

This PR adds companion frame fragmentation for firmware-to-app frames.

The companion protocol version is increased from 13 to 14. Fragmentation is enabled only when the app advertises target protocol version 14 or newer via `CMD_DEVICE_QUERY`. Apps that continue to advertise protocol version 13 keep the old behavior, with no breaking protocol changes.

A new firmware-to-app frame type is added:

- `PACKET_FRAME_FRAGMENT` / `PUSH_CODE_FRAME_FRAGMENT` (`0x91`)

The fragment frame wraps chunks of the original companion frame. The app reassembles all fragments and then parses the reconstructed bytes as the original companion frame.

Fragmentation is applied to:

- oversized queued contact/channel message frames delivered via `CMD_SYNC_NEXT_MESSAGE`
- oversized live raw data push frames
- oversized live control data push frames
- oversized RF log push frames

Frames that fit into `MAX_FRAME_SIZE = 176` are still sent unchanged. Oversized frames are fragmented only for apps that negotiated protocol v14+. Each fragment frame is also capped to `MAX_FRAME_SIZE`; with the 10-byte fragment header, the maximum chunk size is 166 bytes.

For queued frames delivered through `CMD_SYNC_NEXT_MESSAGE`, protocol v14 adds fragment ACK fields. The app ACKs each received queued fragment in the next `CMD_SYNC_NEXT_MESSAGE` request. If the ACK is missing or does not match the last sent fragment, the firmware repeats that fragment. After the final fragment is ACKed, the queued frame is removed and the firmware proceeds to the next queued item.

Live push fragments are best-effort and are not ACKed through `CMD_SYNC_NEXT_MESSAGE`.

## Compatibility

Protocol v13 behavior is preserved:

- apps that do not advertise v14 never receive `PACKET_FRAME_FRAGMENT`
- frames that fit into the old limit are sent exactly as before
- oversized queued frames keep the previous legacy truncated single-frame delivery behavior for old apps

## How to reproduce problem on original firmware

1. Use original firmware and any current available app.
2. Set a short node name, for example "AAA" for 3 bytes-length.
3. Send a channel message.
4. In my setup, a 134-byte text produced retranslation counter updates, while 135 bytes and longer did not.
5. Verify on another node that the 135-144byte message is still received and have retranslations.
6. Observe that the sending app does not show retranslation confirmation for such long messages.

## How To Test This PR

1. Build the firmware with the changes from the PR.
2. Build the app with the changes to the companion protocol (I'm using my fork of "meshcore_open" app, and attaching the apk at https://hdden.ru/mcoa-180-frames-patch-v2.apk , and you can build it yourself from https://github.com/HDDen/meshcore-open/tree/mcoa-framed-messages ).
3. In the app settings, enable "Settings - App settings - Enable Message Tracing."
4. Connect to the node and send a message to the channel with the maximum allowed limit (in my case, it was 155 bytes for the text). Wait and watch the retranslation counter increase.
5. Also verify that an app using protocol v13 (official MeshCore app any current available) can still connect and receive normal non-fragmented frames.

## Changes from original PR#2885

Redundant incorrect APP_SAFE_FRAME_SIZE mechanism, which had been added in error, has been removed; fulfilled code with comments, updated documentation, fixed potential  «queued_fragment_id == 0» bug.

### DIFF
--- a/docs/companion_protocol.md
+++ b/docs/companion_protocol.md
@@ -1,7 +1,7 @@
 # Companion Protocol
 
-- **Last Updated**: 2026-03-08
-- **Protocol Version**: Companion Firmware v1.12.0+
+- **Last Updated**: 2026-07-04
+- **Protocol Version**: Companion Firmware v1.16.0+
 
 > NOTE: This document is still in development. Some information may be inaccurate.
 
@@ -107,6 +107,12 @@ The default BLE MTU is 23 bytes (20 bytes payload). For larger commands like `SE
     - iOS: `peripheral.maximumWriteValueLength(for:)`
     - Python (bleak): MTU is negotiated automatically
 
+### Companion Frame Size Limits
+
+Firmware transports use `MAX_FRAME_SIZE = 176` as the companion frame buffer size.
+
+When an app advertises target protocol version 14 or newer in `CMD_DEVICE_QUERY`, firmware emits `PACKET_FRAME_FRAGMENT` for original firmware-to-app companion frames larger than `MAX_FRAME_SIZE`. Frames up to 176 bytes are sent unchanged. Each emitted fragment is also kept at or below `MAX_FRAME_SIZE`. Apps that do not advertise version 14 keep the legacy behavior based on the older single-frame transport limit.
+
 ### Command Sequencing
 
 **Critical**: Commands must be sent in the correct sequence:
@@ -187,13 +193,15 @@ Bytes 8+: Application name (UTF-8, optional)
 **Command Format**:
 ```
 Byte 0: 0x16
-Byte 1: 0x03
+Byte 1: App target protocol version
 ```
 
 **Example** (hex):
 ```
-16 03
+16 0E
 ```
+
+Set byte 1 to the highest companion protocol version the app understands. Firmware uses this value to decide whether it may emit newer optional frame formats. Apps that support `PACKET_FRAME_FRAGMENT` should send `0x0E` (14) or higher.
 
 **Response**: `PACKET_DEVICE_INFO` (0x0D) with device information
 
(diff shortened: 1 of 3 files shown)


IoTThinks:
I would prefer to keep the message short at 176 bytes max or even less.
So we don't need to deal with oversized frames at all.
Additionally,  we will have lower collission due to shorter frame too.

I believe the chat message is very short nowsaday.

> Frames that fit into `MAX_FRAME_SIZE = 176` are still sent unchanged. Oversized frames are fragmented only for apps that negotiated protocol v14+. Each fragment frame is also capped to `MAX_FRAME_SIZE`; with the 10-byte fragment header, the maximum chunk size is 166 bytes.

HDDen:
@IoTThinks Hello!
By and large, this PR addresses a different issue. It does not increase the size of the LoRa packet nor does it affect it in any way; its purpose here is different – it enables more data to be transmitted from the node to the connected application, without being constrained by the strict 176-byte limit. This PR is specifically designed to enable the node to send more data to the connected application.

In the long term, this could prove useful if we learn to apply compression algorithms to the text being sent (particularly useful for multibyte text – for example, Cyrillic, where one character takes up two bytes). Node could decompress such text itself and pass the final UTF-8 text to the connected application. Consequently, a firmware modification would suffice for such support, without the need to wait for an application update – because node would send plain text directly to the application via BLE/TCP/USB, just as it does now.

Examples of text compression/optimisation include BOCU-1, UCF-encoding and mesh-compressor (the latter being rather niche, as it requires powerful hardware). At the moment, we have to resort to makeshift optimisation for compression – for example, for Cyrillic I use cyr2lat – replacing visually similar Cyrillic characters with Latin ones, which provides around 20–30 per cent compression without any loss of readability or compatability with the official Meshcore app.

As for increasing the message length: it has long been possible to send LoRa messages that are slightly longer than the official limit – for example, via the well-known meshcore_open – so this PR isn’t necessary for that. It merely ensures, as a consequence, that we can correctly hear relays of this messsages.

IoTThinks:
I got it now.
The idea looks useful for applications.

> it enables more data to be transmitted from the node to the connected application, without being constrained by the strict 176-byte limit. This PR is specifically designed to enable the node to send more data to the connected application.

PR #2889, opened 2026-07-04
Fix ESP32 Wi-Fi companion STA MAC drift by pinning STA MAC to eFuse

maciek01:
## Summary
In Wi-Fi companion mode on ESP32, observed MAC values can differ depending on interface/state. This change pins the STA MAC to the eFuse factory MAC before connect so the companion path uses a deterministic STA address.

## Changes
- Include ESP-IDF MAC/Wi-Fi headers in companion startup path.
- Set Wi-Fi mode to `WIFI_STA` before connect.
- Read factory MAC using `esp_efuse_mac_get_default`.
- Apply factory MAC to STA interface via `esp_wifi_set_mac(WIFI_IF_STA, ...)`.
- Add debug prints for factory MAC and active STA MAC.

## Scope
- File: `examples/companion_radio/main.cpp`
- ESP32 + `WIFI_SSID` companion build path only.

## Motivation
This avoids confusion from seeing multiple MAC identities and keeps companion STA identity stable and predictable.


maciek01:
wrong content

PR #2897, opened 2026-07-05
Added MCU temp sensor to the companion

### LINKED ISSUE #2896: MCU temperature telemetry for companion
Repeaters show internal MCU temperature which can be handy, please add the same for companions.

yarda:
Fixes #2896

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -657,6 +657,11 @@ uint8_t MyMesh::onContactRequest(const ContactInfo &contact, uint32_t sender_tim
       // query other sensors -- target specific
       sensors.querySensors(permissions, telemetry);
 
+      float temperature = board.getMCUTemperature();
+      if(!isnan(temperature)) { // Supported boards with built-in temperature sensor. ESP32-C3 may return NAN
+        telemetry.addTemperature(TELEM_CHANNEL_SELF, temperature); // Built-in MCU Temperature
+      }
+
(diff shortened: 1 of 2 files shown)


IoTThinks:
Yeah, great.
Thanks a lot.

pemensik:
Ah, nice! I would recommend moving different way to obtain the temperature to different (static) functions for increased readability, because they differ significantly. But yes, if my companion is able to measure the temperature without extra sensor, I want to have it readable.

yarda:
> Ah, nice! I would recommend moving different way to obtain the temperature to different (static) functions for increased readability, because they differ significantly. But yes, if my companion is able to measure the temperature without extra sensor, I want to have it readable.

It is consistent with the existing coding style, i.e. there is usually check for the softdev and two code paths in the function.

liamcottle:
Hello! Thanks for the PR.

You are trying to merge changes into `main`.
Can you please change the base branch to `dev`. 

Also, you may need to rebase your changes on your local branch against `dev`, and force push, and not just change it in the dropdown on the GitHub PR UI.

Otherwise it will pull in a bunch of commits from `main`, which we don't want.

yarda:
Hmm, it seems I hit the wrong button in the UI :D But no problem, I will fix it.

yarda:
@liamcottle I hope it's OK now.

liamcottle:
> @liamcottle I hope it's OK now.

Tested and working on my Wio Tracker L1 Pro. Thanks for contributing :)

PR #2911, opened 2026-07-09
Add power source (battery/solar/mains) to node adverts

ViezeVingertjes:
Right now people stick emojis in their node names to show whether a device is on battery, solar, or mains. That's not machine-readable and it counts towards their message length leaving less room for actual messages. This change adds a power.source capability field to the advert so power source gets signaled properly instead.

Did not add it to companions adverts although that might be a future improvement if there is consensus on this.

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -323,6 +323,20 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 
 ---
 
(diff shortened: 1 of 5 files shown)


ViezeVingertjes:
More people working on this; will check if i can bring them together to find a proper unified solution first.

PR #2839, opened 2026-06-25
Add support for Elecrow ThinkNode M7

liamcottle:
This PR adds support for the [Elecrow ThinkNode M7](https://www.elecrow.com/thinknode-m7-wireless-communication-gateway-for-meshtastic-support-poe-powered-powered-by-esp32-s3-and-lr1110.html).

I have tested the following firmwares:
- companion_radio_ble
- companion_radio_usb
- companion_radio_wifi
- repeater
- room_server

The kiss_modem firmware has also been added, not tested, but compilation was successful.

Please note, the ThinkNode M7 board was shipping with an LR1110 radio with older firmware.
See https://github.com/jgromes/RadioLib/issues/1814 for more information.

This PR will be blocked until a new RadioLib version is released, with https://github.com/jgromes/RadioLib/pull/1815 merged in.

Without the above RadioLib upgrade, the LR1110 (if running older firmware) will fail to init, due to missing features.
This means the MeshCore firmware will hang at startup, and won't be usable.

The new RadioLib changes resolve this, but it's still recommended that users upgrade the firmware on the LR1110 radio chip if using in production or for critical applications. We will look to provide a tool, or firmware binary to make it easy for end users to update the LR1110 firmware.

_Note: PR is in draft state until RadioLib is updated and we bump the library dependency in MeshCore_

PR #2912, opened 2026-07-09
bump radiolib to 7.7.1

weebl2000:
It includes some nice fixes. [Full release notes here](https://github.com/jgromes/RadioLib/releases#release-7.7.0)

Build here: https://mcimages.weebl.me?commitId=radiolib-7.7.1

Changed the init calls to use config instead, this prepares us for radiolib 8.0.0 removal of the now deprecated init methods.

### DIFF
--- a/library.json
+++ b/library.json
@@ -4,7 +4,7 @@
     "dependencies": {
         "SPI": "*",
         "Wire": "*",
-        "jgromes/RadioLib": "^7.6.0",
+        "jgromes/RadioLib": "^7.7.1",
         "rweather/Crypto": "^0.4.0",
         "adafruit/RTClib": "^2.1.3",
         "melopero/Melopero RV3028": "^1.1.0",
--- a/library.json
+++ b/library.json
@@ -4,7 +4,7 @@
     "dependencies": {
         "SPI": "*",
         "Wire": "*",
-        "jgromes/RadioLib": "^7.7.1",
+        "RadioLib": "https://github.com/jgromes/RadioLib.git#6d8934836678d8894e3d556550475b37dce3e2b6",
         "rweather/Crypto": "^0.4.0",
         "adafruit/RTClib": "^2.1.3",
         "melopero/Melopero RV3028": "^1.1.0",
--- a/src/helpers/radiolib/CustomLLCC68.h
+++ b/src/helpers/radiolib/CustomLLCC68.h
@@ -18,16 +18,23 @@ class CustomLLCC68 : public LLCC68 {
   #endif
     {
   #ifdef SX126X_DIO3_TCXO_VOLTAGE
-      float tcxo = SX126X_DIO3_TCXO_VOLTAGE;
+      tcxoVoltage = SX126X_DIO3_TCXO_VOLTAGE;
   #else
-      float tcxo = 1.6f;
+      tcxoVoltage = 1.6f;
   #endif
 
+      ConfigLoRa_t cfg;
+      cfg.frequency = LORA_FREQ;
+      cfg.bandwidth = LORA_BW;
+      cfg.spreadingFactor = LORA_SF;
   #ifdef LORA_CR
-      uint8_t cr = LORA_CR;
+      cfg.codingRate = LORA_CR;
   #else
-      uint8_t cr = 5;
+      cfg.codingRate = 5;
   #endif
+      cfg.syncWord = RADIOLIB_LORA_SYNC_WORD_PRIVATE;
+      cfg.power = LORA_TX_POWER;
+      cfg.preambleLength = 16;
 
   #if defined(P_LORA_SCLK)
     #ifdef NRF52_PLATFORM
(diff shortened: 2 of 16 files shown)


liamcottle:
Hey, just FYI, haven't read your code changes yet;

We are currently waiting for v7.7.2+ of RadioLib to be released, so we can merge in PR https://github.com/meshcore-dev/MeshCore/pull/2839

This PR might clash with an incoming PR from @oltaco which I believe bumps the repo to the latest git commit hash instead of a release version. Tagging him for awareness.

weebl2000:
Apart from the version bump the code is functionally the same, just using the newer patterns that radiolib will switch to starting version 8.

library.json
```diff
@@ -4,7 +4,7 @@
     "dependencies": {
         "SPI": "*",
         "Wire": "*",
-        "jgromes/RadioLib": "^7.6.0",
+        "jgromes/RadioLib": "^7.7.1",
```
liamcottle: not sure if we should tag against the git commit, as used in latest dev commits, hoping 7.7.2 or 7.8 will be out soon anyway.

src/helpers/radiolib/CustomSX1268.h
```diff
@@ -42,11 +49,11 @@ class CustomSX1268 : public SX1268 {
       if (spi) spi->begin(P_LORA_SCLK, P_LORA_MISO, P_LORA_MOSI);
     #endif
   #endif
-      int status = begin(LORA_FREQ, LORA_BW, LORA_SF, cr, RADIOLIB_SX126X_SYNC_WORD_PRIVATE, LORA_TX_POWER, 16, tcxo);
+      int status = begin(cfg);
       // if radio init fails with -707/-706, try again with tcxo voltage set to 0.0f
       if (status == RADIOLIB_ERR_SPI_CMD_FAILED || status == RADIOLIB_ERR_SPI_CMD_INVALID) {
-        tcxo = 0.0f;
-        status = begin(LORA_FREQ, LORA_BW, LORA_SF, cr, RADIOLIB_SX126X_SYNC_WORD_PRIVATE, LORA_TX_POWER, 16, tcxo);
+        tcxoVoltage = 0.0f;
```
liamcottle, line 55: `txco` was originally passed to `begin`, but now it's renamed to `tcxoVoltage` and not used at all? Assume this should be set on the radio object before the first call to `begin` and if it fails, needs to be updated again before the next call to `begin`. Seems it's not used at all here.

src/helpers/radiolib/CustomSX1262.h
```diff
@@ -17,22 +17,27 @@ class CustomSX1262 : public SX1262 {
...
+      cfg.codingRate = LORA_CR;
   #else
-      constexpr bool useRegulatorLDO = false;
+      cfg.codingRate = 5;
   #endif
+      cfg.syncWord = RADIOLIB_LORA_SYNC_WORD_PRIVATE;
+      cfg.power = LORA_TX_POWER;
+      cfg.preambleLength = 16;
 
       MESH_DEBUG_PRINTLN("SX1262 regulator requested: %s", useRegulatorLDO ? "LDO" : "DC-DC");
```
liamcottle, line 44: `useRegulatorLDO` was originally passed to `begin`, but now it's not used anywhere..?

src/helpers/radiolib/CustomSX1262.h
```diff
@@ -51,12 +56,12 @@ class CustomSX1262 : public SX1262 {
...
     #endif
   #endif
-      int status = begin(LORA_FREQ, LORA_BW, LORA_SF, cr, RADIOLIB_SX126X_SYNC_WORD_PRIVATE, LORA_TX_POWER, 16, tcxo, useRegulatorLDO);
+      int status = begin(cfg);
       // if radio init fails with -707/-706, try again with tcxo voltage set to 0.0f
       if (status == RADIOLIB_ERR_SPI_CMD_FAILED || status == RADIOLIB_ERR_SPI_CMD_INVALID) {
         MESH_DEBUG_PRINTLN("SX1262 init failed with error %d, retrying with TCXO at 0.0V", status);
-        tcxo = 0.0f;
-        status = begin(LORA_FREQ, LORA_BW, LORA_SF, cr, RADIOLIB_SX126X_SYNC_WORD_PRIVATE, LORA_TX_POWER, 16, tcxo, useRegulatorLDO);
+        tcxoVoltage = 0.0f;
```
liamcottle, line 65: Same issue with `tcxoVoltage` as noted in other comments. Doesn't seem to be used/passed in anyway.

src/helpers/radiolib/CustomLLCC68.h
```diff
@@ -42,11 +49,11 @@ class CustomLLCC68 : public LLCC68 {
       if (spi) spi->begin(P_LORA_SCLK, P_LORA_MISO, P_LORA_MOSI);
     #endif
   #endif
-      int status = begin(LORA_FREQ, LORA_BW, LORA_SF, cr, RADIOLIB_SX126X_SYNC_WORD_PRIVATE, LORA_TX_POWER, 16, tcxo);
+      int status = begin(cfg);
       // if radio init fails with -707/-706, try again with tcxo voltage set to 0.0f
       if (status == RADIOLIB_ERR_SPI_CMD_FAILED || status == RADIOLIB_ERR_SPI_CMD_INVALID) {
-        tcxo = 0.0f;
-        status = begin(LORA_FREQ, LORA_BW, LORA_SF, cr, RADIOLIB_SX126X_SYNC_WORD_PRIVATE, LORA_TX_POWER, 16, tcxo);
+        tcxoVoltage = 0.0f;
```
liamcottle, line 55: Same problem with `tcxoVoltage`. should be set on radio before first attempt, and updated before second attempt

liamcottle (changes requested):
Howdy! Thanks for the PR!

Would be nice to migrate to the new config object, but currently not a requirement.
I've just reviewed the code, but there seems to be a lot of issues.
So I assume this wasn't tested at all :)

Seems that variables for LDO and TCXO have been renamed, but not passed in at all with new config etc. So I guess in some cases, radios that expect TCXO voltage would not longer work, since it's not configured?

weebl2000:
I didn't check after updating from the 7.7.1 pin to the tag, will have a look.

weebl2000:
> Seems that variables for LDO and TCXO have been renamed, but not passed in at all with new config etc. So I guess in some cases, radios that expect TCXO voltage would not longer work, since it's not configured?

They're not locals - code should do the same as before.

I think this will clear it up

- https://github.com/jgromes/RadioLib/blob/7.7.1/src/modules/SX126x/SX126x.h#L76-L82
- https://github.com/jgromes/RadioLib/blob/7.7.1/src/modules/SX126x/SX1262.cpp#L72-L84

src/helpers/radiolib/CustomSX1262.h
weebl2000: It's in the base class now.

PR #2916, opened 2026-07-09
Dynamic CAD sensitivity

weebl2000:
Probe every 4 seconds to measure actual channel activity. If channel seems busy we adjust CAD to be less sensitive. If channel seems empty maximum CAD sensitivity.

Build here: https://mcimages.weebl.me/?commitId=dynamic-cad

Based on conversation in #1727 - this should make a busier node be less senitive to preambles. Tagging @recrof and @syssi would be interesting to see how this changes behavior.

You can use get cad and it will show stats for actual preambles detected during idle periods.

### DIFF
--- a/examples/simple_repeater/MyMesh.h
+++ b/examples/simple_repeater/MyMesh.h
@@ -213,6 +213,7 @@ public:
   void removeNeighbor(const uint8_t* pubkey, int key_len) override;
   void formatStatsReply(char *reply) override;
   void formatRadioStatsReply(char *reply) override;
+  mesh::Radio* getRadio() override { return _radio; }
   void formatPacketStatsReply(char *reply) override;
   void startRegionsLoad() override;
   bool saveRegions() override;
--- a/examples/simple_repeater/MyMesh.h
+++ b/examples/simple_repeater/MyMesh.h
@@ -216,6 +216,7 @@ public:
   void removeNeighbor(const uint8_t* pubkey, int key_len) override;
   void formatStatsReply(char *reply) override;
   void formatRadioStatsReply(char *reply) override;
+  mesh::Radio* getRadio() override { return _radio; }
   void formatPacketStatsReply(char *reply) override;
   void startRegionsLoad() override;
   bool saveRegions() override;
--- a/examples/simple_room_server/MyMesh.h
+++ b/examples/simple_room_server/MyMesh.h
@@ -215,6 +215,7 @@ public:
   }
   void formatStatsReply(char *reply) override;
   void formatRadioStatsReply(char *reply) override;
+  mesh::Radio* getRadio() override { return _radio; }
   void formatPacketStatsReply(char *reply) override;
   void startRegionsLoad() override;
   bool saveRegions() override;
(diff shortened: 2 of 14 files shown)


usrflo:
@weebl2000 : I tested in my setup with 5 weakly connected nodes, 50 direct messages in 10s intervals, interference by a LoRa node sending local adverts every 6 seconds and a successful delivery counted if the last node ACKs.

Looks good: the problems I saw with https://github.com/meshcore-dev/MeshCore/commit/7c8e09245760c71a64f14ca46b4ff6b8dcdd3aa9 disappeared.

**My results:**

- cad off with this branch: 37/38 ACKs for 50 messages
- cad on with this branch: 39/40 ACKs for 50 messages
- cad on with this branch plus the latest dev commits: 42/36 ACKs for 50 messages

Attached you can find the monitoring logs.

[dev-50-wo-cad.txt](https://github.com/user-attachments/files/30196971/dev-50-wo-cad.txt)
[dev-50-with-cad.txt](https://github.com/user-attachments/files/30196970/dev-50-with-cad.txt)
[dev-50-wo-cad-2.txt](https://github.com/user-attachments/files/30196969/dev-50-wo-cad-2.txt)
[dev-50-with-cad-2.txt](https://github.com/user-attachments/files/30196968/dev-50-with-cad-2.txt)
[dev-50-with-cad-latest-dev.txt](https://github.com/user-attachments/files/30196967/dev-50-with-cad-latest-dev.txt)
[dev-50-with-cad-2-latest-dev.txt](https://github.com/user-attachments/files/30196965/dev-50-with-cad-2-latest-dev.txt)

usrflo:
I wonder why using dynamic CAD together with repeated sending https://github.com/meshcore-dev/MeshCore/pull/2670 decreases deliverability significantly. Resending happens relatively often, but following nodes seem not to react. May the CAD analysis create too much deafness? I wonder why I didn't see this effect when running the same test without the sending repetition.

[dyn-cad-on-rep.txt](https://github.com/user-attachments/files/30201825/dyn-cad-on-rep.txt)

ignisf:
I've been running this for a while now on a companion in a network with 80+ repeaters in Sofia. I've noticed no ill-effects and I think it should be merged sooner rather than later to complement the release of CAD.

Magalex2x14:
I've also been using this on my repeater since publishing. For now, I can only say that the repeater seems to have become more responsive, but flag 2 is still set quite quickly for 8 seconds CADFailMaxDuration (about 40 neighbors, 15-19 of them with positive SNR). I increase CADFailMaxDuration and observe. Raised it to 16 seconds (4->8->16) just now. CoreScope shows that in our network the packet propagation time reaches 20 seconds...

ignisf:
@IoTThinks consider adding this patch to your PS 17.

IoTThinks:
@ignisf We will explore this PR.

How does it help you in your usecase?

IoTThinks:
IMO, all devices should have CAD on always.
Else, everyone will try to have more TX power and pollute the air.

Since CAD is an option, eventually, people will set it off.

recrof:
> IMO, all devices should have CAD on always. Else, everyone will try to have more TX power and pollute the air.

before we set CAD on for everyone, we need to have good auto-detection for dtPeak/dtMin that won't run every 4 seconds. then we need to have option to manually override those 2 parameters for corner cases. when testing this with radios that have very sensitive LNA, it caused lot of trouble - that's why we still have it disabled by default.

PR #2932, opened 2026-07-12
stm32wl: implement powersaving for repeaters

h0lad:
This is a first attempt of getting powersaving running on the STM32WL (RAK3x72) devices.

First measurements:

- RAK3x72-T (TXCO) has around ~8-9mA in Stop. The TXCO eats around 2mA.
- RAX3x72 (w/o TXC0) should be around 5-6.5mA in Stop. Was not able to test this yet due to lack of devices.

Setup: Nordic Power Profiler @ 3700mV -> BQ25570 -> RAK3172-T at 868MHz.

image

24h test on the balcony now running.

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -442,7 +442,7 @@ void CommonCLI::handleCommand(uint32_t sender_timestamp, char* command, char* re
       _prefs->powersaving_enabled = 1;
       savePrefs();
       strcpy(reply, "on - Immediate effect");
-#elif defined(ESP32) && !defined(WITH_BRIDGE)
+#elif (defined(ESP32) || defined(STM32_PLATFORM)) && !defined(WITH_BRIDGE)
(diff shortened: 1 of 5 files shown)


fdlamotte:
thanks, will look into it ...

for the tcxo part it really depends if the board has tcxo or not, I have both and they don't behave the same

h0lad:
> thanks, will look into it ...
> 
> for the tcxo part it really depends if the board has tcxo or not, I have both and they don't behave the same

Good point. Radiolib has autoprobing for the TXCO I accidentally nuked. Reverted this.

fdlamotte:
what was the issue ? 
didn't have time to dig into it but it seemed promising !

h0lad:
Oh... deleted  it by accident! Damn.

Wanted to add a comment with "Runs for >10 days with no problems" and somehow fat fingered the close button.

Currently testing this patch with the new RX Preamble Detection   implementation.

PR #2936, opened 2026-07-12
Feed external hardware watchdog on Heltec Tower V2 (aka MeshTower V2) Implementation 2

beala:
This addresses the same issue as https://github.com/meshcore-dev/MeshCore/pull/2904 and only one of these should be merged. If you haven't already, please read that PR's description first.

In the other fix, I simply started a `SoftwareTimer` that would feed the watchdog every 60 seconds. That was the smallest change I could make to fix the issue. I know the core devs are very busy, so I wanted to make the fix as small as possible.

This PR contains a more comprehensive fix, but is also more code. It starts with the changes @Quency-D made in https://github.com/meshcore-dev/MeshCore/pull/1446 (which fixes the issue in MeshTower V1, but was never merged) and updates it to also fix the issue in MeshTower V2. In 1446, Quency added an API for feeding external watchdogs, and wired into the main loops of the different node types. With the correct `#define`s, this same code works with MeshTower V2. I updated the `#define`s and added an instance of `ExternalWatchdogManager` to the V2's variant.

## Testing
I flashed this to my device and verified the device hasn't rebooted.

If you'd like to test this yourself, here are build instructions for a repeater:

```bash
# Checkout my PR
git fetch origin pull/2936/head:pr-2936
git checkout pr-2936

# Build with a custom version string and serial debug logging turned on
PLATFORMIO_BUILD_FLAGS='-D FIRMWARE_VERSION=\"v1.16.0-watchdog-fix-2\"' pio run -e Heltec_tower_v2_repeater
```

Head over to https://meshcore.io/flasher and use the "Custom firmware" option to flash `.pio/build/Heltec_tower_v2_repeater/firmware.zip`

Here's the compiled firmware for repeaters if you don't want to build it yourself: [firmware.zip](https://github.com/user-attachments/files/29946758/firmware.zip)


cwichura:
While @2904 has the description of why, this is the PR that should get merged.  It creates a common framework for any node type with a hardware watchdog, and integrates it properly into the main loop.  devteam, please use this PR.

Quency-D:
here you go!

liamcottle:
Looks good, thanks! Will merge.

If there's any issues it would only affect the boards that have opted in via the watchdog build flag.

I'll close the original #1446 PR by @Quency-D, thanks for your work on that :)

Apologies for delay with original PR, lots has been going on behind the scenes!

PR #2942, opened 2026-07-13
Add support for Elecrow ThinkNode M9

liamcottle:
This PR adds initial support for the Elecrow ThinkNode M9.

Currently, the default companion UI isn't too interesting for this device, and the keyboard has not been wired up, so navigation is not possible.

For now, it's being added in a disabled state, where firmware won't be auto built by our release flow on GitHub Actions. However, users can build the firmware themselves if they want it.

The intention is to provide the base structure that's successfully compiling and running, to allow users to tinker and pave the way for a new companion UI.

I have tested the following firmwares:

- companion_radio_ble
- companion_radio_usb
- companion_radio_wifi
- repeater
- room_server

Note that this device uses a keyboard that sends key events over I2C. You would need to poll I2C each loop, and handle the key events that have been provided. For now, we don't have any input source framework in place for injecting key events from custom keyboards into the existing UITasks.

It would be best to avoid having to setup drivers for all keyboards inside the UITask, and instead have a way to register input sources on the Board object, and allow UITasks to register key event callbacks via the provided Board instance.

This will be left for the future :)

### DIFF
--- a/boards/thinknode_m9.json
+++ b/boards/thinknode_m9.json
@@ -1,0 +1,57 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "esp32s3_out.ld",
+      "memory_type": "qio_opi",
+      "partitions": "default_16MB.csv"
+    },
+    "core": "esp32",
+    "extra_flags": [
+      "-DBOARD_HAS_PSRAM",
+      "-DARDUINO_USB_CDC_ON_BOOT=0",
+      "-DARDUINO_USB_MODE=0",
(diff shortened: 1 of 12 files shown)


ripplebiz:
You don't need that 'libqst.a' static lib.  Apparently, that just has some accelerometer driver in it.

liamcottle:
> You don't need that 'libqst.a' static lib. Apparently, that just has some accelerometer driver in it.

Removed, and force pushed with commits squashed.

PR #2959, opened 2026-07-15
Add USB serial bridge transport for repeater firmware

### LINKED ISSUE #2958: Add USB serial bridge transport for repeater firmware (RS232Bridge over USB CDC)
## Use case

Some repeater deployments are connected to a host over USB rather than a physical UART.  Providing the existing serial bridge protocol, but over native USB CDC, removes the need for separate RX/TX wiring while retaining the bridge packet format.

A real-world example of this use case is [Raven (AREDN messaging) integration](https://github.com/kn6plv/Raven/wiki/MeshCore).

## Proposal

Extend `RS232Bridge` (using a new `WITH_USB_SERIAL_BRIDGE` build flag), so repeater firmware targets can expose the binary bridge stream through a second USB CDC interface.

This moves the bridge frames from UART to USB serial:

- USB CDC 0 remains the normal MeshCore CLI/debug `Serial` port.
- USB CDC 1 carries the existing bridge frames.
- Existing UART bridge targets remain unchanged.
- Additional bridge targets (e.g., `RAK_4631_repeater_bridge_usbserial`) can be added as needed.

The bridge wire format is unchanged, only the transport.

## Initial targets

I've bench-tested these successfully:

- `RAK_4631_repeater_bridge_usbserial`
- `SenseCap_Solar_repeater_bridge_usbserial`

darrylhodgins:
Adds native USB CDC support to `RS232Bridge` for selected repeater targets with a second USB CDC interface.

Initially, includes these targets:

- RAK4631
- SenseCAP Solar

The existing serial bridge frame format is unchanged; only its physical transport changes.

- USB CDC 0 remains the normal MeshCore CLI/debug `Serial` port.
- USB CDC 1 carries the serial bridge stream.

For all other bridge targets, existing UART behaviour remains unchanged.

Closes #2958

## Testing

Built and bench-tested on physical hardware:

- `RAK_4631_repeater_bridge_usbserial`
- `SenseCap_Solar_repeater_bridge_usbserial`

Verified that the CLI and bridge enumerate as separate USB serial ports and that bridge traffic works over the second port.

### DIFF
--- a/examples/simple_repeater/MyMesh.h
+++ b/examples/simple_repeater/MyMesh.h
@@ -14,6 +14,10 @@
 #endif
 
 #ifdef WITH_RS232_BRIDGE
+#ifdef WITH_USB_SERIAL_BRIDGE
+#include <Adafruit_TinyUSB.h>
+extern Adafruit_USBD_CDC bridgeSerial;
+#endif
 #include "helpers/bridges/RS232Bridge.h"
 #define WITH_BRIDGE
 #endif
--- a/examples/simple_repeater/main.cpp
+++ b/examples/simple_repeater/main.cpp
@@ -3,6 +3,10 @@
 
 #include "MyMesh.h"
 
+#ifdef WITH_USB_SERIAL_BRIDGE
+Adafruit_USBD_CDC bridgeSerial;
+#endif
+
 #ifdef DISPLAY_CLASS
   #include "UITask.h"
   static UITask ui_task(display);
@@ -36,6 +40,10 @@ static unsigned long userBtnDownAt = 0;
 #endif
 
 void setup() {
+#ifdef WITH_USB_SERIAL_BRIDGE
+  // Register the second CDC interface before Serial mounts the USB device.
+  bridgeSerial.begin(115200);
+#endif
   Serial.begin(115200);
   delay(1000);
 
--- a/src/helpers/bridges/RS232Bridge.cpp
+++ b/src/helpers/bridges/RS232Bridge.cpp
@@ -2,6 +2,10 @@
 
 #include <HardwareSerial.h>
 
+#ifdef WITH_USB_SERIAL_BRIDGE
+#include <Adafruit_TinyUSB.h>
+#endif
+
 #ifdef WITH_RS232_BRIDGE
 
 RS232Bridge::RS232Bridge(NodePrefs *prefs, Stream &serial, mesh::PacketManager *mgr, mesh::RTCClock *rtc)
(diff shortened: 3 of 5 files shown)


enigmaspb:
Will it work if I connect two boards RAK4631 / RAK19007  with a USB cable?

darrylhodgins:
> Will it work if I connect two boards RAK4631 / RAK19007  with a USB cable?

It may be possible to have one of the boards act as a USB host, but as it stands at this point, both boards will just sit there waiting for a host to appear. At least for now, the UART approach is the way to go for that.

This initial PR is meant to provide the serial stream to a computer (in my case it's a simple Raspberry Pi).

I'll look into whether USB OTG or having one of the boards act as a USB host is possible, but I think that would belong in a separate PR.

enigmaspb:
I see, that's a pity, I was hoping to avoid soldering. Thank you for your response.

hickey:
I have just published #3020 which is a similar effort. Instead of trying shoehorn the code in to the repeater code base I have created a USBSerial bridge code that is in `src/helpers/bridges` that gets activated by defining `WITH_USB_SERIAL_BRIDGE`. There is secondary define that specifies the serial stream to use for the bridge transport. In theory this would support this author's use case, but I am not familiar enough with platforms that support multiple USB serial ports to confirm if this is true or not. 

In addition, I have added the TCP bridge to #3020 in a similar manner. Further details are described in #3020. 

I have also added code to return more descriptive values for the `get bridge.type` CLI command that is not present in this PR.

hickey:
Killed off #3020 as it was not based on the `dev` branch. 

Reopened the PR as #3029 and there is a feature request issue created as #3028.

synalysis:
@darrylhodgins @hickey In the past days I've created a server that uses your bridge firmware to create tunnels between MeshCore islands: https://isthmus.channel

The transport is done transparently via Reticulum or Nostr connections, so the MeshCore networks ideally behave like one.

Any feedback welcome!

Talie5in:
> In the past days I've created a server that uses your bridge firmware to create tunnels between MeshCore islands: https://isthmus.channel
> 
> The transport is done transparently via Reticulum or Nostr connections, so the MeshCore networks ideally behave like one.
> 
> Any feedback welcome!

The community don't want packets to potentially go over IP, and you cant enforce Reticulum packets to only be RF either.
See https://github.com/meshcore-dev/MeshCore/pull/3030#issuecomment-5053378182

@ripplebiz - duplicate of the other one you closed (issue and older PR here)

Issue #2747, opened 2026-06-11
Proposal: Hierarchical Region Gating via TX Duty Cycle


khagele:
### Problem
MeshCore networks are vulnerable to disruption from intentional spammers, misconfigured nodes, or high-traffic events. See https://mc-spamdetector.nl/incidents?status=all&type=all&hours=0 for recent events on the NL mesh. For the Dutch mesh a volunteer group (NOC) is monitoring and trying to coordinate the mesh through such events. 

Current mitigation relies on in-person triangulation and activation of the regional social clusters. When mitigation of such disruptive events using adjustments to the repeater configuration is possible, this either requires local repeater admin access or that the mesh is still functional enough to send/receive commands to the repeater remotely. During such events, the mesh is both not functional and at the same time wired access to repeaters is not easily obtained locally. To make the mesh more resilient to high-traffic events, this gate mechanism based on regions is proposed for discussion. 

### Proposed Mechanism
Repeaters autonomously gate inter-region traffic based on locally measured TX duty cycle. Each repeater has a configured region list, where order defines priority:
```
nl-ge-nij (local cluster, highest priority)
 └─ nl-ge
    └─ nl
      └─ eu
```
When TX duty cycle exceeds a configurable threshold T1, the repeater progressively drops traffic from the back of the list forward:

1. Duty cycle > T1 → stop forwarding `eu` traffic
2. Still above threshold → stop forwarding `nl` traffic
3. Still above threshold → stop forwarding `nl-ge` traffic
4. Last `nl-ge-nij` cluster always protected

**Recovery**
As duty cycle drops below T_recover (where T_recover < T1), the repeater re-enables regions from the innermost outward, one step at a time. The gap between T1 and T_recover provides hysteresis, preventing rapid oscillation. One known risk: when an event ends, repeaters independently observe the drop at roughly the same time and begin re-enabling simultaneously, potentially causing a brief surge of traffic that re-triggers gating. A small randomised per-repeater recovery delay (e.g. 0-30s jitter per step) would reduce this risk.

**TX duty cycle**
TX duty cycle is a locally observable metric. Each repeater acts independently based on its own radio behaviour, preserving the decentralised nature of the mesh.

**Failure behaviour**
Under sustained disruption, the mesh degrades gracefully: each repeater converges on its smallest sustainable local cluster. The network self-partitions into isolated but functional local cells rather than collapsing entirely. Once disruption subsides, the mesh recovers outward, restoring broader connectivity step by step.

[...]

PR #2960, opened 2026-07-15
Hierarchical region gating via TX duty cycle (simple_repeater)

khagele:
Implements the proposal in #2747: autonomous, per-repeater gating that sheds inter-region traffic when a repeater's own TX duty cycle is high, so the local cluster keeps working during high-traffic/disruption events without needing admin access.

This is a reference implementation of the proposal, opened to move the discussion to concrete code. Credit to @fizzyfuzzle, whose POC in the issue thread this tracks and generalizes, and to the NL/CZ operators (@LennertVA, @DreitTheDragon, @onnozweers, @titusn) whose feedback shaped the design.

### How it works
Reuses the existing `RegionMap` parent hierarchy and `Dispatcher` airtime budget. When TX duty cycle exceeds a threshold, regions are gated from the outermost layer inward — wildcard `*` first, then the broadest named regions — always protecting the innermost cluster. Recovery re-enables regions inside-out once duty cycle falls below `threshold − hysteresis`, with per-step random jitter to desync recovery across repeaters. Multiple regions on the same hierarchy level gate together; their leaves stay protected.

### Design / safety
- **Transient overlay:** a new non-persisted `rt_flags` field on `RegionEntry` is OR'd with config `flags` in the forward path, so a `region save` or reboot mid-event can never make a deny permanent.
- **Self-healing state:** the gate level is re-asserted every check interval, so region edits or a bulk `region load` commit (which rebuild entries with cleared `rt_flags`) cannot desync the gate; disabling the feature while gated lifts the gate immediately.
- **Idle-safe recovery:** `Dispatcher::getTxDutyCyclePercent()` refreshes the airtime budget before reading, so recovery still fires when traffic (and thus TX) stops.

### Config (opt-in, off by default; persisted with upgrade-safe append)
`set dc.gate <0|1>` · `set dc.gate.thresh <1-100>` (default 70) · `set dc.gate.hyst <0-50>` (default 10) · `get dc.gate[.thresh|.hyst]` · `get dc.gate.status` (live duty % + gate level).

Gate state is intentionally **not** added to the `region` tree dump for now — that format is parsed and round-tripped by deployed apps, so surfacing per-region gate state in "Manage Regions" is left as an open question. Live state is available via `get dc.gate.status`.

### Verification
- Full `Heltec_E290_repeater` firmware builds & links.
- Native gtest suite (`test/test_region_gating/`) covering wildcard-first ordering, leaf protection, inside-out recovery, config/runtime separation, same-level regions, and edge cases; native mocks (`Arduino.h`/`Stream.h`/`SHA256.h`) extended so `RegionMap` builds off-device.

### Open questions / out of scope
- Duty-cycle metric uses the existing fixed 1-hour window → strong hysteresis but slow recovery; the window could be made configurable later (override `getDutyCycleWindowMs()`, mirroring `getAirtimeBudgetFactor()`).
- On an *unbalanced* tree, a whole-depth gate can drop a shallow branch's leaf before the deepest branch's; a per-branch "never gate a childless leaf" refinement is a possible follow-up.
- Reserved admin airtime (@titusn) and a flat priority-number model (@onnozweers) are noted as separate follow-ups.
- Only wired into `simple_repeater`; `simple_room_server` would be a trivial follow-up.

### DIFF
--- a/test/test_region_gating/test_region_gating.cpp
+++ b/test/test_region_gating/test_region_gating.cpp
@@ -3,120 +3,195 @@
 
 // Unit tests for duty-cycle region gating (issue #2747).
 // These exercise the pure hierarchy/level logic; no crypto or filesystem is touched.
+// Each test covers one gating scenario (one level, one transition, or one invariant)
+// so a failure points directly at the behaviour that broke.
 
 static bool gated(RegionMap& rm, const char* name) {
   RegionEntry* r = rm.findByName(name);   // findByName("*") returns the wildcard
   return r && (r->rt_flags & REGION_DENY_FLOOD);
 }
 
-// Build:  * > eu > nl > nl-ge > nl-ge-nij   (a single deep chain)
-static void buildNlChain(RegionMap& rm) {
-  RegionEntry* eu   = rm.putRegion("eu", 0);
-  RegionEntry* nl   = rm.putRegion("nl", eu->id);
-  RegionEntry* nlge = rm.putRegion("nl-ge", nl->id);
-  rm.putRegion("nl-ge-nij", nlge->id);
-}
-
-// Build:  * > cz > { cz-ulk, cz-stc, cz-lbk }   (three regions on the same level)
-static void buildCzFlat(RegionMap& rm) {
-  RegionEntry* cz = rm.putRegion("cz", 0);
-  rm.putRegion("cz-ulk", cz->id);
-  rm.putRegion("cz-stc", cz->id);
-  rm.putRegion("cz-lbk", cz->id);
-}
-
-TEST(RegionGating, DepthAndMaxGateLevel) {
+// Fixture:  * > eu > nl > nl-ge > nl-ge-nij   (a single deep chain)
+class NlChainGating : public ::testing::Test {
+protected:
   TransportKeyStore ks;
-  RegionMap rm(ks);
-  buildNlChain(rm);
+  RegionMap rm{ks};
+
+  void SetUp() override {
+    RegionEntry* eu   = rm.putRegion("eu", 0);
+    RegionEntry* nl   = rm.putRegion("nl", eu->id);
+    RegionEntry* nlge = rm.putRegion("nl-ge", nl->id);
+    rm.putRegion("nl-ge-nij", nlge->id);
+  }
+};
+
+// Fixture:  * > cz > { cz-ulk, cz-stc, cz-lbk }   (three regions on the same level)
+class CzFlatGating : public ::testing::Test {
+protected:
+  TransportKeyStore ks;
+  RegionMap rm{ks};
+
+  void SetUp() override {
+    RegionEntry* cz = rm.putRegion("cz", 0);
+    rm.putRegion("cz-ulk", cz->id);
+    rm.putRegion("cz-stc", cz->id);
+    rm.putRegion("cz-lbk", cz->id);
+  }
+};
+
+// ---------- depth model ----------
+
+TEST_F(NlChainGating, MaxDepthCountsChainDepth) {
   EXPECT_EQ(rm.getMaxDepth(), 4);
+}
+
+TEST_F(NlChainGating, MaxGateLevelEqualsMaxDepth) {
   EXPECT_EQ(rm.getMaxGateLevel(), 4);
 }
 
-TEST(RegionGating, GatesWildcardFirstThenOutermostInward) {
-  TransportKeyStore ks;
-  RegionMap rm(ks);
-  buildNlChain(rm);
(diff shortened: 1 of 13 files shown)


test/test_region_gating/test_region_gating.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+TEST(RegionGating, GatesWildcardFirstThenOutermostInward) {
+  TransportKeyStore ks;
+  RegionMap rm(ks);
+  buildNlChain(rm);
+
+  rm.applyDutyGate(0);
+  EXPECT_FALSE(gated(rm, "*"));
+  EXPECT_FALSE(gated(rm, "eu"));
```
titusn, line 57: /nitpick: should this not expect all four regions to not be gated and would splitting this test up into multiple tests make it more readable, especially for later debugging if the test ever fails? If this is according to the current coding guidelines, no problem of course. However a basic rule is: a test should contain one expect. Same can be said about the tests below.

khagele: Good point, thanks - addressed in 13935ba. The suite is now one-scenario-per-test: each gate level, each recovery transition, and each invariant is its own test, and the baseline/level tests assert the state of every region rather than a subset. I kept a handful of expects per test where they jointly define a single behaviour (e.g. "level 2 gates `*` and `eu`, and nothing else"), since a strict one-expect split would obscure that. Failure output still points at the exact region and level that broke.

khagele:
Pushed 13935ba, which adds a home region exemption to the gating, following a question raised in the Dutch community discussion: should the home region or the innermost layer have the highest priority?

Previously gating protected only the deepest layer of the hierarchy, so a repeater with home set to `nl-ge` in `* > eu > nl > nl-ge > nl-ge-nij` would gate its own home region at max level while protecting the deeper `nl-ge-nij`. Since the home region is an explicit operator statement of which cluster the repeater serves (a better signal than the depth heuristic), it is now never gated, in addition to the always-protected innermost layer. Repeaters without a home region keep the pure depth-based behaviour.

One trade-off worth discussing: if an operator sets home to a broad region (e.g. `home = eu`), that region stays ungated and load-shedding is largely defeated on that repeater. The depth rule can't be misconfigured that way; the home rule can. Guidance would be "set home to your local cluster region", but if people feel this is too footgun-prone, the exemption could be restricted (e.g. only exempt home if it's within the deepest N layers).

The same commit also restructures the unit tests into one-scenario-per-test cases with complete per-region assertions, and adds coverage for the exemption.

LennertVA:
> One trade-off worth discussing: if an operator sets home to a broad region (e.g. `home = eu`), that region stays ungated and load-shedding is largely defeated on that repeater. The depth rule can't be misconfigured that way; the home rule can.

But then the behaviour reverts to the current default behaviour (right?), which I would think is not a problem; it has no active negative effects, only a lack of positive effects. Not getting a bonus because you configured your machines badly is acceptable - there are far worse things that can happen if you screw up other settings.

Nice work! Hope to see this in stable release soon.

khagele:
Can also be tested here: https://mcimages.weebl.me/?commitId=13935ba

DreitTheDragon:
At this point, wouldn't it be simpler to implement what @onnozweers mentioned in https://github.com/meshcore-dev/MeshCore/issues/2747#issuecomment-4688993246 ?
It's not complaint, just reminder/suggestion.

khagele:
Fair reminder! I see it this way: the core of this PR (gate outward-in on TX duty cycle, with hysteresis, recovery jitter, and transient state that can't be persisted by accident) is independent of how region priority is determined. Right now priority is derived implicitly from the hierarchy depth, which costs zero extra configuration for meshes that already maintain a region hierarchy, with the home region exemption covering the main case depth can't express.

An explicit per-region priority number would be more flexible for unbalanced or flat setups, but it also means every operator has to assign and maintain numbers, plus a new field, CLI command, and persistence. My suggestion: field-test the depth-based version first, and if real topologies turn out to need it, add an optional priority override later ("priority = depth unless explicitly set"). That would be a small, compatible change on top of this mechanism rather than a rewrite. Curious what others think.

LennertVA:
> My suggestion: field-test the depth-based version first, and if real topologies turn out to need it, add an optional priority override later ("priority = depth unless explicitly set"). That would be a small, compatible change on top of this mechanism rather than a rewrite.

Best approach indeed, and as mentioned before - we're talking about tweaking values to get a bonus, with the absolute worst case you can get by not doing that or doing it wrong = getting the current behaviour. 

No conflicts, that looks promising. Hopefully merged in the next release. :smile:

PR #2962, opened 2026-07-15
Nrf52 softdevice checks reduction

pemensik:
Reduce repeating check for softdevice mode. Use common function for it. Reduce also repetitions in recently added  getMCUTemperature


pemensik:
This should improve some unnecessary repetitions seen in #2897. It does not add any new functionality, just polishes the code.

PR #2966, opened 2026-07-16
Add Heltec rc32 board

Quency-D:
Adds support for the Heltec RC32 board, including:

- SX1262 radio and GPS configuration
- NV3001B display support
- Rotary encoder input
- PlatformIO environments for supported firmware variants


examples/companion_radio/ui-new/UITask.cpp
```diff
@@ -750,6 +750,16 @@ void UITask::loop() {
     c = handleTripleClick(KEY_SELECT);
   }
 #endif
+#if defined(UI_HAS_ROTARY_INPUT)
+  if (c == 0) {
+    RotaryInputEvent ev = rotary_input.poll();
+    if (ev == RotaryInputEvent::Next) {
+      c = checkDisplayOn(KEY_NEXT);
```
ripplebiz: Just wondering if it would be better to _not_ have the screen woken from rotary up/down events? Could easily get bumped while in pocket, and waste battery(?)

variants/heltec_rc32/HeltecRC32Board.cpp
```diff
@@ -0,0 +1,78 @@
+#include "HeltecRC32Board.h"
+#if defined(UI_HAS_ROTARY_INPUT)
+#include "HeltecRC32RotaryInput.h"
+#endif
+
+#if defined(UI_HAS_ROTARY_INPUT)
```
ripplebiz: I recommend moving this to target.cpp

variants/heltec_rc32/HeltecRC32Board.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include <Arduino.h>
+#include <driver/rtc_io.h>
+#include <helpers/ESP32Board.h>
+#include <helpers/RefCountedDigitalPin.h>
+#if defined(UI_HAS_ROTARY_INPUT)
```
ripplebiz: recommend moving this to target.h

examples/companion_radio/ui-new/UITask.cpp
Quency-D: You are right. The rotary encoder's wake-up function has been disabled.

PR #2968, opened 2026-07-16
Allow .envrc to integrate into existing direnv environments

hickey:
The existing `.envrc` sets up the Nix environment which is nice, but not all of us use Nix.

This change will make setting up the Nix environment conditional upon finding `nix-shell`. This way if the `direnv` is being executed outside of the Nix environment `direnv` will not fail. 

In addition, the `source_up` stdlib call is used to process all parent `.envrc` files.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,2 @@
-use nix
+source_up 
+nix-shell --version >/dev/null 2>&1 && use nix


hickey:
I should have read `CONTRIBUTING.md` a while ago. I will close this PR and open a new one to fix my faux pas.

PR #2971, opened 2026-07-16
Fix Git error during project configuration

YoshiWalsh:
Previously project configuration would fail with "fatal: couldn't find remote ref d541301" (git v2.55.0)

This affected almost all platformio commands too. E.g.

```
 *  Executing task: platformio run --environment RAK_3401_repeater 

Processing RAK_3401_repeater (board: rak3401; platform: nordicnrf52; framework: arduino)
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Tool Manager: Installing git+https://github.com/meshcore-dev/Adafruit_nRF52_Arduino#d541301
git version 2.55.0
hint: Using 'master' as the name for the initial branch. This default branch name
hint: will change to "main" in Git 3.0. To configure the initial branch name
hint: to use in all of your new repositories, which will suppress this warning,
hint: call:
hint:
hint:   git config --global init.defaultBranch <name>
hint:
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint:
hint:   git branch -m <name>
hint:
hint: Disable this message with "git config set advice.defaultBranchName false"
Initialized empty Git repository in /root/.platformio/.cache/tmp/pkg-installing-wokqua5c/.git/
fatal: couldn't find remote ref d541301
VCSBaseException: VCS: Could not process command ['git', 'fetch', '--depth=1', 'origin', 'd541301']
```

Per https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt-refspec, the <src> argument needs to be either a ref (read: branch or tag) or a "**fully spelled** hex object name". The error appears to have been caused by the partial hex.

This requirement is also mentioned in this SO answer: https://stackoverflow.com/a/30701724

After this change, project configuration now succeeds.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -84,7 +84,7 @@ platform_packages =
   ; use internal fork that includes patch to ble stack to prevent firmware lockup during rapid connect/disconnect
   ; https://github.com/meshcore-dev/MeshCore/pull/1177
   ; https://github.com/meshcore-dev/MeshCore/pull/1295
-  framework-arduinoadafruitnrf52 @ https://github.com/meshcore-dev/Adafruit_nRF52_Arduino#d541301
+  framework-arduinoadafruitnrf52 @ https://github.com/meshcore-dev/Adafruit_nRF52_Arduino#d541301665b40959682252911e57b11df3ee651a
 extra_scripts = create-uf2.py
 build_flags = ${arduino_base.build_flags}
   -D NRF52_PLATFORM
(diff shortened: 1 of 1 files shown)


liamcottle:
Please rebase onto `dev` branch.

YoshiWalsh:
Done :)

liamcottle:
Can you rebase your local branch and force push please. Otherwise it pulls in those main branch changes.

If you create your local branch from dev instead of main, it doesn't pull those in :)

YoshiWalsh:
Oh I didn't realise there were commits reachable from main that aren't reachable from dev. Should be fixed now.

liamcottle:
> Oh I didn't realise there were commits reachable from main that aren't reachable from dev. Should be fixed now.

Yeah, it's mostly just documentation that gets opened against `main`, since the website etc builds from `main` branch. But all firmware/code related changes should be branched from `dev` and PR opened against `dev`.

Thanks for the fix :) Seems that this issue only affects some build environments, as it works fine on my end. But merging as I have seen the same issue for another dependency in the past.

YoshiWalsh:
My hunch is that it's probably related to which version of Git you have installed.

PR #2977, opened 2026-07-17
Add IRQ timeout logic for SX1262 and LR1121

oltaco:
This PR clears stale Rx IRQ bits when they have been active for too long without an RxDone happening, preventing transmissions needing to wait for the full 4 second timeout under certain conditions.

### DIFF
--- a/src/helpers/radiolib/CustomLR1110.h
+++ b/src/helpers/radiolib/CustomLR1110.h
@@ -72,7 +72,7 @@ class CustomLR1110 : public LR1110 {
       return false;
     }
 
-    PacketMillis setMaxPacketMillis(PacketMillis maxPacketMillis) {
+    void setMaxPacketMillis(PacketMillis maxPacketMillis) {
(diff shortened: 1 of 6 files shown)


src/helpers/radiolib/CustomLR1110.h
```diff
@@ -32,9 +36,46 @@ class CustomLR1110 : public LR1110 {
...
+
+          return false;
+        }
+        return true;
+      }
+      _activityAt = 0; _headerSeen = false;
+      return false;
+    }
+
+    PacketMillis setMaxPacketMillis(PacketMillis maxPacketMillis) {
```
ripplebiz: should this be void return type?

src/helpers/radiolib/CustomSX1262.h
```diff
@@ -99,9 +101,46 @@ class CustomSX1262 : public SX1262 {
...
+
+          return false;
+        }
+        return true;
+      }
+      _activityAt = 0; _headerSeen = false;
+      return false;
+    }
+
+    PacketMillis setMaxPacketMillis(PacketMillis maxPacketMillis) {
```
ripplebiz: same here? (void)

oltaco:
@ripplebiz spot on, fixed 👍

PR #2980, opened 2026-07-18
Add next-hop retry for direct-path relays

benallfree:
This is a draft for visibility. *Note: this intro section will stay updated, so it may not match comments below*

This is a wire-compatible PR for both passive and active direct hop ACKs.

Right now, repeaters relay-and-forget. Any missed link in the path results in a timeout and the user has to retry the entire path. 

With this enhancement, repeaters will listen for an echo from the next hop. If they don't hear it, they will resend. If the receiving repeater hears a resend after it has already processed a message, it will send a hard ACK to prevent further spurious repeats.

We also add several CLI settings, including a volatile `ignore` setting to force link repeats for testing:

| Command | Type | Parameter | Range | Default | Persisted | Purpose |
|---------|------|-----------|-------|---------|-----------|---------|
| `get hop.retry` | read | — | — | `2` | yes | Extra retransmits if next hop echo or HOP_ACK is missed (`0` = off) |
| `set hop.retry <count>` | write | `count` | 0–5 | `2` | yes | Up to `count + 1` total TX (initial + retries) |
| `get hop.retry.ms` | read | — | — | `1500` | yes | Base listen window (ms) before each retry |
| `set hop.retry.ms <ms>` | write | `milliseconds` | 200–10000 | `1500` | yes | Base window; actual deadline also includes `direct.txdelay` and packet airtime |
| `get hop.ignore` | read | — | — | `0` | no | Remaining forward drops (test mode) |
| `set hop.ignore <count>` | write | `count` | 0–255 | `0` | no | Silently drop next N direct-forward opportunities (no forward, no HOP_ACK, no dedup mark); resets on reboot |

Prior art:

* https://github.com/meshcore-dev/MeshCore/pull/2367
* https://github.com/meshcore-dev/MeshCore/pull/2501
* https://github.com/meshcore-dev/MeshCore/pull/2670


usrflo:
Hi @benallfree , 

_> From what I can tell, these listen for the message echo._

that's right, at least https://github.com/meshcore-dev/MeshCore/pull/2670 and https://github.com/meshcore-dev/MeshCore/pull/2367 do.

_> That was my first instinct too, but after more thinking I feel this may actually make the problem worse on the links that need it most. It is good because it costs zero extra airtime, but it is risky on weak/problematic/busy links because the airtime needed to hear the echo dramatically increases the chance of collisions and errors._

The approach of PR2670 is to listen as long as possible and keep deafness low (prevent TX, prevent calibration that blocks RX) on the sending node. If a node is stressed with TX it may not hear a downstream package repetition so it will resend its package which is unnecessary. So yes, in busy environments the error correction remains but the package frequency becomes higher. I tried to reduce the probability of collisions by checking for noise before repeating a package but yes, the more traffic the worse the quality gets. But still better for deliverability than with common MeshCore without the duplication.

_> In contrast, a hard ack is about 12 bytes regardless of the original message size. So for 12 extra bytes of airtime, you get the most reliable ack's possible._

That means to at least double the amount of TX for a direct message plus possible repetitions if the ACK was not received by the sender. Plus special collision problems when the ACK packages cannot be heard and the sender repeats its package.
In contrast to that I think "less is more": within PR2670 the regular package sent by a downstream repeater is considered as the ACK you're focusing on; in my view a duplicate confirmation does not change the situation but increases TX. Do I miss something?

Side note: package resending like with PR2670 is downwards compatible; so it works without updating all other nodes.
According to your explanation IMHO a repeater with your draft reacts on "no ACK received?" with a re-sending: this would multiply all direct packages as long as the surrounding repeaters are not updated, isn't it?

benallfree:
Hi @usrflo :)

> The approach of PR2670 is to listen as long as possible and keep deafness low (prevent TX, prevent calibration that blocks RX) on the sending node.

I like this! Very smart 🔥 

> That means to at least double the amount of TX for a direct message plus possible repetitions if the ACK was not received by the sender. Plus special collision problems when the ACK packages cannot be heard and the sender repeats its package.

Maybe we can both think about this, I'd like to hear your opinion. 

You have probably noticed like I have that smaller packets are delivered far more reliably than larger packets. Hard ACK takes advantage of this idea and is 14 bytes:

```
10:57:30 - 15/5/2024 U: TX, len=14 (type=11, route=D, payload_len=12)
```

If the sender can't hear a 14 byte ACK, it seems even less likely that it would hear a >14 byte echo of the original message. A separate 14 byte ACK stays small even with large 200-byte messages. For 14 extra bytes of airtime, we buy the most reliable ACK possible.

At first, it does seem like increased TX compounds the problem. However, I think spending some airtime on the most reliable ACK possible will ultimately lead to the fewest resends between hops and also the fewest full path resends from the user (the worst problem), which far outweigh 15 byte ACKs.

I imagine we both agree that best ACK = fewest resends. So the question is passive vs active ACK. Maybe it could be a setting, like `hop.ack.mode: passive|active`.

> this would multiply all direct packages as long as the surrounding repeaters are not updated, isn't it?

Sorry I didn't mention it, but this PR will not attempt resends until the first ACK is heard from the receiver. Graceful upgrade. I added the note.

Thanks for the interesting conversation. Any way we improve or solve this is a win for mesh 🙏

benallfree:
@usrflo I have thought about your points more and made an update.

Now we start with passive echo listening. If A does not hear an echo, it will resend. If B detects a resend, it knows A did not hear the original echo and will send a hard ACK back to hopefully prevent A from spamming more retries when B has already relayed the message.

So now hard ack is reserved for only situations where passive echo fails.

Do you like it?

usrflo:
Hi @benallfree, I advised AI to compare both of our PRs via mcsim. It accumulated lots of knowledge in the last weeks about MeshCore. After checking it I rate this information as a profound analysis.

# Simulation results: possible ACK-return regression with per-hop retry (PR #2980)

Hi @benallfree — thanks for the PR, the per-hop echo retry idea is a nice fit for
multi-hop paths. We ran `feature/next-hop-retry` (PR defaults: `hop.retry=2`,
`hop.retry.ms=1500`) through our discrete-event simulator and want to share one
observation that looks like it may be a regression, in case it's useful.

## Test setup

- **Simulator:** [mcsim](https://github.com/usrflo/mcsim) — it compiles the **actual firmware C++** (same
  `Mesh.cpp` / `Dispatcher.cpp` code path as on-device) into a simulated radio
  environment with a configurable PHY model. So the mesh/retry logic under test
  is the real firmware logic, not a re-implementation.
- **Topology ([retry_showcase](https://github.com/usrflo/mcsim/blob/main/examples/topologies/retry_showcase.yaml)):** Alice → R1 → R2 → R3 → Bob with **three
  marginal forward hops** (−3 dB SNR ≈ 13% per-packet loss each) and a
  **reliable return path** (Bob → R3 → R2 → R1 → Alice, all ≈100%). SF7,
  BW 62.5 kHz.
- **Behavior ([retry_stress](https://github.com/usrflo/mcsim/blob/main/examples/behaviors/retry_stress.yaml)):** Alice sends **20 direct messages** to Bob,
  10 s apart (90 s startup).
- **Config:** PR defaults (`hop.retry=2`, `hop.retry.ms=1500`) together with the
  topology's `max_resend_attempts=3`.
- **10 random seeds**, 450 s each.

## Results (mean over 10 seeds)

| Metric (mean of 10 seeds) | this PR (`hop.retry=2`) | [reference PR 2670](https://github.com/meshcore-dev/MeshCore/pull/2670)¹ (sender-resend only) |
|---|---:|---:|
| DMs sent (of 20 intended) | 19.0 | 20.0 |
| DMs delivered | 17.4 (**87 %** of 20) | 19.6 (**98 %** of 20) |
| **ACKs received back by sender** | **~1.0 (!)** | **~19.6** |
| `rx_collided` | 5.3 | 3.3 |
| TX airtime (total) | 22.3 s | 23.7 s |
| TX airtime per delivered DM | 1.28 s | 1.21 s |

¹ Same firmware **without** per-hop echo retry (sender-side `max_resend_attempts` only), same topology/behavior/seeds.

## Observations

**1. ACK return appears to break almost completely.** The return path is
reliable by construction, so the reference build gets ~98% of ACKs back to
Alice. With this PR, only ~1 ACK per run makes it back. Because mcsim runs the
real firmware logic, this looks like it may originate in the per-hop
echo / `HOP_ACK` path (e.g. dedup marking, half-duplex deafness during the echo
listen window, or the extra forward transmissions colliding with the returning
ACK) rather than in the PHY model.

**2. More collisions, and one DM goes unsent.** Collisions rise ~60%
(5.3 vs 3.3), and in most seeds only 19/20 DMs are sent — consistent with the
per-hop retry activity keeping the channel busy and deferring Alice's later
sends past the end of the run.

## Caveats

- This is **simulation, not hardware** — worth confirming on real devices. (We
  have seen cases before where sim and HW diverge on LBT/timing behaviour.)
- `hop.retry` ran **on top of** the topology's `max_resend_attempts=3`. We have
  not yet isolated whether the ACK loss persists with `max_resend_attempts=0`;
  if it does, that would point squarely at the per-hop path rather than at the
  two retry layers fighting.
- We may be missing intended context on how `HOP_ACK` is meant to interact with
  the existing direct ACK.

benallfree:
@usrflo That's very cool!!! 

I am going to try to repro this, thank you so much!

HOP_ACK would only be expected if a downstream repeater had already relayed the message but the upstream repeater resent it. In that sole case, the downstream would send a zero-hop HOP_ACK so the upstream repeater would know to stop spamming retries. A rare corner case, but possible.

Do you happen to know if [mcsim](https://github.com/usrflo/mcsim) can simulate asymmetric links where repeater A can relay to repeater B, but not hear repeater B's echo?

usrflo:
Hi @benallfree, regarding HOP_ACK versus "ACKs going back to Alice": I think mcsim focused on the latter but I am not sure how to differ. The interesting thing would be to find out if ACKs really collided and therefore could not be received by Alice.

Regarding the question about asymmetric links: looking at examples/topologies/retry_showcase.yaml you can see definitions in both directions below "edges:". So yes you can define asymmetry.

PR #2994, opened 2026-07-20
Add setter methods for preambleMillis and maxPayloadMillis on SX1262 and LR1110

oltaco:
Fixes broken builds due to header inclusion ordering, shifts responsibility back to the wrapper layer.


liamcottle:
Confirming it fixed the build issues: https://github.com/liamcottle/MeshCore/actions/runs/29733172545

PR #3003, opened 2026-07-20
Initialize external RTC after power loss

### LINKED ISSUE #2996: RTC jumps to year 2042 on every power cycle and cannot be corrected
### Hardware:
- Board: Seeed Studio XIAO ESP32S3 + Wio-SX1262 (Meshcore kit)
- Power: USB-C (no external RTC / backup battery)

### Firmware:
- Firmware: Meshcore (Web Flasher build) Version: 1.16.0 (latest available from the Web Flasher at the time of writing) Bluetooth Companion

### Environment:
- OS used for CLI: Debian 12
- Meshcore CLI version: v1.5.7
- Web Flasher: latest as of July 2026 Version: 1.16.0
- Node role: Bluetooth Companion

### Description:
- On my XIAO ESP32S3 + Wio-SX1262 node, the device clock consistently jumps into the future after a hard power cycle.
- After each power cycle (USB power removed and reconnected), the device reports a “Current time” of: 2042-04-27 10:20:18 (Unix epoch 2282199618)
- Once this incorrect time has been set, it cannot be brought back to the real current time using normal means (Android App, CLI). The node keeps thinking it is in the year 2042.
- This looks like a persistent RTC / time-handling bug specific to the ESP32-S3/XIAO hardware combination, and it is reproducible on my device.

### Steps to reproduce:
- Flash firmware 1.16 onto a XIAO ESP32S3 + Wio-SX1262 using the Web Flasher.
- Set the correct time (e.g. via CLI: meshcore clock sync)
- Verify that the time is correct.
- Perform a hard power cycle:
- Disconnect USB power completely.
- Wait a few seconds.
- Reconnect USB power and let the node boot.
- Check the reported device time (CLI or App UI).

### Observed behavior:
- After the hard power cycle, the device clock is no longer at the previously set time.
- Instead, it consistently jumps to: Current time : 2042-04-27 10:20:18 (2282199618)
- Any attempt to set the time back:
   - clock sync: No time sync needed
   - clkreboot: ERROR:meshcore:Unknown command : clkreboot, ['clkreboot'] not executed ...
- The node continues to use this future time for logs / packets, which breaks “last heard” timestamps and time-related features in the App.

### Expected behavior:
- After a power cycle, the device should either:
   - retain the previously set time (if hardware/RTC supports it), or
   - start at the Unix epoch (1970-01-01) or some neutral baseline, and then accept new time values from BLE, CLI or GPS.
   - Time sources with a valid current timestamp (phone, GPS, CLI) should always be able to override a clearly invalid future time.

### Additional observations / hypotheses:
- The timestamp 2282199618 consistently corresponds to 2042-04-27 10:20:18, suggesting a deterministic misinterpretation of RTC memory, not random garbage.
- This may be related to ESP32-S3 RTC / RTC_SLOW_CLK behavior after losing power on the RTC domain (XIAO ESP32S3 does not have a backup battery for RTC), combined with how Meshcore interprets internal time on boot.

### Questions / requested changes:
- Can the firmware detect obviously invalid times (e.g. far in the future compared to build date or phone time) and reset to a safe baseline?
- Always accept a new time value from App / CLI / GPS if the current device time differs by more than a configurable threshold?
- Is there any known ESP32-S3/XIAO-specific workaround (e.g. different RTC init path) that should be applied at boot for this hardware?
- If this is a hardware limitation (no backed-up RTC), could the firmware treat “uninitialized” RTC time more defensively on this platform?

Thank you very much!

axhoff:
## Summary

- detect lost-power state for discovered PCF8563 and DS3231 RTCs
- initialize an invalid external RTC from the already-initialized fallback clock
- preserve valid battery-backed RTC time
- add regression tests for both seed and preserve behavior

## Root cause

`AutoDiscoverRTCClock` selected a PCF8563 after successful I2C initialization without checking its voltage-low/lost-power flag. On the XIAO expansion board without an RTC backup battery, a hard power loss could therefore leave untrustworthy clock contents (for example, a date in 2042). The companion's existing backward-time guard would then correctly reject the lower current time, leaving the node stuck on the invalid future timestamp.

## Scope

- no protocol, packet, or storage changes
- existing backward-time guards remain unchanged
- valid external RTCs are unchanged
- DS3231 is included because it exposes the same `lostPower()` state and follows the same discovery path
- RV3028 and RX8130 behavior is unchanged

## Validation

- `pio test -e native` — 15/15 tests passed
- `pio run -e Xiao_S3_WIO_companion_radio_ble` — passed
- `git diff --check` — passed

Fixes #2996

### DIFF
--- a/src/helpers/AutoDiscoverRTCClock.cpp
+++ b/src/helpers/AutoDiscoverRTCClock.cpp
@@ -1,4 +1,5 @@
 #include "AutoDiscoverRTCClock.h"
+#include "RTCValidity.h"
 #include "RTClib.h"
 #include <Melopero_RV3028.h>
 #include "RTC_RX8130CE.h"
@@ -30,6 +31,10 @@ void AutoDiscoverRTCClock::begin(TwoWire& wire) {
   #if !defined(DISABLE_DS3231_PROBE)
   if (i2c_probe(wire, DS3231_ADDRESS)) {
     ds3231_success = rtc_3231.begin(&wire);
+    if (ds3231_success && mesh::seedRTCIfLostPower(rtc_3231.lostPower(), _fallback->getCurrentTime(),
+                                                   [](uint32_t time) { rtc_3231.adjust(DateTime(time)); })) {
+      MESH_DEBUG_PRINTLN("DS3231: Lost power, initialized from fallback clock");
+    }
   }
   #endif
 
@@ -44,6 +49,10 @@ void AutoDiscoverRTCClock::begin(TwoWire& wire) {
   if (i2c_probe(wire, PCF8563_ADDRESS)) {
     MESH_DEBUG_PRINTLN("PCF8563: Found");
     rtc_8563_success = rtc_8563.begin(&wire);
+    if (rtc_8563_success && mesh::seedRTCIfLostPower(rtc_8563.lostPower(), _fallback->getCurrentTime(),
+                                                     [](uint32_t time) { rtc_8563.adjust(DateTime(time)); })) {
+      MESH_DEBUG_PRINTLN("PCF8563: Lost power, initialized from fallback clock");
+    }
   }
 
   if (i2c_probe(wire, RX8130CE_ADDRESS)) {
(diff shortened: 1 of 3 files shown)


IoTThinks:
So this PR will help RTC module to keep the time for the board after power loss?
Thanks a lot.

PR #3007, opened 2026-07-20
Synchronize the ESP32 BLE receive queue

### LINKED ISSUE #2923: SerialBLEInterface receive queue is shared across tasks without synchronization
`SerialBLEInterface::onWrite` appends to `recv_queue` from the BLE callback task while `checkRecvFrame` reads, decrements, and shifts the same array from the Arduino loop task. `recv_queue_len` and records have no critical section or SPSC publication discipline.

This can lose or corrupt closely spaced companion frames. A focused fix should protect receive queue publication/drain with an ESP32 critical section (including clear/reset), while keeping send queue behavior unchanged, and add a bounded queue model/stress test where feasible.

Discovered while auditing hermes-gadget/SigurdOS-tdeck (tracking issue hermes-gadget/SigurdOS-tdeck#813).

axhoff:
## Summary

- replace the ESP32 BLE receive array with a statically allocated FreeRTOS queue
- use non-blocking queue send/receive operations and reset the queue when buffers are cleared
- keep send-queue behavior unchanged

## Root cause

The BLE callback task appended to `recv_queue` while the Arduino loop task read, decremented, shifted, and cleared the same array without synchronization. Closely spaced frames could therefore observe partially published state or race with draining and clearing.

## Scope

Only the receive queue is changed. `xQueueSend(..., 0)` retains the existing drop-newest behavior when all four slots are occupied, `xQueueReceive(..., 0)` keeps reads non-blocking, and `xQueueReset` handles clear/reset. Static allocation avoids introducing heap allocation or queue-lifetime management.

## Validation

- `git diff --check`
- `pio test -e native` — 13/13 tests passed
- `pio run -e Xiao_S3_WIO_companion_radio_ble`

Fixes #2923


ripplebiz:
Wow, I did a bit of research and it's news to me that the Arduino BLE classes _don't_ wrap the operations/callbacks into thread-safe accessors. Yoiks!  So, in that case, we should just use an RTOS Queue, eg
  dataQueue = xQueueCreate(FRAME_QUEUE_SIZE, sizeof(Frame));
  ...
  onWrite() {  ...  xQueueSend(dataQueue, ...)   }
  ...
  in checkRecvFrame() {  ... if (xQueueReceive(dataQueue,  ... }

axhoff:
Thanks @ripplebiz — good call on using the RTOS primitive. I replaced the custom queue and critical section with a statically allocated FreeRTOS queue using `xQueueSend`, `xQueueReceive`, and `xQueueReset` in 4d4d7c37. The BLE target and native tests both pass.

PR #3012, opened 2026-07-21
fix: recover from corrupt littlefs

benallfree:
Hey guys, this is a fix for a rare corner case where littlefs becomes corrupted somehow.

Currently, the device will get locked in a loop the first time it attempts to *write* to a corrupted littlefs. 

This PR does an integrity check at boot. I'm not sure if this is the best approach but here to discuss.

### DIFF
--- a/examples/companion_radio/DataStore.cpp
+++ b/examples/companion_radio/DataStore.cpp
@@ -44,6 +44,48 @@ static File openWrite(FILESYSTEM* fs, const char* filename) {
 
 #if defined(NRF52_PLATFORM) || defined(STM32_PLATFORM)
   static uint32_t _ContactsChannelsTotalBlocks = 0;
+
+/* Boot-time filesystem integrity check.
+ *
+ * A littlefs whose superblock is intact but whose interior metadata is damaged (e.g. flash
+ * region partially overwritten, or interrupted writes) mounts successfully, then wedges the
+ * device on the FIRST write after boot: lfs_deorphan() -- run lazily by lfs_remove()/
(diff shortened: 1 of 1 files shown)


IoTThinks:
So this PR will format FS in case of FS corruption?

My T1000e corrupted FS a few times in last 3 months.

benallfree:
@IoTThinks that’s correct. As far as I know, there is no way to recover from a corrupted file system, so you might as well just format it and start again. But if I’m wrong about that, and there is someway to run a file system repair, we should consider that first.

IoTThinks:
It is still better to recover and start over instead of stuck in bootloop.
I guess there is a major flaw in LittleFS. It seems very fragile.

Thanks a lot for the PR.

benallfree:
@IoTThinks does this one seem ready to move out of draft?

I think LittleFS v2 is better but upstream seems slow to adopt it.

IoTThinks:
@benallfree I haven't tested it.
I dont have 350+ contacts to simulate the corruption and test.

I guess slowness to go live as we need to export backup, fresh install and import the backup.

benallfree:
@IoTThinks I'm not actually sure how LittleFS becomes corrupt. Do you know how to repro?

Or maybe we could add a `fs corrupt` CLI for testing purposes.

IoTThinks:
In another PR, Otalco (sorry for spelling) explained the case.

Have big contact list like 350 contacts.
During sync after connect or edit contact, quickly remove the power.

benallfree:
Yes I see now, https://github.com/meshcore-dev/MeshCore/pull/2964

PR #3014, opened 2026-07-21
fix: Restrict bridge.source args to rx and tx

hickey:
This change updates the CLI so that the `set bridge.source` will only accept `rx` and `tx`. Otherwise an error is produced that specifies that these are the only two choices. Previous experience would accept `rx` and anything else was silently ignored and would not switch back to `tx`.


hickey:
This PR is superseded by the fix that is incorporated into #3020.

PR #3025, opened 2026-07-22
Allow .envrc to integrate into existing direnv environments

### LINKED ISSUE #3256: Existing .envrc is broken for most users
The existing `.envrc` makes the assumption that Nix is being used and breaks a user's `direnv` setup. To fix this problem and make a working environment for everyone the current `.envrc` needs to check if Nix is installed before attempting to execute any Nix specific code. If not found, it should exit gracefully without errors. 

In addition, the `.envrc` needs to source any parent '.envrc' so that the rest of the user's environment is not broken. It is customary to setup tool env variables at the parent directory of all the development projects instead of specifying all the same env vars in every project. The current `.envrc` does not allow this and this can be fixed by adding the `source_up` directive.

hickey:
The existing .envrc sets up the Nix environment which is nice, but not all of us use Nix.

This change will make setting up the Nix environment conditional upon finding nix-shell. This way if the direnv is being executed outside of the Nix environment direnv will not fail.

In addition, the source_up stdlib call is used to process all parent .envrc files.

Closed #3256

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,2 @@
-use nix
+source_up 
+nix-shell --version >/dev/null 2>&1 && use nix


hickey:
Given the comments in #2699, I wonder if it would be beneficial to add `source_env_if_exists .envrc-local` to the `.envrc` to allow a developer to add any personal customization without having it get pushed back into the repo.

PR #3026, opened 2026-07-22
fix: Restrict bridge.source args to logRx and logTx 🧁 🧁 

hickey:
This change updates the CLI so that the `set bridge.source` will only accept `logRx` and `logTx`. Otherwise an error is produced that specifies that these are the only two choices. 

Previous experience would accept `rx` and anything else was silently ignored and would not switch back to `logTx`. This was contradictory to what is documented in `cli_commands.md` which specified `logRx`/`logTx` and not `rx`.


hickey:
ARG.... hate that I created the PR merging to the wrong branch!

Issue #3028, opened 2026-07-22
[Feature Request] Bridge over USB and IP


hickey:
There are several scenarios where it would be beneficial to be able to bridge from the LORA radio over USB serial or IP connections. Pretty much all of these scenarios open the door to allow more than just MeshCore nodes to interact on the MeshCore network. 

The proposed addition of adding 2 new bridge types would transport the raw packets over the chosen connection using the following frames:

- USB: packets would use the RS232 framing so that they could be detected and extracted from any console messages if using the same USB stream. 
- IP: packets would be transmitted as raw octets given there would be no other data transmitted over the connection.

## Use Cases

1. Bridging MeshCore traffic from one band to another band (i.e. 915 MHz to 433 MHz)
3. Transporting MeshCore traffic around barriers (e.g. mountains, forests, deserts)
4. Connecting virtualized services

In use case 1, this might take the form of a point to point link to extend the network footprint or allow users running equipment on other frequencies to communicate across both bands. 

Use case 2 is where this idea started from. Where I live here in Florida the county is divided right down the middle by about a 15 mile wide state forest. Trying to communicate across the forest has been problematic at times although it is better now with the 1W transmitters. Even with the higher power repeaters, we are still dropping packets at times. Being able to transport the MeshCore traffic over the internet or a private network connection would allow reliable communication around physical barriers until better infrastructure is in place. 

Use case 3 is a solution where there does not need to be any middleware between the MeshCore node and a software service running. My need for this use case is to have virtualized Companion nodes running at home. Right now I have one node running on WIFI so that I can connect to the MeshCore network from anywhere in my house or 14 acre property. 

Very early on I found it was going to be a bit unreasonable to always be grabbing my Companion to carry with me as I worked around the house or property and then to find out that the Companion node was not charged just added injury to insult. Now add a significant other that is not as technical and will almost never carry a Companion node with them. 

So my solution for this use case is to build a Companion node that will run in a Docker container that will provide a network connection to the Companion app. I can then spin up a virtual Companion for each member of the family without having to maintain hardware nodes. 

I have a third bridge type that I have build to make use case 3 more feasible which is bridging over MQTT. The MQTT bridge will be submitted under a separate PR. Using MQTT actually simplifies the use case where the MQTT topic is a virtual extension to the MeshCore network. This would allow all the virtual Companions to subscribe to the same topic and all the Companions would go through the same repeater to get on and off the actual MeshCore network. 

[...]

PR #3029, opened 2026-07-22
[Feature] Allow repeater firmware to bridge over USB and IP

hickey:
This PR adds support for bridging packets over USB serial and TCP connections in the same manner as the RS232 bridging works. It uses the same encoding as the RS232 bridge code.

USB serial bridge code is enabled with defining `WITH_USB_SERIAL_BRIDGE` and setting the serial stream by defining `USB_SERIAL_STREAM` (usually to `Serial`). Defining the stream in theory should allow the bridging output to be directed to a second USB port if the platform has another USB connection; although I do not have a solid way to test this configuration with the hardware available to me. 

TCP bridge code is enable by defining `WITH_TCP_BRIDGE`. The port that the repeater will listen on is defined by `TCP_BRIDGE_PORT`and will default to `4403` if not specified. As a result when using the TCP bridge code the WIFI credential definitions need to be specified also in the build. 

I have added build environments for the `lilygo_tbeam_SX1262` named `Tbeam_SX1262_repeater_bridge_usbserial` and `Tbeam_SX1262_repeater_bridge_wifi` to demonstrate how the firmwares can be built. These environments should be able to be transported to other platforms that have compatible hardware. 

In addition, I have added the `usbserial` and `tcp` bridge types to the `get bridge.type` CLI command. I am not certain if this will have any effects on the remote management in the companion app. There is nothing I see that is obvious. I have also updated the CLI documentation to note the values that can be returned by `get bridge.type`.

Closes: #3028

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -198,9 +198,13 @@ uint8_t MyMesh::handleAnonClockReq(const mesh::Identity& sender, uint32_t sender
     memcpy(&reply_data[4], &now, 4);     // include our clock (for easy clock sync, and packet hash uniqueness)
     reply_data[8] = 0;  // features
 #ifdef WITH_RS232_BRIDGE
-    reply_data[8] |= 0x01;  // is bridge, type UART
+    reply_data[8] |= RS232_BRIDGE;  // is bridge, type UART
 #elif WITH_ESPNOW_BRIDGE
-    reply_data[8] |= 0x03;  // is bridge, type ESP-NOW
+    reply_data[8] |= ESPNOW_BRIDGE;  // is bridge, type ESP-NOW
+#elif WITH_USB_SERIAL_BRIDGE
+    reply_data[8] |= USB_SERIAL_BRIDGE;  // is bridge, type USB serial
+#elif WITH_TCP_BRIDGE
+    reply_data[8] |= TCP_BRIDGE;  // is bridge, type TCP
 #endif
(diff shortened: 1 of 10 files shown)


ripplebiz:
Similar here, as with MQTT. We aren't going to officially support internet bridges.

hickey:
But you do already indirectly and the precedence has been set. By officially supporting bridging over RS232 you have enabled opening the MeshCore network to other networks. This work just builds on that foundation.

hickey:
And further thinking about your response and the immediate shutdown of any discussion just clearly shows the true intention of keeping MeshCore a closed community. There are lots of references out there that promote MeshCore as an open community but you are definitely showing the indications that this is just marketing spin and that you don't want to look at new use cases. 

Too bad for the community.

recrof:
> open community

Open community doesn't mean we are open to any changes without limits. Allowing easy IP bridging means that the community will need to deal with irresponsible users, who will cluelessly merge meshes data over internet. There is no really stopping it happening, however we'll die trying. There is precedence in this on other mesh network, their MQTT bridges caused lot of headaches for mesh communities and actual devs. We won't repeat that.

hickey:
@recrof Your definitions are completely wrong. Open community has nothing to do with code changes. Open community has everything to do with discussions and listening to ideas from diverse perspectives. 

That simply was not the case here and it was a complete shutdown. 

BTW, this PR has nothing to do with the MQTT PR (#3030) that I produced.

recrof:
> BTW, this PR has nothing to do with the MQTT PR (#3030) that I produced.

the idea is the same. our stance is clear: no internet downlinks.

PR #3030, opened 2026-07-22
[Feature] Add MQTT bridge 

hickey:
As discussed in #3028, this is the MQTT bridge code that I referenced there. 

Just as in #3029, this creates a new bridge module that allows the MeshCore traffic to be published to an MQTT topic to allow virtualized services to communicate on the MeshCore network. 

There is a `lilygo_tbeam_SX1262` environment included to show how the node can be configured for MQTT bridging. That environment is named  `lilygo_tbeam_SX1262_repeater_bridge_mqtt`. 

**Note:** The last commit on this PR incorporates #3027 and extends that PR by adding the `logBoth` parameter. This allows the bridge to publish both the received packets and the transmitted packets from the repeater. This is necessary for situations where the a virtualized Companion is remote managing the physical repeater. With out this option, the virtual Companion will never receive responses to the remote management requests.


ripplebiz:
The community have made it pretty clear we don't want MQTT bridges. Won't be something we admit into official firmware.

PR #3051, opened 2026-07-28
Fixes for GPS support on RAK boards

YoshiWalsh:
This PR adds support for GPS on the RAK3401, improves GPS acquisition speed for RAK12500 and RAK12501 GPS modules, and adds support for syncing the clock from the RAK12500.

Behind the scenes, it fixes some additional logic errors and slightly improves code cleanliness. 

## Code changes

 - Replace opaque `RAK_BOARD` and `RAK_WISBLOCK_GPS` constants with `ENV_INCLUDE_RAK12500` (consistent with other sensor constants like `ENV_INCLUDE_RAK12035`)
 - Fixes UART GPS support on RAK3401 due to transposed RX/TX pins \*
 - Ensures RAK12500 support is compiled into RAK3401 builds \*
 - Replace faulty reset pin logic with PIN_3V3_EN logic (see below)
 - Ensure 3V3 rail remains powered if `ETHERNET_ENABLED` or `SKY66122` constants are set \*\*
 - Add support for clock syncing from RAK12500
 - Optimise RAK12500 communication & acquisition times
 - Set constellation for L76K to improve accuracy and acquisition times
 - Code cleanliness

\* This was fixed in #3268 after this PR was created, but this PR still has code cleanliness improvements
\*\* This was partially fixed in #3275 after this PR was created, but this PR also fixes ethernet module (if present) being erroneously disabled on RAK4631. 

## Details on faulty reset pin logic

The previous code only worked by accident.

1. Regardless of which slot the GPS module was present in, it would always be detected as Slot A
2. Slot A reset pin (WB_IO2) overlaps with PIN_3V3_EN.
3. Per u-blox ZOE-M8Q documentation: "In reset state, the SiP consumes a significant amount of current. It is therefore recommended to use RESET_N only as a reset signal and not as an enable/disable."
So using reset PIN to control GPS status is not desirable anyway.

Due to the way these logic errors interacted, the RAK12500 sensor was actually able to work correctly in Slot A and Slot C, but for the wrong reasons.

This PR explicitly just controls the 3V3 bus, which was what the previous code was doing by accident.

Note: 3V3 control is disabled on the RAK3401 as this board is most commonly used in the "1W LoRa Booster Kit", and turning 3V3 off would also disable 5V power supply to the SKY66122 PA.

Changes tested with RAK3401 with a RAK12500 in both Slot A and Slot C, as well as a RAK12501 in Slot A. (RAK12501 should work in Slot D too, I just haven't tested it)

## Thanks

I'd like to give a special thanks to @Mizman102 for sponsoring this PR (bought me a RAK4631 for testing)


IoTThinks:
RAK12501 (L76K) works RAK4631 on MeshCore 1.16.

IoTThinks:
You may want to rebase to dev instead. 
The PR will be merged to dev, then to main.

Have you tested the combination of RAK4631/RAK3401 and RAK12500/RAK12501?

YoshiWalsh:
@IoTThinks I have tested on RAK3401 with RAK12500 and RAK12501. For the RAK12500, I also tested with I2C disabled (to ensure UART works as well), and I tested in both Slot A and Slot C.

I do not have a RAK4631 to test with.

re: RAK12501 working with RAK4631 on MeshCore 1.16, I'm pretty confused as to how. Is that running the repeater-bridge RS232 firmware? These appear to be the only variants that define the ENV_INCLUDE_GPS feature flag.

I did just see that the RAK4631 config [transposes the GPS RX/TX pins](https://github.com/meshcore-dev/MeshCore/blob/03b6ef4b0de98fc70b49ef10a6d0d61f8381fb7a/variants/rak4631/platformio.ini#L15-L16) which cancels out the fact they are transposed in [EnvironmentSensorManager](https://github.com/meshcore-dev/MeshCore/blob/03b6ef4b0de98fc70b49ef10a6d0d61f8381fb7a/src/helpers/sensors/EnvironmentSensorManager.cpp#L779). I'm not quite sure why this is done, but it also seems to be the convention for other boards, so I'll adjust my code to follow this convention too.

IoTThinks:
GPS 12501 (L76K) works on my RAK4631.

Yah, let me push this PR for some of my friends to test on RAK3401 and Rak12500 (i2c).

Thanks for the PR.
We can share the information as we do not have all required boards.

YoshiWalsh:
- Rebased to dev branch
- Incorporated ethernet check from #1983
- Restored RAK4631 transposition of RX/TX pins for consistency with other boards + applied transposition to RAK3401

I have __not__ yet tested these latest changes on real hardware. I'm sick right now. Once I've recovered I'll test this on RAK3401 with RAK12500 and RAK12501. In the meantime, I'd appreciate if someone could test on RAK4631.

IoTThinks:
> if someone could test on RAK4631.

I will test RAK4631 and RAK12501 tomorrow.

variants/rak4631/platformio.ini
```diff
@@ -9,11 +9,10 @@ build_flags = ${nrf52_base.build_flags}
   -I variants/rak4631
   -D RAK_4631
   -D RAK_BOARD
+  -UENV_INCLUDE_GPS
```
IoTThinks: Why we need to exclude GPS here?
GPS codes of other boards is protected by this flag too.

IoTThinks: This kills all GPS code for RAK4631.
Image

IoTThinks: Well, we need to fix this PR to make GPS work for RAK4631.
Currently, GPS (RAK12500 works perfectly on my RAK4631).

Let me see how to fix.
Image

IoTThinks:
I have just done some more research.
WB_IO2 is 3.3V_S controlling power to all sensor slot A-D including GPS 12500, 12501 if any.
* So if we set off WB_IO2, all sensor slots and IO Slot will be off.
* RAK does not have GPS_EN to cut off GPS alone as many other boards.

Impact of WB_IO2 to LoRa:
* For RAK4601, LoRa is inside RAK4601. So set WB_IO2 will not impact LoRa. => GPS PowerSaving works. Yes, it works now.
* For RAK3401, LoRa 1W is mounted on IO Slot. So if we off WB_IO2 to off GPS, LoRa will be powered off too. => So GPS can not be off for RAK3401. => GPS PowerSaving will skip this board.

image

YoshiWalsh:
> I have just done some more research.
> WB_IO2 is 3.3V_S controlling power to all sensor slot A-D including GPS 12500, 12501 if any.
> * So if we set off WB_IO2, all sensor slots and IO Slot will be off.

Correct, and I explained this in the PR description above. It is safe to use PIN_3V3_EN to control GPS power state *only* if other modules do not require the 3V3 rail. For this reason, if an ethernet module is present, or if the SKY66122 is present, my code prevents turning off the GPS.

variants/rak4631/platformio.ini
YoshiWalsh: Oh huh. This should be -DENV_INCLUDE_GPS. I'm not sure how I managed to make this mistake, oops.

IoTThinks:
Btw, this PR doesnot work for RAK4631 yet.
Let me see if I can fix.

Btw, when you do "gps off" and "gps", do you see it still "on, deactivated..."?

IoTThinks:
Still have hope to have GPS PowerSaving for RAK3401 and I2C RAK12500. :)

image

YoshiWalsh:
> Still have hope to have GPS PowerSaving for RAK3401 and I2C RAK12500. :)


If you do not have either the ETHERNET_ENABLE or SKY66122 constants defined, power saving works the same as it always has: by turning off the 3V3 rail. So for the default configuration of the RAK4631, the GPS can still be turned off.

I don't think the RAK12500 exposes the correct pin on the ZOE-M8 to allow for proper warm-start power savings.

variants/rak4631/variant.h
```diff
@@ -182,7 +182,9 @@ extern "C"
 #define EXTERNAL_FLASH_DEVICES IS25LP080D
 #define EXTERNAL_FLASH_USE_QSPI
 
-#define PIN_GPS_1PPS      17  //GPS PPS pin
+#define PIN_GPS_TX PIN_SERIAL1_RX
```
IoTThinks, line 206: Please add 

```
// enables 3.3V periphery like GPS and the IO Module
#define PIN_3V3_EN (34)
```

Like this
```
// enables 3.3V periphery like GPS and the IO Module
#define PIN_3V3_EN (34)

#define PIN_GPS_TX PIN_SERIAL1_RX
#define PIN_GPS_RX PIN_SERIAL1_TX
```

IoTThinks (commented):
Please help to fix as in the comment.
I have tested it with RAK4631 + UART 12501 GPS.

Image

Then I will test again after your changes.
Well, a few lines take me quite some time.

Thanks a lot.

variants/rak4631/variant.h
IoTThinks: PIN_3V3_EN is currently missing in RAK4631.

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -784,14 +782,14 @@ void EnvironmentSensorManager::rakGPSInit(){
...
 
-  //search for the correct IO standby pin depending on socket used
-  if(gpsIsAwake(WB_IO2)){
-  }
-  else if(gpsIsAwake(WB_IO4)){
-  }
-  else if(gpsIsAwake(WB_IO5)){
-  }
-  else{
+  #ifdef PIN_3V3_EN
```
IoTThinks, line 829: Except for RAK3401 with 1W with required PIN_3V3_EN at start, most of the RAK boards should have PIN_3V3_EN as INPUT to save power.
This is same as start_gps() too.

And we only need to set it HIGH as this is Power switch, not a GPS RESET.

Please change this:
```
  #ifdef PIN_3V3_EN
  digitalWrite(PIN_3V3_EN,LOW);
  delay(1000);
  digitalWrite(PIN_3V3_EN,HIGH);
  delay(2000);
  #endif
```

to this
```
  #ifdef PIN_3V3_EN
  pinMode(PIN_3V3_EN, OUTPUT);
  digitalWrite(PIN_3V3_EN, HIGH);
  delay(500);
  #endif
```

enigmaspb: > Except for RAK3401 with 1W with required PIN_3V3_EN at start, most of the RAK boards should have PIN_3V3_EN as INPUT to save power.

And RAK Ethernet (RAK13800) ?

IoTThinks: > And RAK Ethernet (RAK13800) ?

I have no idea.
However, putting  pinMode(PIN_3V3_EN, OUTPUT); in front is better than without it.
We can call it as many times without impact.

If the boards have PIN_3V3_EN as INPUT such as RA4631 and may be other RAK-based boards,  digitalWrite(PIN_3V3_EN,HIGH); will not be effective.

We can see in start_gps(), we also call pinMode(PIN_3V3_EN, OUTPUT); before digitalWrite(PIN_3V3_EN,HIGH);

variants/rak4631/variant.h
YoshiWalsh: ah good point! I'll fix it

cwichura:
The 12501 is powered off via a `$` command.  Not by GPIO enable pin.  Look at the ZephCore code for it.  ZephCore also sends a couple `$` commands on first init to help the GPS start up faster.

Waking the GPS when it is powered down is done by sending a 0xFF to it.

YoshiWalsh:
Ok, I dove into the wonderful world of RAK pinouts again. The canonical reference for pinouts is [here](https://github.com/RAKWireless/WisBlock/blob/ae4137bbff8f003cfbf77c14eecd22a03cd37bb1/PlatformIO/RAK4630/WisCore_RAK4631_Board/variant.h#L44-L60).

Confusingly, this doesn't really agree with any of the pinouts listed on any of RAK's datasheets. For example, while the variant definition for the RAK4631 board shows IO2 as being pin 34, the datasheet for both the [RAK4631 module](https://docs.rakwireless.com/product-categories/wisblock/rak4631/datasheet/#rak4631-module-pinout) and the [RAK19007 baseboard](https://docs.rakwireless.com/product-categories/wisblock/rak19007/datasheet/?utm_source=rak_store&utm_medium=products&utm_campaign=rak19007-wisblock-base-board-2nd-gen&utm_content=link&utm_term=datasheet#connector-for-wisblock-core) shows IO2 as being pin 30. This is also sometimes referred to as SW2, and it then connects to pin 26 on the RAK4630, which appears to connect to pin 1.02 on the nRF52840, but I can't find how that matches up against the value of 34 in the WB_IO2 constant.

In theory, all boards that are based on the RAK WisBlock system should use the same pinouts (at least as long as they are using a RAK4631-compatible MCU module, and a RAK19007-compatible base). But when I went to make sure they all agreed, I noticed that some of the GAT562 products were using 33 as their `GPS_EN` pin.

Ok, back to digging. Did I miss something somehow? Well, just as 34 is WB_IO2, 33 is WB_SW1. Looking at [this schematic](https://github.com/gat-iot/GAT562-family/blob/main/GAT562%20Mesh%20Tracker%20Pro%20SCH.pdf) we can see where SW1 comes out of the GAT562 module. (We can also see that P1.02/SW2 connects to a line labelled IO2, which agrees with what we saw earlier and gives us a nice sanity-check)

The SW1 line then connects to an external header called "P1.1". It does not go anywhere near the GPS.

Looking at the GPS on the schematic (the familiar L76K), there aren't many things connected to it. The only one that would allow turning it on/off is the 3V3_GPS rail. And if we look at the IRLML5203TR MOSFET, we can see that the 3V3_GPS rail is controlled by... IO2!

In summary, I am confident that the previous specification of 33 for GPS_EN for the GAT562 family was a mistake, and I have now updated it to use WB_IO2 as I believe was intended. I have also removed further unused definitions for GPS PPS pins.

src/helpers/sensors/EnvironmentSensorManager.cpp
YoshiWalsh: Can you link documentation that states that keeping PIN_3V3_EN as INPUT saves power? My understanding is that setting the pin as HIGH shouldn't take power because it's pulled high by default.

re: removing the code to pulse PIN_3V3_EN low for 1s at startup, that code appears to be [recommended by RAK](https://github.com/RAKWireless/WisBlock/blob/ae4137bbff8f003cfbf77c14eecd22a03cd37bb1/examples/common/sensors/RAK12500_GPS_ZOE-M8Q/RAK12500_GPS_ZOE-M8Q_IIC/RAK12500_GPS_ZOE-M8Q_IIC.ino#L21) so I'm hesitant to remove it. I think it ensures that the GPS is in a consistent state when our code initialises.

YoshiWalsh:
> The 12501 is powered off via a `$` command. Not by GPIO enable pin. Look at the ZephCore code for it. ZephCore also sends a couple `$` commands on first init to help the GPS start up faster.

Oh wow, that's very interesting! Unfortunately it's unclear what license ZephCore is released under, and MeshCore's MIT license is permissive enough to allow basically any license to be used. So I don't think we can use code from ZephCore unless @liquidraver makes it clear what license ZephCore is released under (which would also require the consent of all ZephCore contributors).

That said, now that I know where to look in the ZOE-M8Q and L76K documentation, I might look to implement my own UART GPS optimisations. I believe that may be best be handled in a second pull request though.

liquidraver:
> > The 12501 is powered off via a `$` command. Not by GPIO enable pin. Look at the ZephCore code for it. ZephCore also sends a couple `$` commands on first init to help the GPS start up faster.
> 
> Oh wow, that's very interesting! Unfortunately it's unclear what license ZephCore is released under, and MeshCore's MIT license is permissive enough to allow basically any license to be used. So I don't think we can use code from ZephCore unless @liquidraver makes it clear what license ZephCore is released under (which would also require the consent of all ZephCore contributors).
> 
> That said, now that I know where to look in the ZOE-M8Q and L76K documentation, I might look to implement my own UART GPS optimisations. I believe that may be best be handled in a second pull request though.

I don’t know how much clearer I can state zephcore’s licence in the repo. It’s in license.md and in all the code headers.
You can use anything from it, if it helps anyone with anything, I’m happy.

YoshiWalsh:
> I don’t know how much clearer I can state zephcore’s licence in the repo. It’s in license.md and in all the code headers.

My apologies, I just looked in the repository root, I didn't think to look in the zephcore directory. Thanks for clarifying!

Looking into it a bit, I think the GPS support in ZephCore might have some errors. The L76K protocol used appears to be based on [documentation for the L76-LB](https://quectel.com/content/uploads/2024/04/Quectel_L76-LBL26-LBLC86L_GNSS_Protocol_Specification_V1.1.pdf), which uses a different protocol from the [L76K](https://files.waveshare.com/upload/d/dd/Quectel_L76K_GNSS_Protocol_Specification_V1.1.pdf). I could be wrong about this, but I can't find any documentation online suggesting that the L76K supports the $PMTK353 message, or other messages used by ZephCore.

liquidraver:
> > I don’t know how much clearer I can state zephcore’s licence in the repo. It’s in license.md and in all the code headers.
> 
> My apologies, I just looked in the repository root, I didn't think to look in the zephcore directory. Thanks for clarifying!
> 
> Looking into it a bit, I think the GPS support in ZephCore might have some errors. The L76K protocol used appears to be based on [documentation for the L76-LB](https://quectel.com/content/uploads/2024/04/Quectel_L76-LBL26-LBLC86L_GNSS_Protocol_Specification_V1.1.pdf), which uses a different protocol from the [L76K](https://files.waveshare.com/upload/d/dd/Quectel_L76K_GNSS_Protocol_Specification_V1.1.pdf). I could be wrong about this, but I can't find any documentation online suggesting that the L76K supports the $PMTK353 message, or other messages used by ZephCore.

You are right, I could not verify it actually does anything on L76K, but worst case it does nothing so I left it there to remember :)

YoshiWalsh:
More digging, more fun facts uncovered. The RAK12501 is built around the L76K, which does not have any UART codes for power saving / warm off. Instead, based on the [schematic diagram](https://docs.rakwireless.com/product-categories/wisblock/rak12501/datasheet/?utm_source=rak_store&utm_medium=organic&utm_campaign=rak12501_page&utm_term=gnss_gps_location_module_quectel_l76k_datasheet&utm_content=documentation_link#schematic-diagram) this appears to be achieved by cutting power to GPS_3V3/3V3_S while retaining power to 3V3. This is exactly what pulling PIN_3V3_EN low does.

So my code currently implements this correctly, pulling 3V3_EN low to turn the GPS into power saving mode if this doesn't interfere with other components that rely on 3V3_S.

Time to look into the 12500!

EDIT:

12500 can be put into Hardware Backup mode via the same procedure. So for this PR, I think I'm happy.

But the 12500 does also support a Software Backup mode via the UBX-RXM-PMREQ which can be activated without deactivating the 3V3_S bus. This would be pretty nice to have, I'll probably try to implement it in a future PR. The ZOE-M8Q NMEA protocol is specified [here](https://content.u-blox.com/sites/default/files/products/documents/u-blox8-M8_ReceiverDescrProtSpec_UBX-13003221.pdf).

IoTThinks:
I have tried the UART command to sleep L76K. Saved a few mA only.

May be you have more better Uart commands.

Nothing beats a full power cut to save 20mA.
In clear sky, a cold start takes around 1 minute only.

src/helpers/sensors/EnvironmentSensorManager.cpp
IoTThinks: 1. You can check the code for T096 and see initVariantShutdown.

RAK boards are no magic, but NRF52 boards. They follow the same rules.

2. You can check the code for RAK4631 and other RAK-based boards, then imaging what digitalWrite does without pinMode OUT?

Currently, RAK4631 doesnot have pinMode for 3v3 en.

Assume you can turn off LoRa for RAK3401, try to remove pin Mode OUT for 3v3 en.
Then try to bring up GPS via Uart.

IoTThinks: At least, we have to put pinMode OUT.
RAK4631 does not have it.

You can keep the delay or you can explore what are best values.

RAK is a hardware vendor. Their code is sample and on safest side only.

YoshiWalsh:
As far as I can tell, L76K doesn't have any UART sleep commands. It only has Power Saving state (activated via setting VCC low, V_BCKUP high), which uses 7μA. So 17.993mA savings compared to normal operation.

For ZOE-M8Q, typical operational usage is 9-25mA, Software Backup is 20μA, and Hardware Backup is 15μA. So the difference between software & hardware power saving is trivial.

IoTThinks:
> As far as I can tell, L76K doesn't have any UART sleep commands. It only has Power Saving state (activated via setting VCC low, V_BCKUP high), which uses 7μA. So 17.993mA savings compared to normal operation.

I think so too. So I tried the UART commands and save a little only.
So I just cut the power for L76K. 

> For ZOE-M8Q, typical operational usage is 9-25mA, Software Backup is 20μA, and Hardware Backup is 15μA. So the difference between software & hardware power saving is trivial.

After this PR, you can explore this opportunity.
Many RAK3401 friends love to have this feature.

YoshiWalsh:
Ok I couldn't resist fixing a few more things:

1. Add support for syncing time from RAK12500
2. Enable AssistNow Autonomous feature for RAK12500
3. Enable use of GLONASS constellation for L76K, including RAK12501
4. Remove unused / unnecessary u-blox GNSS library dependency from variants which don't need it
5. Some code refactoring

Tested working on RAK3401 with RAK12500 and RAK12501. @IoTThinks could you test RAK4631 again please?

src/helpers/sensors/EnvironmentSensorManager.cpp
```diff
@@ -160,38 +160,71 @@ static Adafruit_VL53L0X VL53L0X;
...
+class UbloxLocationProvider : public LocationProvider {
   long _lat = 0;
   long _lng = 0;
   long _alt = 0;
   int _sats = 0;
   long _epoch = 0;
+  bool _initialised = false;
   bool _fix = false;
 public:
+  void setRTC(mesh::RTCClock* clock) {
```
YoshiWalsh, line 181: This feels gross but I don't really know how it can be done cleaner. Because I can only provide rtc_clock to an instance of the EnvironmentSensorManager class, but UbloxLocationProvider is initialised statically outside of that class.

IoTThinks:
Let me test tomorrow.

src/helpers/sensors/MicroNMEALocationProvider.h
```diff
@@ -40,20 +40,17 @@
 class MicroNMEALocationProvider : public LocationProvider {
     char _nmeaBuffer[100];
     MicroNMEA nmea;
-    mesh::RTCClock* _clock;
```
IoTThinks, line 43: GPS L76K and GPS clock are working perfectly.
And I may expect to have a few lines to fix GPS for RAK3401 and RAK12500.

GPS module is fragile for big changes.

IoTThinks:
> Tested working on RAK3401 with RAK12500 and RAK12501. @IoTThinks could you test RAK4631 again please?

It is broken again on RAK4631 and RAK12501.
image

Latest code from this PR.
image

This is GPS PowerSaving (based on working GPS on MeshCore 1.16).
image

IoTThinks:
May be we don't need to fix for RAK12500 anymore :D 
image

YoshiWalsh:
I'll look into why this isn't working for the RAK4631 soon. Thought I had it this time but I must've missed something.

Tbh this PR is my first time working on embedded and my first contribution to a project in C++. I feel a bit like a junior dev again lol. Thanks for your patience.

And I think we should still fix it for RAK12500 since the board is still being sold + I think a lot of people already have it.

IoTThinks:
Gps toggle not found means during init the board cannot detect L76K.

Likely Io2 is still low.

YoshiWalsh:
Just tested and this is definitely working for me with RAK3401 + RAK12501. L76K is correctly detected.

The code path up until the `gpsIsAwake` should be identical between RAK3401 and RAK4631. In both cases PIN_3V3_EN is high for 2000ms before we look for data from the GPS.

Now *after* we detect the GPS, the `stop_gps()` function will set PIN_3V3_EN low on RAK4631 but *not* on RAK3401. But by that point the `is_detected` flag has already been set, so this shouldn't have any impact on the GPS commands.

Since I don't have a RAK4631 I'm not sure how I can test this further.

IoTThinks:
There are too much code changes in this PR than I expect.
And I don't like the idea to replace RAK_WISBLOCK_GPS and RAK_BOARD by an end-of-sale ENV_INCLUDE_RAK12500.

IoTThinks:
For completeness, I will test again this PR later on my RAK4631 and RAK12501.

YoshiWalsh:
> There are too much code changes in this PR than I expect. And I don't like the idea to replace RAK_WISBLOCK_GPS and RAK_BOARD by an end-of-sale ENV_INCLUDE_RAK12500.

The only featured gated by the `RAK_WISBLOCK_GPS` and `RAK_BOARD` constants is RAK12500 support though. And I don't think removing support for this is a good option because that would be a regression. Like, should all currently-deployed repeaters using that GPS module just never be updated?

`ENV_INCLUDE_RAK12500` would be consistent with other modules like `ENV_INCLUDE_INA260` or `ENV_INCLUDE_RAK12035`. I think it would be a cleanliness improvement to use a single variable with naming consistent with other similar variables, instead of two variables with unclear / arbitrary names.

If you'd like me to split this into 3 PRs I can. But I kinda see that as just needing to test it 3 times instead of once. And I do think all these changes are pretty closely related.

YoshiWalsh:
I have ordered a RAK4631 so I can test using that. A big thanks to @Mizman102 for funding this purchase.

YoshiWalsh:
I've noticed that #3268 and #3275 have been merged, which implement some (but not all) of the fixes from this PR.

It is currently my belief that my implementation is more robust and adheres more closely to the conventions of the codebase.

On this topic, I don't want to be overly negative but I think it's important to say:

It is a bit discouraging to me that #3268 was merged with seemingly no testing after how much work I have had to put in to try to get this PR merged, including needing to buy a RAK4631. Especially because that PR introduced a bug (later fixed in #3275) relating to an issue I had already documented here.

It feels like the efforts that I have taken to implement this change cleanly & correctly have got this PR stuck, while a "quick and dirty" PR got rubberstamped despite containing a showstopping bug. I also note that liamcottle is a member of the MeshCore project.

Given that my efforts on this PR are active & ongoing, this has left a bit of a sour taste in my mouth as a new contributor, that a maintainer would take my PR, do a worse job of it, and skip the review process that I have been subjected to. I have not experienced anything like this before in any other open source project.

Sincerely,
Yoshi Walsh

YoshiWalsh:
@IoTThinks Fixed on RAK4631. Would appreciate your testing, thanks :)

IoTThinks:
> @IoTThinks Fixed on RAK4631. Would appreciate your testing, thanks :)

Since you have RAK4631, you can test it right?

YoshiWalsh:
Yes, I have tested all combinations of RAK4631 and RAK3401, with RAK12500 and RAK12501. In this latest round of testing I *only* tested Slot A. I previously also tested RAK12500 in Slot C and RAK12501 in Slot D, haven't re-tested these.

My preference is for this PR to be reviewed and merged "whole", since this means merging everything I've tested together and doesn't require me testing every combination for multiple smaller PRs. I also think that broadly all the changes I've made fit within "Fixes for GPS support on RAK boards". Since I ended up refactoring some things, it's also kinda unclear how this would best be split into smaller PRs. But I can give it a go if necessary.

src/helpers/sensors/MicroNMEALocationProvider.h
YoshiWalsh: GPS clock has previously not been supported on RAK12500. RAK12500 uses UbloxLocationProvider instead of MicroNMEALocationProvider, both inherit from LocationProvider. Since clock functionality is now supported by both providers, it makes sense to extract this logic to the shared base class, no?

src/helpers/sensors/EnvironmentSensorManager.cpp
YoshiWalsh: This is now resolved by setting the pin to write mode in RAK4631Board.cpp.

YoshiWalsh:
I have updated the PR description to accurately reflect the changes included within this PR.

PR #3058, opened 2026-07-29
Attempt self-recovery on radio_init() failure

entr0p1:
If radio_init() fails on the first attempt, currently the board will halt immediately with no feedback. This commit changes the behaviour to make 3 attempts to call the init. On a 4th failure, the board reboots to attempt to clear the fault and try again.

Debug logs are also written if MESH_DEBUG happens to be active.

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -135,7 +135,16 @@ void setup() {
   }
 #endif
 
-  if (!radio_init()) { halt(); }
+  int radioinit_attempts = 0;
+  while (!radio_init()) {
+    ++radioinit_attempts;
+    MESH_DEBUG_PRINTLN("Radio init failed! (attempt %d)", radioinit_attempts);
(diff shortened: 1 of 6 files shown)


examples/companion_radio/main.cpp
```diff
@@ -135,7 +135,15 @@ void setup() {
   }
 #endif
 
-  if (!radio_init()) { halt(); }
+  int radioinit_attempts = 0;
+  while (!radio_init()) {
+    MESH_DEBUG_PRINTLN("Radio init failed! (attempt %d)", ++radioinit_attempts);
```
ripplebiz: if debugging is off, you'll get infinite loop (I think)

entr0p1: I think you're right.. Better play it safe, I'll move the counter increment before the debug print and force push the branch.

entr0p1:
@ripplebiz could you take another glance at this one and let me know if its good now?

PR #3061, opened 2026-07-29
do not bootstrap companion rtc

### LINKED ISSUE #3050: Persisted contact lastmod timestamps can permanently corrupt the RTC and prevent posting to Room Servers
_Note that the following report is AI generated but I have personally verified that a messed up Companion clock prevents posting in Room Servers and choosing a new private key, (so that the room server doesn't think the next login is a replay), fixing the exported/imported contact "last_modifed" stamp and the Companion clock re-enables it. I have updated this report since yesterday because it turns out, after some further analysis, that this doesn't "Spread across the mesh"_

**Environment**

- Hardware: LilyGo T-Echo (nRF52840) and also Seeed Studio Xiao S3 WIO
- Firmware: companion_radio (USB), v1.16.0-07a3ca9
- Build environment: PlatformIO
- Related server: simple_room_server

**Summary**
MeshCore companions bootstrap their real-time clock at startup from the lastmod field of stored contacts. lastmod is supposed to record “when did we last hear from this contact by our clock”, but it is persisted in /contacts3 and loaded across reboots. If a device’s clock ever jumps into the future, every contact touched while the clock is wrong is stamped with that future time. On the next boot, bootstrapRTCfromContacts() sees those future timestamps and sets the clock even further forward, making the corruption self-perpetuating.

Because the companion then uses that future clock for sender_timestamp in logins and posts, Room Servers record the future value as client->last_timestamp and reject all later, correctly-timed messages from that identity as replay attacks. Once the clock is corrupt it cannot be corrected through normal clock sync / time commands because the firmware refuses to move the clock backwards, leaving a reflash or filesystem erase as the only recovery.

The corruption can also move between devices if a contact store or filesystem backup containing the bad lastmod values is restored on another companion.

**Affected code**
- BaseChatMesh.cpp — bootstrapRTCfromContacts() uses contacts[i].lastmod to set the RTC.
- src/helpers/DataStore.cpp — loadContacts() / saveContacts() persist lastmod.
- BaseChatMesh.cpp — populateContactFromAdvert() and message handlers stamp lastmod = getRTCClock()->getCurrentTime().
- MyMesh.cpp — onPeerDataRecv() and anonymous login reject sender_timestamp <= client->last_timestamp.

**Why this is hard to recover from**
- clock sync and time commands only allow the clock to move forward:
```
cpp
if (secs > curr) { getRTCClock()->setCurrentTime(secs); }
```
- bootstrapRTCfromContacts() runs at every boot and actively restores the future time from the corrupted contact list.

**How it spreads (corrected)**
Normal mesh adverts do not spread this. last_advert_timestamp records the sender’s claimed time and is only used for advert replay detection; it is not used to set the RTC. The corruption spreads through:
1. Self-reinforcement on the same device: a future clock stamps future lastmods, which bootstrap the clock back to the future on every boot.
2. Filesystem / contact-store backups: /contacts3 stores lastmod verbatim. Restoring that file (or a full filesystem image) to another device causes the second device to bootstrap its RTC from the same bad timestamps.
3. Closed-source app backups: if a phone/web backup includes lastmod metadata and restores it, the same effect occurs.

It does not propagate through routine meshcore:// contact shares, because exportContact() only returns the raw advert packet.

**Impact**
- A single corrupt companion can no longer post to any Room Server it logs into, because its timestamps are always in the future relative to the server’s recorded last_timestamp.
- If the corrupt contact store is restored to multiple companions, multiple users are affected.
- The only current recovery is to reflash, erase the filesystem, or manually remove /contacts3.

**Suggested fixes**
1. Do not bootstrap the RTC from lastmod at all. lastmod is a local “last heard” timestamp, not a network time source. Devices without a battery-backed RTC should fall back to a known-safe epoch and rely on manual/GPS sync.
2. Sanitize on load. When loading contacts, reject or clamp lastmod values that are far in the future (e.g., more than a few days ahead of the current RTC or a compile-time anchor).
3. Allow safe backwards correction. The companion serial/BLE interface should allow setting the clock backwards when the current time is clearly wrong, so a user can recover without reflashing.
4. Separate lastmod from RTC bootstrap. If clock bootstrap from contacts is kept, use a dedicated, validated field rather than a general-purpose local timestamp that is also used for contact eviction and sorting.

**Minimal reproduction**
1. Set a companion’s clock to a future date.
2. Receive or import a few contacts while the clock is future.
3. Reboot the device.
4. Observe that bootstrapRTCfromContacts() restores the future clock.
5. Log into a Room Server and attempt to post; subsequent posts with correct time are rejected as replays.

kujeger:
Bootstrapping companion RTC from contact lastmod time meant that if/when the companion clock strayed into future time because of clock drift, contacts would get a future/corrupted timestamp -- which the companion would happily restore upon reboot. And since the firmware rejects setting time backwards, it could never be fixed even when rebooting, causing the companion clock to ever drift further ahead.

Resolves #3050

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -957,7 +957,6 @@ void MyMesh::begin(bool has_display) {
 
   resetContacts();
   _store->loadContacts(this);
-  bootstrapRTCfromContacts();
   addChannel("Public", PUBLIC_GROUP_PSK); // pre-configure Andy's public channel
   _store->loadChannels(this);
 
--- a/src/helpers/BaseChatMesh.cpp
+++ b/src/helpers/BaseChatMesh.cpp
@@ -55,18 +55,6 @@ void BaseChatMesh::sendAckTo(const ContactInfo& dest, const uint8_t* ack_hash, u
   }
 }
 
-void BaseChatMesh::bootstrapRTCfromContacts() {
-  uint32_t latest = 0;
-  for (int i = 0; i < num_contacts; i++) {
-    if (contacts[i].lastmod > latest) {
-      latest = contacts[i].lastmod;
-    }
-  }
-  if (latest != 0) {
-    getRTCClock()->setCurrentTime(latest + 1);
-  }
-}
-
 ContactInfo* BaseChatMesh::allocateContactSlot(bool transient_only) {
   if (num_contacts < MAX_CONTACTS) {
     return &contacts[num_contacts++];
@@ -166,7 +154,7 @@ void BaseChatMesh::onAdvertRecv(mesh::Packet* packet, const mesh::Identity& id, 
       MESH_DEBUG_PRINTLN("onAdvertRecv: unable to allocate contact slot for new contact");
       return;
     }
-    
+
     populateContactFromAdvert(*from, id, parser, timestamp);
     from->sync_since = 0;
     from->shared_secret_valid = false;
@@ -773,7 +761,7 @@ void BaseChatMesh::checkConnections() {
       memcpy(data, &now, 4);
       data[4] = REQ_TYPE_KEEP_ALIVE;
       memcpy(&data[5], &contact->sync_since, 4);
-    
+
       // calc expected ACK reply
       mesh::Utils::sha256((uint8_t *)&connections[i].expected_ack, 4, data, 9, self_id.pub_key, PUB_KEY_SIZE);
 
@@ -781,7 +769,7 @@ void BaseChatMesh::checkConnections() {
       if (pkt) {
         sendDirect(pkt, contact->out_path, contact->out_path_len);
       }
-    
+
       // schedule next KEEP_ALIVE
       connections[i].next_ping = futureMillis(connections[i].keep_alive_millis);
     }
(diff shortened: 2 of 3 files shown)


kujeger:
Tested and verified on my T114 BLE companion. With the patched firmware flashed, the android app (and meshcore-cli) was able to set the clock correctly.

liamcottle:
While it's fresh in my mind, a side effect of not bootstrapping the clock will result in the firmware deleting the wrong contacts when you have auto overwrite oldest enabled. Tagging @oltaco as there was a reason this was implemented.

kujeger:
Possibly a better way to resolve these issues is to allow the companion firmware's clock to be set backwards, but I'd guess that has some other implications as well (but maybe not as bad as for a repeater?)

oltaco:
Unfortunately this would break the overwrite oldest contact feature when nodes have been restarted and haven't had their clock set yet. It would cause incoming adverts to cycle in the oldest slot instead and so you only ever get the last contact added to the list in the oldest slot.

I think allowing the companion clock to be set backwards would be the more correct fix. Allowing clock to be set backwards probably solves more problems than it creates?

kujeger:
an alternative approach to overwriting oldest contact could be to give each contact an auto-incrementing serial whenever it gets added/updated, and simply remove the one with the lowest id when storage is full, but I'd wager that's a big rewrite :sweat_smile:

todd-herbert:
As a hotfix I've been allowing the time to move back and then updating any bad (future) lastmod times.
https://codeberg.org/todd-herbert/InkCore/commit/6ce7b5cf24368a975fccceab90718aba7b0f76eb

PR #3066, opened 2026-07-29
ThinkNode M6: button shutdown/wake for repeater and room server

### LINKED ISSUE #2313: Shutdown for ThinkNode M6
Since the battery in the ThinkNode M6 is not accessible, the device cannot be turned off. Therefore, the option to turn off the device (put it into sleep mode) should be added to conserve battery life when the device is not in use. 

Implementation as done in #1873:

Allow shutdown and wakeup by button:
- add long-press user-button shutdown (hold ~1.5s).
- add long-press user-button weakeup (hold ~1.5s).

codemonkeybr:
## Problem

Fixes #2313. The ThinkNode M6's battery isn't physically accessible, so there was no way to power the device off to conserve battery when idle.

## Fix

Hold the Function Button for ~5s to power off (matches the timing of Elecrow's stock firmware for this board). A single press wakes it back up — the chip is fully halted in `SYSTEMOFF` while off, so there's no way to measure a hold duration on wake, only on the way down. The power LED (red) blinks 3 times on both shutdown and wake/boot as visual confirmation the action registered.

`ThinkNodeM6Board::powerOff()` now arms the button pin as a GPIO wake source (pull-up + `SENSE_LOW`) before entering `SYSTEMOFF` via `NRF52Board::powerOff()`.

## Scope

Limited to `simple_repeater` and `simple_room_server`. `companion_radio` is intentionally left out — #2837 is already moving that build's button handling into `UITask` (running headless via `NullDisplayDriver`), and duplicating a second independent button-read loop there would conflict with that approach rather than complement it.

## Testing

Built and flashed on a physical ThinkNode M6 repeater:
- Hold ~5s → LED blinks 3x, device powers off, mesh goes unreachable
- Single press → LED blinks 3x, device reboots, mesh reachable again
- Repeated the cycle multiple times, consistent behavior each time
- `ThinkNode_M6_room_server` and `SenseCap_Solar_repeater` (shared guard, unaffected) build clean; room_server not flash-tested on hardware

### DIFF
--- a/examples/simple_repeater/main.cpp
+++ b/examples/simple_repeater/main.cpp
@@ -30,9 +30,13 @@ static char ethernet_command[160];
 // For power saving
 unsigned long POWERSAVING_FIRSTSLEEP_SECS = 120; // The first sleep (if enabled) from boot
 
-#if defined(PIN_USER_BTN) && defined(_SEEED_SENSECAP_SOLAR_H_)
+#if defined(PIN_USER_BTN) && (defined(_SEEED_SENSECAP_SOLAR_H_) || defined(THINKNODE_M6))
 static unsigned long userBtnDownAt = 0;
+#if defined(THINKNODE_M6)
+#define USER_BTN_HOLD_OFF_MILLIS 5000   // matches Elecrow's stock firmware UX
+#else
 #define USER_BTN_HOLD_OFF_MILLIS 1500
+#endif
 #endif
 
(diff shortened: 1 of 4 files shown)


codemonkeybr:
Btw i used the 5000ms from the electrow website.. i could swap to use the same as sensecap 1500ms if the team prefers. nbd for me

PR #3068, opened 2026-07-29
Power off ThinkNode M3: wake on button, cut peripheral rails

ebaschiera:
The long press already reached board.powerOff(), but nothing armed a wake source, so the board stayed dead until a reset or a USB reconnect. Arm the user button (pull-up + SENSE_LOW) before entering SYSTEMOFF, after waiting for the release: SENSE is level triggered and a still pressed button wakes the board up immediately. If the button is never released, reboot instead: SYSTEMOFF would wake up right away anyway, so make the reset explicit.

GPIO output latches survive SYSTEMOFF, so shut the switched rails down as well (GNSS, RGB led supply, EEPROM, battery divider, EN1/EN2, PWR_EN), in a shutdownPeripherals() override so they go down after the radio and the GNSS receiver. Give pins 36 and 34 a name while there.

Tested on a ThinkNode M3 companion_radio_ble: a long press powers the board off (the USB device disappears and stays away), a short press brings it back, and holding the button past the timeout reboots.

### DIFF
--- a/variants/thinknode_m3/ThinkNodeM3Board.cpp
+++ b/variants/thinknode_m3/ThinkNodeM3Board.cpp
@@ -13,6 +13,42 @@ void ThinkNodeM3Board::begin() {
   delay(10);   // give sx1262 some time to power up
 }
 
+void ThinkNodeM3Board::shutdownPeripherals() {
+  NRF52Board::shutdownPeripherals();   // display, LoRa radio and GNSS receiver
+
+  // GPIO output latches survive SYSTEMOFF, so every switched rail still enabled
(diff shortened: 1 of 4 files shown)


ebaschiera:
Scope note: this only covers the powerOff() path, which on the M3 is reached by the companion long press (and by the poweroff/shutdown CLI command on the repeater/room server builds). I deliberately left simple_repeater/simple_room_server alone: they don't poll the button on this board, and the lines that would need it are the same ones #3066 is changing for the ThinkNode M6 — happy to follow up once that one lands.

Issue #2122, opened 2026-03-21
[Feature Request] Allow room server admin to pin one message


jschrempp:
When the message queue is full older messages are erased. In some situations companions might post messages that get lost before all other room participants log in to receive them. If a room server admin wants to get a message out to all participants, they have to re-post the message periodically to keep it in the queue.

Idea: Allow room server admin to mark one message as "persistent". This message would never get deleted. And when a companion retrieves messages it would always receive the persistent message.

This mechanism gives the room server admin a way to have an important message delivered to all participants. 

The admin could log in and remove the persistent message, or just mark it not persistent and let normal queue management take care of deleting the message.

MGJ520:
That's the function of the bulletin board.

jschrempp:
I'm sorry, I don't understand your comment - I'm new to MC. I thought the room-server was considered a bulletin board. Are you saying that there is some other function that will allow the admin to pin a particular message? If you can post a link to that documentation I would be happy to read up on it. 

> That's the function of the bulletin board.

MikesAllotment:
I agree - this would be a great idea.

briels:
This could be implemented as a room topic, add a "topic message" to a room thats displayed at the top of the room. It would function like a pinned message, and could be applied to hash channels as well. Seems like a good idea, and would probably only need a field adding to the room/channel setup which would probably be easier than figuring how to pin a specific message?

benbatt:
I have created a [draft pull request](https://github.com/meshcore-dev/MeshCore/pull/3070) that adds support for room topics - does this seem like it would suit your needs?

briels:
> I have created a [draft pull request](https://github.com/meshcore-dev/MeshCore/pull/3070) that adds support for room topics - does this seem like it would suit your needs?

The draft pull looks like it would cover that off pretty well, and in my mind should effectively hit the "pinned message" request that started this with the least complexity. It could be interesting to extend this to channels although not sure how much value that would add.

benbatt:
Ah cool, thank you! Yeah, it seems like channel topics could be useful too. I'm afraid I don't have a deep enough understanding of the protocol to picture how they would work though.

MikesAllotment:
This sounds like it should do the trick.  I own a RoomServer which is used daily by multiple people, so I'm happy to help with testing if you can build me a version for the Faketec ProMicro nRF variant?

benbatt:
[...]

PR #3070, opened 2026-07-30
Add support for room topics

benbatt:
Adds support for setting a room server topic via a new `set topic` command.

* The topic is saved in a new `/room_prefs` file
* A new `topic_timestamp` member variable is used to determine which clients have seen the current topic
* For clients that haven't seen the current topic, `MyMesh::loop()` pushes the topic before any new posts
* The topic is pushed as a post from the room server itself
  * This seems like it should be backwards compatible
  * Maybe the app could be updated to display the latest post from the room server as a persistent header in the room UI?


benbatt:
This implements a version of #2122.

Issue #3072, opened 2026-07-30
[Feature request] Enable BME680 and BMP085 in default builds for ProMicro (FakeTec) firmware


glutesha:
Since FakeTec is a pretty popular hardware for DIY repeaters it would be nice if it supported more environment sensors, like RAK4631 does. The board still has a plenty of onboard memory available for more libraries and i think it would be a great addition to the firmware.

PR #3073, opened 2026-07-30
Add BME680 and BMP085 sensors to the firmware builds for ProMicro

glutesha:
Added support for BMP085 and BME680 environment sensors for the ProMicro (Faketec) builds of MeshCore.


glutesha:
#3072

PR #3076, opened 2026-07-30
LR1110: fix startReceive() passing an IRQ bit as the RX timeout 🤖🤖

### LINKED ISSUE #3075: LR1110 boards cannot receive on dev: startReceive() passes an IRQ bit as the RX timeout
### Summary

On `dev`, `CustomLR1110::startReceive()` passes `RADIOLIB_LR11X0_IRQ_PREAMBLE_DETECTED` as RadioLib's **first** argument. That parameter is the RX **timeout**, not an IRQ mask. The constant is `1 << 4` = `16`, so the receiver is armed with a 16-tick timeout — at the LR11x0's 30.52 µs tick that is a **~488 µs receive window**. The radio leaves RX before any packet can arrive, so the node never receives anything, while transmitting normally.

https://github.com/meshcore-dev/MeshCore/blob/dev/src/helpers/radiolib/CustomLR1110.h#L38-L41

```cpp
int16_t startReceive() override {
  // include the PREAMBLE_DETECTED irq bit in reported flags
  return LR1110::startReceive(RADIOLIB_LR11X0_IRQ_PREAMBLE_DETECTED, RADIOLIB_IRQ_RX_DEFAULT_FLAGS | (1UL << RADIOLIB_IRQ_PREAMBLE_DETECTED), RADIOLIB_IRQ_RX_DEFAULT_MASK, 0);
}
```

RadioLib's own implementation passes the timeout correctly:

```cpp
// RadioLib LR11x0.cpp
int16_t LR11x0::startReceive() {
  return(this->startReceive(RADIOLIB_LR11X0_RX_TIMEOUT_INF, RADIOLIB_IRQ_RX_DEFAULT_FLAGS, RADIOLIB_IRQ_RX_DEFAULT_MASK, 0));
}
```

### Affected

All variants built with `RADIO_CLASS=CustomLR1110`:

`t1000-e`, `wio_wm1110`, `minewsemi_me25ls01`, `thinknode_m3`, `thinknode_m7`, `thinknode_m9`

`main` is **not** affected — the introducing commit is not an ancestor of `main`, and `main`'s `CustomLR1110.h` has no `startReceive()` override.

### This looks like a copy-paste slip

The sibling wrappers added in the same series all pass the timeout correctly. LR1110 is the only one passing an IRQ bit:

| Wrapper | first arg to `startReceive()` |
|---|---|
| `CustomLLCC68` | `RADIOLIB_SX126X_RX_TIMEOUT_INF` |
| `CustomSTM32WLx` | `RADIOLIB_SX126X_RX_TIMEOUT_INF` |
| `CustomSX1262` | `RADIOLIB_SX126X_RX_TIMEOUT_INF` |
| `CustomSX1268` | `RADIOLIB_SX126X_RX_TIMEOUT_INF` |
| `CustomLR1110` | `RADIOLIB_LR11X0_IRQ_PREAMBLE_DETECTED` |

Introduced in ea5d7c8b ("LR1110: add PREAMBLE_DETECTED to reported irq flags", 2026-07-25).

### Why it is easy to miss

The failure is silent and misdirects. The radio initialises cleanly, `stats-core` reports `errors: 0`, TX counters and airtime increment normally, and the noise floor reads a believable value. Nothing points at the firmware — it looks like a bad antenna, a region/preset mismatch, or a dead unit.

### Evidence

Two SenseCAP T1000-E units, matched at 910.525 MHz / BW 62.5 / SF 8 / CR 5, a few feet apart, both on `dev`. Neither could receive anything:

```
repeater : {"recv":0,"sent":9,"flood_tx":8,"direct_tx":1,"flood_rx":0,"direct_rx":0,"recv_errors":0}
           {"noise_floor":-120,"last_rssi":0,"last_snr":0.00,"tx_air_secs":8,"rx_air_secs":0}
companion: {"recv":0,"sent":7,"flood_tx":7,"direct_tx":0,"flood_rx":0,"direct_rx":0,"recv_errors":0}
           {"noise_floor":-120,"last_rssi":0,"last_snr":0.00,"tx_air_secs":3,"rx_air_secs":0}
```

Both transmit (`tx_air_secs` rising) but `rx_air_secs` and `recv_errors` stay at 0 — the receiver is never open long enough to detect even a failed preamble. (`-120` is the clamp floor; it is also close to the true thermal floor at BW 62.5 kHz, which is part of why this reads as normal.)

Applying the fix to **only one** of the two units isolates it. The fixed repeater immediately started receiving while the unfixed companion stayed deaf:

```
repeater  RX delta: {'recv': 1, 'flood_rx': 1, 'recv_errors': 0}
          last_rssi -29, last_snr 17.0, noise_floor -104
companion RX delta: {'recv': 0, 'flood_rx': 0, 'recv_errors': 0}
```

After fixing both, packets flow in each direction (`-29`/`-31` dBm, SNR ~15).

Ruled out beforehand: frequency/BW/SF/CR mismatch (identical on both, also applied live via `tempradio`), TX power and near-field desense (fails identically at 0 dBm), USB brownout, and radio init failure (`stats-core` `errors: 0`).

### Fix

```diff
-      return LR1110::startReceive(RADIOLIB_LR11X0_IRQ_PREAMBLE_DETECTED, ...);
+      return LR1110::startReceive(RADIOLIB_LR11X0_RX_TIMEOUT_INF, ...);
```

The PREAMBLE_DETECTED flag the original commit intended is unaffected — it is in the second argument, where it already is. PR to follow.

### Caveat

Confirmed by behaviour on T1000-E only. The other five variants are affected by inspection (same override, same constant), not by testing.

MDamon:
Fixes #3075

`CustomLR1110::startReceive()` passes `RADIOLIB_LR11X0_IRQ_PREAMBLE_DETECTED` as RadioLib's first argument, which is the RX **timeout**, not an IRQ mask. The constant is `1 << 4` = `16`, so the receiver is armed with a 16-tick timeout — ~488 µs at the LR11x0's 30.52 µs tick. The radio drops out of RX before a packet can arrive, so LR1110 boards on `dev` transmit normally but never receive anything.

This passes `RADIOLIB_LR11X0_RX_TIMEOUT_INF` instead, which is what `LR11x0::startReceive()` itself uses. The `PREAMBLE_DETECTED` flag that ea5d7c8b set out to add is untouched — it lives in the second argument, where it already was — so the intent of that commit is preserved.

The sibling wrappers added in the same series (`CustomLLCC68`, `CustomSTM32WLx`, `CustomSX1262`, `CustomSX1268`) all pass `*_RX_TIMEOUT_INF` correctly; LR1110 was the only one passing an IRQ bit.

### Testing

Two SenseCAP T1000-E units at 910.525 MHz / BW 62.5 / SF 8 / CR 5.

Before: both transmit (`tx_air_secs` rising) but `rx_air_secs`, `recv_errors`, `last_rssi` all stay at 0 — the receiver never opens long enough to log even a failed preamble.

Applying the fix to only one unit isolates the cause. The fixed repeater began receiving immediately while the unfixed companion stayed deaf:

```
repeater  RX delta: {'recv': 1, 'flood_rx': 1}   last_rssi -29, last_snr 17.0, noise_floor -104
companion RX delta: {'recv': 0, 'flood_rx': 0}
```

With both fixed, traffic flows in each direction (`-29`/`-31` dBm, SNR ~15), verified with adverts and public-channel messages.

### Build coverage

Builds clean on `dev` for `t1000e_repeater`, `ThinkNode_M7_repeater`, `ThinkNode_M3_repeater`, `wio_wm1110_repeater`.

`Minewsemi_me25ls01_repeater` does not build, but that is pre-existing on `dev` and unrelated to this change — `examples/simple_repeater/UITask.cpp` fails on `user_btn` / `BUTTON_EVENT_*` not being declared for that variant. It fails identically on a clean `dev` checkout without this commit.

### Caveat

Behaviour confirmed on T1000-E hardware only. The other `CustomLR1110` variants (`wio_wm1110`, `minewsemi_me25ls01`, `thinknode_m3/m7/m9`) are affected by inspection — same override, same constant — not by testing.

### DIFF
--- a/src/helpers/radiolib/CustomLR1110.h
+++ b/src/helpers/radiolib/CustomLR1110.h
@@ -36,8 +36,15 @@ class CustomLR1110 : public LR1110 {
     bool getRxBoostedGainMode() const { return _rx_boosted; }
 
     int16_t startReceive() override {
-      // include the PREAMBLE_DETECTED irq bit in reported flags
-      return LR1110::startReceive(RADIOLIB_LR11X0_IRQ_PREAMBLE_DETECTED, RADIOLIB_IRQ_RX_DEFAULT_FLAGS | (1UL << RADIOLIB_IRQ_PREAMBLE_DETECTED), RADIOLIB_IRQ_RX_DEFAULT_MASK, 0);
+      // include the PREAMBLE_DETECTED irq bit in reported flags.
+      //
+      // NOTE: the first argument is the RX *timeout*, not an IRQ mask. Passing
+      // RADIOLIB_LR11X0_IRQ_PREAMBLE_DETECTED (1<<4 = 16) here armed the receiver
+      // with a 16-tick timeout -- at the LR11x0's 30.52us tick that is a ~488us
+      // receive window, so the radio dropped out of RX before any packet could
+      // arrive and the node never received anything. It must stay RX_TIMEOUT_INF
+      // (continuous RX), which is what LR11x0::startReceive() itself passes.
+      return LR1110::startReceive(RADIOLIB_LR11X0_RX_TIMEOUT_INF, RADIOLIB_IRQ_RX_DEFAULT_FLAGS | (1UL << RADIOLIB_IRQ_PREAMBLE_DETECTED), RADIOLIB_IRQ_RX_DEFAULT_MASK, 0);
     }
 
     bool isReceiving() {
--- a/src/helpers/radiolib/CustomLR1110.h
+++ b/src/helpers/radiolib/CustomLR1110.h
@@ -37,13 +37,6 @@ class CustomLR1110 : public LR1110 {
 
     int16_t startReceive() override {
       // include the PREAMBLE_DETECTED irq bit in reported flags.
-      //
-      // NOTE: the first argument is the RX *timeout*, not an IRQ mask. Passing
-      // RADIOLIB_LR11X0_IRQ_PREAMBLE_DETECTED (1<<4 = 16) here armed the receiver
-      // with a 16-tick timeout -- at the LR11x0's 30.52us tick that is a ~488us
-      // receive window, so the radio dropped out of RX before any packet could
-      // arrive and the node never received anything. It must stay RX_TIMEOUT_INF
-      // (continuous RX), which is what LR11x0::startReceive() itself passes.
       return LR1110::startReceive(RADIOLIB_LR11X0_RX_TIMEOUT_INF, RADIOLIB_IRQ_RX_DEFAULT_FLAGS | (1UL << RADIOLIB_IRQ_PREAMBLE_DETECTED), RADIOLIB_IRQ_RX_DEFAULT_MASK, 0);
     }
 


oltaco:
Thanks, you were correct it was a copy paste error on my part. I have tidied up and merged 👍

Issue #2229, opened 2026-04-02
Help! M5stack Unit C6l firmwares not work


comsorg:
I have tried all the firmware for the M5stack Unit C6l (with and without cleaning the device, and not one of them does not work - after any firmware there is nothing on the screen and the device cannot be connected via the console via usb (it does not connect), while the USD is determined by the system. Can anyone help, maybe who knows who did the firmware?

recrof:
screen on this device is not implemented. not sure which role you flashed, but you won't be able to connect on Companion Bluetooth via USB.

comsorg:
There is no output on the screen, understood. I'm flashing M5Stack_Unit_C6L_companion_radio_usb-v1.14.1-467959c-merged.bin and M5Stack_Unit_C6L_companion_radio_usb-v1.14.1-467959c.bin I need USB access. I also flashed the router, but I can't connect in any way. There were no problems with the firmware for Meshtastic. Firmware log `esptool.js
Serial port WebSerial VendorID 0x303a ProductID 0x1001
Connecting...

Detecting chip type... ESP32-C6
Chip is ESP32-C6 (revision 2)
Features: Wi-Fi 6,BT 5,IEEE802.15.4
Crystal is 40MHz
MAC: 58:8c:81:4f:f2:4c
Uploading stub...
Running stub...
Stub running...
Manufacturer: 0
Device: 00
Detected flash size: undefined
Erasing flash (this may take a while)...
Chip erase completed successfully in 36.182s
Not changing the image
Compressed 612752 bytes to 335441...
Writing at 0x0... (4%)
Writing at 0x1229d... (9%)
Writing at 0x1e711... (14%)
Writing at 0x2302e... (19%)
Writing at 0x27e52... (23%)
Writing at 0x31f7d... (28%)
Writing at 0x390da... (33%)
Writing at 0x3f33e... (38%)
Writing at 0x46155... (42%)
Writing at 0x4c6ba... (47%)
Writing at 0x527e6... (52%)
Writing at 0x57fd2... (57%)
Writing at 0x5d695... (61%)
Writing at 0x648a2... (66%)
Writing at 0x6b3d9... (71%)
Writing at 0x71e32... (76%)
Writing at 0x788c2... (80%)
Writing at 0x7f458... (85%)
Writing at 0x8514e... (90%)
Writing at 0x8b435... (95%)
Writing at 0x91d10... (100%)
[...]

PR #3088, opened 2026-08-01
fix: M5Stack Unit C6L crash loop on boot

jirogit:
## Summary

Fixes a repeatable crash loop on the M5Stack Unit C6L (ESP32-C6 + SX1262) where the USB serial port cycles connect/disconnect continuously on boot (`/dev/ttyACM0` present/absent, several times per minute). ROM download mode was stable, confirming the hardware/USB wiring was fine and the crash was in application firmware.

Four changes, all required together:

- `SX126X_DIO3_TCXO_VOLTAGE`: `1.8` → `3.0`
- I2C pins (`PIN_BOARD_SDA`/`PIN_BOARD_SCL`): `16/17` → `10/8`
- Add a PI4IO I/O-expander init sequence in `UnitC6LBoard::begin()`, including a write to the `OUT_H_IM` (output high-impedance) register. Upstream's board file never initializes the PI4IO expander at all, so `P_LORA_RESET=-1` (SX1262 NRST wired through PI4IO P7) was left floating. `radio_init()` would then hang waiting on the SX1262 BUSY pin, and the watchdog would reset the MCU — the crash loop.
- Remove `P_LORA_TX_LED=15` (GPIO15 is an ESP32-C6 strapping pin)

Root cause for the PI4IO piece specifically: on this I/O expander, `IO_DIR` (pin direction) and `OUT_H_IM` (output driver enable) are independent registers. Writing `IO_DIR=output` and `OUT_SET` alone has no effect on the physical pin until the corresponding `OUT_H_IM` bit is also cleared — otherwise the pin stays in its power-on Hi-Z state. That's why NRST was never actually released.

## Credit

The TCXO voltage, I2C pin assignment, and PI4IO register sequence are adapted from [TheRealHaoLiu/MeshCore](https://github.com/TheRealHaoLiu/MeshCore) (`main-m5stack-unit-c6l` branch), a working M5Burner distribution build for this board.

## Testing

All testing on real M5Stack Unit C6L hardware, `companion_radio_usb` build.

- Stacked the 4 changes incrementally (PI4IO fix alone, then + I2C pins, then + TCXO, then + TX_LED removal), reflashing between steps. The crash loop persisted at every intermediate step and was only resolved once all 4 were combined. This confirms the combination is sufficient; it does not rule out that some subset smaller than 4 might also work, since not every subset was isolated separately.
- Verified with a full chip erase (`esptool erase_flash`) + a from-scratch clean build (`.pio/build` removed, no incremental cache) + reflash, to rule out stale build artifacts or leftover NVS state as a factor in the earlier failed steps.
- Confirmed stable: zero USB disconnect/reconnect events over multiple sustained monitoring windows (dmesg-level USB re-enumeration tracking, 10+ minutes total with zero drops after the fix), versus a 2-3 second disconnect/reconnect cycle beforehand.
- Confirmed a USB client (companion app protocol) can connect and exchange data over the stabilized port.

contribute to  #2229.

**Update:** Confirmed working via `pio run -t upload` (USB env,
companion_radio_usb). I haven't been able to get it working via the
official web flasher (flasher.meshcore.io) with the merged .bin yet.
Also haven't tested the BLE companion env specifically — the fix
targets the crash-loop-on-boot issue, not connectivity beyond that.

If anyone tries this via the web flasher, would be curious whether you
hit the same wall or if it's something specific to my setup.

### DIFF
--- a/variants/m5stack_unit_c6l/UnitC6LBoard.h
+++ b/variants/m5stack_unit_c6l/UnitC6LBoard.h
@@ -1,12 +1,96 @@
 #pragma once
 
 #include <Arduino.h>
+#include <Wire.h>
 #include <helpers/ESP32Board.h>
 
+#define PI4IO_ADDR 0x43
+
+// PI4IO registers
+#define PI4IO_REG_CHIP_RESET 0x01
+#define PI4IO_REG_IO_DIR 0x03
+#define PI4IO_REG_OUT_SET 0x05
+#define PI4IO_REG_OUT_H_IM 0x07
+#define PI4IO_REG_IN_DEF_STA 0x09
+#define PI4IO_REG_PULL_EN 0x0B
+#define PI4IO_REG_PULL_SEL 0x0D
+#define PI4IO_REG_IRQ_STA 0x13
+#define PI4IO_REG_INT_MASK 0x11
+
 class UnitC6LBoard : public ESP32Board {
+private:
+  bool i2c_write_byte(uint8_t addr, uint8_t reg, uint8_t value) {
+    Wire.beginTransmission(addr);
+    Wire.write(reg);
+    Wire.write(value);
+    return Wire.endTransmission() == 0;
+  }
+
+  bool i2c_read_byte(uint8_t addr, uint8_t reg, uint8_t *value) {
+    Wire.beginTransmission(addr);
+    Wire.write(reg);
+    if (Wire.endTransmission() != 0) return false;
+    if (Wire.requestFrom(addr, (uint8_t)1) != 1) return false;
+    if (!Wire.available()) return false;
+    *value = Wire.read();
+    return true;
+  }
+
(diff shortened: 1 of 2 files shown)


jirogit:
Closing this — the four changes here turned out not to be the actual
fix. I re-tested with a cleaner methodology (full erase, clean build,
consistent flashing method) and found none of them were necessary;
stock upstream/dev boots fine under the same conditions.

The real cause was `flash_mode: qio` in the board manifest
(esp32-c6-devkitm-1.json), which this chip doesn't handle reliably —
confirmed by bisecting flash_mode specifically (reproduced the crash
loop twice with qio, resolved it with dio, on otherwise identical
firmware).

Opening a new PR with the actual fix: a one-line
`board_build.flash_mode = dio` override. Sorry for the churn — details
in the new PR and in #2229.

PR #3103, opened 2026-08-03
fix(jp-lbt): reduce LBT backoff parameters for SF10 airtime

jirogit:
## Summary

JP LBT backoff parameters were tuned during SF12/CR4:8 development.
Following the SF10/CR4:6 migration (jp-v1.8.0), airtime decreased by ~4-5x,
making the original backoff values unnecessarily conservative.

## Changes

```
base_ms:     2000 -> 500 ms
max_backoff: 16000 -> 4000 ms
```

jitter is already airtime-scaled (`airtime_ms / JP_LBT_JITTER_DIVISOR`)
and requires no change.

## Rationale

Backoff duration should reflect how long a competing transmission actually
occupies the channel. At SF10/BW125/CR4:6, a 255-byte packet takes ~2854ms
(vs ~14s at SF12). Keeping SF12-era backoff values at SF10 adds unnecessary
latency with no regulatory benefit — ARIB STD-T108 specifies the sensing
threshold (-80 dBm / 5 ms) and TX time limit (4s), not backoff duration.

These are provisional values. Precise tuning will follow once traffic
statistics are available from deployed nodes.

## Dependency

Depends on #2218 (JP LBT implementation). This PR targets upstream/dev
and should be merged after #2218.

### DIFF
--- a/src/helpers/radiolib/RadioLibWrappers.cpp
+++ b/src/helpers/radiolib/RadioLibWrappers.cpp
@@ -2,6 +2,12 @@
 #define RADIOLIB_STATIC_ONLY 1
 #include "RadioLibWrappers.h"
 
+#ifdef NRF52_PLATFORM
+  #define YIELD_TASK() vTaskDelay(1)
+#else
+  #define YIELD_TASK() delay(1)
+#endif
+
 #define STATE_IDLE       0
 #define STATE_RX         1
 #define STATE_TX_WAIT    3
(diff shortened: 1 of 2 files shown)


jirogit:
Folded into #2218 (same commit, cherry-picked onto feature/jp-lbt) since
#2218 already depends on this and having the fix live separately was
creating confusion about which constants are current. Closing as superseded.

PR #2054, opened 2026-03-17
Muzi Base Duo Support

lbibass:
Added support for the Muzi Base Duo. I didn't buy the SuperIO board, so those features aren't tested. But still, everything other than the display, IMU, and the toggle switch should work? (no support for the SH1107 in MC yet, and no use for an IMU yet either). 

So really, it's just the GPS and Buzzer that would work on the SuperIO. 

This has in fact been tested on actual hardware, and isn't just some AI slop where a majority of the pins are defined incorrectly. 

I may try and add support for the Uno as well, but I don't own one of those (yet). 

EDIT: the display is working now. Still no IMU or toggle switch support yet.

PR #3108, opened 2026-08-04
Add Muzi Base board support (Duo + Uno + SuperIO)

### LINKED ISSUE #2779: Firmware Support for Muzi Works SuperBase Kit
Muzi Works SuperBase Kit is a standalone companion device powered by Muzi Work's [Super IO](https://muzi.works/products/super-io) and [Base Uno](https://muzi.works/products/base-uno) boards.

https://muzi.works/products/superbase-kit

The Base Uno is an nRF52840 MCU combined with SX1262. More details here: https://muzi.works/products/base-uno

As the Base Uno/Duo boards can be used separately to build custom devices, it would be nice to support each of them individually as well. The [Base Duo](https://muzi.works/products/base-duo) board is an nRF52840 combined with a LR1121 radio.

tjdownes:
Adds board support for the Muzi Base (muzi.works).

It's one nRF52840 variant shared by both radio flavors, selected per build env: the Duo uses an LR1121, the Uno an SX1262. They're the same PCB apart from the radio module and its DIO1 pin.

The SuperIO add-on brings the SH1107 OLED, GPS, buzzer and joystick. GPS is switched on and off by the 3-position mode switch ("Mode 2"), polled live the way thinknode_m1 already does it. Five quick presses of the user button power the unit off, and a press wakes it back on. Tested on Duo and Uno with SuperIO.

Build envs, per radio: `repeater`, `room_server`, `companion_radio` usb/ble, and `companion_radio` ble superIO.

Pin map was cross-checked against the Meshtastic muzi_base variant.

This consolidates #2054 (base duo, @lbibass) with @andyshinn's shared-base + uno work, rebased onto current dev, so it supersedes #2054. Original authorship is kept via co-author trailers on the commit.

Closes #2779

### DIFF
--- a/variants/muzi_base/muzi_baseBoard.h
+++ b/variants/muzi_base/muzi_baseBoard.h
@@ -1,0 +1,48 @@
+#pragma once
+
+#include <MeshCore.h>
+#include <Arduino.h>
+#include <helpers/NRF52Board.h>
+
+// The Muzi Base PCB ships in two radio flavors. Keep the user-facing identity
+// (OTA/DFU name and reported manufacturer) distinct per radio even though the
+// board logic is shared.
+#if defined(USE_SX1262)
+  #define MUZI_BASE_OTA_NAME  "MUZI_BASE_UNO_OTA"
+  #define MUZI_BASE_MFR_NAME  "Muzi Base Uno"
+#else
+  #define MUZI_BASE_OTA_NAME  "MUZI_BASE_DUO_OTA"
+  #define MUZI_BASE_MFR_NAME  "Muzi Base Duo"
+#endif
+
+class muzi_baseBoard : public NRF52BoardDCDC {
+protected:
+#ifdef NRF52_POWER_MANAGEMENT
+  void initiateShutdown(uint8_t reason) override;
+#endif
+
+public:
+  muzi_baseBoard() : NRF52Board(MUZI_BASE_OTA_NAME) {}
+  void begin();
+
+  #define BATTERY_SAMPLES 8
+
+  uint16_t getBattMilliVolts() override {
+    analogReadResolution(12);
+    analogReference(AR_INTERNAL_3_0);
+    delay(1);
+
+    uint32_t raw = 0;
+    for (int i = 0; i < BATTERY_SAMPLES; i++) {
+      raw += analogRead(PIN_VBAT_READ);
+    }
+    raw = raw / BATTERY_SAMPLES;
+
+    // ADC_MULTIPLIER is the voltage divider ratio
+    return (raw * ADC_MULTIPLIER * AREF_VOLTAGE) / 4.096;
+  }
+
+  const char* getManufacturerName() const override {
+    return MUZI_BASE_MFR_NAME;
+  }
+};
--- a/variants/muzi_base/platformio.ini
+++ b/variants/muzi_base/platformio.ini
@@ -1,0 +1,326 @@
+; ============================================================================
+; Muzi Base  (MuziWorks)
+;
+; One nRF52840 board, shipped in two radio flavors. The radio is selected per
+; environment; all board logic lives in variants/muzi_base and is shared:
+;     muzi_base_duo_*  -> LR1121  (USE_LR1121)
+;     muzi_base_uno_*  -> SX1262  (USE_SX1262)
+; ============================================================================
+
+[muzi_base_common]
+extends = nrf52_base
+board = muzi_base
+board_build.ldscript = boards/nrf52840_s140_v6.ld
+build_flags = ${nrf52_base.build_flags}
+  -I src/helpers/nrf52
+  -I lib/nrf52/s140_nrf52_6.1.1_API/include
+  -I lib/nrf52/s140_nrf52_6.1.1_API/include/nrf52
+  -I variants/muzi_base
+  -I src/helpers/ui
+  -D muzi_base
+  -D NRF52_POWER_MANAGEMENT
+  -D PIN_USER_BTN=PIN_BUTTON1
+  -D USER_BTN_PRESSED=LOW
+  -D PIN_STATUS_LED=35
+  -D LORA_TX_POWER=22
+  -D P_LORA_BUSY=LORA_BUSY
+  -D P_LORA_SCLK=LORA_SCLK
+  -D P_LORA_NSS=LORA_NSS
+  -D P_LORA_DIO_1=LORA_DIO_1
+  -D P_LORA_MISO=LORA_MISO
+  -D P_LORA_MOSI=LORA_MOSI
+  -D P_LORA_RESET=LORA_RESET
+  -D QSPIFLASH=1
+build_src_filter = ${nrf52_base.build_src_filter}
+  +<helpers/*.cpp>
+  +<helpers/sensors>
+  +<../variants/muzi_base>
+debug_tool = jlink
+upload_protocol = nrfutil
+lib_deps =
+  ${nrf52_base.lib_deps}
+  ${sensor_base.lib_deps}
+
+; ---------------------------------------------------------------------------
+; Radio selection (Duo = LR1121, Uno = SX1262). The matching radio pins / TCXO
+; / RF-switch defines are picked up in variant.h via USE_LR1121 / USE_SX1262.
+; ---------------------------------------------------------------------------
+[muzi_base_duo]
+extends = muzi_base_common
+build_flags = ${muzi_base_common.build_flags}
+  -D USE_LR1121
+  -D RADIO_CLASS=CustomLR1121
+  -D WRAPPER_CLASS=CustomLR1121Wrapper
+  -D RF_SWITCH_TABLE
+  -D RX_BOOSTED_GAIN=true
+  -D LR11X0_DIO_AS_RF_SWITCH=true
+  -D LR11X0_DIO3_TCXO_VOLTAGE=3.0
+
+[muzi_base_uno]
+extends = muzi_base_common
+build_flags = ${muzi_base_common.build_flags}
+  -D USE_SX1262
+  -D RADIO_CLASS=CustomSX1262
+  -D WRAPPER_CLASS=CustomSX1262Wrapper
+  -D SX126X_RX_BOOSTED_GAIN=1
+  -D SX126X_CURRENT_LIMIT=140
+  -D SX126X_DIO2_AS_RF_SWITCH=1
+  -D SX126X_DIO3_TCXO_VOLTAGE=3.3
+
+; ===========================================================================
+; Duo (LR1121) environments
+; ===========================================================================
+[env:muzi_base_duo_repeater]
+extends = muzi_base_duo
+build_flags = ${muzi_base_duo.build_flags}
+  -I examples/companion_radio/ui-new
+  -D ADVERT_NAME='"Muzi Base Repeater"'
+  -D ADVERT_LAT=0.0
+  -D ADVERT_LON=0.0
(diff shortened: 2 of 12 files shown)


notbenh:
👍 Awesome thank you.

tjdownes:
I want to add that I have tested these scenarios:

## Companion
- Base Uno without Super IO (one month)
- Base Uno with Super IO (one week)
- Base Duo without Super IO (one week)
- Base Duo with Super IO (one week)

## Repeater
- Base Uno (1 day)
- Base Duo (1 day)

I have not tested room servers, but I assume they will operate fine based on the performance/results of testing repeaters

notbenh:
+1 again, rebased to the 0.17.1 tag with one minor merge conflict that was easy to follow and built new firmware.zip and loaded successfully via https://apps.meshamerica.com/

tjdownes:
@liamcottle just an FYI, as I notice you added the "New Hardware Support" label to #2054 , this PR should supersede that one. It's more complete, and more thoroughly tested across the devices and roles it supports. Thank you.

jessenine:
One note: this work has been rebased onto current `dev` and submitted fresh as **PR #3457** (https://github.com/meshcore-dev/MeshCore/pull/3457) so it can be merged cleanly. This PR supersedes it; the author credit is preserved in the co-author trailers.

variants/muzi_base/muzi_baseBoard.h
```diff
@@ -0,0 +1,48 @@
...
+// board logic is shared.
+#if defined(USE_SX1262)
+  #define MUZI_BASE_OTA_NAME  "MUZI_BASE_UNO_OTA"
+  #define MUZI_BASE_MFR_NAME  "Muzi Base Uno"
+#else
+  #define MUZI_BASE_OTA_NAME  "MUZI_BASE_DUO_OTA"
+  #define MUZI_BASE_MFR_NAME  "Muzi Base Duo"
+#endif
+
+class muzi_baseBoard : public NRF52BoardDCDC {
```
liamcottle: Please make sure class names start with capital letters. I believe the company/brand name uses lowercase, but would prefer to stick to uppercase convention for class names, so it's not confused as being a variable name in some places.

`muzi_baseBoard` -> `MuziBaseBoard`

variants/muzi_base/platformio.ini
```diff
@@ -0,0 +1,326 @@
...
+  +<../examples/simple_room_server>
+lib_deps = ${muzi_base_duo.lib_deps}
+
+[env:muzi_base_duo_companion_radio_usb]
+extends = muzi_base_duo
+board_build.ldscript = boards/nrf52840_s140_v6_extrafs.ld
+board_upload.maximum_size = 708608
+build_flags = ${muzi_base_duo.build_flags}
+  -I examples/companion_radio/ui-new
+  -D MAX_CONTACTS=500 ;can increase number of contacts since we have a ton of extra flash.
```
liamcottle: 350 is the highest we support in all the other variants. So keep in mind this will make contact syncing much slower, and could lead to unintended side effects where we haven't expected contacts to be allowed this high.

variants/muzi_base/platformio.ini
```diff
@@ -0,0 +1,326 @@
...
+  densaugeo/base64 @ ~1.4.0
+  stevemarple/MicroNMEA @ ^2.0.6
+  end2endzone/NonBlockingRTTTL@^1.3.0
+  adafruit/Adafruit SH110X@^2.1.14
+  artronshop/ArtronShop_RX8130CE@1.0.0
+  adafruit/Adafruit GFX Library @ ^1.12.1
+debug_tool = jlink
+upload_protocol = nrfutil
+
+[env:muzi_base_duo_companion_radio_ble_superIO]
```
liamcottle: what is `superIO`? this board wouldn't be included in automated builds as the env/target name is unexpected format. `superIO` should be _before_ `companion_radio_ble`.

variants/muzi_base/platformio.ini
```diff
@@ -0,0 +1,326 @@
...
+  densaugeo/base64 @ ~1.4.0
+  stevemarple/MicroNMEA @ ^2.0.6
+  end2endzone/NonBlockingRTTTL@^1.3.0
+  adafruit/Adafruit SH110X@^2.1.14
+  artronshop/ArtronShop_RX8130CE@1.0.0
+  adafruit/Adafruit GFX Library @ ^1.12.1
+debug_tool = jlink
+upload_protocol = nrfutil
+
+[env:muzi_base_uno_companion_radio_ble_superIO]
```
liamcottle: same here, this board won't get built due to `superIO` suffix.

variants/muzi_base/target.h
```diff
@@ -0,0 +1,64 @@
+#pragma once
+
+#define RADIOLIB_STATIC_ONLY 1
+#include <RadioLib.h>
+#include <helpers/radiolib/RadioLibWrappers.h>
+#include "muzi_baseBoard.h"
```
liamcottle: should be uppercase class name

variants/muzi_base/target.h
```diff
@@ -0,0 +1,64 @@
...
+#endif
+#include <helpers/ArduinoHelpers.h>
+#include <helpers/SensorManager.h>
+#include <helpers/sensors/LocationProvider.h>
+#include <helpers/AutoDiscoverRTCClock.h>
+#include <helpers/sensors/EnvironmentSensorManager.h>  // Added: Include for EnvironmentSensorManager
+#include <helpers/ui/MomentaryButton.h>
+
+
+#ifdef muzi_base_superIO
```
liamcottle: build flags should generally be fully uppercased. ie `MUZI_BASE_SUPERIO`

variants/muzi_base/variant.cpp
```diff
@@ -0,0 +1,90 @@
...
+    41,
+    42,
+    43,
+    44,
+    45,
+    46,
+    47,
+};
+
+void initVariant()
```
liamcottle: Just a note. As far as I am aware, we don't use `initVariant` anywhere in the code base, so the code in this section will never run. I believe these variant.cpp files have come from the manufacturer, so if the code below is required, it will need to be done in the begin section of the target class.

liamcottle (changes requested):
Hey there! Thanks for the PR, since it was mentioned that this PR should supersede the original, I have done my review on this one.

variants/muzi_base/muzi_baseBoard.h
notbenh: Do we have documented code conventions for things like this?

variants/muzi_base/platformio.ini
notbenh: it's the carrier board https://muzi.works/products/super-io

variants/muzi_base/muzi_baseBoard.h
liamcottle: > Do we have documented code conventions for things like this?

Not documented in MeshCore itself, but it's generally standard practice in all programming languages that class names usually start with uppercase letters.

variants/muzi_base/platformio.ini
notbenh: again are there docs for conventions? Also given that we are seeing a lot more boards with way more memory are plans to continue artificially capping to 350 out of momentum?

tjdownes: I propose for now we'll standardize on the current 350... it can always be bumped later

liamcottle: > I propose for now we'll standardize on the current 350... it can always be bumped later

Sounds good to me, thanks!

liamcottle: > again are there docs for conventions? Also given that we are seeing a lot more boards with way more memory are plans to continue artificially capping to 350 out of momentum?

There's not really any docs for things like this, but if you search for `MAX_CONTACTS=` in the repo, you'll see that the majority of them are 350.

The memory isn't really the issue, it's more so the file system and how contacts are stored. We still have issues with file system corruption, due to erase page sizes not being the same as filesystem block sizes.

So for now it's probably safer to use the same as all other boards, until we get that sorted.

variants/muzi_base/muzi_baseBoard.h
tjdownes: it's MuziBaseBoard now. I renamed the files to match too (MuziBaseBoard.h/.cpp)

variants/muzi_base/platformio.ini
tjdownes: Renamed to `muzi_base_duo_superIO_companion_radio_ble` so the suffix matches and build.sh picks it up

variants/muzi_base/platformio.ini
tjdownes: renamed to `muzi_base_uno_superIO_companion_radio_ble`

variants/muzi_base/target.h
tjdownes: `MUZI_BASE_SUPERIO` now. I also renamed the board define from `muzi_base` to `MUZI_BASE`

variants/muzi_base/variant.cpp
tjdownes: Moved into `MuziBaseBoard::begin()`. One FYI: on nRF52 the Arduino core does call it (`cores/nRF5/main.cpp` calls `initVariant()` right after `init()`), and 37 other variants in the repo define one, so it was running. Keeping it in `begin()` is fine though, it keeps the setup in one place.

variants/muzi_base/target.h
tjdownes: Done

liamcottle:
Looks good, thanks for all the requested changes.

I don't have any of these devices, but I'm happy to merge as is if you're all happy that it's working as expected.

Only one other note I had was that I see you're defining `-D BLE_TX_POWER=0`, instead of leaving it undefined, which would use the default of `4`. Is there an issue with these devices running it at the higher power? Or is this just to save a small amount of battery life?

Wondering if users would prefer the further BLE range. But I do see that we have `-D BLE_TX_POWER=0` in `meshtracker_x1`, `t1000e` and `thinknode_m3` variants. Unsure why those were changed to be lower.

tjdownes:
> I don't have any of these devices, but I'm happy to merge as is if you're all happy that it's working as expected.

Several of us have been testing the variants for a few months. I've been using one as as daily for two months.I haven't heard any complaints in some time, so I am happy to merge as is.

> Only one other note I had was that I see you're defining `-D BLE_TX_POWER=0`, instead of leaving it undefined, which would use the default of `4`. Is there an issue with these devices running it at the higher power? Or is this just to save a small amount of battery life?

To be honest, I'd hesitate to reduce the power. My real world experience is that it needs what it's currently using.

liamcottle:
> To be honest, I'd hesitate to reduce the power. My real world experience is that it needs what it's currently using.

Sorry, I meant that this PR intentionally uses less power for bluetooth than all the other firmwares. Removing that build flag would give it more bluetooth transmit power, likely leading to more range.

But if it's working fine as is, I can merge...

tjdownes:
> Sorry, I meant that this PR intentionally uses less power for bluetooth than all the other firmwares. Removing that build flag would give it more bluetooth transmit power, likely leading to more range.

Hah, I see what you meant. Sorry, very early here!

Yes, I agree, the BLE is probably too low. I will push another commit to set it to the default.

liamcottle:
There's some extra discussion here. Looks like it was introduced to save a bit of extra battery life on the t1000e.
https://github.com/meshcore-dev/MeshCore/pull/736

So my guess is this muzi variant copied some of those build flags over.

Will merge in, thanks to all that have contributed!

PR #3120, opened 2026-08-06
ThinkNode M6: Remove QSPIFLASH=1 build flag to fix companion roles

recrof:
M6 doesn't have flash chip available, so companion roles fail to start.


liamcottle:
Can't find any schematics for this device, but looking at internal board, I couldn't see any flash chips.

recrof:
Adding photo of the board for future reference. Note the unpopulated area above nRFLR1262:
IMG20260806125142

PR #3112, opened 2026-08-04
Add support for Seeed SenseCAP MeshTracker X1 (LR2021)

Hacuchino-hash:
## Summary

Adds support for the **Seeed SenseCAP MeshTracker X1**, a card-sized tracker built on the nRF52840 with Semtech's new **LR2021** radio and a dual-band (L1+L5) Airoha GNSS.

This is (as far as I can tell) the first LR2021-based target in MeshCore. RadioLib gained LR2021 support in the 7.6/7.7 line, and the existing `jgromes/RadioLib @ ^7.6.0` dependency resolves to 7.7.1, which includes the driver plus the datasheet v2.1 updates, so no dependency changes were needed.

## What's included

- `src/helpers/radiolib/CustomLR2021.h` + `CustomLR2021Wrapper.h`: LR2021 wrappers modeled closely on the LR1110 ones (header-error recovery via `standby()`, preamble/SF handling, instant RSSI for noise-floor sampling, RX boosted gain mapped to the LR2021's 0 to 7 boost levels)
- `variants/meshtracker_x1/`: variant pin map, board class, GPS sensor manager (Airoha @ 115200, same power-pin sequencing style as the T1000-E), and envs for `repeater`, `room_server`, `companion_radio_usb`, `companion_radio_ble`
- `boards/seeed-mesh-tracker-x1.json`: board definition (S140 7.3.0, same layout as the T1000-E)

Pin assignments were cross-checked against Seeed's wiki and the Meshtastic `seeed_mesh_tracker_X1` variant. One LR2021-specific detail: the chip's IRQ line to the MCU is its **DIO8** pin (routed to P1.01), configured via RadioLib's `irqDioNum` before `begin()`.

The four env names follow the standard suffixes, so `build.sh` firmware discovery picks them up automatically.

## Testing

- All four environments build clean (RAM 11-58%, flash 44-52%).
- Flashed `MeshTracker_X1_companion_radio_ble` on real hardware via serial DFU: boots, USB enumerates, BLE advertises, and a phone running the MeshCore app connects and interacts normally. Since `radio_init()` halts the firmware before BLE starts on failure, this confirms the LR2021 initializes with the TCXO/IRQ config in this PR.
- Messaging is fully functional: texts send and deliver to other MeshCore nodes over the air, with delivery ACKs received back. TX and RX both confirmed on the LR2021.
- GPS is fully functional: enabling the `gps` sensor setting powers the Airoha receiver, the node gets a fix, and location populates in the companion app.
[...]

PR #3122, opened 2026-08-06
Add RGB LED and haptic notifications for SenseCAP MeshTracker X1

Hacuchino-hash:
## Summary

Follow-up to #3112, adding notification and sensor support for the SenseCAP MeshTracker X1's remaining onboard hardware.

**RGB status LED** (the X1 has a discrete RGB LED on P0.03/P0.24/P0.28):
- Color-coded status blinks with the same stock duty cycle: green heartbeat, blue when a phone is connected over BLE, amber when messages are waiting, red when battery is low
- A short rainbow sweep on boot and on message arrival
- Implemented in ui-orig behind `PIN_STATUS_LED_R/G/B` defines, so boards with a single status LED keep the stock code path (verified by building t1000e_companion_radio_ble)

**DRV2605 haptics** (LRA motor behind the DRV2605 on I2C):
- New `DRV2605Vibration` helper with the same interface as `GenericVibration`, using the register config and effect that Meshtastic ships for this same chip
- Vibrates on incoming messages, and fires whether or not the app is connected (the buzzer keeps its stock disconnected-only behavior)
- Triple-press now cycles four alert modes: Buzz+Vibe, Buzz only, Vibe only, Silent. Each mode confirms by tone and/or vibration since the X1 has no screen, and both flags persist (`buzz_q`, new `vibe_q` pref)

**SPA06-003 barometer**:
- Temperature and pressure reported through `TELEM_PERM_ENVIRONMENT` telemetry, using the same library and settings as Meshtastic (8x oversampling, 1 Hz continuous so reads never block)

**Charge status on the LED**:
- While docked: amber breathing during charge, dim green when the charger signals done. The charger asserts both status pins at full charge, so the done pin (P1.04) takes priority, and both open-drain status lines get internal pullups
- Message indications always outrank the charge display

## Testing

All verified on real hardware (X1 + phone app + other MeshCore nodes):
- Status colors and both flourishes behave as described, and messaging/BLE are unaffected
- Vibration fires on messages while connected and disconnected, honoring mode and a 5 s cooldown
- Alert mode cycling persists across reboots
- Barometer readings appear in telemetry alongside GPS
- Charge LED transitions amber to green at charge termination, and the battery reads 4.40 V at the charger's CV point (high voltage cell)
- All four X1 envs build clean, plus t1000e_companion_radio_ble as a regression check for the shared ui-orig changes

### DIFF
--- a/examples/companion_radio/NodePrefs.h
+++ b/examples/companion_radio/NodePrefs.h
@@ -28,6 +28,7 @@ public:
   uint32_t ble_pin = 0;
   uint8_t  advert_loc_policy = 0;
   uint8_t  buzzer_quiet = 0;
+  uint8_t  vibe_quiet = 0;
   uint8_t  gps_enabled = 0;      // GPS enabled flag (0=disabled, 1=enabled)
   uint32_t gps_interval = 0;     // GPS read interval in seconds
   uint8_t autoadd_config = 0;    // bitmask for auto-add contacts config
@@ -101,6 +102,7 @@ private:
       def("defs_key", (void *) _parent->default_scope_key, sizeof(_parent->default_scope_key));
       def("pin", _parent->ble_pin);
       def("buzz_q", _parent->buzzer_quiet);
+      def("vibe_q", _parent->vibe_quiet);
       def("auto_add", _parent->autoadd_config);    // bitmask for auto-add contacts config
       def("man_add", _parent->manual_add_contacts);
       def("tel_base", _parent->telemetry_mode_base);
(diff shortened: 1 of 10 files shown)


ripplebiz:
Thanks for this.

Hacuchino-hash:
Small correction after more bench testing on the charge status LED. The charger exposes two status lines and I was giving priority to the second one (P1.4) as a charge complete signal, but on this board it reads low regardless of state, so the LED sat on the charged colour the whole time. Charge state now comes from P1.3 alone, which behaves correctly, and P1.4 is left unused with a note (Meshtastic leaves it out of their X1 variant for the same reason). Verified on hardware: amber while charging at 4.15 V, and all four X1 envs still build clean.

PR #3124, opened 2026-08-07
Fix X1 charge status LED always showing charged

Hacuchino-hash:
## Summary

Follow up to #3122, correcting the charge complete handling on the MeshTracker X1.

The charging indication itself works, the LED shows the charging animation while power is applied. What is wrong is the charge complete branch: it gives priority to the charger's second status line (P1.4) as a "done" signal, and on this board that pin reads low regardless of charge state, so the complete branch always wins and the LED sits on the charged colour.

Charge state now comes from P1.3 alone, which behaves correctly (low while charging, high once charging stops). P1.4 is left unused with a comment explaining why, the same choice Meshtastic makes in their X1 variant where the equivalent define is commented out.

## Testing

Measured on hardware at partial charge (4.15 V, actively charging): P1.3 reads low as expected, and P1.4 also reads low, which is what sends it down the wrong branch. With this change the LED correctly shows the charging animation at that state. All four X1 environments build clean.

### DIFF
--- a/examples/companion_radio/ui-orig/UITask.cpp
+++ b/examples/companion_radio/ui-orig/UITask.cpp
@@ -344,11 +344,7 @@ void UITask::userLedHandler() {
   static bool ext_powered = false, ext_charging = false;
   if (cur_time > next_pwr_check) {
     ext_powered = _board->isExternalPowered();
-    bool chrg = digitalRead(EXT_CHRG_DETECT) == LOW;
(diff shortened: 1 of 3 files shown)


Hacuchino-hash:
Closing this for now. The charge complete state needs more work than this change covers: on this board neither charger status line gives a usable done signal, so the charged state is still not reachable. I will retest properly on hardware and open a complete fix.

PR #3125, opened 2026-08-07
Fix X1 charge status LED always showing charged

Hacuchino-hash:
## Summary

Follow up to #3122, correcting the charge complete handling on the MeshTracker X1.

The charge status logic gives priority to the charger's second status line (P1.4) as a "done" signal, but on this board that pin reads low regardless of charge state, so the complete branch always wins and the LED sits on the charged colour from the moment it is plugged in.

Charge state now comes from P1.3 alone, which behaves correctly. P1.4 is left unused with a comment explaining why, the same choice Meshtastic makes in their X1 variant where the equivalent define is commented out.

## Testing

Verified on hardware across a full charge cycle: with this change the LED shows the charging animation while charging and switches to the charged colour when the charger terminates. Pin readings taken during charging confirm the cause, P1.3 reads low as expected while charging and P1.4 also reads low, which is what sends the merged code down the wrong branch. All four X1 environments build clean.

### DIFF
--- a/examples/companion_radio/ui-orig/UITask.cpp
+++ b/examples/companion_radio/ui-orig/UITask.cpp
@@ -344,11 +344,7 @@ void UITask::userLedHandler() {
   static bool ext_powered = false, ext_charging = false;
   if (cur_time > next_pwr_check) {
     ext_powered = _board->isExternalPowered();
-    bool chrg = digitalRead(EXT_CHRG_DETECT) == LOW;
-  #ifdef EXT_CHRG_DONE
-    if (digitalRead(EXT_CHRG_DONE) == LOW) chrg = false;   // charge-done wins
-  #endif
-    ext_charging = ext_powered && chrg;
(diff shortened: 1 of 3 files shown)


oltaco:
I tried this out and can confirm that the LED breathes amber when charging now but it never seems to complete charge? Voltage is showing 4.41v when plugged in and ~4.3v when unplugged but the device still hasn't changed to dim green LED.

I'll leave it on charge overnight, maybe it'll be green in the morning?

edit: No luck, unless I'm missing something this just seems to swap one behaviour (always green when plugged in) for another (always breathing orange when plugged in).

oltaco:
It turns out that the BQ24045 only signals charge complete on the *first* charge cycle, which explains the behaviour I was seeing. 

> The CHG signal is active only during the first charge cycle. Exiting TTDM or the OUT voltage falling below VRCH
will not activate the CHG signal if it is already in the open-drain (off) state.

Thanks for your work on this @Hacuchino-hash 👍

Hacuchino-hash:
Happy to help! Sorry for the late reply!

Issue #3130, opened 2026-08-08
[Feature request] Hardware support: Inhero MR-2 (CE-certified BQ25798/RAK4630 based solar repeater board)


liekmarflow:
The Inhero MR-2 is a purpose-built solar repeater board that went on sale in
August 2026
[(https://shop.inhero.de/en/products/inhero-mr-2-solar-mesh-repeater-board-rak4630-sx1262-mppt-red-ce-gepruft)](https://shop.inhero.de/en/products/inhero-mr-2-solar-mesh-repeater-board-rak4630-sx1262-mppt-red-ce-gepruft)
I am the hardware manufacturer and would like to contribute first-class
MeshCore support for it.

**Hardware:**
- RAK4630 core module (nRF52840 + SX1262), 45 x 40 mm, CE-certified
  (RED 2014/53/EU, tested by an accredited lab)
- BQ25798 buck/boost charger: universal solar input 3.6-24 V with MPPT,
  JEITA temperature-controlled charging
- INA228 coulomb counter: real SOC tracking and 7-day energy statistics
- Supported chemistries: Li-ion, LiFePO4, LTO, Na-ion
- RV-3028 RTC: low-voltage system sleep (<500 uA) with autonomous recovery
- BME280 environment sensor

The MeshCore variant is complete and has been running in the field for months
on my own installations. Production boards ship with only the bootloader
preinstalled — users flash the MeshCore firmware themselves (prebuilt images
are available from my fork's releases until this is merged). I plan to submit
it as two pull requests:

1. A minimal set of board hooks in the core: two no-op virtual methods on `mesh::MainBoard` — `tick()` for periodic housekeeping and a watchdog, and `queryBoardTelemetry()` for board-specific telemetry channels. Defaults are no-ops; existing boards are unaffected. The `board.*` CLI needs nothing new — it rides on the `MainBoard::handleCommand()` hook that `dev` already provides.

2. The variant itself (`variants/inhero_mr2/` + board JSON), which uses those hooks. Self-contained; the full EN/DE documentation set is maintained in my fork and linked from the variant README.

As the manufacturer I can test changes on real hardware — feel free to tag me on anything affecting this variant.

Transparency note: the firmware for this variant was developed with AI
assistance under my continuous engineering direction and review; it has been
through a full pre-submission review pass and months of field operation.
I am stating this upfront in line with this project's position on code
provenance.

PR #3131, opened 2026-08-08
Add board loop hook for periodic work

liekmarflow:
Adds a default no-op `mesh::MainBoard::loop()` and calls it from the
`simple_repeater` and `simple_sensor` main loops. Board variants can override
it to feed a watchdog and run periodic housekeeping.

The patch is eight added lines in three files. The MR2 variant in #3132
uses its SensorManager subclass for additional telemetry and the existing
`MainBoard::handleCommand()` for its CLI commands.

Verified locally: `RAK_4631_repeater` and `Heltec_v3_repeater` both build
successfully. Needed for #3130 (Inhero MR2 hardware support).

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -249,6 +249,11 @@ int MyMesh::handleRequest(ClientInfo *sender, uint32_t sender_timestamp, uint8_t
     }
     sensors.querySensors(perm_mask, telemetry);
 
+    // Board-specific telemetry (boards can override queryBoardTelemetry in their Board class)
+    if (perm_mask & TELEM_PERM_ENVIRONMENT) {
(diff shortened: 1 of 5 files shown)


ripplebiz:
Am ok with the Board::tick(), but the queryBoardTelemetry has a code smell.  Subclassing the SensorManager, or configuring particular sensors via conditionals is where that belongs.

liamcottle:
Could we also make it board.loop instead of board.tick pls. So it's the same as other classes. Only exception I believe is the RTC class for time.

liekmarflow:
Thanks @ripplebiz and @liamcottle, that makes sense. Renamed tick() to loop() and removed the board telemetry hook entirely. MR2 telemetry will use the existing SensorManager subclass in #3132 instead. This PR now only adds the default no-op MainBoard::loop() and calls it from the repeater and sensor loops.

PR #3134, opened 2026-08-08
Add other Adafruit environment sensors to Pro Micro (Faketec) hardware variant builds

glutesha:
NRF Pro Micro is a pretty popular board for building repeaters and companions, and since it already includes BME280 and BMP280 sensors why not add other ones, that are included in, for example, RAK4631 builds?

Uses #2634 code for BME680 part by @NickDunklee because his library integration works better than the standard Adafruit library, at least on the similar hardware.


glutesha:
#3072

PR #3140, opened 2026-08-09
Lilygo T-Echo Lite Pin Fixes

entr0p1:
- Aligned to Lilygo's updated schematic and pin map as per https://github.com/Xinyuan-LilyGO/T-Echo-Lite/issues/14

Edit: tested on a USB companion.


oltaco:
Looks good, thanks for the legwork on this one 👍

PR #3154, opened 2026-08-09
nrf52: call nRFCrypto.begin()/end() only once

fdlamotte:
This follows discussion from #2824 

Still think this has to be checkout profundly to asses the hw is used as it should (this is a touchy thing)

Needs some tests on different boards too

don't know if there is a difference energy wise

### DIFF
--- a/src/Identity.cpp
+++ b/src/Identity.cpp
@@ -27,11 +27,9 @@ bool Identity::verify(const uint8_t* sig, const uint8_t* message, int msg_len) c
   // needs much less, around 600-700bytes. The CC310 workspace is static, faster,
   // should save power at scale as well.
   static CRYS_ECEDW_TempBuff_t cc310_tmp;
-  nRFCrypto.begin();
   CRYSError_t rc = CRYS_ECEDW_Verify((uint8_t*)sig, CRYS_ECEDW_SIGNATURE_BYTES,
                                      (uint8_t*)pub_key, CRYS_ECEDW_MOD_SIZE_IN_BYTES,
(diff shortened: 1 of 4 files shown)


fdlamotte:
for RNG we should probably use the work already made in #2280 already uses HW on most architectures ...

IoTThinks:
> don't know if there is a difference energy wise

Let me test on RAK4631 and T096.
The original PR has little power impact. Therefore, the fix will likely have little power impact too.

NickDunklee:
Nice work!


Back-story for fun (not judgement): The previous flavor cycled the CC310 on/off the way it did when hardware crypto wasn't being used when the node sleeps and traffic is idle. I figured potential issues would shake out in testing one way or another. I've been using the previous variation for 2+ months without issue with varying traffic load on multiple nodes. Documentation said it must be disabled when not needed to ensure lowest power consumption. (Cited below.)


Ref for that original decision (https://docs.nordicsemi.com/r/bundle/ps_nrf52840/page/cryptocell.html) :
>CRYPTOCELL must be disabled to ensure lowest possible power consumption when the subsystem is not needed.


THAT being said, it seems power consumption is likely not an issue just leaving the CC310 turned on all the time, which is great, and much less thrashy under load than my brutalist interpretation. :)


I'm sure there are further changes that can be made to improve the handling - wanted to keep that initial change small so everyone could iterate and improve, as it were.


@IoTThinks did you get a chance to test this variant out of curiosity? I doubt it will matter, but really interesting seeing the numbers.

PR #3099, opened 2026-08-02
Fix M5Stack Unit C6L kiss_modem env missing USB CDC flags

jirogit:
## Problem

`M5Stack_Unit_C6L_kiss_modem` env did not include the USB CDC build flags present in `M5Stack_Unit_C6L_companion_radio_usb`:

- `ARDUINO_USB_CDC_ON_BOOT=1`
- `ARDUINO_USB_MODE=1`
- `ENABLE_USB_INTERFACE`

On ESP32-C6, native USB serial requires these flags at boot. Without them, output routes to UART0, which has no physical connector on this board, making the KISS modem unresponsive over USB.

## Fix

Added the same three flags used in the companion_radio_usb env for this board.

## Testing

Hardware-verified on M5Stack Unit C6L. Built with `pio run -t upload -e M5Stack_Unit_C6L_kiss_modem`, confirmed device enumerates as `/dev/ttyACM0`, confirmed KISS modem operational via openhop_repeater.

PR #3156, opened 2026-08-09
fix(m5stack_unit_c6l): add USB CDC flags to repeater and room_server envs

jirogit:
Same root cause as #3099 (kiss_modem): without `ARDUINO_USB_CDC_ON_BOOT=1` / `ARDUINO_USB_MODE=1`, Serial routes to UART0, which has no physical connector on this board — so the USB admin console was unreachable on the repeater and room_server envs.

Tested on real hardware: USB admin console now responds on both envs, confirme configuration via serial.


jirogit:
Closing - this was opened against the wrong base, pulling in the entire JP fork history instead of just the CDC flag fix. Will reopen with a clean branch based on upstream/dev.

PR #1439, opened 2026-01-21
Added support of cyrillic characters output on companion devices with…

ivang78:
This pull request is an improvement of this closed PR: https://github.com/meshcore-dev/MeshCore/pull/1237 . 
All changes sqashed into one commit branched from current dev state. 
Fixed bug with "glueing" next message preview with current message preview for messages ended with trimmed UTF-8 sequence. 
Also message previews with UTF-8 characters truncated for correct length. 
Works only with SSD1306 display. Use compile flag -D OLED_RU=1 (for variants with -D DISPLAY_CLASS=SSD1306Display) for enable feature. Example of usage compile flag added to variants/heltec_v3/platformio.ini file.

PR #3175, opened 2026-08-11
Add Cyrillic/UTF-8 text support to SH1106Display

strasharo:
Switches `SH1106Display` from Adafruit_SH110X/GFX to U8g2 so it can actually render Cyrillic (and other non-Latin UTF-8) text instead of falling back to block characters. Picks up on #1930 and the still-open #1439, where the author got SSD1306 working but didn't have SH1106 hardware to test against - I do, so this covers that gap.

What changed:
- font rendering now goes through U8g2 instead of Adafruit_GFX. Using the X11 "fixed" fonts (public domain, ISO10646/Unicode-encoded, full Basic Latin + Cyrillic coverage) at two sizes to roughly match the old font's small/large text sizes
- `print()`/`getTextWidth()` now use U8g2's UTF-8-aware `drawUTF8()`/`getUTF8Width()` instead of the old byte-per-char path
- overrode `translateUTF8ToBlocks()` to just pass the string through unchanged, since the display can render it properly now and doesn't need the block-character fallback
- fixed `drawXbm()` along the way - the icon/logo bitmaps in this codebase are packed MSB-first (Adafruit_GFX::drawBitmap() convention), but U8g2's own drawXBM() expects LSB-first, so icons came out completely garbled until I switched to U8g2's u8glib-compatible drawBitmap() instead, which uses the same MSB-first order

SH1106Display is shared by 7 boards (nano_g2_ultra, thinknode_m2, station_g2, wio-tracker-l1, lilygo_tbeam_1w, station_g3_esp32, lilygo_tbeam_supreme_SX1262), swapped the SH110X/GFX lib_deps for U8g2 on all of them. Built all 7 clean. Actually flash-tested on Wio Tracker L1 Pro and T-Beam Supreme since that's what I have on hand - confirmed Cyrillic text renders correctly (including a real message received over a channel), icons render correctly, and existing Latin-script screens still look right.

Not in scope here: long messages still don't wrap or scroll on this display, that's a separate thing I'll follow up with once this lands.

### DIFF
--- a/src/helpers/ui/SH1106Display.cpp
+++ b/src/helpers/ui/SH1106Display.cpp
@@ -1,6 +1,4 @@
 #include "SH1106Display.h"
-#include <Adafruit_GrayOLED.h>
-#include "Adafruit_SH110X.h"
 
 bool SH1106Display::i2c_probe(TwoWire &wire, uint8_t addr)
 {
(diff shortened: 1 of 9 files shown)


strasharo:
Closing this for now - found a cleaner way to do this (wrapping the existing Adafruit_GFX driver with U8g2_for_Adafruit_GFX instead of swapping the whole rendering backend), going to rework it that way instead. Will open a new PR once it's ready.

PR #3179, opened 2026-08-11
fix: split radio.rxgain, radio.fem.rxgain def

### LINKED ISSUE #3172: T096 - 1.17 repeater radio.fem.rxgain not working
**TL;DR**:  `radio.fem.rgain` is not persisted across boots and does not seem to have an effect at all.
Fixed instantly by flashing back to 1.16

# Hardware / firmware

- Heltec Mesh Node T096 (nRF52840), repeater role
- 868 MHz, cavity filter, 3 dBi omni, solar powered, urban environment
- Clean 1.17.0 flash with data erased (or really any method I tested).

# First problem

`set radio.fem.rxgain off` applies at runtime but does not survive a reboot. After restart the value reads on again.

## Steps to reproduce:
- `get radio.fem.rxgain`  -> on
- `set radio.fem.rxgain off`
- `get radio.fem.rxgain`  -> off
- `reboot`
- `get radio.fem.rxgain`  -> on

Which, in the context of a solar repeater, makes the setting pretty much useless.

# Second problem 

Setting `radio.fem.rxgain` off (without a reboot) does not seem to have any effect (tested over the course of a few hours)

Admittedly I can't test directly wether or not the LNA is enabled, my only evidence is the degradaed noise floor (~7dbm) & RSSI (~20dbm) associated with the RX gain.

## Steps to reproduce (on 1.17):

- `set radio.fem.rxgain off` ->  *nothing*
- Flash 1.16 (keep settings) ->  Noise floor and RSSI go back to normal

alustenberg:
fixes #3172 with radio.rxgain and radio.fem.rxgain being written to the same config variable.

### DIFF
--- a/examples/companion_radio/NodePrefs.h
+++ b/examples/companion_radio/NodePrefs.h
@@ -33,6 +33,7 @@ public:
   uint32_t gps_interval = 0;     // GPS read interval in seconds
   uint8_t autoadd_config = 0;    // bitmask for auto-add contacts config
   uint8_t rx_boosted_gain = 0; // SX126x RX boosted gain mode (0=power saving, 1=boosted)
+  uint8_t radio_fem_rxgain = 0; // Front End Module LNA mode (0=off, 1=on)
   uint8_t _client_repeat = 0;  // DEPRECATED -> use repeat.disable_fwd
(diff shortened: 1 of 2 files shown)


alustenberg:
this is already covered in PR #3137 as well, with tx PA extended.

PR #3184, opened 2026-08-12
Fix FEM RX gain preference persistence.

oltaco:
This PR fixes the radio.fem.rxgain preference tracking the rx_boosted_gain setting.


cwichura:
This should be closed in favor of #3137 which is a more complete fix.

Issue #3151, opened 2026-08-09
T096 no power output 1.17 & fem force on


towerviewcams:
There is no transmit power on T096 with new 1.17 firmware. confirmed on 3 boards.  also, fem will force on when flashed

ellisgl:
I'm also seeing this issue. Even when I'm a couple feet from each other, nothing is see from it.

HighVoltage73:
I also observed this behavior. Companion node works fine on 1.17.  I flashed the repeater back to 1.16 and all is well. I also notice you can set the tx power pretty much arbitrarily and it doesn't make a difference. I could log into the repeater with the companion beside the repeater but move it away say 30-40' and couldn't even ping it.

CincyRon:
Having the same issue here. 2 different T096 boards and pretty much zero output power with this latest firmware. Flashed back to 1.16 and showing 28.8 dBm output power on the current board I have hooked up.

dg1tal:
I know this is cumbersome with repeaters deployed in the field, but a useful next step would be to test what happens if you factory-reset your known-bad v1.17.0 T096 and then restore its config line-by-line. 

So if anyone got a lab spare that already showed this issue, please try that.

CincyRon:
> I know this is cumbersome with repeaters deployed in the field, but a useful next step would be to test what happens if you factory-reset your known-bad v1.17.0 T096 and then restore its config line-by-line.
> 
> So if anyone got a lab spare that already showed this issue, please try that.

I did this multiple times last night with 2 different boards. Erase flash, flash with 1.17, reconfig line by line. No xmit power. Erase flash, flash with 1.16, reconfig line by line. Full xmit power.  Erase flash, flash with 1.17 bluetooth companion, reconfigure. Full Xmit power. It's definitely something with 1.17 repeater firmware. I just put my repeater back up with v1.16, but I have my second T096 sitting here on the bench if you have any other ideas to try.

oltaco:
I'm looking into this, confirmed no power on my T096 running v1.17 main.

oltaco:
Small update, when ``MESH_DEBUG`` is enabled T096 repeater works properly on v1.17. My first thought was that the 5 second delay at startup on ``MESH_DEBUG`` builds is what allows it to work but forcing the 5 second delay also doesn't fix it.

Will keep digging.

dg1tal:
[...]

PR #3188, opened 2026-08-12
Fix Heltec T096 Tx issue

oltaco:
This PR fixes the Heltec T096 transmit issue which was found to be caused by out of bounds read of the pin map.

Reverted the previous workarounds, changed ``PIN_SPI1_MISO`` pin to 0, which in this variant maps to 0xFF and ends up as ``NRFX_SPIM_PIN_NOT_USED``.

Closes https://github.com/meshcore-dev/MeshCore/issues/3151

### DIFF
--- a/src/helpers/CommonCLI.h
+++ b/src/helpers/CommonCLI.h
@@ -79,8 +79,7 @@ private:
       def("bw", _parent->bw);
       def("sf", _parent->sf);
       def("cr", _parent->cr);
-      static const char radio_cad_key[] = {'c', 'a', 'd', '\0'}; // workaround for T096, don't touch without testing T096 repeater still transmit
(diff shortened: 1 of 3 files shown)


jschrempp:
Is it appropriate to clap for all the investigation you all have done? good job!

IoTThinks:
@oltaco Very nice find.
There are two potentiall more boards to check.

They don't have the FEM btw.
image

oltaco:
> @oltaco Very nice find. There are two potentiall more boards to check.
> 
> They don't have the FEM btw. image

All fixed in #3189 👍
Note that even without FEM using -1 for any SPI or I2S pin will lead to out of bounds read and potential for undefined behaviour.

PR #3191, opened 2026-08-13
Add companion_radio_serial env for XIAO nRF52840 🤖🤖

jpmartineau:
This change lets a Seeed XIAO nRF52840 radio talk to a host computer over a simple two-wire serial connection instead of USB or Bluetooth — which is what's needed to use it inside a Clockwork PicoCalc or any other project that requires a connection to the nRF via GPIO instead of USB.

Two problems stood in the way, both now fixed:

It wouldn't compile. The code borrowed a serial-port setup written for a different chip family. Swapping in the nRF52's own equivalent fixed it without touching anything else.
Two pins were being used twice. The same two pins were also claimed by the sensor bus, which grabbed them back a moment after the serial link was set up — so the connection went dead with no error message anywhere. The sensors now use different pins.

Moving the sensor bus rather than the serial link was the only option: with the radio board mounted, just three pins are left exposed, and the two connections need four between them — and the board's own sensor was never on those pins anyway.

This board has several firmware variants — Bluetooth companion, USB companion, repeater, room server, modem, and now this serial one — sharing a common block of settings. The pin change is written as an override inside the new variant's own section rather than into that shared block, so it applies to this firmware alone: flash any of the other five and their pins behave exactly as before. Nothing else needs re-testing.

Tested on real hardware wired to a Luckfox Lyra: the radio has stayed on the mesh since flashing and handles everything expected of it. The two neighbouring variants were rebuilt as a check and still work. My PicoCalc has been working great with the nRF using this firmware for 3 weeks now. Super stable.

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -61,7 +61,13 @@ MultiSerialInterface interface_manager;
 #if defined(SERIAL_RX)
   #include <helpers/ArduinoSerialInterface.h>
   ArduinoSerialInterface hardware_serial_interface;
-  HardwareSerial companion_serial(1);
+  #if defined(NRF52_PLATFORM)
+    // nRF52 has no numbered-constructor HardwareSerial; Serial1 is the concrete UARTE0
+    // instance the core always defines. setPins() below moves it to the requested pads.
+    Uart& companion_serial = Serial1;
+  #else
+    HardwareSerial companion_serial(1);
+  #endif
 #endif
 
 // platform file system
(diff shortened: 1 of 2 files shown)


jpmartineau:
I updated the PR's description for clarity and readability.

jpmartineau:
Related: #2499 proposed moving the sensor bus off D6/D7 for every XIAO build, because those pins are also Serial1's. The discussion there found the overlap has no effect in those builds, since none of them ever starts Serial1, and that moving the sensor pins for everyone would break anyone with sensors on D6/D7.

This PR is the one build where the overlap is real: it starts Serial1 on D6/D7, so the sensor bus has to move, but only in this environment. The shared settings, and the pins of every existing environment, are unchanged. The new pins (SCL 16, SDA 17) are the values already written, commented out, in `variants/xiao_nrf52/variant.h`.

PR #3195, opened 2026-08-13
Allow companion BLE/USB CMD_SET_DEVICE_TIME to set clock backwards

### LINKED ISSUE #3173: FW1.17.0 - time setting not correct nRF52
After updating and clkreboot and clock sync time still wrong on repeater and compagnions with nRF52 boards.

estebanscdev:
CMD_SET_DEVICE_TIME only ever arrives from the locally-paired companion app/CLI, not from untrusted mesh peers, so the owner should be able to correct a drifted device clock in either direction. Previously a fast- running clock (e.g. software fallback clock on nRF52 boards with no GPS fix or RTC chip, after long uptime) could never be corrected once it drifted ahead, since the firmware rejected any earlier timestamp.

Fixes #3173

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -1240,13 +1240,8 @@ void MyMesh::handleCmdFrame(size_t len) {
   } else if (cmd_frame[0] == CMD_SET_DEVICE_TIME && len >= 5) {
     uint32_t secs;
     memcpy(&secs, &cmd_frame[1], 4);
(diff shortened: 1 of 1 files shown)


ErikBrown2:
Thanks for this modification. I hope that it will be accepted because it really solves a headache. My client always checks the time at startup and tries to correct it if it is more off than 10 seconds. But this never works due to the restriction of not being able to set an earlier time. I can see some logic in this rule but I think that it makes overall more sense to allow to set any time.

PR #3196, opened 2026-08-13
fix(repeater): prevent stack overflow out of RX handler

benallfree:
Hey team, this is a bug I stumbled across as I was customizing the firmware and noticed a stack overflow happening in an advert action from a remote admin UI. It is not reproducible on the stock firmware, but the bug/potential is there nonetheless.

It can be fixed by queuing remote commands and executing them as part of the loop instead of handling them inline.


mikecarper:
https://github.com/meshcore-dev/MeshCore/pull/3101

PR #3202, opened 2026-08-13
ThinkNode M6: expose external-power and charger state

F4FPR:
## What

- adds `MainBoard::isChargerActive()` with a safe default of `false`;
- defines the ThinkNode M6 external-power and charger-status lines using the
  existing `EXT_PWR_DETECT` / `EXT_CHRG_DETECT` conventions;
- overrides `isExternalPowered()` and `isChargerActive()` on the M6;
- adds an exact `power` CLI command returning battery millivolts, external-power
  state and charger-source state.

Example:

```text
batt:4012 mV ext:yes charger:yes
```

## Why

The generic nRF52 implementation only checks MCU USB VBUS. The ThinkNode M6 also
has dedicated board signals for external input (P0.13, active HIGH) and its solar
charger (P0.15, active LOW). Exposing both makes USB/powerbank and solar behavior
observable without changing the charging path.

## Semantics

- `ext:yes` means that the dedicated external-input line, the active charger
  line, or the nRF52 USB-VBUS fallback detects an external source;
- `charger:yes` means that the solar charger signal is active or an external USB
  charging source is present;
- with USB input, the available board signals cannot distinguish active battery
  current from charge completion.

## Scope

- other boards retain the default `isChargerActive() == false`;
- no change to GPS, ADC conversion, charger control, sleep policy or radio
  configuration;
- the command uses an exact `strcmp`, so it does not overlap with `poweroff`;
- no Meshtastic code is copied; only documented board pin facts are reused.

## Testing

The board-state logic was compiled and exercised in an isolated C++11 harness
for no source, external USB input and active solar charging. Full target build
and on-device validation remain for CI/hardware.

Expected hardware checks:

1. no external source: `ext:no charger:no`;
2. USB-C/powerbank connected: `ext:yes charger:yes`;
3. solar charging in sufficient light: `ext:yes charger:yes`;
4. unplugging the source restores `ext:no charger:no` once the charger line is
   inactive.

### DIFF
--- a/src/MeshCore.h
+++ b/src/MeshCore.h
@@ -74,6 +74,8 @@ public:
 
   // Power management interface (boards with power management override these)
   virtual bool isExternalPowered() { return false; }
+  // True while a board can identify an active battery-charging source.
+  virtual bool isChargerActive() { return false; }
   virtual uint16_t getBootVoltage() { return 0; }
   virtual uint32_t getResetReason() const { return 0; }
   virtual const char* getResetReasonString(uint32_t reason) { return "Not available"; }
--- a/src/MeshCore.h
+++ b/src/MeshCore.h
@@ -76,6 +76,14 @@ public:
   virtual bool isExternalPowered() { return false; }
   // True while a board can identify an active battery-charging source.
   virtual bool isChargerActive() { return false; }
+
+  // Optional, source-specific power detection. Boards that can distinguish a USB
+  // supply from a solar charger override these; defaults keep every other board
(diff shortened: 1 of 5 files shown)


F4FPR:
Updated to report USB and solar-charger states separately (usb / solar_chg), in addition to the existing ext / charger indication.
Source-specific detection is gated behind capability methods that default to false, so other boards are unaffected.
Validated on real hardware in solar autonomy (usb:no solar_chg:yes on battery+solar only).

PR #3204, opened 2026-08-14
Add selectable UI color theme for color TFT displays

### LINKED ISSUE #3169: T096 - Bright theme. Difficult to read. Battery wasting.
Heltec T096 - Bright background theme on 1.17. Difficult to read and energy wasteful. An unfortunate decision considering the work having gone into battery saving.

estebanscdev:
Adds a runtime-selectable "legacy" dark theme (restoring the pre-refactor black background / green-yellow text palette) as an opt-in alongside the current "default" light theme, for boards using ST7735Display, ST7789LCDDisplay, NV3001BDisplay or LGFXDisplay (e.g. Heltec T096).

Fixes #3169

Selectable via Serial CLI: "set theme legacy" / "set theme default". DisplayDriver::applyTheme() returns false for an unrecognized/unsupported theme name, so the command rejects typos instead of silently falling back.

Persisted in NodePrefs.ui_theme (survives reboots and full power loss, written to flash via DataStore::savePrefs), reapplied at boot and live without reboot. DisplayDriver::getThemeCount() reports per-driver theme support so the command only works on boards with a color-capable display.

### DIFF
--- a/examples/companion_radio/AbstractUITask.h
+++ b/examples/companion_radio/AbstractUITask.h
@@ -43,4 +43,6 @@ public:
   virtual void newMsg(uint8_t path_len, const char* from_name, const char* text, int msgcount) = 0;
   virtual void notify(UIEventType t = UIEventType::none) = 0;
   virtual void loop() = 0;
+  virtual bool applyTheme(const char* theme) { return false; }
+  virtual int getThemeCount() { return 0; }
 };
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -2050,6 +2050,19 @@ void MyMesh::checkCLIRescueCmd() {
         _prefs.ble_pin = atoi(&config[4]);
         savePrefs();
         Serial.printf("  > pin is now %06d\n", _prefs.ble_pin);
+      } else if (memcmp(config, "theme ", 6) == 0) {
+        if (_ui != NULL && _ui->getThemeCount() > 0) {
+          if (_ui->applyTheme(&config[6])) {
+            strncpy(_prefs.ui_theme, &config[6], sizeof(_prefs.ui_theme) - 1);
+            _prefs.ui_theme[sizeof(_prefs.ui_theme) - 1] = 0;
+            savePrefs();
+            Serial.printf("  > theme is now %s\n", _prefs.ui_theme);
+          } else {
+            Serial.println("  Error: theme not found or not supported by this display");
+          }
+        } else {
+          Serial.println("  Error: this device has no theme-capable display");
+        }
       } else {
         Serial.printf("  Error: unknown config: %s\n", config);
       }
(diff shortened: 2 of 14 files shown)


IoTThinks:
> set theme legacy" / "set theme default

This is cool.
However, there is no CLI in companions.

shleepong:
<p align="center"></p>

Tested this on real hardware (Heltec Wireless Tracker V2, `heltec_tracker_v2_companion_radio_ble`), a few findings plus one suggestion.

**Tested and working**: `set theme legacy` / `set theme default` both work exactly as described, immediate effect, persists across a full power cycle (not just a soft reset).

**One thing not obvious from the diff**: on `companion_radio_ble`, the USB serial port only accepts plain-text `set` commands once "CLI Rescue" mode is entered (long-press the board's user button within the first 8 seconds after boot, see `UITask::handleLongPress()`). Outside that mode the port carries the binary companion protocol and silently drops plain text. Worth a one-line mention somewhere, since it cost us a while to figure out why `set theme legacy` had no effect over USB.

**Current merge conflict**, reproduced against today's `dev`: two files. In `NodePrefs.h`: trivial, another PR added `cad_enabled`/`tz_offset` at the same spot as `ui_theme`/`theme`, just keep both. In `MyMesh.cpp`: real conflict. `checkCLIRescueCmd()`'s old inline `set`-command chain (where the `theme` case lives here) has since been refactored into a shared `handleCommand()`, called from three places: the serial rescue console, `CMD_RUN_CLI_COMMAND` (the binary companion protocol, an app could expose this without any CLI), and `onCLICommandRecv()` (remote CLI over the mesh from a trusted contact). Moving `theme` into `handleCommand()` alongside `set pin`/`set name` resolves the conflict and makes it reachable from all three, not just serial, which also addresses @IoTThinks's point above about companions having no CLI.

**Suggestion, a third theme.** We like the mechanism a lot (persisted, runtime-selectable, minimal diff), but `legacy`'s green-on-black wasn't to our taste. Since `getThemeCount()` already anticipates more than two themes, we prototyped a `dark` option: identical to `default`, only `window_bkg`/`primary_txt` flipped, nothing else touched. Tested side by side on the same board:

<table cellspacing="12" cellpadding="0" border="0">
<tr>
<td align="center"><br><sub>legacy</sub></td>
<td align="center"><br><sub>default</sub></td>
<td align="center"><br><sub>dark</sub></td>
</tr>
<tr>
<td align="center"><br><sub>legacy</sub></td>
<td align="center"><br><sub>default</sub></td>
<td align="center"><br><sub>dark</sub></td>
</tr>
</table>

One thing that stood out testing side by side: on `default`, the unused part of the screen shows up as a stark white rectangle; on `dark` it just blends into the panel's black bezel, easier to read, and probably a small power-draw win. Happy to open a PR against your branch with the rebase plus this addition if that's useful, or just leave the diff here if you'd rather do it yourself.

Separately, we also prototyped driving `set theme` from an ordinary chat message (no CLI at all) via a synthetic local contact intercepting `CMD_SEND_TXT_MSG` before it touches the radio, see #3347 for the general idea. Might be worth keeping in mind as an app-side answer to the "no CLI in companions" gap, independent of this PR.

PR #3206, opened 2026-08-14
combine radio's entropy with CC310 RNG

ripplebiz:


src/helpers/radiolib/RadioLibWrappers.h
nextgens, line 104: aka:

```suggestion
void random(uint8_t* dest, size_t sz) override {
#ifdef USE_CC310_HW_CRYPTO
    nRFCrypto.Random.generate(dest, (uint16_t)sz);
#endif
    for (int i = 0; i < sz; i++) {
      dest[i] ^= _radio->randomByte() ^ (::random(0, 256) & 0xFF);
    }
  }
```

nextgens (changes requested):

src/helpers/radiolib/RadioLibWrappers.h
ripplebiz: You've missed the ' ^= '

nextgens: Indeed; fixed it now. I'm not usually coding in a browser :)

oltaco: Thanks for your suggestion

PR #3207, opened 2026-08-14
Rpi PicoW: Fix i2c pins and enable environment sensors

reinforce:
* Arduino-Pico SDK uses setSDA and setSCL for i2c pins assignment instead of setPins
* Define preffered i2c pins for PicoW targets
* Build PicoW targets with sensors enabled
* Use environment sensor manager instead of just sensor manager


oltaco:
All PRs must be based on dev branch, please reopen a new PR against dev or rebase and force push.

PR #3214, opened 2026-08-14
Companion USB interface: report real connection state, add flow control 🤖🤖

notbucki:
Two related fixes for the companion USB interface, found while running the `*_companion_radio_usb` / multi-interface builds on a small fleet of devices.

## 1. The device always believes a client is connected

`ArduinoSerialInterface::isConnected()` returns `true` unconditionally, so from boot onwards the companion thinks somebody is attached to the USB port. That leaks into two places:

* `MyMesh` only calls `_ui->notify()` while no client is connected, so **new message notifications (display, buzzer) never fire** on a USB companion build.
* `UITask` shows `< Connected >` in place of the BLE pairing PIN. On a build that exposes both BLE and USB this means **a freshly flashed device cannot be paired at all** - the PIN is simply never shown.

The commit adds an optional `setConnectedCheck()` hook and wires it up per platform:

| target | source of truth |
|---|---|
| TinyUSB CDC (nRF52, RP2040) | `(bool)Serial` = real DTR, ie. the host has the port open |
| ESP32 USB-OTG (`ARDUINO_USB_MODE=0`, CDC on boot) | `tud_cdc_n_get_line_state()` bit 0 = DTR (see the edit note below) |
| ESP32 USB-Serial-JTAG (`ARDUINO_USB_MODE=1`) | frame activity, see below |
| plain UART | unchanged, still assumes connected |

The USB-Serial-JTAG peripheral has no DTR concept at all - `HWCDC::isCDC_Connected()` is driven by SOF frames plus an IN-EMPTY interrupt, so it goes true as soon as the cable sits in a powered port, whether or not anything opened the port. There is no register that exposes the CDC line state, so the only honest signal left is "a client has actually sent us a frame recently" (`USB_CLIENT_IDLE_TIMEOUT`, 10 minutes, overridable per build).

## 2. No flow control on the USB write path

`CMD_GET_CONTACTS` streams up to `MAX_CONTACTS` frames back to back (~150 bytes each, so ~50 KB with the companion default of 350). `MyMesh::checkSerialInterface()` does gate this on `!_serial->isWriteBusy()`, but `ArduinoSerialInterface::isWriteBusy()` returns `false` unconditionally, and `writeFrame()` calls `write()` without ever looking at `availableForWrite()`.

On ESP32 that meets a 256 byte TX ring: when the host stalls for longer than the write timeout, HWCDC latches itself disconnected and `flushTXBuffer()` silently discards queued bytes, ie. it tears a frame in half. The framing is a length prefix with no checksum and no resync marker, so the client stays desynchronised for the rest of the session. In practice this shows up as "the app disconnects while syncing contacts", and only on devices that already have a decent contact list - a fresh device sends a few hundred bytes and never hits it.

The commit makes flow control an opt-in feature of `ArduinoSerialInterface` (the `Print` default for `availableForWrite()` is 0, so it must not be on by default for arbitrary streams) and enables it for the companion USB interface: report busy until a whole frame fits, and drop a frame as a unit rather than tearing it. It also gives HWCDC a larger TX buffer and a short write timeout, the same reasoning as the existing `kiss_modem` tuning in #2819.

Note the ordering: while no client is connected, writes are dropped and `isWriteBusy()` returns false. Otherwise a port that nobody drains keeps its buffer full forever, and the busy flag would stall the paced streams on *all* interfaces, including BLE.

## Testing

* Built for `heltec_v4_r8_companion_radio_usb` and `GAT562_30S_Mesh_Kit_companion_radio_usb` (ESP32-S3 HWCDC and nRF52 TinyUSB).
* Running in daily use on Heltec V4-R8, MTools GAT562 30s, RAK WisMesh Pocket, LilyGo T-Beam Supreme, LilyGo T-Echo, Heltec T114, Seeed T1000-E and Meshnology W12 in combined BLE+USB builds.
* Verified on device: BLE pairing PIN is shown again on a freshly flashed unit, and a full contact sync from a third party client (KiekR) that previously hung part way through now completes.

The two commits are independent in spirit but the second one uses the connection state from the first, so they are submitted together. Happy to split or rework either of them.

---

*Disclosure: this pull request was prepared by an AI agent (Claude) working together with the repository owner of the fork, who runs the affected devices. The bugs were found in daily use on real hardware, the analysis and patches are the agent's work and were reviewed and tested by a human before submission. Marked with 🤖🤖 per CONTRIBUTING.md. Happy to answer any questions or rework the patches.*

---

**Edited 2026-09-05.** A cross-model adversarial review of this PR's own code found four defects in it. All are fixed in the current head, and two of them change what the patch does to other people's boards, so the description above has been corrected rather than left standing.

* **It broke three existing builds.** The HWCDC block was guarded on `ARDUINO_USB_MODE == 1` alone, but `Heltec_v3_companion_radio_usb`, `Heltec_Wireless_Paper_companion_radio_usb` and `Heltec_WSL3_companion_radio_usb` set that while leaving `ARDUINO_USB_CDC_ON_BOOT` unset — their `Serial` is a `HardwareSerial`, which has neither `setTxTimeoutMs()` nor `onEvent()`. Every HWCDC guard now requires CDC-on-boot as well, and automatic flow control is scoped to native CDC transports, so those three keep exactly the behaviour they had before this PR. Verified by building all three.
* **`(bool)Serial` is not a DTR test on ESP32 USB-OTG.** `USBCDC::operator bool()` only becomes true once the host asserts DTR *and* RTS, so a client that deliberately leaves RTS low — the usual way to keep these boards' auto-reset circuit quiet — had every reply silently discarded. That path now reads DTR directly from `tud_cdc_n_get_line_state()`. Affects `Station_G2`, `Station_G3_ESP32`, `Heltec_E213`, `Heltec_T190`, `heltec_tracker_v2` and `LilyGo_TBeam_1W`.
* **Clients that only listen stopped receiving pushes.** Gating notifications on the ten minute activity window cut off a client that sends its opening commands and then only reads. Notification eligibility is now a predicate of its own — link up and a frame seen at least once — while pacing, PIN visibility and the connection indicator keep the stricter test. `MultiSerialInterface` aggregates it, otherwise it never reaches `MyMesh` at all.
* **The TX buffer was resized after `Serial.begin()`.** `HWCDC::setTxBufferSize()` frees the old ring without masking the interrupt, so resizing a live buffer can hand a TX-empty ISR freed memory. It is now sized before `begin()`, while the ISR is still inactive.

**Removed since the previous version of this branch:** the bus-reset event hook and the RX drain that went with it. They existed to stop a dead session's leftover bytes from confusing the next one, and four different implementations of that idea each turned out worse than the problem — the last of them could splice a frame together across a session boundary and hand `MyMesh` a contact record with framing bytes spliced into it. Without them the parser behaves as it does with no patch at all, which is the behaviour that actually recovers.

Known limits, deliberately not addressed here:

* On link loss the receive parser is left untouched, exactly as it is without this PR: the continuation of an interrupted transfer is absorbed by the in-flight frame and the stream resynchronises by itself. Several attempts to do better each produced a worse failure than the one they removed.
* An attached client that only listens can still see message-driven screen wakes after ten minutes, because that display flag also gates PIN visibility — and restoring the pairing PIN is the point of this change.
* `CMD_SYNC_NEXT_MESSAGE` takes a message out of the offline queue and then ignores `writeFrame()`'s return value. That is pre-existing and outside this PR; it deserves a separate fix.

*The review was run with Codex (`gpt-6-astra`, reasoning effort `max`) as an independent cross-provider second opinion over several rounds; the patches and this description are the work of Claude Code (an AI coding agent), and a human reviewed and approved this edit before it was posted.*

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -290,7 +290,7 @@ uint8_t MyMesh::getExtraAckTransmitCount() const {
 }
 
 void MyMesh::logRxRaw(float snr, float rssi, const uint8_t raw[], int len) {
-  if (_serial->isConnected() && len + 3 <= MAX_FRAME_SIZE) {
+  if (_serial->isSessionEstablished() && len + 3 <= MAX_FRAME_SIZE) {
     int i = 0;
     out_frame[i++] = PUSH_CODE_LOG_RX_DATA;
     out_frame[i++] = (int8_t)(snr * 4);
@@ -342,7 +342,7 @@ uint8_t MyMesh::getAutoAddMaxHops() const {
 
 void MyMesh::onContactOverwrite(const uint8_t* pub_key) {
     _store->deleteBlobByKey(pub_key, PUB_KEY_SIZE); // delete from storage
-  if (_serial->isConnected()) {
+  if (_serial->isSessionEstablished()) {
     out_frame[0] = PUSH_CODE_CONTACT_DELETED;
     memcpy(&out_frame[1], pub_key, PUB_KEY_SIZE);
     _serial->writeFrame(out_frame, 1 + PUB_KEY_SIZE);
@@ -350,14 +350,14 @@ void MyMesh::onContactOverwrite(const uint8_t* pub_key) {
 }
 
 void MyMesh::onContactsFull() {
-  if (_serial->isConnected()) {
+  if (_serial->isSessionEstablished()) {
     out_frame[0] = PUSH_CODE_CONTACTS_FULL;
     _serial->writeFrame(out_frame, 1);
   }
 }
 
 void MyMesh::onDiscoveredContact(ContactInfo &contact, bool is_new, uint8_t path_len, const uint8_t* path) {
-  if (_serial->isConnected()) {
+  if (_serial->isSessionEstablished()) {
     if (is_new) {
       writeContactRespFrame(PUSH_CODE_NEW_ADVERT, contact);
     } else {
@@ -511,7 +511,7 @@ void MyMesh::queueMessage(const ContactInfo &from, uint8_t txt_type, mesh::Packe
   i += tlen;
   addToOfflineQueue(out_frame, i);
 
-  if (_serial->isConnected()) {
+  if (_serial->isSessionEstablished()) {   // an idle listener still gets the tickle
     uint8_t frame[1];
     frame[0] = PUSH_CODE_MSG_WAITING; // send push 'tickle'
     _serial->writeFrame(frame, 1);
@@ -522,7 +522,7 @@ void MyMesh::queueMessage(const ContactInfo &from, uint8_t txt_type, mesh::Packe
   bool should_display = txt_type == TXT_TYPE_PLAIN || txt_type == TXT_TYPE_SIGNED_PLAIN;
   if (should_display && _ui) {
     _ui->newMsg(path_len, from.name, text, offline_queue_len);
-    if (!_serial->isConnected()) {
+    if (!_serial->isSessionEstablished()) {   // an idle listener gets the push instead
       _ui->notify(UIEventType::contactMessage);
     }
   }
@@ -633,7 +633,7 @@ void MyMesh::onChannelMessageRecv(const mesh::GroupChannel &channel, mesh::Packe
   i += tlen;
   addToOfflineQueue(out_frame, i);
 
-  if (_serial->isConnected()) {
+  if (_serial->isSessionEstablished()) {   // an idle listener still gets the tickle
     uint8_t frame[1];
     frame[0] = PUSH_CODE_MSG_WAITING; // send push 'tickle'
     _serial->writeFrame(frame, 1);
@@ -681,7 +681,7 @@ void MyMesh::onChannelDataRecv(const mesh::GroupChannel &channel, mesh::Packet *
   }
   addToOfflineQueue(out_frame, i);
 
-  if (_serial->isConnected()) {
+  if (_serial->isSessionEstablished()) {   // an idle listener still gets the tickle
     uint8_t frame[1];
     frame[0] = PUSH_CODE_MSG_WAITING; // send push 'tickle'
     _serial->writeFrame(frame, 1);
@@ -858,7 +858,7 @@ void MyMesh::onControlDataRecv(mesh::Packet *packet) {
   memcpy(&out_frame[i], packet->payload, packet->payload_len);
   i += packet->payload_len;
 
-  if (_serial->isConnected()) {
+  if (_serial->isSessionEstablished()) {
     _serial->writeFrame(out_frame, i);
   } else {
     MESH_DEBUG_PRINTLN("onControlDataRecv(), data received while app offline");
@@ -878,7 +878,7 @@ void MyMesh::onRawDataRecv(mesh::Packet *packet) {
   memcpy(&out_frame[i], packet->payload, packet->payload_len);
   i += packet->payload_len;
 
-  if (_serial->isConnected()) {
+  if (_serial->isSessionEstablished()) {
     _serial->writeFrame(out_frame, i);
   } else {
     MESH_DEBUG_PRINTLN("onRawDataRecv(), data received while app offline");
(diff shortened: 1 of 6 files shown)


IoTThinks:
I will test this PR.
Serial in Heltec v4 USB companion seems broken.

notbucki:
Pushed two follow-up commits fixing defects found in a cross-model adversarial review of this PR's own flow-control code:

1. **Flow control deadlocked transports with TX buffers under 179 bytes.** `isWriteBusy()` paces against `MAX_FRAME_SIZE+3` = 179 bytes of free TX space, but classic ESP32 UART bridges cap at 128 (HW FIFO, no TX ring), ESP32-S3 TinyUSB CDC (`ARDUINO_USB_MODE=0`) at 64, and STM32 (wio-e5) serial at 63 — on those boards the paced contact sync never emitted and `writeFrame()` dropped every frame larger than the buffer. Fixed with a runtime high-water probe of `availableForWrite()`: pacing and whole-frame drops only engage once the buffer has proven it can hold a max-size frame; small-buffer transports keep exactly the unpaced writes they had before this PR. HWCDC (4KB) and nRF52/RP2040 (256) are unchanged.

2. **HWCDC client activity survived the USB session.** `_last_frame_ms` was never cleared, so a re-plug within the idle window counted as "client connected" before the new session sent a frame. Fixed with a `resetActivity()` (clears timestamp, parser state and unread RX bytes), a level-triggered clear while the link is down, and a bus-reset event hook for re-enumerations that begin and end inside one long loop iteration. Documented trade-off: host suspend reads as link-down (no DTR on HWCDC), so after resume the client must send one frame before mirrored pushes restart — failing toward "disconnected" keeps device-side notifications and the BLE pairing PIN working.

Verified by building representatives of each family (heltec_v4/HWCDC, Heltec_v2/UART-bridge, Station_G2/TinyUSB-S3, wio-e5/STM32, nRF52 control); size cost is single-digit bytes of RAM and tens of bytes of flash (~3KB flash on HWCDC targets for the event hook).

This update was analyzed and implemented with Claude Code (agent); a human reviewed and approved the changes. Hardware testing on real boards is pending on our side.

---

*Edited 2026-09-05: a further adversarial review of this branch found four defects in it, including one that broke three existing builds (`Heltec_v3`, `Heltec_Wireless_Paper` and `Heltec_WSL3` set `ARDUINO_USB_MODE=1` without CDC-on-boot). All four are fixed, the branch has been rebased onto current `dev`, and the bus-reset hook this comment's second item relied on has been removed in favour of leaving the parser alone. The pull request description carries the details and the remaining known limits.*

notbucki:
@IoTThinks Thanks for offering to test, and sorry for the silence on this thread. The branch was reworked on 2026-09-05, but a force-push and an edited description do not trigger a GitHub notification, so you would not have heard about it. The current head is `2764e892`, rebased onto current `dev`; the PR description lists everything that changed and the known limits.

For a Heltec V4 (`heltec_v4_companion_radio_usb`: USB-Serial-JTAG/HWCDC, the same USB path as the V4-R8 the description talks about), this is what differs from the version you looked at on 2026-08-18:

* The 4 KB TX buffer is now sized **before** `Serial.begin()`. Resizing the live HWCDC buffer could hand a TX-empty interrupt freed memory.
* Link state is debounced (500 ms), and a confirmed loss retires the client's activity mark. After a re-plug or a host resume the device starts as "no client" until the client sends one frame.
* Push traffic (new-message notifications, mirrored frames) is gated on "link up and at least one frame seen" instead of the ten minute activity window, so a client that only listens keeps receiving.
* The bus-reset hook and RX drain from the 2026-08-31 version are gone again. On link loss the receive parser is left alone, exactly as it is without this PR.

Everything else on HWCDC is as described: activity-based connection state with the ten minute idle window, whole-frame writes and pacing against the 4 KB buffer, 5 ms write timeout.

`heltec_v4_companion_radio_usb` and `heltec_v4_tft_companion_radio_usb` build cleanly at the current head. On-device testing of this head is still pending on our side; our V4-R8 units run an earlier build of this series.

Could you say which side "serial seems broken" was on: plain `dev` at the time (the symptom this PR targets is the contact sync stalling and desynchronising on a device with a larger contact list), or this branch as of 2026-08-18? If the branch: which client (app over USB, `meshcore-cli`, web serial), what you saw (no reply at all, or a reply that stopped part way), and whether it persists on the current head. That would tell us whether one of the fixed defects explains it or whether something is still wrong.

*Written with Claude Code (Claude Fable 5.1), which also did the analysis above; a human reviewed and approved this comment before it was posted. The reviews mentioned were run with Codex (`gpt-6-astra`, reasoning effort `max`).*

PR #3241, opened 2026-08-18
Implement cold sleep for onboard SX1262 on RAK4630 when 13302 booster board is installed and report sleep state

agessaman:
## feat(rak3401): cold-sleep the RAK4631's onboard SX1262 on RAK13302 builds

A RAK4631 core paired with a RAK13302 leaves the core's own SX1262 powered but never initialized—it idles in `STDBY_RC` for the life of the board. Firmware can't cut its supply (P1.05 is the RF switch rail, not the radio's), so this brings it up on its own pins just long enough to issue a cold sleep, before `std_init()` repins SPI onto the RAK13302.

Gated behind `RAK4631_ONBOARD_SX1262_SLEEP` and exposed through seven `RAK_4631_13302_*` envs. RAK_3401_*` builds are untouched.

### Verified on RAK19007 + RAK4631 + RAK13302
- `begin=0 sleep=0` — chip present, cold sleep accepted
- 5/5 reboots repeatable; no spurious wake across a 40 min soak
- RAK13302 unaffected: `device_errors=0x0000`, TX and RX both confirmed
- RAK3401 artifact unchanged (preprocessed compiler input byte-identical)
- All seven envs build

### Not verified
The current saving is **datasheet-derived, not measured**: 600 µA `STDBY_RC` → 160 nA cold sleep. That's ~0.5 mA at a 5 V input, below what was resolvable here—the WisBlock LiPo charger dominates any USB-input reading. No current reduction is claimed.

### Notes
- The sleep doesn't survive an MCU reset (GPIOs revert, NSS falls, chip wakes); the helper re-sleeps it each boot, so it's self-healing.
- BUSY reads low in `STDBY_RC` and high while asleep, making the state observable—without it a failed sleep would be silent.

### DIFF
--- a/variants/rak3401/platformio.ini
+++ b/variants/rak3401/platformio.ini
@@ -134,3 +134,128 @@ build_src_filter = ${rak3401.build_src_filter}
 extends = rak3401
 build_src_filter = ${rak3401.build_src_filter}
   +<../examples/kiss_modem/>
+
+; ------------------------- RAK4631 core + RAK13302 -------------------------
+; Same RF hardware as the RAK3401 1W kit, but the RAK4631 core carries a second,
+; unused SX1262 that would otherwise idle in standby forever. These envs match
+; the RAK_3401_* ones and additionally cold-sleep that radio at boot.
+; Do not flash them to a genuine RAK3401 -- there is no onboard radio to sleep.
+
+[rak4631_13302]
+extends = rak3401
+board = rak4631
+build_flags = ${rak3401.build_flags}
+  -D RAK4631_ONBOARD_SX1262_SLEEP
+build_src_filter = ${rak3401.build_src_filter}
+lib_deps = ${rak3401.lib_deps}
+
+[env:RAK_4631_13302_repeater]
+extends = rak4631_13302
+build_flags =
+  ${rak4631_13302.build_flags}
+  -D DISPLAY_CLASS=SSD1306Display
+  -D ADVERT_NAME='"RAK4631 1W Repeater"'
+  -D ADVERT_LAT=0.0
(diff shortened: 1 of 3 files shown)


agessaman:
I implemented this with new envs for discoverability—a user with a 4631 wouldn't necessarily go looking for a 3401 firmware and know that it would work. But it could be implemented as a conditional check on the RAK3401 initialization.

dmwarren:
Incredible work—thank you so much. I crammed a RAK13302 into a WisMesh Pocket v2 with the stock 3200mAh battery and…
Battery life without this patch: 18 hours
Battery life with this patch: 50 hours

jaas666:
Great work. I guess, just like me, a lot of users don't even know about the 3401. I've always thought the 13302 was an "upgrade" to the 4631.

Today I accidentally flashed the 3401+13302 firmware without even realizing it. Luckily it just works but then I found about the onboard SX1262 being left on.

I assumed it worked exactly like this PR. It would detect the 13302 and switch to using that.

Can't wait for this to be merged. I love this community.

agessaman:
I've been running this without issue on a very busy live repeater. @ripplebiz, I'm happy to make changes to this if you would prefer this to be a conditional check on the 3401 variant instead of a separate variant.

jaas666:
Something to consider. I used the 4631+13302 on mine because it was the only option that said RAK and 1W. I didn't even know there was a 3401, nor did I pay attention when selecting it.

I think both options should be a good idea: a new variant for 4631+13302 and a conditional check on the 3401+13302.

Having both options should prevent someone from making the same mistake I made, which luckily didn't cause any issue.

Issue #2469, opened 2026-05-04
Raspberry Pi 5 VSCode Compile Error


KPrivitt:
I am not a platformio.ini expert, but I found that I could not compile the MeshCore firmware on my new Raspberry Pi OS based system using VSCode.  The Raspberry Pi uses an ARM core processor.  I do not have any issues with my Windows 11 Intel CPU based system.

I received this error:
Tool Manager: Installing platformio/toolchain-gccarmnoneeabi @ >=1.60301.0,<1.80000.0
UnknownPackageError: Could not find the package with 'platformio/toolchain-gccarmnoneeabi @ >=1.60301.0,<1.80000.0' requirements

I searched for this and I could not find it in the Git Hub code base, so I was not sure why it is loading this.  The only place I found that "toolchain" existed was in a comment and not in the code.

It turns out this is an older version that is apparently no longer available and that is why it could not be found.

The solution was to add:  toolchain-gccarmnoneeabi @ 1.400201.0 to the platformio.ini file.

[nrf52_base]
platform_packages =
toolchain-gccarmnoneeabi @ 1.1400201.0

VSCode then downloaded it and the MeshCore firmware would now compile with no errors.

**Should this line be added to the main GitHub MeshCore codes platformio.ini file?**

Or do I just need to add this each time to my local copy that I clone to my Raspberry Pi based system.

KPrivitt:
I did some additional research and found that:
Windows PlatformIO chose GCC 9.x because that’s the default toolchain bundled with the nordicnrf52 platform (I'm not sure where this is...) on the Windows version.

Raspberry Pi’s PlatformIO tries to install toolchain-gccarmnoneeabi @ >=1.60301.0,<1.80000.0
That version range no longer exists in the registry
PlatformIO on ARM Linux simply does not ship the GCC 9.x toolchain anymore.

Both are "compatible" with MeshCore code, but I found one incompatibility with my code:

> **time_t is now undefined.**

This is because GCC 14.x:

Implements C++20 and large parts of C++23.
Much stricter header hygiene.
Requires explicit includes for:
   `<cstdint>`
   `<cstring>`
   `<cstdlib>`
   **<time.h> (your case)**
More aggressive constexpr evaluation.
More aggressive dead‑code elimination.

If MeshCore explicitly adds the GCC 14.x toolchain it will be compatible with both versions of VSCode.

[...]

PR #3246, opened 2026-08-19
Update GCC toolchain for nRF52 boards to fix builds on Raspberry Pi

liamcottle:
This PR updates the GCC toolchain for nRF52 boards to `^1.140201.0`.

Without this change, attempting to compile firmware on a Raspberry Pi 4B or 5 results in the following error:

```
Tool Manager: Installing platformio/toolchain-gccarmnoneeabi @ >=1.60301.0,<1.80000.0
UnknownPackageError: Could not find the package with 'platformio/toolchain-gccarmnoneeabi @ >=1.60301.0,<1.80000.0' requirements
```

Updating the toolchain resolves this error, and allows the firmware to compile successfully.

This however, introduces a few warning messages about unimplemented methods. Screenshots can be seen in the discussion on issue https://github.com/meshcore-dev/MeshCore/issues/2469.

This PR includes a new `armstubs.cpp` file, which weakly implements the unimplemented methods, which silences those warnings during compilation. As far as I'm aware, none of those methods are used in the code base.

I can now successfully compile firmware on my Raspberry Pi 4B.

Closes https://github.com/meshcore-dev/MeshCore/issues/2469

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -86,6 +86,7 @@ platform_packages =
   ; https://github.com/meshcore-dev/MeshCore/pull/1177
   ; https://github.com/meshcore-dev/MeshCore/pull/1295
   framework-arduinoadafruitnrf52 @ https://github.com/meshcore-dev/Adafruit_nRF52_Arduino#d541301
+  platformio/toolchain-gccarmnoneeabi@^1.140201.0
 extra_scripts = create-uf2.py
 build_flags = ${arduino_base.build_flags}
   -D NRF52_PLATFORM
--- a/src/armstubs.cpp
+++ b/src/armstubs.cpp
@@ -1,0 +1,9 @@
+// https://github.com/meshcore-dev/MeshCore/issues/2469
+// armstubs.cpp exists to silence warnings introduced after upgrading platformio/toolchain-gccarmnoneeabi
+extern "C" __attribute__((weak)) int _close(int fd)  { return -1; }
+extern "C" __attribute__((weak)) int _lseek(int fd, int offset, int whence) { return 0; }
+extern "C" __attribute__((weak)) int _read(int fd, char *buf, int len)  { return 0; }
+extern "C" __attribute__((weak)) int _fstat(int fd, struct stat *st) { return 0; }
+extern "C" __attribute__((weak)) int _isatty(int fd) { return 1; }
+extern "C" __attribute__((weak)) int _getpid(void) { return 1; }
+extern "C" __attribute__((weak)) int _kill(int pid, int sig) { return -1; }


KPrivitt:
I reverted back to dev, verified it would not compile. Fetched #3246 and it compiled with SUCCESS.

Thanks! It's working.

KPrivitt:
FYI,
I reverted back to dev and I manually put in just the toolchain with @1.140201.0 (not the arm stubs.coo) and it does compile but has the warnings you saw.

However, If I do not put in the specific version (after gccarmnoneeabi) it does not get any warnings and compiles with no errors.

liamcottle:
Thanks for confirming.

> However, If I do not put in the specific version (after gccarmnoneeabi) it does not get any warnings and compiles with no errors.

My guess is it ends up using some older cached version on your machine, which doesn't have those warnings enabled. But we might as well migrate to the latest if we're upgrading.

YoshiWalsh:
This seems to break builds for RAK3401 and RAK4631. Building the current dev commit (#65aa11) yields the following: 

```
Linking .pio/build/RAK_4631_repeater/firmware.elf
/home/yoshi/.platformio/packages/toolchain-gccarmnoneeabi/bin/../lib/gcc/arm-none-eabi/14.2.1/../../../../arm-none-eabi/bin/ld: .pio/build/RAK_4631_repeater/variants/rak4631/target.cpp.o: in function `MicroNMEALocationProvider::MicroNMEALocationProvider(Stream&, mesh::RTCClock*, int, int, RefCountedDigitalPin*)':
/mnt/d/git/MeshCore/src/helpers/sensors/MicroNMEALocationProvider.h:63:(.text._ZN25MicroNMEALocationProviderC2ER6StreamPN4mesh8RTCClockEiiP20RefCountedDigitalPin[_ZN25MicroNMEALocationProviderC5ER6StreamPN4mesh8RTCClockEiiP20RefCountedDigitalPin]+0x88): undefined reference to `vtable for LocationProvider'
/home/yoshi/.platformio/packages/toolchain-gccarmnoneeabi/bin/../lib/gcc/arm-none-eabi/14.2.1/../../../../arm-none-eabi/bin/ld: .pio/build/RAK_4631_repeater/libFrameworkArduino.a(Uart.cpp.o): in function `Uart::Uart(NRF_UARTE_Type*, IRQn_Type, unsigned char, unsigned char)':
/home/yoshi/.platformio/packages/framework-arduinoadafruitnrf52/cores/nRF5/Uart.cpp:48:(.text._ZN4UartC2EP14NRF_UARTE_Type9IRQn_Typehh+0x64): undefined reference to `vtable for HardwareSerial'
collect2: error: ld returned 1 exit status
*** [.pio/build/RAK_4631_repeater/firmware.elf] Error 1
```

Removing the manual specification of `toolchain-gccarmnoneeabi` (and thus reverting to 7.2.1) removes this error.

EDIT: I think it's worth noting the `toolchain-gccarmnoneeabi` package squeezes the entire gcc toolchain version into the "minor" version field.  Since the `^` version prefix in a platformio dependency declaration indicates that it allows minor & patch version upgrades of that package, this means that using `^` allows for gcc major version updates. I'm not sure this is wise.

EDIT2: I have done some further testing, and 1.90301.200702 is the newest available version that builds successfully. 1.100301.220327 and newer produce this error.

PR #3252, opened 2026-08-20
feat(cli): doctor diagnostics commands

benallfree:
This PR adds a `doctor` CLI command for repeaters to perform on-device diagnostics. I was hitting LittleFS errors and needed more insight. 

| Command | Remote admin | Description |
|---------|--------------|-------------|
| `doctor stat` | No (USB) | Partition free/block headroom (`FS_STAT` on serial) |
| `doctor gc` | Yes | Remove common cruft (`/packet_log`, legacy prefs paths, doctor temp files) |
| `doctor check` | Yes | Probe atomic prefs write, report prefs/id/acl/regions presence |
| `doctor ls` | No (USB) | Recursive file listing (`FS_LS` on serial) |
| `doctor probe` | No (USB) | Write-size probe + prefs JSON write test (`FS_PROBE` on serial) |
| `doctor dump` | No (USB) | Hex dump raw flash region backing InternalFS |

Also appends `(try: doctor gc)` hints to the FS save error replies, and upgrades `erase` to format+remount when the FS callback is available.

**Stacked on #3253** (FS error replies), which stacks on #3254 (atomic saves) — merge those first, then rebase this onto `dev`. The first two commits here are theirs.


benallfree:
@dotspencer Curious if you have any feedback on the `doctor dump` that does a hex dump of the partition. Would that have helped you with https://github.com/meshcore-dev/MeshCore/pull/3254#issuecomment-5372144983 in any meaningful way?

PR #3261, opened 2026-08-20
Fix LR11x0 and LR2021 packet loss when a stale SPI reply is read as the length

fkallay1:
**Type: bug**

### Symptom

On an LR2021 repeater, the raw receive log occasionally reported a length of 4 for
frames that were 38, 50, 126 or 133 bytes long - always exactly 4, regardless of the
real length. The frames were then rejected as corrupt and never reached the mesh
layer. A second repeater on the same channel, an SX1262 board, logged the same frames
with correct lengths at the same moment, so it was not an RF problem.

It came in bursts. When several repeaters re-flood the same packet at once and the
chip is busy, roughly a third of the frames were lost this way. In quiet periods it
did not happen for hours.

The loss is silent: `readData()` returns success, so it is not counted as a receive
error, only as a frame that was received and then thrown away.

### Cause

On the LR11x0 and LR2021 families a read command is two SPI transactions
(`LRxxxx::SPIcommand()`): send the opcode, then read the reply.
`Module::SPItransferStream()` waits `delayMicroseconds(1)` after the first
transaction and then polls BUSY for a low level - so if BUSY has not risen within
that microsecond, the wait is skipped and the reply is read before the chip has
prepared it.

The chip then sends its default `[stat 2B][irq 4B]` stream, and the length is parsed
straight out of that. The status byte is valid, so nothing reports an error.

On LR2021 the length is built from two bytes, so `getRxPktLength()` returns
`irq[31:16]`. With RX_DONE (bit 18) set that is exactly 4 - the observed value. With
TX_DONE or CRC_ERROR also set it becomes 12 or 68, which is plausible enough to pass
as a real length and produce a garbage packet instead of an obviously short one.

On LR11x0 the length is a single byte, so `getRxBufferStatus()` returns `irq[31:24]`,
and every RX-relevant flag of that family lives in the two low bytes - so the length
comes back as **0**. `recvRaw()` skips the packet on its `len > 0` test, and the frame
is dropped when Rx is re-armed. No log line, no error counter, nothing but a gap in
the `isr` versus `recv` counts.

`getIrqStatus()` cannot be caught by this race, because it *is* that default stream.
That also makes the top bytes of the IRQ word an exact fingerprint of a stale reply,
which is what the fix uses.

### Fix

The chip already says whether a reply is on its way: the command status field of stat1
is `CMD_DAT` ("successfully processed, data is being transmitted") for the read half of
a get, and `CMD_OK` ("nothing to collect") when the status stream comes back instead.
RadioLib consumes that byte internally and only rejects `CMD_FAIL` and `CMD_PERR`, so
the distinction is lost. Reading the length here directly keeps the status byte visible
- the same trick `LRxxxx::getIrqStatus` uses to read the IRQ word - and the length is
trusted only when the status says the reply is ours. The Rx FIFO is still intact at that
point, so re-reading recovers the frame; if it never settles, the plain library read is
used so this can never end up worse.

Keeping the status byte is the only part that needs care. Setting the status width to 0
tells the transfer layer there is nothing to strip, so the whole reply lands in our own
buffer:

```cpp
mod->SPIwriteStream(RADIOLIB_LR2021_CMD_GET_RX_PKT_LENGTH, NULL, 0, wait, false);
mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_STATUS] = Module::BITS_0;
mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_CMD]    = Module::BITS_0;
mod->SPIreadStream(RADIOLIB_LRXXXX_CMD_NOP, buff, sizeof(buff), wait, false);
// buff[0..1] = status word, buff[2..3] = the length
```

The guard itself is then just the status test:

```cpp
size_t getPacketLength(bool update = true) override {
  uint8_t  stat = 0;
  uint16_t val  = 0;
  for (int i = 0; i < 3; i++) {
    readRxPktLenWithStatus(true, &stat, &val);
    if ((stat & 0x0E) == RADIOLIB_LRXXXX_STAT_1_CMD_DAT) return val;
  }
  return LR2021::getPacketLength(update);   // never worse than the plain read
}
```

An earlier version of this compared the returned length against `irq[31:16]` instead,
which needs no status byte. That turned out to misfire: 68 is both a common frame length
here and what `irq[31:16]` reads when RX_DONE and CRC_ERROR are set together, so it
flagged genuine frames. The status reported `CMD_DAT` on exactly those, which is why it
decides now.

On LR11x0 the same check applies, for the same reason. The first version of this used
`len == 0 && RX_DONE` there instead, on the assumption that a zero length is unambiguous
on that family. Hardware says otherwise: on a T1000-E that state occurs 25 times in three
minutes, roughly one per two and a half received frames, and re-reading recovers nothing
in any of them - 0 out of 25. So the zero is usually honest, and only the status can tell
it apart from a stale reply. The existing `len == 0 && HEADER_ERR` handling is left
untouched.

Both overrides use only public API and protected members of the base class, so no
`RADIOLIB_GODMODE` is needed.

### What this does not cover

This guards the path through `recvRaw()`. `LR11x0::readData()` reads the length and
the buffer offset again through the two-argument
`getPacketLength(bool update, uint8_t* offset)`, which is not virtual and therefore
cannot be intercepted from a subclass. If the race is lost there, the payload comes
out empty, or shifted by a bogus offset - which may well be the "packets shifted"
symptom the existing comment in `CustomLR1110::getPacketLength()` refers to.

The same race also affects every other read on these chips - `getRSSI()`, `getSNR()`,
`getRssiInst()`, `getVbat()`, `getTemp()` - so a wrong RSSI or SNR can reach
`packetScore()` without any indication. Those cannot be guarded from here either,
because a wrong RSSI has no fingerprint to test against. A proper fix belongs in the
driver's transport layer.

That part has been reported to RadioLib as an issue
(https://github.com/jgromes/RadioLib/issues/1857), with the reproduction and the
measurements above - not as a patch, since where the check belongs is a decision about
the transport layer of the whole LR11x0/LR2021 family. If the driver starts rejecting a
reply whose status is not `CMD_DAT`, every get command is covered at once and this guard
becomes redundant. It would still be harmless - our own read never goes through
`SPIcommand()` - so it can simply be dropped then, though not before the pinned RadioLib
is bumped: the pin currently sits on a commit that predates any such fix, so a driver-side
fix changes nothing here until then. Until that happens this guard is what actually keeps
the frames.

### Affected boards

LR2021: `meshtracker_x1`, `meshnology_w12`.

LR11x0: `t1000-e`, `wio_wm1110`, `thinknode_m3`, `thinknode_m7`, `thinknode_m9`,
`minewsemi_me25ls01`.

### Testing

LR2021 on hardware: Seeed XIAO nRF52840 with a NiceRF LoRa2021 module, 869.618 MHz,
SF7, BW 62.5, in a live mesh with several repeaters, against a second receiver (SX1262)
on the same channel as a reference.

The failure is a timing race, so rather than wait for it, it was forced: a debug build
skips the BUSY wait on every fourth length read, in the real receive path.

- 11 of 11 sabotaged reads returned exactly 4 - the original symptom - and every one of
  them reported `CMD_OK`.
- The re-read returned the true length every time (63, 65, 67, 73, ...).
- Over the same window the node logged the same frames as the reference receiver, with
  no bogus lengths and no receive errors, although a quarter of its length reads had
  been broken on purpose.
- On genuine frames whose length happened to equal `irq[31:16]`, the status reported
  `CMD_DAT` and the guard correctly stayed out of the way.

LR11x0 verified on hardware as well, on a Seeed T1000-E on the same channel, the same
way - the BUSY wait skipped on every fourth length read:

- 4 of 4 sabotaged reads came back as 0 and reported `CMD_OK`; the re-read returned the
  true length each time, recovering 84, 20 and 196 byte frames.
- Genuine reads reporting a length of 0 - 25 of them in three minutes, with the IRQ word
  at 0x38, 0x78 (header error) or 0xB8 (CRC error) - all reported `CMD_DAT`, and a probe
  doing three extra reads on each recovered nothing. That is what ruled out the
  value-based rule for this family.

Builds clean: `t1000e_repeater`, `wio_wm1110_repeater`, `MeshTracker_X1_repeater`.

### DIFF
--- a/src/helpers/radiolib/CustomLR1110.h
+++ b/src/helpers/radiolib/CustomLR1110.h
@@ -13,8 +13,48 @@ class CustomLR1110 : public LR1110 {
   public:
     CustomLR1110(Module *mod) : LR1110(mod) { }
 
+    // Two SPI transactions with the status word kept - mirror of the LR2021 helper.
+    int16_t readRxPktLenWithStatus(bool wait, uint8_t* stat, uint16_t* val, uint8_t* off = NULL) {
+      int16_t st = mod->SPIwriteStream(RADIOLIB_LR11X0_CMD_GET_RX_BUFFER_STATUS, NULL, 0, wait, false);
+      Module::BitWidth_t sw = mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_STATUS];
+      Module::BitWidth_t cw = mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_CMD];
+      mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_STATUS] = Module::BITS_0;
+      mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_CMD]    = Module::BITS_0;
+      uint8_t buff[4] = { 0 };
+      st = mod->SPIreadStream(RADIOLIB_LRXXXX_CMD_NOP, buff, sizeof(buff), wait, false);
+      mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_STATUS] = sw;
+      mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_CMD]    = cw;
+      if (stat) *stat = buff[0];
+      if (val)  *val  = buff[2];
+      if (off)  *off  = buff[3];
+      return st;
+    }
+
+    // Guard against a stale SPI reply being parsed as the received length. A "get" is
+    // two transactions (LRxxxx::SPIcommand): send the opcode, then read the reply.
+    // SPItransferStream() waits 1 us before polling BUSY, so if BUSY has not risen yet
+    // the reply is read too early and the chip answers with its default [stat 2B][irq 4B]
+    // stream. getRxBufferStatus() then takes the length from irq[31:24] and the buffer
+    // offset from irq[23:16] - and that offset is what shifts a payload.
+    //
+    // The value cannot decide this on its own here: a length of 0 while RX_DONE is set is
+    // a state the chip reaches routinely. Measured on a T1000-E, 25 times in three
+    // minutes - roughly one per two and a half received frames - with the IRQ word
+    // reading 0x38, 0x78 (header error) or 0xB8 (CRC error). Re-reading recovers nothing
+    // there, 0 out of 25, so treating every zero as suspect only burns SPI reads. The
+    // status separates the two cleanly: those reads report CMD_DAT, while a reply read
+    // too early reports CMD_OK and re-reading then returns the true length - 4 of 4 with
+    // the BUSY wait skipped on purpose, recovering 84, 20 and 196 byte frames.
     size_t getPacketLength(bool update) override {
-      size_t len = LR1110::getPacketLength(update);
+      uint8_t  stat = 0;
+      uint16_t val  = 0;
+      size_t   len  = 0;
+      for (int i = 0; i < 3; i++) {
+        readRxPktLenWithStatus(true, &stat, &val);
+        len = val;
+        if ((stat & 0x0E) == RADIOLIB_LRXXXX_STAT_1_CMD_DAT) break;
+        if (i == 2) len = LR1110::getPacketLength(update);   // never worse than the plain read
+      }
       if (len == 0 && getIrqStatus() & RADIOLIB_LR11X0_IRQ_HEADER_ERR) {
         // we've just received a corrupted packet
         // this may have triggered a bug causing subsequent packets to be shifted
--- a/src/helpers/radiolib/CustomLR2021.h
+++ b/src/helpers/radiolib/CustomLR2021.h
@@ -75,6 +75,54 @@ class CustomLR2021 : public LR2021 {
       return LR2021::startReceive(RADIOLIB_LR2021_RX_TIMEOUT_INF, RADIOLIB_IRQ_RX_DEFAULT_FLAGS | (1UL << RADIOLIB_LR2021_IRQ_PREAMBLE_DETECTED), RADIOLIB_IRQ_RX_DEFAULT_MASK, 0);
     }
 
+    // Read the received length ourselves, keeping the status word.
+    // A "get" on this family is two SPI transactions (LRxxxx::SPIcommand): send the
+    // opcode, then read the reply. SPItransferStream() waits 1 us before polling
+    // BUSY, so when BUSY has not risen yet the reply is read too early and the chip
+    // answers with its default [stat 2B][irq 4B] stream instead. RadioLib strips the
+    // status and hands the rest back as data, so getRxPktLength() returns irq[31:16]
+    // - exactly 4 with RX_DONE set. Setting the status width to 0 keeps both status
+    // bytes in our own buffer, the same trick LRxxxx::getIrqStatus uses.
+    int16_t readRxPktLenWithStatus(bool wait, uint8_t* stat, uint16_t* val) {
+      int16_t st = mod->SPIwriteStream(RADIOLIB_LR2021_CMD_GET_RX_PKT_LENGTH, NULL, 0, wait, false);
+      Module::BitWidth_t sw = mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_STATUS];
+      Module::BitWidth_t cw = mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_CMD];
+      mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_STATUS] = Module::BITS_0;
+      mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_CMD]    = Module::BITS_0;
+      uint8_t buff[4] = { 0 };
+      st = mod->SPIreadStream(RADIOLIB_LRXXXX_CMD_NOP, buff, sizeof(buff), wait, false);
+      mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_STATUS] = sw;
+      mod->spiConfig.widths[RADIOLIB_MODULE_SPI_WIDTH_CMD]    = cw;
+      if (stat) *stat = buff[0];
+      if (val)  *val  = ((uint16_t)buff[2] << 8) | (uint16_t)buff[3];
+      return st;
+    }
+
+    // Trust the length only when the chip says the reply is ours. The command status
+    // field of stat1 has four values and CMD_DAT means "successfully processed, data
+    // is being transmitted" - the only correct one for the read half of a get. A
+    // reply produced by the race reports CMD_OK instead, i.e. "nothing to collect",
+    // which is exactly the case where the status stream comes back. The Rx FIFO is
+    // still intact at this point (readData() runs later), so re-reading recovers the
+    // frame instead of losing it.
+    // Measured on hardware: with the BUSY wait skipped on purpose in the live RX
+    // path, 11 of 11 bogus reads reported CMD_OK and every frame was recovered; on
+    // genuine frames whose length happened to equal irq[31:16] the status correctly
+    // reported CMD_DAT. Judging by that value alone - as an earlier version did -
+    // therefore misfires on real frames, which is why the status decides here.
+    size_t getPacketLength(bool update = true) override {
+      uint8_t  stat = 0;
+      uint16_t val  = 0;
+      for (int i = 0; i < 3; i++) {
+        readRxPktLenWithStatus(true, &stat, &val);
+        if ((stat & 0x0E) == RADIOLIB_LRXXXX_STAT_1_CMD_DAT) return val;
+      }
+      // never end up worse than the plain library read
+      return LR2021::getPacketLength(update);
+    }
+
+
+
     bool isReceiving() {
       uint32_t irq = getIrqStatus();
       bool preamble = irq & RADIOLIB_LR2021_IRQ_PREAMBLE_DETECTED;  // bit 5


fkallay1:
Updated after testing this on an LR11x0 board as well, and it changed the LR11x0 half of
the patch - worth flagging rather than leaving it in the force-push.

The first version guarded LR11x0 with `len == 0 && RX_DONE`, on the assumption that a
zero length cannot be genuine there, so no status byte was needed. That assumption does
not hold. On a T1000-E that state occurs 25 times in three minutes, roughly one per two
and a half received frames, with the IRQ word reading 0x38, 0x78 (header error) or 0xB8
(CRC error). A probe doing three extra reads on each of them recovered nothing, 0 out of
25 - so the zero is usually honest, and the old rule would have burned three SPI reads
every time without saving a frame.

Both chips now use the same check, and it holds on both: with the BUSY wait skipped on
purpose, 4 of 4 sabotaged reads on the T1000-E came back as 0 reporting CMD_OK and the
retry recovered 84, 20 and 196 byte frames, while the genuine zeros above all reported
CMD_DAT. Same shape as the 11 of 11 measured on LR2021.

`t1000e_repeater` and `MeshTracker_X1_repeater` build clean.

oltaco:
Nice work, have you opened an issue/PR at RadioLib?

edit: I see you've already brought it to @jgromes attention (https://github.com/jgromes/RadioLib/issues/1857) and it looks like a fix is already in the works upstream (https://github.com/jgromes/RadioLib/pull/1858).

It's probably best to just wait for a fix to land upstream in RadioLib rather than papering over the issue down here. I guess it applies to more than just this command anyway.

fkallay1:
Yes — https://github.com/jgromes/RadioLib/issues/1857. The description mentioned it only in words, which was not much use; I have put the link in there as well.

jgromes agreed and put the cause more sharply than I had: the read transaction does not wait for BUSY between sending the opcode and reading the reply, which the LR1110 datasheet requires. He has pushed a candidate fix to the `1857-lrxxxx-read-busy` branch and asked me to verify it on hardware without the SPI sabotage the measurements used. I will report back there once it has run in a live mesh.

What that means for this PR is in the description already: the proper fix belongs in the driver's transport layer, and once it lands this guard becomes redundant and can be dropped — but not before the pinned RadioLib commit is moved past it, and the pin currently sits on 11 July.

carlhodder:
This conversation really belongs to this so I'm adding it here rather the other ticket. 

I'm starting to be more confident that there is something here to be found, but not sure if CMD_DAT is it. I've been logging with some extra bits and code changes - I added a requirement for CMD_DAT on replies for these commands (logging and returning an error if CMD_OK is encountered):

- getRssiInst
- getPacketType
- getLoRaPacketStatus
- getRxPktLength

I haven't seen this occur in the logs yet, but I am still seeing things like:

> DEBUG: IRQ: 262512 (RX_TIMESTAMP, PREAMBLE, HEADER_VALID, SYNCWORD_VALID, HDR_TIMESTAMP, RX_DONE)
>  RAW: 1570D0CD444F880F121EA2036BF23ADE39A581F4E2E6A786301FAAE72EC8B768FC9A80C1C7F87A1F925E77E75CEDF5D886946721BE2F462CAA1BF4708AEDCBC291B4DC1AEDDE4ED88CE00E2A
> DEBUG:  Dispatcher::checkRecv(): partial or corrupt packet received, len=76

We have correct IRQ, the response was CMD_DAT, but the data out is not right. The next reception is good:

> DEBUG: IRQ: 262512
>  RAW: 1507D0CD444F880FDE1EA2036BF23ADE39A581F4E2E6A786301FAAE72EC8B768FC9A80C1C7F87A1F925E77E75CEDF5D886946721BE2F462CAA1BF4708AEDCBC291B4DC1AEDDE4ED88CE00E2A
> DEBUG:  Dispatcher::checkRecv(), score delay below threshold (0)

Or:

> DEBUG: IRQ: 262512
>  RAW: 587D9EABCE2B688E6312F2E05EA8F6DBE8DBAB4F73F2D55E4354
> DEBUG:  Dispatcher::checkRecv(): unsupported packet version

Again, no obvious issues I can see from IRQ. Radiolib basic debugging is enabled and nothing from there either. Next RX is good (different packet though):

> DEBUG: IRQ: 262512
>  RAW: 1541B64E596A7EABB29C7D417CD26AACB1D1677ADB3F2C
> DEBUG:  Dispatcher::checkRecv(), score delay below threshold (0)

I've dug around the RadioLib code and it doesn't jump out as incorrect BUSY handshaking - SPIwriteStream and SPIreadStream ultimately use SPItransferStream with waitForGpio which wraps the SPI transaction with a BUSY line check either side. We also shouldn't need to wait for it to rise as it should be high well before the SPI completes - the BUSY line is raised when the CS line is asserted so the process is:

- Check/wait till not BUSY
- Assert CS, (BUSY line goes high here)
- Transfter data
- Deassert CS
- Wait till BUSY line falls


While poking around I altered RadioLib readData so that the call to getPacketLength is immediately followed by getRxFifoLevel - it's intermittent but I do also see instances where getPacketLength reports data but getRxFifoLevel says there is nothing in the buffer?
e.g:

> DEBUG: getPacketLength differs from getRxFifoLevel: 126, 0

(Seen overnight with packet lengths 140, 133, 126, 74, 73, 61, 56, 49, 43, 28. getRxFifoLevel reports 0 in all instances)

I'd forgotted I'd altered to return an error code when these mismatched so I hadn't been logging what the packet would look like if you pulled the data anyway, ~~I changed that and captured one this morning:~~

**Correction**
The packet I had here previously was not correct - I added the hex output but was still returning before reading from the fifo. After correcting that I think I might have found the trail for the odd packets I mentioned in the RadioLib ticket that were just '05' repeated. I've seen packets made all of 15's, or 19's since, and this is what the next one looked like::

> getPacketLength: 130, getRxFifoLevel: 0
> DEBUG: IRQ: 262512
>  (-30) RAW: E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4E4
> DEBUG: RadioLibWrapper: error: readData(-30)
> 

NOTE: 

- I'm returning -30 myself, this isn't a code RadioLib returns here
- This is in addition to the packets failing checkRecv checks above, those packets report same packet length and fifo length

It could be my CMD_DAT code isn't right, I did reverse the check to require CMD_OK and it fails and logs so that suggests it's OK.


Something smells here, and it's starting to smell like an undocumented race condition :/



In the meantime I'm going to continue digging. I figure I might as well switch things around tonight to use one-shot RX mode so we always read the fifo/packet registers in standby mode - nothing async should write to the fifo registers in standy, so should help prove or exclude that idea.

jschrempp:
> Something smells here, and it's starting to smell like an undocumented race condition :/

There has been some chatter about something similar in Bay Area Meshcore Discord #general.

nullrouten0:
Some data from Bay Area... We've been measuring MeshCore corruption from the MQTT side on the mesh  ~2.2M relay transmissions over 33 days and it is similar to what you're both seeing.

  About 0.33% of relay transmissions corrupt the frame. Sounds small, but each message gets retransmitted ~50x, so ~18% of group text ends up damaged somewhere. Below 6 hops the mesh is basically clean (~0.6%) …  it's the amplification that hurts.
  
  And the amplifier is calculatePacketHash(): it deliberately (as designed) excludes the path, so one flipped bit is a brand-new hash, wasSeen() misses it, and every node that already forwarded the clean copy re-floods the corrupt one. On 282 messages we caught both ways, damaged copies reached a median of 17 hops vs 10 for their own clean twin, deeper in 93% of cases. So a corrupt frame isn't just unreadable, it costs the whole mesh airtime.

  The damage is also present before the radio CRC :  16-32 observers report byte-identical corrupt payloads, which independent RF errors can't produce. Shape is 1-2 flipped bits, ~83% in the last 10% of the frame. That doesn't match the failure mode here, so this PR isn't the fleet-wide cause. But path and length corruption are exactly what archive analysis can't see (everything matches on equal-length payloads), so receive-side faults like this one sit right in our blind spot  which is why the bench work in this thread is worth more than our numbers are.   
  
We did chase the SPI path on one bad relay of our own, Python-stack node (openhop) running ~34x the fleet rate with readback probes either side of the FIFO write, and caught it mangling 2.6% of its transmissions: always one contiguous 3-4 byte run, and 68 of 88 were an exact copy of bytes from elsewhere in the same frame, so displacement rather than bit flips (reads themselves were clean across 125k probe lines). Different stack and a different mechanism from this one, but it does confirm the MCU→FIFO window as a real place for a frame to break and whatever breaks there gets a valid CRC computed over it, so every receiver downstream accepts it.

carlhodder:
UPDATE: I think I'm honing in on the issue but there's still a part I can't explain. I figured I'd keep updating this on the off chance something here helps anyone else investigating!

I'm starting to think there may be a chip issue that is the root cause, but the result is also partly our fault as we're using continuous RX mode without thinking of races.

### Evidence/Hypothesis:

OK, so what I've discovered is that during these busy moments (bunch of simultaneous repeats including distant repeaters) we appear to get an RX_DONE interrupt that returns a valid packet length but there is no data in the buffer. This has two main fail modes:

**One-shot mode (TIMEOUT_NONE)**

This is where the module automatically returns to standby once a single RX has completed. And yes - we see a tiny number of RX_DONE interrupts in this mode with a packet length and (always) 0 getRxFifoLevel which surprised me (I initially assumed this was purely a race - it could still be, I don't know how their internal code does single and continuous modes).

The buffer level isn't checked in readData, and if you continue to read data out you get one character repeated, e.g.:

> 5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A5A

Which mostly get rejected by various checkRecv checks - but 05 and 15 repeated bytes make it out which was what was jumping out at me in the wild.


**Continuous RX mode (TIMEOUT_INF, the default mode)**

Firstly this mode seems to prove that it is not that we're reading the buffer too early. If I make this condition exit before reading or clearing the buffer the following RX_DONE event returns a match between packet length and fifo level - so no extra data from this previous event. Also it doesn't look like a 'bounce' - the timing between RX_DONE events suggests these should be valid RX events.

The one that jumped out at me was for the 21 instances I logged overnight there was one that went:

> getPacketLength: 134, getRxFifoLevel: 10
> getPacketLength: 134, getRxFifoLevel: 124

This looks like the same no-buffer-data issue as above except by the time we got around to reading we had already started getting data for the next packet! In this failure mode we currently have one packet that starts with the first 10 bytes of the next packet before repeated garbage bytes, and the next packet now has data missing from the start and rubbish at the end. Both are marked as valid / no CRC error as there doesn't seem to be a way to detect the modem is mid-write.

Thinking about it this passes a sanity check - in continuous RX mode there is a natural race between reading the data after an interrupt and when the next header has been decoded (which updates packet len and later starts writing to the buffer). I'm just surprised there isn't a busy state status or something somewhere? I've combed the datasheet but can't see anything.

(I also have code checking if the getPacketType and getRxPktLength calls are returning RADIOLIB_ERR_NONE as the return status is not checked in getPacketLength, but nothing has been logged).


### Request for confirmation

So I have logged this behaviour on two devices, and seen the repeated '05' packet in the wild on another 3. These are a mix of different boards and modules, with the only common things being the NRF52840 as the MCU and the LR2021 as the radio chip.

Can anyone else confirm they observe this behaviour? If you could add this kludge block to CustomLR2021.h with a build that enables MESH_DEBUG do you capture the log showing DEBUG: RadioLibWrapper: error: readData(-30)? (I hijacked this error code for testing)

```
    int16_t readData(uint8_t* data, size_t len) {
      // Check we have enough data in buffer - infrequent RX_DONE events seen with packet len and empty buffer.
      // Also observed one instance where getRxFifoLevel return part of next packet, reading the first 10 bytes before repeated chars,
      // and corrupting next packet.
      uint16_t fifoLevel = 0;
      int16_t state = getRxFifoLevel(&fifoLevel);
      if (state != RADIOLIB_ERR_NONE) {
        MESH_DEBUG_PRINTLN("readData getRxFifoLevel call returned an error: %d", state);
      }
      
      if (fifoLevel != len) {
        MESH_DEBUG_PRINTLN("FIFO len (%u) does not match packet (%u)", fifoLevel, len);
        // Just in case if we've reached buffer overflow then we have to read to clean up.
        if (fifoLevel >= 256) {
          state = LR2021::readData(data, len);
        }
        if (state == RADIOLIB_ERR_NONE) {
          // We can't be sure we're not mid RX so we should return an error for this, with any readData errored having priority.
          state = RADIOLIB_ERR_PACKET_TOO_SHORT;
        }
      } else {
        state = LR2021::readData(data, len);
      }
      // Always clear IRQ flags for next loop, readData can exit without clearing and MC does not expect that.
      clearIrqState(RADIOLIB_LR2021_IRQ_ALL);
      return state;
    }
```

### Request for test

I don't really have a pared back setup to test this but if someone does could they confirm what the IRQ flags are set if you receive a 0 length packet and if it does set the RX_DONE IRQ flag?

The reason I ask is that this behaviour COULD be explained by the SetLoraPacketParams RadioLib default that sets the packet length to max (255). The datasheet indicates:

> Device only accepts payload lengths from 1 to “xx”. Payloads with a length of 0 or greater than “xx” are rejected, resulting in a header error interrupt

This doesn't specify whether the RX_DONE irq still occurs or not (I don't think it would, but it would be nice to eliminate this possibility). If this only 'filters out' the packet this would make sense - it wouldn't update getPacketLength so I'd expect that to return the previous packet size, and there would be nothing added to the buffer.

Long shot, but worth excluding if possible?


### Final note

What I am unable to explain is how I'm seeing an RX_DONE, a valid packet length in getPacketLength, and no data in the buffer (that never arrives). This is what it looks like from the logs, but it doesn't make sense which makes my hypothesis inconclusive.


### Next steps

1) If someone else confirms the RX_DONE with no buffer data event I will suggest this a is added to RadioLib readData that would cover the one-shot mode. 

2) I'll keep testing continuous RX mode - a sample size of 1/21 isn't exactly conclusive. I'd prefer to keep continuous RX to avoid what seems an unnecessary transition to standby and back - if it comes to that I'll be suggesting a single line tweak to RadioLib anyway so we can default to STANDBY_XOSC and at least avoid the TCXO startup time penalty (which can be longer than 5ms).

carlhodder:
Some more updates - after the last few days of testing I'm getting things sorted to put a PR up to discuss/test some changes for continuous RX mode. Here's a bit of a brain dump/notes around it:

**The race goes both ways**

One thing I've proved (and the datasheet mentions vaguely) is that clearing the FIFO does not stop the current read. The assumption that partial data in the buffer could be a current RX event is not always true - if we cleared the buffer half way through a packet it can be the end of the previous packet.

e.g:

> getPacketLength: 46, getRxFifoLevel: 67
> 09443B00C4D780A30D8E27BC600B1C60AAED6E425DE3408D6C5FC8C9FE493D4A4FBF353EA4011FA3506C3DDD31B2 

So if we're halfway through the next packet the bytes out are what we expected. We interrupted a new read so they bytes we wanted are at the start. The RX fifo was cleared here. 

> getPacketLength: 46, getRxFifoLevel: 25

We just got the RX done from the half-read/half-deleted packet (my code at the time was set to exit without clearing so we can see):

> getPacketLength: 132, getRxFifoLevel: 157
> E3408D6C5FC8C9FE493D4A4FBF353EA4011FA3506C3DDD31B2 114288D20FC28FF355777C2169B407AB8AE6236837324D60D2E40458A481334A45A150305437AD29A16A87D4F41B05BE33EF4517FEC2420ED93A92C2C148A6A9B4130B1CFCBDD8FC7AFF61EF208D932B6EEB1FFA9A05F3C870448B1EA1D32654009F80D5567602D56D0E92

I added a gap to highlight it: This time the last 25 bytes from the previous packet are at the start. 

- With the first oversized packet the old data was at the end of the buffer
- With the second, the old data was at the start 

So we should have some care when clearing the FIFO in continuous mode.

**Continuous vs one-shot RX - is there a difference**

I set up a device running each mode with an SX1262 as a control. Over ~90K RX events in continuous mode I saw a consistent ~5% additional packets compared to one-shot mode so it would be nice if we could keep continuous RX mode. Also while it's most likely the RX sensitivity, the two LR2021's were approx 30% and 35% above the SX1262 in receives.


**Previous-state approach**

My first implementation was to clear the preamble flag immediately, and use this check to guard the read. This is good, but in practise a bit overkill.

We can simplify this a bit more by just tracking the last state:

- If have more or less data than expected set state and wait for more (we can't read this data as the packet length we have is most likely associated with the half-read packet)
- If too much data and this flag is set then this data is old - read and dump it, then read the packet we have now

Taking this method we still need to be careful with preserving IRQ states, so there are some other tweaks required.

This brings the checkReceive error rate down to match one-shot RX, and at a similar rate to the SX1262 (0.09% vs 0.07% over 90K packets), and some of these may be real packets as you sometimes see it on more than one device.


**Reject if not expected size**

Or we could just reject any RX where packet len != buffer. This is the simpler solution - the rate at which I've seen reads mid-write has been from 28/20K to 16/1566 so it may be worthwhile. Seems to increase in occurrence when there is a lot of activity and it's not that busy around here.



### Other changes

**Reduce CRC header to timing only**

I'm seeing this clear the header valid flag on valid packets (0.8%, one run was 132/17154) which drops them as -24, probably another race condition. This flag is only needed for timing - we need a valid header to begin decoding the payload so this has no effect on the data out so I pulled the clear back to just the header CRC IRQ.

**No return/continue to check IRQs**

Similar thing - removed the return statements in the headErr and !headerSeen if blocks, in continuous RX we should fall through as we may have a preamble. 


**Intermittent deafness**

I have also observed this in the wild - randomly my repeater stopped responding to messages for hours and then recovers with advert. I managed to capture one device in this state - somehow we'd missed a RX DONE event or something. The state was still STATE_RX, but the interrupt pin was high. The device healed itself when clearing the IRQs switching to TX mode.

My suggested "fix" for this lockup is to add an explicit RX_DONE IRQ clear in RadioLibWrapper.cpp's recvRaw within the existing #define as it is possible for readData to return an error code and skip clearing interrupts. It only occurs for a few hours maybe once a month so it's not yet a "fix" as it's only been a bit less than a week.


### Misc other things


**SPI2 vs SPI3**

This made no noticeable difference - I discovered that SPI3 has had a bunch of anomalies with the NRF52 platform and even though none match the issues I've observed it was worth a check.

**Odd corruption**

I started noticing that a handful of checkRecv errors usually had 1-4 bytes at the start altered - this also occurs on the Seeed XIAO NRF52850 (SX1262), so I suspect this is some SPI corruption occurring. Or a did until I saw a packet like this received by two units at the same time.

I'll have to set up another device as an SPI logger to check at some point.

PR #3262, opened 2026-08-21
New Variant: Modular Mesh Maker

KR4DIO:
This adds support for several build options of the Modular Mesh Maker DIY platform: Promicro NRF52 and Xiao NRF52 MCUs with NiceRF LR2021 and EByte E22P 1W radio modules. I've only included the repeater envs for the four possible combinations, as those are what I've personally tested, but I can add the room server, bridge, etc if required.

More info on the board is here: https://posiwid.net/projects/modular-mesh-maker/


bplein:
Tested with LR2021+Promicro, LR2021+Xiao_nRF, and for bonus points, tested with LR2021+ Xiao_nRF_Plus on special carrier board. 
image
IMG_2684

saunto:
IMG_3693
Confirmed repeater working on a xiao nrf52 and NiceRF LR2021

PR #3268, opened 2026-08-22
Add support for RAK12500 and RAK12501 GPS on RAK3401

liamcottle:
This PR adds support for RAK12500 (u-blox ZOE-M8Q) and RAK12501 (Quectel L76K) on the RAK3401 1W boards.

We already have support for these GPS modules in RAK4631 builds, however these were not working on RAK3401.

There's a couple of other PRs looking to add support for GPS on RAK3401, but they're significantly more complex.

- https://github.com/meshcore-dev/MeshCore/pull/2640
- https://github.com/meshcore-dev/MeshCore/pull/2927
- https://github.com/meshcore-dev/MeshCore/pull/3051

For now, I'm opting for the most simplistic change to get these modules working.

Changes have been tested with both modules on RAK3401, and I've verified satellite lock on both repeater and companion firmwares.

> Note: GPS modules need to be in Slot A on the RAK4631 and RAK3401 boards. Connecting them to Slot D will result in the user button thinking it's being long pressed, which on companion firmware results in the node booting into CLI Rescue Mode, and Bluetooth Companion connections will not work. For now, users should avoid using Slot D.

### DIFF
--- a/variants/rak3401/platformio.ini
+++ b/variants/rak3401/platformio.ini
@@ -6,6 +6,7 @@ build_flags = ${nrf52_base.build_flags}
   ${sensor_base.build_flags}
   -I variants/rak3401
   -D RAK_3401
+  -D RAK_BOARD
   -D NRF52_POWER_MANAGEMENT
   -D RADIO_CLASS=CustomSX1262
   -D WRAPPER_CLASS=CustomSX1262Wrapper
--- a/variants/rak3401/variant.h
+++ b/variants/rak3401/variant.h
@@ -188,8 +188,8 @@ static const uint8_t AREF = PIN_AREF;
 // Power is on the controllable 3V3_S rail
 #define PIN_GPS_PPS (17) // Pulse per second input from the GPS
 
-#define PIN_GPS_RX PIN_SERIAL1_RX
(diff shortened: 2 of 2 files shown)


netprince17:
I'm using a RAK3401 with no GPS module, I no longer have working LORA radio after this commit.

I rolled back to the commit before this, and it works again.

liamcottle:
> I'm using a RAK3401 with no GPS module, I no longer have working LORA radio after this commit.
> 
> I rolled back to the commit before this, and it works again.

Hey @netprince17 thanks for the report!

My mistake... I've pushed fix for this in https://github.com/meshcore-dev/MeshCore/pull/3275. It would be great if you can also confirm it's working on your side.

Cheers!

netprince17:
https://github.com/meshcore-dev/MeshCore/pull/3275 seems to correct the issue, thanks!

liamcottle:
> #3275 seems to correct the issue, thanks!

Thanks for confirming! I've merged it in :)

IoTThinks:
Yes, I have put a fix for GPS on RAK3401 too.

The fix is also having few lines.

I agree simplicity is best.

PR #3277, opened 2026-08-23
add variant for Adafruit Feather RP2040 with RFM95 LoRa Radio

weebl2000:
Adds a variant for the [Adafruit Feather RP2040 RFM95](https://www.adafruit.com/product/5714), which pairs an RP2040 with an SX1276-based RFM95 module.

The RFM95 is wired to the RP2040's spi1 block, which this board's Arduino variant exposes as the default SPI object. Pin mapping is taken from the framework-arduinopico adafruit_feather_rfm variant: NSS=16, RST=17, DIO0=21, DIO1=22, SCK=14, MISO=8, MOSI=15.

TX power is capped at 20dBm as the RFM95 is PA_BOOST only.

Provides repeater and USB companion environments. BLE and WiFi companion environments are not included: the board has no wireless hardware other than the RFM95, and SerialBLEInterface/SerialWifiInterface are only implemented for esp32 and nrf52.

Credits to Matt Hixson for testing this.

### DIFF
--- a/variants/adafruit_feather_rp2040_rfm/FeatherRP2040RFMBoard.cpp
+++ b/variants/adafruit_feather_rp2040_rfm/FeatherRP2040RFMBoard.cpp
@@ -1,0 +1,33 @@
+#include "FeatherRP2040RFMBoard.h"
+
+#include <Arduino.h>
+#include <Wire.h>
+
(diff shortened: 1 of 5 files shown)


mhixson:
I can confirm that the repeater firmware works on two Adafruit RP2040 LoRa boards.

PR #3287, opened 2026-08-24
ThinkNode M6: fix GPS and status LED in the repeater build

andyhoobing:
## What this fixes

On the Elecrow ThinkNode M6 (nRF52840 + SX1262 + L76K GNSS, sealed IP65 solar
enclosure) the `ThinkNode_M6_repeater` build shipped with an unusable GNSS and
one of its two LEDs completely dead. Pin definitions were already correct; the
problems were behavioural.

### GPS

1. `simple_repeater` defaults `gps_enabled = 0`, so `sensors.begin()` detected
   the L76K and then immediately powered it back down via `_location->stop()`
   (which drives `PIN_GPS_EN` low). The M6 is a sealed box with no button or
   screen, so there was no way to switch it on short of a USB console.
2. Even with GPS on, the repeater defaults `advert_loc_policy` to
   `ADVERT_LOC_PREFS`, so adverts carried the fixed `ADVERT_LAT`/`ADVERT_LON`
   (0,0) and ignored the GNSS entirely.
3. The 1-second "is a GPS attached?" probe in `initBasicGPS()` gates the `gps`
   setting; losing that race leaves GPS unreachable until reboot. The M6 always
   has the module fitted, so the probe is skipped for this board.

Rather than changing behaviour for everyone, the two `simple_repeater` defaults
are now build-time overridable:

```c
#ifndef GPS_ENABLED_DEFAULT
  #define GPS_ENABLED_DEFAULT 0
#endif
#ifndef ADVERT_LOC_POLICY_DEFAULT
  #define ADVERT_LOC_POLICY_DEFAULT ADVERT_LOC_PREFS
#endif
```

**Every other board keeps its existing defaults**; only the M6 repeater env opts
in. Both remain adjustable at runtime via `gps on`/`gps off` and
`gps advert prefs|share|none`.

### LEDs

Only the blue LED was driven (LoRa TX activity). The red LED — the enclosure's
power LED, on the bottom face next to the USB-C port — was configured as an
output in `initVariant()` and then never touched again.

It now indicates state: solid through boot, a three-flash signature at
`onBootComplete()`, then a ~3% duty heartbeat — one blink per 5s while running,
two once the GNSS has a fix. That makes a deployed node diagnosable through the
case without a laptop. `powerOff()` clears it too.

The heartbeat is driven from a small `EnvironmentSensorManager` subclass — the
same pattern `thinknode_m1` already uses — since its `loop()` already runs every
iteration and holds the live GNSS state. **No core or shared-example changes
were needed for the LED work.**

## Testing

- `ThinkNode_M6_repeater` builds against `dev` (flash 44.5%, RAM 14.5%).
- `RAK_4631_repeater` builds unchanged, confirming the shared `simple_repeater`
  defaults are untouched for other boards.
- Verified on real M6 hardware over the serial CLI: `gps -> on, active, fix,
  10 sats`, `gps advert -> share`, and the settings survive a reboot.

## Note for maintainers

One behavioural caveat worth being aware of, which cost me some debugging: these
compiled defaults only apply to a *fresh* install. A UF2 update does not erase
InternalFS, so a device that has run MeshCore before keeps its saved prefs and
silently ignores the new defaults — it has to be reconfigured at runtime or
factory reset. That is pre-existing behaviour, not something this PR changes.

Opened as a draft: happy to split the shared `simple_repeater` change out from
the M6-only variant changes if you would prefer two PRs per the one-fix-per-PR
guideline.

---
Authored with assistance from Claude (see `Co-Authored-By` trailers on the
commits).

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -29,6 +29,16 @@
   #define ADVERT_LON 0.0
 #endif
 
+// GPS defaults for a NEW install. Boards with a permanently-fitted GNSS (and no
+// UI to switch it on) override these in their platformio.ini; every other board
+// keeps the historical "GPS off, advertise the configured coordinates" defaults.
+#ifndef GPS_ENABLED_DEFAULT
+  #define GPS_ENABLED_DEFAULT 0
+#endif
+#ifndef ADVERT_LOC_POLICY_DEFAULT
+  #define ADVERT_LOC_POLICY_DEFAULT ADVERT_LOC_PREFS
+#endif
+
 #ifndef ADMIN_PASSWORD
   #define ADMIN_PASSWORD "password"
 #endif
@@ -919,9 +929,9 @@ MyMesh::MyMesh(mesh::MainBoard &board, mesh::Radio &radio, mesh::MillisecondCloc
   StrHelper::strncpy(_prefs.bridge_secret, "LVSITANOS", sizeof(_prefs.bridge_secret));
 
   // GPS defaults
-  _prefs.gps_enabled = 0;
+  _prefs.gps_enabled = GPS_ENABLED_DEFAULT;
   _prefs.gps_interval = 0;
-  _prefs.advert_loc_policy = ADVERT_LOC_PREFS;
+  _prefs.advert_loc_policy = ADVERT_LOC_POLICY_DEFAULT;
 
   _prefs.adc_multiplier = 0.0f; // 0.0f means use default board multiplier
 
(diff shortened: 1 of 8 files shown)


andyhoobing:
Superseded — split into two self-contained PRs per the one-fix-per-PR guideline in CONTRIBUTING.md:

- #3288 — GPS defaults (repeater + room server examples, M6 env flags)
- #3289 — M6 status LED (variant files only)

They share no files and are independent, so they can be reviewed or merged in any order. Verified the split reconstructs this branch's tree exactly (41 + 116 = 157 insertions, 4 + 3 = 7 deletions across the same 8 files), and each branch builds standalone against `dev`.

Closing this one to avoid duplicate review.

PR #3294, opened 2026-08-24
BME680 BSEC for RAK3401 variant

Magalex2x14:
This is a copy of the current implementation of BME680 environment sensor support using the Bosch BSEC library for the RAK4631 #2634.

Tested with RAK3401+RAK19001+RAK13302 and RAK1906 environment sensor for repeater, sensor, companion builds.


Magalex2x14:
Oh, I noticed the presence of #2716, but this request does not contain other changes than those described.

Issue #3299, opened 2026-08-25
Cyrillic characters are not rendered on the device screen (Heltec Mesh Node T114)


alexex1993:
Title: Cyrillic characters are not rendered on the device screen (Heltec Mesh Node T114)

Device: Heltec Mesh Node T114 (nRF52840 + ST7789 TFT, 160×80)
Firmware version: 1.17.1

Description

Text containing Cyrillic characters is not displayed correctly on the built-in screen of the T114. Instead of the actual letters, the screen shows [blank spaces]. Latin characters, digits and punctuation in the same message render normally.

PR #3300, opened 2026-08-25
[3299] add cyrilic to t114

alexex1993:
https://github.com/meshcore-dev/MeshCore/issues/3299


enigmaspb:
It seems to me that it should be developed together with https://github.com/meshcore-dev/MeshCore/pull/3222

PR #3301, opened 2026-08-25
Fix clock on Heltec T1: Do not misidentify IMU as an RTC

beala:
The T1 has an ICM-42607-P IMU with the same address as a DS3231 RTC. This completely breaks the clock on the T1. Reading the time results in a garbage value, causing all sorts of strange issues. I noticed this because the time never advances, which breaks logging in to repeaters (due to their time-based replay protection).

This simply disables the probe on the I2C line. The DS3231 doesn't have a WHO_AM_I register that could be used to prevent misidentifying it, so this simpler fix is used.

Here is the schematic showing the IMU: https://resource.heltec.cn/download/Mesh_Node_T1/schematic/Mesh_Node_T1_V1.0.pdf
image

Pin 1 is tied to GND, which sets the address to 1101000 or 0x68 ([datasheet](https://www.lcsc.com/datasheet/C5129967.pdf)), same as the DS3231.

image

https://github.com/meshcore-dev/MeshCore/blob/0679dbeffc504d562d2f09eb072fdc223f8ffc2a/src/helpers/AutoDiscoverRTCClock.cpp#L18

# Testing

Without this fix, the time doesn't update:
```
alex@thinkpad ~/w/MeshCore> meshcore-cli -S
40C140D1|* clock
Current time : 1999-12-31 18:09:00 (946688940)
40C140D1|* clock sync
Time synced
40C140D1|* clock
Current time : 1999-12-31 18:09:00 (946688940)
```

With this fix:
```
40C140D1|* clock
Current time : 2024-05-15 05:22:46 (1715772166)
40C140D1|* clock sync
Time synced
40C140D1|* clock
Current time : 2026-08-25 15:29:19 (1787693359)
```

Here's the ble companion firmware if anyone wants to try flashing it at https://meshcore.io and testing it.
[firmware.zip](https://github.com/user-attachments/files/31439182/firmware.zip)

### DIFF
--- a/variants/heltec_t1/variant.h
+++ b/variants/heltec_t1/variant.h
@@ -9,6 +9,12 @@
 #define USE_LFXO
 #define VARIANT_MCK (64000000ul)
 
+// This board has no external RTC chip -- the ICM-42607-P IMU sits at 0x68,
+// the same I2C address DS3231 probing checks, so without this the IMU gets
+// misidentified as a DS3231 and its unrelated registers get read/written
+// as if they were time/status registers, producing a bogus, unfixable clock.
+#define DISABLE_DS3231_PROBE
+
 ////////////////////////////////////////////////////////////////////////////////
 // Number of pins
 


beala:
cc @Quency-D who PR'd initial support for this https://github.com/meshcore-dev/MeshCore/pull/2520

benaltair:
I tested this on a Heltec Mesh Node T1 using BLE companion firmware based on v1.17.1 / `d929643`, paired with MeshCore One. The original symptoms were incorrect dates—including April 30, 2043—and rejected local clock-sync requests.

I used Astra from OpenAI through Codex to look into this and produce test firmware versions. I flashed the builds and performed the device testing myself.

It was restoring a future time even through reboots and flashes. 

Here's the analysis:

> Applying the DS3231 probe fix alone did not recover the device. A second diagnostic build using `VolatileRTCClock` directly still reported April 18, 2068 after restart, and local time sync was still rejected. Both custom version strings were confirmed on the device.
> 
> The investigation found an additional recovery problem. `MyMesh::begin` loads persisted contacts and calls `bootstrapRTCfromContacts`, which sets the RTC to `max(contact.lastmod) + 1`. Since `contact.lastmod` comes from the local RTC and is persisted in `/contacts3`, an earlier clock fault can survive a firmware update and reboot through saved contact metadata.
> 
> A third T1-specific diagnostic build kept the software clock, skipped bootstrapping from contacts, and allowed local companion time-setting to move the clock backwards.
> 
> On hardware, local clock sync then succeeded. Correct UTC time advanced normally, and restart, reconnect, and re-sync tests passed with GPS both off and on. The normal May 2024 software-clock baseline appeared before post-restart sync, and future dates did not return during these tests.
> 
> Regression fixtures also confirmed that the tested non-T1 behaviour was unchanged. A full GPS resync interval, messaging, longer-term behaviour, and all GPS lock-state transitions have not yet been verified.
> 
> The custom workaround intentionally skips offline contact-based time estimation on the T1, so it is not intended as a finished general policy change. The useful upstream finding is that preventing incorrect RTC detection may not be enough for already-affected devices. Recovery also needs to account for future-dated contact metadata and a local clock already ahead of the paired app.
> 
> It may be useful to document a recovery path alongside #3301 and consider a supported local time-correction mechanism, along with a policy for validating or limiting RTC bootstrapping from persisted contact times.
> 
> I can provide the minimal source changes, build details, and host regression results. I also have CLI screenshots from the hardware tests, but have not uploaded or shared them publicly.

PR #3332, opened 2026-08-31
Feature: Add live channel-health metrics to device displays

usrflo:
## Motivation

The statistics a MeshCore device reports to the app are cumulative lifetime totals - they say nothing about the current situation. Channel stats over a short time window let a user quickly check the health of the mesh at their location.

This adds three small, windowed channel-health metrics rendered directly on the device display (companion `ui-new`/`ui-tiny` and `simple_repeater`) and exposed via the `stats-radio` JSON.

channel-stats

## The three metrics

| Line | Question it answers | Window | What feeds it |
|---|---|---|---|
| **CH free** | How much airtime is free? | ~5 s | own TX + mid-receive + RSSI above the quiet-channel reference (see below) with a fixed +15 dB margin, sampled at least every 50 ms |
| **RX ready** | How much of the time was the radio actually in RX mode? | ~5 s | `isInRecvMode()` sampling in the radio loop - catches duty-cycle deafness, blocking code, sleep misconfig |
| **RX quality** | Of the packets that should decode, how many did? | ~10 min | good vs total decodes; CRC failures only counted when SNR ≥ per-SF threshold + 3 dB guard, and reported as 0 % while reception is jammed dead (see below) |

All three render as uniform health bars. A radio that cannot measure them (e.g. ESP-NOW builds) reports no data, and the UI shows "-- %" with a dimmed bar instead of a false "all healthy".

## Sustained interference

The channel-busy verdict needs a reference for "quiet". Using the adapted noise floor directly has a failure mode: a floor that tracks a sustained interferer hides that interferer behind its own margin, so a jammed channel reads as free. **CH free** therefore measures against a *quiet-channel reference*: the 10th percentile of the last 64 published noise-floor values (covering several minutes), absolutely capped at −100 dBm. A sustained broadband interferer keeps CH free near 0 % for as long as it is present — even for hours — and the reference recovers on its own within minutes once the channel is quiet again. During warm-up (first 8 blocks after boot) the current noise floor is used as reference.

**RX quality** cannot drop on its own when reception dies: a jammed channel produces no decode attempts at all, so the windowed ratio would freeze at its last healthy value. It therefore reports 0 % when ambient energy stayed above the quiet reference for ≥ 80 % of the last 5 s **and** no decode attempt occurred for 2 minutes — meaning reception is dead, not merely "no traffic". While packets still decode through the interference, the honest good/total ratio is shown; genuine quiet stays "-- %".

This is display-only: LBT (`interference_threshold`, CAD, resend deferral) keeps using the adapted noise floor. The verdicts reuse the existing ≥ 50 ms RSSI poll — no extra SPI traffic; the percentile is recomputed once per ~3.2 s calibration block.

Field observation (site with an intermittent broadband interferer ~37 dB above the −113 dBm ambient): while present, CH free reads 0 % and RX quality falls to 0 % within ~2 min, with the noise-floor row showing −76; once it disappears, all three rows recover within minutes.

## `stats-radio` JSON

Short keys `util` / `deaf` / `good` / `tot` alongside the existing `chan_util_pct` / `rx_deaf_pct` / `rx_good` / `rx_total` (the reply buffer is 160 bytes).

## Testing

- mcsim end-to-end: utilization/deafness rise with traffic, `good == tot` in clean runs, JSON reply ~106 B
- Builds: Heltec_v3_repeater, RAK_4631_repeater, LilyGo_T-Echo_Card, Heltec_t1 companion, WioTrackerL1 companion
- Hardware field testing on a WioTrackerL1 companion (observation above) and Heltec V3

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -170,6 +170,20 @@ class HomeScreen : public UIScreen {
 #endif
   }
 
+  // Channel-health mini bar (battery-indicator pattern): a small bar with a
+  // right-aligned "NN%" value at the display's right edge. Positive framing:
+  // a full bar is good; it turns warning-coloured below 'warn_below'.
+  void drawHealthBar(DisplayDriver& display, int y, uint8_t pct, uint8_t warn_below) {
+    display.setColor(pct < warn_below ? UIColor::warning_txt : UIColor::primary_txt);
+    char val[8];
+    sprintf(val, "%u%%", pct);
+    display.drawTextRightAlign(display.width(), y, val);
+    const int bar_w = 24;
+    int bar_x = display.width() - display.getTextWidth(val) - 3 - bar_w;
+    display.drawRect(bar_x, y + 1, bar_w, 7);
+    display.fillRect(bar_x + 1, y + 2, (pct * (bar_w - 2)) / 100, 5);
+  }
+
   CayenneLPP sensors_lpp;
   int sensors_nb = 0;
   bool sensors_scroll = false;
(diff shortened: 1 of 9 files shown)


usrflo:
The channel statistics illustrate the limited channel resources to the user:

wiol1stats

PR #3344, opened 2026-09-02
docs: correct stale defaults and stale references across docs 🤖🤖

robekl:
Corrects documentation that had drifted from the firmware. Docs-only — no source changes. One commit per topic.

Each value was re-verified against `dev` (not `main`) before being changed, and the replacement values were verified against source rather than carried over from the old text.

### `docs/cli_commands.md` — stale default values

| Setting | Was | Now | Cause |
|---|---|---|---|
| `radio` / `freq` | `869.525,250,11,5` | `869.618,62.5,8,5` | default preset moved to EU/UK (Narrow) in b777a7c6 |
| `flood.advert.interval` | `12` | `47` (repeater + room server), `0` (sensor) | raised in 40180b8f |
| `advert.interval` | `0` | `2` fresh, `0` once configured | see below |
| `direct.txdelay` | `0.2` | `0.3` repeater, `0.2` room server/sensor | role-dependent |

`advert.interval` needed both states documented: prefs store minutes/2 and the getter doubles on read, so a factory-fresh node reports `2` — but `savePrefs()` zeroes any interval below the 60-minute minimum (`CommonCLI.cpp:162-165`) and is called from all 37 `set` handlers, so it becomes `0` the moment the node is configured. Neither `2` nor the old `0` is the whole story alone.

### `docs/companion_protocol.md` — channel index, secret semantics, payload limit

- **Payload limit.** Was `163`. The naive fix is `167` (`MAX_FRAME_SIZE` went 172→176 in 62f1b11d), but that is *worse* than the stale value: 167 is only the host-frame bound (`MyMesh.cpp:1265`). The radio-side bound is `MAX_GROUP_DATA_LENGTH = 184 - 16 - 3 = 165` (`MeshCore.h:21`, enforced at `BaseChatMesh.cpp:544`). A 166–167 byte payload passes the frame check, fails the radio check, and returns `ERR_CODE_TABLE_FULL` — which this same doc describes as "retry later", so a conforming client would retry a permanently-failing send forever. Now documents 165 as the limit to enforce and calls out the 166–167 band.
- **Channel index.** Was hardcoded `0-7` in seven places. The bound is `MAX_GROUP_CHANNELS`, which is 40 on 176 of the current variant configs, 8 on 31 and 1 on 26. Clients should read `max_channels` from byte 3 of `PACKET_DEVICE_INFO`.
- **Secret field.** Was documented as "all zeros" for public channels. The firmware always hashes a real 16-byte key, and the public channel ships with `izOH6cXN6mrJ5e26oRXNcg==`. An all-zero secret is not private and not inert — SHA256 over 16 zero bytes is a global constant, giving a well-known channel with an all-zero AES key. `searchChannelsByHash` skips *unnamed* slots for exactly this reason (`BaseChatMesh.cpp:392-401`), but a named slot with a zero secret is not skipped and absorbs null-key traffic from any node.

### `docs/nrf52_power_management.md` — pwrmgt on unsupported boards

The doc claimed every `pwrmgt` command except `get pwrmgt.support` returns `ERROR: Power management not supported`. `get pwrmgt.bootreason` is not inside the `#ifdef` (`CommonCLI.cpp:787`) and answers on every board — `getResetReason()`/`getShutdownReason()` are base-class virtuals (`MeshCore.h:71-74`) returning `Not available` when unoverridden, and ESP32 overrides the reset half with `esp_reset_reason()`.

### `docs/terminal_chat_cli.md` — command that does not exist

f6338430 added `get`/`set dutycycle` and updated both CLI docs, but Terminal Chat does not use the shared handler — `simple_secure_chat/main.cpp:479-507` has its own inline `set` supporting only `af|name|lat|lon|tx|freq`. Typing `set dutycycle` there returns `ERROR: unknown config`. Terminal Chat is the *only* firmware without it; Companion Radio does support it via `CommonRadioPrefs`. Also documents `help`, which the client implements but the doc never listed.

### `README.md` / `docs/faq.md` — stale references

- Two links still pointed at `ripplebiz/MeshCore`.
- `pio run -e RAK_4631_Repeater` — no such env; it is `RAK_4631_repeater`.
- `Heltec_V3_companion_radio_ble` → `Heltec_v3_...`; artifacts are named from the env name (`build.sh:147`), and the two neighbouring examples in the same block already used lowercase.
- Build recipe pointed at `LORA_FREQ=867.5`; reworded to reference the flag rather than a value that drifts with the default preset.

### Not included

Found during the same audit, deliberately left out to keep this reviewable:

- `extra.sf` (`set`/`get`) is entirely undocumented.
- The sensor `io` GPIO command is undocumented.
- The Ethernet section omits ThinkNode M7 / CH390.
- The nRF52 power-management board table lists 5 of 13 boards that now define `NRF52_POWER_MANAGEMENT`.
- `companion_protocol.md`'s packet-type table covers 21 of 46 codes.
- `payloads.md` lists `PAYLOAD_TYPE_MULTIPART` without specifying it, and its request-type table stops at `0x02`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,6 @@ There are a number of fairly major features in the pipeline, with no particular 
 
 ## 📞 Get Support
 
-- Report bugs and request features on the [GitHub Issues](https://github.com/ripplebiz/MeshCore/issues) page.
+- Report bugs and request features on the [GitHub Issues](https://github.com/meshcore-dev/meshcore/issues) page.
 - Find additional guides and components on [my site](https://buymeacoffee.com/ripplebiz).
 - Join [MeshCore Discord](https://meshcore.gg) to chat with the developers and get help from the community.
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -211,7 +211,7 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 
 **Set by build flag:** `LORA_FREQ`, `LORA_BW`, `LORA_SF`, `LORA_CR`
 
-**Default:** `869.525,250,11,5`
+**Default:** `869.618,62.5,8,5`
 
 **Note:** Requires reboot to apply
 
@@ -256,7 +256,7 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 **Parameters:**
 - `frequency`: Frequency in MHz
 
-**Default:** `869.525`
+**Default:** `869.618`
 
 **Note:** Requires reboot to apply
 **Serial Only:** `set freq <frequency>`
@@ -537,7 +537,7 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 **Parameters:**
 - `value`: Direct transmit delay factor (0-2)
 
-**Default:** `0.2`
+**Default:** `0.3` (Repeater) - `0.2` (Room Server, Sensor)
 
 **Note:** Same collision-avoidance random window as `txdelay`, but applied to direct (non-flood, routed) traffic. The default is lower because direct packets are addressed to a specific next hop, so far fewer nodes compete to retransmit them.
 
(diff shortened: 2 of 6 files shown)


oltaco:
👍

Issue #2011, opened 2026-03-12
[Feature Request] - Disable LoRa radio in all firmware variants as default


AI7NC:
Problem:
Currently in all firmware variants, the default radio configuration that is set upon flashing a device is 869.618 MHz at full device power. While this is likely a UK legacy design, I have strong concerns about defaulting to any radio-on state prior to user configuration in an international distribution with varying spectrum regulation. 

In the US, 869.618MHz is spectrum dedicated for cellular carrier use. See below search for active licenses in this region. With our users being a general audience interested in off-grid coms with layman knowledge, we're putting them at a disadvantage of properly operating within the compliance of their regional laws. 

Of all the areas of spectrum, cellular is the most monitored with the fastest enforcement for being out of compliance, and our 62.5khz bandwidth will stick out like a sore thumb. I believe it is in the best interest of this community to design our firmware to avoid the eventuality of an enforcement action bringing negative publicity the efforts of this group.

![Image](https://github.com/user-attachments/assets/5ec23835-ddda-4b70-987a-8fedf1aecee3)

How to fix it:
Change the default radio confirmation in the firmware variants to 'disabled' by default. This forces the user to set the regional preset at device confirmation and puts the regional compliance responsibility on the user rather than the firmware originator.

jade-on-mesh:
I concur, this would be a good change and in my opinion is long overdue.

jeroen94704:
Coincidence or not, I was just looking at proposing something similar for the admin password. Can this be combined, or should that be a separate feature request?

AI7NC:
I think that should be a separate discussion. You should open a feature request for that.

jeroen94704:
Thanks, done: https://github.com/meshcore-dev/MeshCore/issues/2012

txkbaldlaw:
This seems like a sound idea.    

It also can help ameliorate people burning out their radio if they don't hook up an antenna before plugging it into their USB for configuration.   With the delicateness of the antenna jacks, minimizing the number of required plug/unplug cycles is good practice in my opinion.  

Recommend for implementation.

AI7NC:
> It also can help ameliorate people burning out their radio if they don't hook up an antenna before plugging it into their USB for configuration.

That is a benefit of this request I hadn't considered.

IoTThinks:
Thinking for a quick fix.
Not sure how RadioLib reacts when set LORA_FREQ to 0.0.
May be the radio will fail to init.

```
[...]

PR #3348, opened 2026-09-02
Documentation change to clarify radio settings and individual country requirements.

KPrivitt:
There has been a lot of discussion about the US regulatory requirements for MeshCore.  Issues #945
In the US, the radio settings are the responsibility of the manufacturer to document and test for compliance in order to manufacture and sell their product.

Changing the documentation further protects the MeshCore firmware by specifically pointing to the manufacturer (they may use MeshCore firmware, but they are responsible for what they ship) and to the Country Requirements.

It is not MeshCore's responsibility to determine or enforce compliance with any countries regulations.

The issue in #945 should be addressed in the Companion Apps, which is where the Presets reside.  They are not in this firmware.


Lazerbeak12345:
I see right now that your suggested change does not make mention of the responsibility of Companion Apps to have a correct preset. Such a mention may not be needed, but it's worth considering. What are your thoughts on this?

KPrivitt:
My first concern is for the MeshCore firmware to be "covered". I'll address the Preset issue last.

I believe there are two areas which need to be addressed:

1) Documentation: This PR.  The primary responsibility first goes to the manufacturer.  They need to address "the setting issue" in their documentation, after that it is the operators responsibility to choose the radio settings for their area (remember this firmware is International, it's not just a US issue). MeshCore is just an ingrediant in the manufacturers product. This PR is covering MeshCore and its documentation.

2) MeshCores default settings. This is issue #2011 

My proposal is to set the default power to 0dBm.  This is low enough that it is complies with most if not all countries rules.The radio is "functional" but not set to communicate on any mesh.  Since it is not functional **any** setting will work, even BW 500, so why not include that too.  To be "useable" it must be set by the user for the configuration of their local mesh.  Responsibility then belongs to them, the operator. (not MeshCore)

I will be preparing a new PR once I get confirmation on the 0dBm direction vs. partially shutting down the radio (there is no "disable" setting).

The default MeshCore setting will be compliant in any country, so MeshCore cannot be "liable", they do not violate any countries rules.  

An additional protection is that there is no UI (display/joystick/keyboard) to interact with (yes, there is the CLI, but without a UI to connect to the CLI it "doesn't work", so another ingredient is necessary, to make it work.  MeshCore is not required to address this, plus if it did it would need to "speak" how many languages and address the intracacies of how many countries individual rules.

Now what to do about Presets:

The App is not the only place they exist.  Custom firmware (only a couple examples: WhisperOS, wadamesh, etc, etc) do include Presets in their firmware and they include MeshCore for the base protocol.  Because they add a UI, this is a place to address this. They are able to do this because their new firmware includes a User Interface.

The Flasher Apps, which provide a UI via BLE, USB-C, WiFi, etc.

Tools, etc, etc...

The companion App, **there are many**. Again with the inclusion of a UI, the responsibility is "transferred".  There will be arguments that it becomes the Apps responsibility, but ultimately it is the operators responsibility. (I am not against doing something in the App, but I am firmly against doing it here in a UI-less "middleware".

Because there are too many places where the "solution" needs to be placed, it seems it may be futile.  But I'm OK with whatever the Apps do, as long as they allow full flexibility by the user (I can legally choose to not use BW 500 if I lower the power).

One thing to note is that a USA Wideband is not the only viable solution: §15.209, §15.249, 15.247 (the first two have been ridiculed as non-starters, but communication is still possible, 3-5kM is definitely possible, sensors use this, etc). But yes, the higher power is better, and remember this is more of a repeater issue (there is definitely no UI)

Yes, there is goodness in helping an ignorant user, but it is not a one size fits all situation.

Ultimately, I don't think it is really any of the Apps responsibility.  It IS the Operators responsibility, and this is "transferred" to the "leaders" of each local mesh helping all mesh members with the settings (like PhillyMesh making their own decisions and directing their users).  Without common settings there is no mesh.

There are too many individual mesh's out there, it will be herding cats...

What are your thoughts?  I'm open to realistic suggestions...

PR #3350, opened 2026-09-03
fix-heltec_v4r8-lna

Quency-D:
Add CLI commands to control the LNA function for heltec_v4r8.


ThomasFarstrike:
https://wiki.heltec.org/news/lna/lna

What is an LNA?

An LNA (Low Noise Amplifier) is a critical RF component located at the receiver front end. Its primary purpose is to amplify weak RF signals from the antenna while maintaining a low noise figure. By providing low-noise amplification and increasing the signal level before further processing, the LNA helps minimize the impact of subsequent stages on overall receiver performance and improves receiver sensitivity.

image

PR #3358, opened 2026-09-05
feat(cli): try confirm-or-revert prefs

benallfree:
Hey guys this is a draft PR for visibility only, curious to hear what everyone thinks.

I've been experimenting with some meshcore settings having to do with radio RX and got to wondering how I would get the node back if the RX setting caused it to go deaf.

My generalized solution/proposal is a `try` command that will flip a setting back after a timeout unless it is committed with `set`. And to avoid footguns, you can also use `try reboot` which will reboot the device after reverting the setting. Some settings auto-arm reboot to further avoid footguns. 

Sample commands:

```text
# Start a 1-hour rxgain trial (reverts unless you commit)
try 3600 radio.rxgain off

# Same, but reboot after timeout revert (optional keyword)
try reboot 3600 radio.rxgain off

# Radio 4-tuple trial (reboot armed on revert is inherited; apply still needs manual reboot)
try 3600 radio 915.0,250,11,5

# Powersaving is a bare verb, not set powersaving
try 1800 powersaving on

# List pending trials (key, trial value, seconds left, reboot suffix if armed)
get try

# Commit before timeout (cancels revert for that key)
set radio.rxgain off

# Snapshot current value (what try would revert to)
get radio.rxgain

# Short bench run
try 10 radio.rxgain off
get try
# wait ~10s, then:
get radio.rxgain
get try
```

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -248,6 +248,30 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 
 ---
 
+#### Try a setting with automatic revert
+**Usage:**
+- `try <secs> <set-args>`
+- `try reboot <secs> <set-args>`
+- `get try`
+
+**Parameters:**
+- `secs`: Trial duration in seconds (1–604800, max 7 days)
+- `set-args`: Same arguments as `set` (e.g. `radio.rxgain off`) or `powersaving on|off`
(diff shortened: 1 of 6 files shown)


LitBomb:
@benallfree are you the guy behind envybot? 

https://www.reddit.com/r/meshcore/comments/1w5whrh/a_local_5013c_organization_is_vibe_coding_a/

https://github.com/MeshEnvy/envybot

PR #3359, opened 2026-09-05
Next hop reliability

OverkillFPV:
This is an attempt at direct path reliability using retries if a repeat is not heard, I reckon the extra airtime implication of retries either for failed packets, or not heard repeats that were actually repeated is overall not as big of an impact of a complete resend. only a draft as i am testing the first revision of it.

### DIFF
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -9,6 +9,7 @@ void Mesh::begin() {
 
 void Mesh::loop() {
   Dispatcher::loop();
+  processNextHopRetries();
 }
 
 bool Mesh::allowPacketForward(const mesh::Packet* packet) { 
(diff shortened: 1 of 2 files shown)


Alain2019:
What's the connection with [Proposal: Use re-transmission for direct messages for more reliability and less floods](https://github.com/meshcore-dev/MeshCore/issues/1342#top)
#1342 ?

OverkillFPV:
Ah, the connection is that I didn't see that one lol. I'll leave this for now as it seems like a simpler implementation but that other one looks to have much more thought into it

Alain2019:
> Ah, the connection is that I didn't see that one lol. I'll leave this for now as it seems like a simpler implementation but that other one looks to have much more thought into it

Maybe do some testing with it?
Also retries with flood msg's is still "open".

PR #3376, opened 2026-09-07
Fix CMD_SEND_CHANNEL_DATA payload validation

### LINKED ISSUE #3345: CMD_SEND_CHANNEL_DATA: 166-167 byte payloads return ERR_CODE_TABLE_FULL instead of ERR_CODE_ILLEGAL_ARG
## Summary

`CMD_SEND_CHANNEL_DATA` (0x3E) gates the outbound payload against a frame-sized bound (167) instead of the radio-side bound (165). Payloads of exactly **166 or 167 bytes** pass the first check, fail the second, and are reported to the host as `ERR_CODE_TABLE_FULL` (3) — the "outbound queue is full, retry later" code. The send can never succeed, so a conforming client retries forever.

## The two bounds

| Constant | Value | Defined | Purpose |
|---|---|---|---|
| `MAX_CHANNEL_DATA_LENGTH` | `MAX_FRAME_SIZE - 9` = **167** | `examples/companion_radio/MyMesh.cpp:103` | host frame budget |
| `MAX_GROUP_DATA_LENGTH` | `MAX_PACKET_PAYLOAD - CIPHER_BLOCK_SIZE - 3` = `184 - 16 - 3` = **165** | `src/MeshCore.h:21` | radio payload budget |

The `- 9` is exactly the `RESP_CODE_CHANNEL_DATA_RECV` header: code, SNR, 2 reserved, channel_idx, path_len, 2 data_type, data_len — nine `out_frame[i++]` writes at `MyMesh.cpp:664-673`. So `MAX_CHANNEL_DATA_LENGTH` is a correct constant for the **inbound** frame, and the bug is reusing it to gate an **outbound** radio send.

## Where it goes wrong

`examples/companion_radio/MyMesh.cpp:1265-1272`:

```cpp
} else if (payload_len > MAX_CHANNEL_DATA_LENGTH) {   // 167 -- wrong bound here
  writeErrFrame(ERR_CODE_ILLEGAL_ARG);
} else if (sendGroupData(channel.channel, path, path_len, data_type, payload, payload_len)) {
  writeOKFrame();
} else {
  writeErrFrame(ERR_CODE_TABLE_FULL);                 // 166-167 lands here
}
```

`BaseChatMesh::sendGroupData` then rejects it (`src/helpers/BaseChatMesh.cpp:544-547`):

```cpp
if (data_len > MAX_GROUP_DATA_LENGTH) {   // 165
  MESH_DEBUG_PRINTLN("sendGroupData: data_len=%d exceeds max=%d", data_len, MAX_GROUP_DATA_LENGTH);
  return false;
}
```

`sendGroupData` returns `bool`, so the handler cannot tell "payload too large" (permanent) from "queue full" (transient) and maps both to `ERR_CODE_TABLE_FULL`.

## Reproduction

Send a `CMD_SEND_CHANNEL_DATA` frame over the companion link with a 166-byte payload:

```
byte 0      : 0x3E   (CMD_SEND_CHANNEL_DATA)
byte 1      : 0x00   (channel_idx -- the built-in Public channel)
byte 2      : 0xFF   (path_len, OUT_PATH_UNKNOWN = flood)
bytes 3-4   : 0xFF 0xFF   (data_type = DATA_TYPE_DEV)
bytes 5-170 : 166 arbitrary payload bytes
```

Total frame length 171 bytes, comfortably under `MAX_FRAME_SIZE` (176), so it is reachable over both BLE and serial.

**Expected:** `ERR_CODE_ILLEGAL_ARG` (6), consistent with how a 168-byte payload is already handled.

**Actual:** `ERR_CODE_TABLE_FULL` (3), plus `sendGroupData: data_len=166 exceeds max=165` on the debug log if `MESH_DEBUG=1`.

167 behaves the same way. 168 and above correctly return `ERR_CODE_ILLEGAL_ARG`. 165 and below succeed.

## Impact

`ERR_CODE_TABLE_FULL` is documented as transient and retryable, so a well-behaved client backs off and retries — and never succeeds. It is a narrow band (2 of 168 lengths), but it fails in the most confusing way available: an infinite retry loop with a misleading diagnostic, rather than a clean rejection.

## Suggested fixes

Either would do; the first is smaller:

1. Gate the outbound path on the radio bound — check `payload_len > MAX_GROUP_DATA_LENGTH` at `MyMesh.cpp:1265` and keep `ERR_CODE_ILLEGAL_ARG`. `MAX_CHANNEL_DATA_LENGTH` stays as-is for the inbound frame check at `MyMesh.cpp:658`, where it is correct.
2. Give `sendGroupData` a distinguishable failure so the handler can map "too large" to `ERR_CODE_ILLEGAL_ARG` and reserve `ERR_CODE_TABLE_FULL` for genuine queue exhaustion.

## Notes

- Not board-specific; affects any `companion_radio` build. Observed by reading `dev` at 65aa1138.
- Found while auditing docs against firmware. meshcore-dev/MeshCore#3344 documents the current behaviour (165 as the effective limit, and the 166-167 band as permanent-failure) but deliberately does not change code. If this is fixed, that doc note should be simplified back to a single limit.

VoxMesh:
Fixes #3345

Validate `CMD_SEND_CHANNEL_DATA` payloads against `MAX_GROUP_DATA_LENGTH` instead of
`MAX_CHANNEL_DATA_LENGTH`.

This rejects 166-167 byte payloads with `ERR_CODE_ILLEGAL_ARG` before they reach
`sendGroupData()`, while leaving the receive-side `MAX_CHANNEL_DATA_LENGTH` check unchanged.

Verified:
- 164-165 byte payloads pass the validation check
- 166-168 byte payloads are rejected by the validation check
- `Heltec_v3_companion_radio_ble` builds successfully

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -1266,8 +1266,8 @@ void MyMesh::handleCmdFrame(size_t len) {
       writeErrFrame(ERR_CODE_NOT_FOUND); // bad channel_idx
     } else if (data_type == DATA_TYPE_RESERVED) {
       writeErrFrame(ERR_CODE_ILLEGAL_ARG);
-    } else if (payload_len > MAX_CHANNEL_DATA_LENGTH) {
-      MESH_DEBUG_PRINTLN("CMD_SEND_CHANNEL_DATA payload too long: %d > %d", payload_len, MAX_CHANNEL_DATA_LENGTH);
+    } else if (payload_len > MAX_GROUP_DATA_LENGTH) {
+      MESH_DEBUG_PRINTLN("CMD_SEND_CHANNEL_DATA payload too long: %d > %d", payload_len, MAX_GROUP_DATA_LENGTH);
       writeErrFrame(ERR_CODE_ILLEGAL_ARG);
     } else if (sendGroupData(channel.channel, path, path_len, data_type, payload, payload_len)) {
       writeOKFrame();


VoxMesh:
#3386 has been merged, so this is no longer needed. Thanks.

PR #3380, opened 2026-09-08
Added button press support for ESP32 repeaters with powersaving on

IoTThinks:
Currently, when ESP32 repeaters enter powersaving on, the screen can not be waken up by button press and long press to power off ESP32 repeaters is not supported.

This is to add button press support for ESP32 repeaters with powersaving on.
* Support short press to wake screen and ESP32 for 30s and off.
* Support long press to power off ESP32 repeaters

Tested working fine for Heltec v3 and V4.2.


IoTThinks:
@liamcottle Hello Liam, 
This is a tiny PR to bring back button press to ESP32 repeaters with powersaving on.
Thanks a lot and have a nice day.

PR #3383, opened 2026-09-09
Add RSSI-based listen-before-talk

jirogit:
## Motivation

MeshCore has two pre-transmit channel checks today, and neither can express the requirement this PR addresses.

`cad` uses the radio's hardware Channel Activity Detection, which looks for a LoRa preamble. In a band shared with equipment using other modulations, that is exactly the traffic it cannot see.

`int.thresh` does sample RSSI, but relative to the tracked noise floor, and only once. A deployment that has to hold to an *absolute* dBm threshold, sense the channel for a defined window rather than instantaneously, or stay quiet for a fixed period after each transmit, has no way to say so.

This PR adds RSSI-based listen-before-talk: modulation-agnostic energy detection with an absolute threshold and a configurable sensing window, plus three transmit-shaping limits that go with it.

## What's new

set rssi.lbt on|off
get rssi.lbt -> "> on" / "> off"

set rssi.lbt.params <thr_dBm>,<sense_ms>,<maxwait_ms>,<txmax_ms>,<pause_ms>
get rssi.lbt.params -> "> -80,5,0,4000,50"

`get rssi.lbt.params` reports the values in the same order `set` takes them and with the same separator, so its output can be pasted straight back into `set` — following the existing `get radio` / `set radio` pair. Its arguments are:

| argument | default | meaning |
|---|---|---|
| `thr_dBm` | `-80` | absolute RSSI threshold; a sample above this marks the channel busy |
| `sense_ms` | `5` | how long RSSI is sampled for. `0` takes a single instantaneous reading |
| `maxwait_ms` | `0` | how long a busy channel may hold off a pending transmit before it is forced out. `0` = unlimited |
| `txmax_ms` | `4000` | max airtime of one transmit; a longer packet is dropped rather than sent. `0` = unlimited |
| `pause_ms` | `50` | quiet period enforced after each transmit |

Throughout, `0` means "unlimited". The defaults are chosen so that a single `set rssi.lbt on` yields a usable configuration without touching `rssi.lbt.params`.

## Behaviour when disabled

This is the first thing worth checking, so to state it plainly:

- **While `rssi.lbt` is `off` — the default — every one of the `rssi.lbt.params` values is ignored.** The sensing loop, the post-transmit pause, the airtime cap and the max-wait are each gated on the enable flag, not merely on their own value being zero. A node that never issues these commands transmits exactly as it does on `dev`.
- `getCADFailMaxDuration()` returns the original hard-coded 4000ms whenever `rssi.lbt` is off, whether or not `cad` is on.
- The legacy binary prefs loader is intentionally not extended. These fields cannot exist in a file written by a build that predates them, so an upgraded node simply starts ith the defaults above.

## Design notes

**Independent of CAD.** The RSSI check is added as a third branch of `RadioLibWrapper::isChannelActive()`, next to `int.thresh` and `_cad_enabled`. The two existing branches are unchanged and the return value stays a plain `bool`, so `isReceiving()` remains the single gate `Dispatcher::checkSend()` consults. Any combination of the three may be active.

**Max-wait.** `rssi.lbt`'s `maxwait_ms` and CAD's fixed 4000ms both bound the same busy state in `checkSend()`. With `rssi.lbt` off the function returns 4000 unchanged. With it on, the stricter — longer — wait applies, so CAD's fixed value never shortens a wait that `rssi.lbt` asked to be longer. Since `0` means "never force the transmit", it is normalised to the largest representable duration before any comparison, so it does not lose to a finite value.

The effective value is derived on every call and never written back to prefs, so a computed number cannot end up stored and disagree with what the CLI reports.

**Airtime cap.** Enforced immediately before `startSendRaw()`, reusing the existing send-failure cleanup path. There is no counter for dropped packets; the drop is visible only through `MESH_DEBUG_PRINTLN`, so release builds stay silent.

**One implementation, four node types.** The prefs live in `CommonRadioPrefs`, which both `NodePrefs::RadioPrefs` implementations derive from, so companion, repeater, room server and sensor are all covered without per-target code.

## Commits

The PR is one change but is split into four commits that can be taken separately if review stalls on any one of them:

1. The RSSI sensing mechanism and post-transmit pause in `RadioLibWrapper`. Inert on its own; nothing enables it yet.
2. The prefs and CLI wiring.
3. The airtime cap and the max-wait handling.
4. Documentation.

## Testing

Firmware builds and basic on-device smoke testing:

- M5Stack Unit C6L (ESP32-C6 / SX1262) — repeater
- Seeed T1000-E (nRF52840 / LR1110) — repeater
- RAK WisMesh Tag (nRF52840 / SX1262) — companion, over both BLE and USB

Nodes come up, advertise and forward packets as before, and with `rssi.lbt` left at its default of `off` I saw no behaviour change from current `dev`. This was smoke testing rather than a measured comparison; I have not run the mechanisms under sustained traffic.

## Open question

`pause_ms` is currently a blocking `delay()` in `onSendFinished()`. At the default of 50ms that seemed acceptable, but I am happy to make it non-blocking if you prefer.

### DIFF
--- a/docs/cli_commands.md
+++ b/docs/cli_commands.md
@@ -613,12 +613,48 @@ This document provides an overview of CLI commands that can be sent to MeshCore 
 - `get cad`
 - `set cad <on|off>`
 
-**Description:** When enabled, the radio performs a hardware Channel Activity Detection scan before transmitting and defers if the channel is busy. Runs independently of `int.thresh` — either, both, or none may be active.
+**Description:** When enabled, the radio performs a hardware Channel Activity Detection scan before transmitting and defers if the channel is busy. Runs independently of `int.thresh` and `rssi.lbt` — any combination may be active.
 
 **Parameters:**
 - `on|off`: Enable or disable hardware CAD
 
 **Default:** `off`
+
+---
+
+#### Enable or disable RSSI listen-before-talk
+**Usage:**
+- `get rssi.lbt`
+- `set rssi.lbt <on|off>`
+
+**Description:** When enabled, the radio samples RSSI continuously for a short window before transmitting and defers if any sample exceeds an absolute threshold. Unlike `cad`, which detects a LoRa preamble, this is energy detection and so is independent of whatever modulation is on air; unlike `int.thresh`, the threshold is absolute rather than relative to the tracked noise floor.
+
+Runs independently of `cad` and `int.thresh` — any combination may be active. When both `cad` and `rssi.lbt` are enabled, the stricter of CAD's fixed 4000ms wait and the `rssi.lbt` `maxwait_ms` applies.
+
+While this is `off`, every value in `rssi.lbt.params` is ignored and the node transmits exactly as it would with this feature absent.
+
+**Parameters:**
+- `on|off`: Enable or disable RSSI listen-before-talk
+
+**Default:** `off`
+
+---
+
+#### View or change the RSSI listen-before-talk parameters
+**Usage:**
+- `get rssi.lbt.params`
+- `set rssi.lbt.params <thr_dBm>,<sense_ms>,<maxwait_ms>,<txmax_ms>,<pause_ms>`
+
+**Description:** All five values are set together and reported together, in the same order, so the output of `get rssi.lbt.params` can be pasted straight back into `set rssi.lbt.params`. None of them have any effect while `rssi.lbt` is `off`.
+
+**Parameters:**
+- `thr_dBm`: Absolute RSSI threshold, `-128`-`0`. A sample above this marks the channel busy.
+- `sense_ms`: How long to sample RSSI for, `0`-`65535`. `0` takes a single instantaneous reading.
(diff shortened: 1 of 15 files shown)


thomasabbott:
A fixed dBm value for RSSI seems risky - it could disable the device entirely if the background level shifts slightly for any reason. Some sort of auto-adjusting threshold might be better.

Measuring the power at the transmitting device is a useful but limited measure - it doesn't really predict the channel power at the intended receiver. I can see some value in delaying transmission if something else is obviously transmitting strongly right now, but no point in waiting ages for the perfect quiet period.

I'd suggest accumulating some RSSI values over maybe 5 seconds, and setting the threshold at say 6 dB above the 50th percentile value. i.e. suppress transmission if the channel RSSI is currently higher than 6 dB more than 50% of the recent samples. This would avoid the strongest bursts but not hold up transmission too much, and not at all if the channel is completely quiet. An EMA is also possible, say in every 5 seconds, feed the 50th percentile to an EMA accumulator, and make the threshold for waiting about 6 dB above that.

PR #3390, opened 2026-09-10
Add isGPSDetected() capability probe to SensorManager

jbrazio:
ENV_INCLUDE_GPS says a build can have GPS. It says nothing about whether one is actually attached, and there is no way for the app to ask.

Adds a virtual `isGPSDetected()` to SensorManager, returning false by default, overridden in EnvironmentSensorManager to return its existing `gps_detected` flag. Builds without GPS keep the default.


oltaco (approved):

oltaco:
Looks good to me, and I think it would give us a cleaner way to do #3391 and could also help solve the issue raised here https://github.com/meshcore-dev/MeshCore/pull/3391#issuecomment-5626321673

PR #3391, opened 2026-09-10
Honour advert location policy in builds without GPS

jbrazio:
In `buildAdvertData()`, ADVERT_LOC_SHARE reads `_sensors->node_lat` and `node_lon`. Builds compiled without GPS never populate those, so a node set to share its location advertises coordinates 0,0.

Gates the function on ENV_INCLUDE_GPS like the rest of the GPS paths: builds without GPS honour ADVERT_LOC_NONE and otherwise fall back to the configured `node_lat`/`node_lon`.

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -168,6 +168,7 @@ void CommonCLI::savePrefs() {
 }
 
 uint8_t CommonCLI::buildAdvertData(uint8_t node_type, uint8_t* app_data) {
+#if ENV_INCLUDE_GPS == 1
   if (_prefs->advert_loc_policy == ADVERT_LOC_NONE) {
     AdvertDataBuilder builder(node_type, _prefs->node_name);
     return builder.encodeTo(app_data);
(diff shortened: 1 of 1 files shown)


f3sty:
Builds without GPS prevent switching between ADVERT_LOC modes, i.e. you can set node_lat/node_lon but you can't get them to be included in adverts because the only way to set ADVERT_LOC_PREFS is via the gps advert command - which isn't available without GPS support.

I think a fix should not only set ADVERT_LOC_NONE if the stored coords are 0/0, but also provide a command to set ADVERT_LOC policy that is still available without GPS support. Otherwise the repeater can never advertise the coords set in prefs.

src/helpers/CommonCLI.cpp
```diff
@@ -178,6 +179,16 @@ uint8_t CommonCLI::buildAdvertData(uint8_t node_type, uint8_t* app_data) {
     AdvertDataBuilder builder(node_type, _prefs->node_name, _prefs->node_lat, _prefs->node_lon);
```
oltaco, line 179: If you dropped the else and moved the guard to here I think it would achieve the same outcome but you could drop the duplicated ``if (_prefs->advert_loc_policy == ADVERT_LOC_NONE)`` branch in the #else below?

oltaco (changes requested):
I think it would be tidier to just guard the ADVERT_LOC_SHARE branch, unless I am missing something?

oltaco:
> Builds without GPS prevent switching between ADVERT_LOC modes, i.e. you can set node_lat/node_lon but you can't get them to be included in adverts because the only way to set ADVERT_LOC_PREFS is via the gps advert command - which isn't available without GPS support.
> 
> I think a fix should not only set ADVERT_LOC_NONE if the stored coords are 0/0, but also provide a command to set ADVERT_LOC policy that is still available without GPS support. Otherwise the repeater can never advertise the coords set in prefs.

@f3sty I think this is a separate but related issue, is there currently an issue open for it?

oltaco:
@jbrazio What do you think about changing this to use isGPSDetected() for the middle branch, gated on lat/lon != 0 so that the advert doesn't report 0 before a fix has acquired?

jbrazio:
@oltaco I don't think those two are interchangeable.

isGPSDetected() answers a runtime question: is a GPS unit attached and talking. ENV_INCLUDE_GPS answers a compile-time one: is any GPS code in this binary at all.

PR #3394, opened 2026-09-10
Drop stale packets from the TX queue to prevent late/duplicate delivery

jbrazio:
When the duty-cycle budget runs out, outbound packets wait in the TX queue until it refills, and nothing ever expires them. A burst can transmit minutes or hours late: ACK/REQ/RESPONSE payloads are useless by then, and channel messages arrive as if the user typed them long ago. While they wait they hold pool slots and dedup entries, starving fresh traffic.

Adds a `queued_at` timestamp to Packet (transient, not serialized) and drops any outbound packet older than `MAX_PACKET_QUEUE_AGE_MS`, default 60 seconds and overridable, chosen above the worst-case budget refill wait. Expiry runs each loop, freeing pool slots before head-of-line blocking, and again in `checkSend()` before a packet goes out.

A virtual `onPacketExpired()` hook lets Mesh clear the dropped packet from the dedup table, so a later copy still circulating via another path gets forwarded instead of suppressed. Drops are counted in `getNumExpired()` and flagged as ERR_EVENT_PKT_EXPIRED.

### DIFF
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -12,6 +12,15 @@ namespace mesh {
 #define MIN_TX_BUDGET_RESERVE_MS   100    // min budget (ms) required before allowing next TX
 #define MIN_TX_BUDGET_AIRTIME_DIV  2      // require at least 1/N of estimated airtime as budget before TX
 
+// Drop an outbound packet that has sat in the TX queue longer than this
+// (e.g. due to duty-cycle exhaustion). Delivering a packet minutes/hours
+// after it was queued is harmful: time-sensitive payloads (ACK/REQ/RESPONSE)
+// are useless, and channel messages show up as if the user typed them long
+// ago. Must sit above the worst-case budget-refill wait (~10-20s at 10%).
+#ifndef MAX_PACKET_QUEUE_AGE_MS
+  #define MAX_PACKET_QUEUE_AGE_MS  60000  // 60 seconds
+#endif
+
 #ifndef NOISE_FLOOR_CALIB_INTERVAL
   #define NOISE_FLOOR_CALIB_INTERVAL   2000     // 2 seconds
 #endif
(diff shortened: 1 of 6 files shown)


oltaco:
@jbrazio I'm interested in this and I assume #3393 is related? What do you see as the the cause for the repeated client messages?

jbrazio:
This edge case takes a couple of prerequisites to show up, but we've seen it happen a few times, because we run duty-cycle limited repeaters at 10% for regulatory reasons.

1. The repeater has to fill its duty cycle bucket and start queuing packets.
2. Overall network traffic has to stay high enough

When the blocked repeater starts releasing its queue, if the hash of a queued packet is no longer in the neighbiours repeater's tables.. it goes again out to the network, gets treated as a new packet, and duplicates.

#3393 is closer to #3395. I added the StaticPoolPacketManager guards while chasing the random crashes. They turned out not to be the cause, but tightening the class assertions doesn't hurt.

ripplebiz:
I'm not really sure this is the best way to go about fixing what you describe.
What could be better is to hook into the onRecvPacket(), calculate the packet-hash then scan the outbound queue for that hash and drop if match(?)

jbrazio:
Having repeater R1 and R2, clients C1 and C2.

- Client C1 send menssage M1.
- Repeater R1 repeat the message, R2 queues the message due to duty-cycle.
- C2 received the message from R1.

10% dutty cycle means 6m/1hr, repeater R2 can be silent for 54m.
MAX_PACKET_HASHES is set to 160.

If within the silent period of R2 160+ messages transverse the network when R2 is ready to offload the it's queue back into the network, R1 no longer holds M1's hash on SimpleMeshTables.

R1 will see it as a new packet, and rebroadcast M1 the network again.

From C2 prespective he will receive M1 from R1 at the proper time and will receive M1 from R2 54m later thus duplicating M1.

PR #3395, opened 2026-09-10
[HIGH] Ensure command buffers stay NUL-terminated to prevent overflow

jbrazio:
The CLI loops in the examples size the command buffer with `strlen()` and keep appending to it. When the buffer fills, the old code wrote `'\r'` over the final byte, the one holding the NUL. From then on `strlen()` can walk past the buffer and the append loop indexes past the end.

Clamps `len` at the top of each pass and, when the buffer is full, places the end-of-line marker at `len-2` so the final byte stays NUL. Applied to companion_radio, simple_repeater, simple_room_server, simple_secure_chat and simple_sensor.

The corruption usually showed up as random reboots on TX, most visible on 433MHz boards. Serial lines left floating pick up RF noise, and that noise fills the command buffer. Most boards are affected; the Heltec v3 just shows it more often.

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -2187,6 +2187,13 @@ bool MyMesh::handleCommand(const char* command, uint32_t sender_timestamp, char*
 
 void MyMesh::checkCLIRescueCmd() {
   int len = strlen(cli_command);
+  // `cli_command` must stay NUL-terminated within its bounds. If it ever isn't,
+  // strlen() above can return >= sizeof(cli_command) and the loop below would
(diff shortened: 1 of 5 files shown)


h0lad:
Can confirm this issue and your bugfix. 

We worked and debugged on same problem in different worlds at the same time :-) I'll need a lot of Portuguese sun to heal the trauma I suffered from analyzing it... it took many days.
 
This bug crashes my HT-CT62 repeater every 24-36h.

RAK3172 is less affected but the implementation shares the same code.

My Analysis: https://github.com/meshcore-dev/MeshCore/issues/3397

PR #3402, opened 2026-09-13
Sensor improvements

ripplebiz:
Adds region_map support to the sensor example.

Adds new binary request types: REQ_TYPE_SUBSCRIBE, REQ_TYPE_UNSUBSCRIBE
These enable a permitted client node to subscribe to push telemetry, with the following guard rails:
* a fallback region _scope_ must be resolvable (either via the SUBSCRIBE request scope, or sensor node's default scope)
* subscription has a timeout specified (and an upper limit), so re-subscribes are necessary
* subscriber _must_ specify which telemetry channels/types to monitor and the _minimum deltas_ which will trigger a push (passed as an LPP-encoded byte array in the SUBSCRIBE request)

### DIFF
--- a/examples/simple_sensor/SensorMesh.cpp
+++ b/examples/simple_sensor/SensorMesh.cpp
@@ -81,7 +81,7 @@ static File openAppend(FILESYSTEM* _fs, const char* fname) {
 
 /* --------------------- Cayenne LPP helpers ----------------------------*/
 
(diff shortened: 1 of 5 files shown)


IoTThinks:
Wow, I like it.
I'm working with sensor example recently.

So we have "subscribe model" now?

PR #3410, opened 2026-09-14
Let companion radios use a bridge (ESP-NOW local lane), and give bridges their own prefs

l33tdawg:
### The problem

A companion radio cannot use a bridge, and it is not an oversight in the companion - it is a type conflict:

```
src/helpers/CommonCLI.h:25:7: error: redefinition of 'class NodePrefs'
```

The bridges take `NodePrefs*` (`BridgeBase.h`), and the `NodePrefs` they mean is the CLI one in `helpers/CommonCLI.h`. The companion has its own, unrelated `NodePrefs` in `examples/companion_radio/NodePrefs.h`. Both are `class NodePrefs : public ConfigSerializer`, so any translation unit that includes a bridge and the companion's prefs defines the class twice. That is why ESP-NOW today exists as a mesh radio (`variants/generic_espnow`) and as a repeater bridge (`heltec_v4_repeater_bridge_espnow`), but never as a companion: a host-connected node cannot be the cheap end of a 2.4 GHz link, even though it is exactly the node with a host behind it to use the bandwidth.

### The fix, in two parts

**1. Give the bridges their own settings type.** `BridgePrefs` holds the six fields the bridges and their examples use. Both `NodePrefs` classes inherit it, so every existing `_prefs.bridge_enabled`, `_prefs.bridge_secret`, ... call site keeps working with no edit, and `BridgeBase` no longer includes `CommonCLI.h` at all. Only the constructor parameter types change.

**2. Wire the bridge into the companion** the way the repeater does it: `logRx` and `logTx` are already virtual on `Dispatcher` and the companion simply never overrode them. The companion gains a bridge member, those two hooks, `bridge.loop()` and `bridge.isRunning()` in `hasPendingWork()` - the same shape as `examples/simple_repeater`.

`examples/companion_radio/NodePrefs.h` persists the bridge settings under new keys, so an existing config file is unaffected; anything unset keeps the `BridgePrefs` defaults.

### Outcome

A companion node keeps LoRa as its long-range transport and mirrors mesh packets onto ESP-NOW for nearby nodes - a fast local lane rather than a replacement radio. New environment:

```
pio run -e heltec_v4_companion_radio_usb_bridge_espnow
RAM:   [=         ]   6.1% (used 128860 bytes from 2097152)
Flash: [==        ]  16.2% (used 1063737 bytes from 6553600)
========================= [SUCCESS] ========================
```

### Regression checks

| environment | result |
| --- | --- |
| `heltec_v4_companion_radio_usb` (no bridge) | SUCCESS - 103,520 B RAM / 650,565 B flash |
| `RAK_4631_companion_radio_ble` (nRF52, no bridge) | SUCCESS |
| `RAK_4631_repeater` (uses the CLI NodePrefs) | SUCCESS |
| `heltec_v4_companion_radio_usb_bridge_espnow` (new) | SUCCESS |

`heltec_v4_repeater_bridge_espnow` also builds here now (2026-09-19): the `ESP32Async/ESPAsyncWebServer` download that failed while this PR was written succeeds, and the merged tree builds the existing ESP-NOW bridge repeater as well as the new companion one.

### Re-validated against current `dev` (2026-09-19)

The branch was cut from `dev` @ 34404309 and that base has moved 24 commits since (#3442 companion CLI commands, #3430, #3393, #3248, and the WIO Tracker variant). It still merges into `dev` @ 166f75c1 with no conflict - upstream's new hunks are in the companion constructor defaults and `handleCommand`, ours are the `logRx`/`logTx` hooks, the constructor member, `loop()` and `hasPendingWork()`. With #3403 and #3450 also merged on top (all three touch `MyMesh.cpp`/`MyMesh.h`), five environments build:

| environment | RAM | flash |
| --- | --- | --- |
| `heltec_v4_companion_radio_usb` | 105,024 B (5.0%) | 651,801 B (9.9%) |
| `heltec_v4_companion_radio_usb_bridge_espnow` | 130,364 B (6.2%) | 1,064,981 B (16.3%) |
| `heltec_v4_repeater` | 61,892 B (3.0%) | 1,172,297 B (17.9%) |
| `heltec_v4_repeater_bridge_espnow` | 63,556 B (3.0%) | 1,182,733 B (18.0%) |
| `RAK_4631_companion_radio_ble` | 151,448 B (64.3%) | 479,096 B (67.2%) |

### Not done

No hardware validation yet - I have no ESP32 boards reachable while writing this, and since 2026-09-19 our two deployment units are reachable but run the plain LoRa companion image. Flashing the bridge companion on one of them and measuring the 2.4 GHz lane is the next step on our side. It is compile-verified only. The bridge itself is unchanged apart from its prefs type, and the ESP-NOW bridge's shared-secret XOR is left exactly as it is (it is documented upstream as network isolation, not security; that is a separate conversation).

### CI

`pr-build-check.yml` had no environment that compiles `src/helpers/bridges` at all, so nothing in that gate would have caught a mistake in this PR - or would catch one later in a bridge or in the shared prefs. The matrix now includes `heltec_v4_companion_radio_usb_bridge_espnow` for that reason.

The environment name follows the existing `heltec_v4_repeater_bridge_espnow`. Note that, like that one, it does not end in `_companion_radio_usb`, so `build.sh get-companion-firmwares-to-build` will not pick it up for releases. That is existing behaviour for bridge variants and this change does not alter it - raising it here in case releasing a bridge build is wanted.


l33tdawg:
Build update (2026-09-19): the environment you could not compile here now builds cleanly from this branch. With network access restored on the build machine, the ESPAsyncWebServer download succeeded and:

```
pio run -e heltec_v4_repeater_bridge_espnow
RAM:   3.0% (used 63468 bytes from 2097152 bytes)
Flash: 18.0% (used 1181637 bytes from 6553600 bytes)
======================== [SUCCESS] Took 103.26 seconds ========================
```

So the only previously-unverified configuration in the PR description now has a local build result, and the existing ESP-NOW repeater bridge compiles unchanged after the BridgePrefs extraction. Still compile-only, no hardware test. If you would like, I can also add `heltec_v4_repeater_bridge_espnow` to the CI matrix alongside the new companion env — I left it out for now because it doubles the bridge CI cost and that is a maintainer call.

l33tdawg:
Hardware validation on two Heltec V4 companions (ESP32-S3), and one fix that the build needs to run at all.

**The companion as submitted cannot start its bridge.** `bridge.begin()` was only ever called from the repeater, so a companion build with `WITH_ESPNOW_BRIDGE` never brought ESP-NOW up and every packet `logRx`/`logTx` handed it was dropped; it also set no bridge prefs, so `bridge_secret` stayed empty and `xorCrypt()` would have taken a modulo over its length. I pushed the fix as `cad00e83` on the branch: the same defaults the repeater sets, `bridge.begin()` in `MyMesh::begin()`, and the hold-after-receive delay as a build flag (`BRIDGE_DELAY_MS`, default 500 ms as before). It builds clean for `heltec_v4_companion_radio_usb_bridge_espnow`, the plain companion, the ESP-NOW bridge repeater and the nRF52 companion.

**The lane works.** Two companion radios flashed with that build (app-only writes, identity and NVS untouched, both restored afterwards) carry real mesh traffic over 2.4 GHz: with the peer's LoRa path physically blocked (its profile moved to 915.000 MHz against 917.375 MHz) 16/25 frames were confirmed, with the receiver's LoRa counters flat and the ACKs returning the same way. Same units, same bench, commissioned LoRa profile for comparison: 25/25 in 1.437 s median.

**A mirror cannot be a fast lane, from your own code path.** `Dispatcher::loop()` calls `logTx` when `_radio->isSendComplete()` is true, so the ESP-NOW copy of every packet leaves after that packet's airtime is already spent — the bridge is additive by construction. Making it a lane is per-packet transport selection (skip the radio for a peer known to be on the bridge), a mirror *before* the transmit rather than after it, and an acknowledgement that travels the lane too: in the first run where only the data frames skipped the radio, the receiver transmitted its ACK on LoRa, was deaf on 2.4 GHz while doing so, and lost every second frame (13/25 confirmed, strictly alternating).

That work is #3454, on top of this one, with the counters (espnow tx/rx/injected/dropped, lane selections, injected-path latency) surfaced as companion stats sub-type 3 so the next measurement reads milliseconds rather than inferring them: 25/25 confirmed at 583 B/s against 89 B/s on the commission profile, with zero radio frames sent by either unit.

PR #3415, opened 2026-09-15
fix(config): allow digits in config keys

und3rX1337:
ConfigSerializer's key tokenizer (is_key_char) accepts letters and _ but not digits. Any preference key containing a digit, for example gr_1hop (grp_relay_first_hop_only), makes loadSerial return false at that key. On device this means the saved prefs no longer reload after a reboot, from that key onward.

Fix: allow 0-9 in is_key_char. Verified against the native test suite: NodePrefs.FemGainSettingsRoundTrip was failing and now passes, all other suites still green.


liamcottle:
Hi there!

Thanks for the PR!

Please rebase your local branch against the `dev` branch, force push and then change the target branch of the PR to `dev` as well.

All code changes should be opened against the dev branch :) thanks!

src/helpers/ConfigSerializer.cpp
```diff
@@ -24,7 +24,9 @@ static bool is_whitespace(char c) {
   return c == ' ' || c == '\t' || c == '\r' || c == '\n';
 }
 static bool is_key_char(char c) {
-  return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || c == '_';
+  // les chiffres sont admis : des cles comme gr_1hop en contiennent, sinon loadSerial
```
liamcottle: please remove french comments

liamcottle (changes requested):

src/helpers/ConfigSerializer.cpp
und3rX1337: Done, comments are in English now. Thanks!

liamcottle:
Thanks for the contribution!

PR #3425, opened 2026-09-16
Treat the GPS enable pin as a pin number, not a truth value 🤖🤖

ptr727:
**Draft.** Opened as a proposed fix to reference from an issue, not as a merge request.
It contains a judgement call about shared-rail power control that is the maintainers'
to make, described under "Deliberately not changed" below. Raised by a reviewer on
#3423 and kept separate because it changes pin handling rather than reporting.

## Problem

```cpp
static uint32_t gpsResetPin = 0;
...
if(PIN_GPS_EN){ gpsResetPin = PIN_GPS_EN; }
```

A truth test on a pin number, assigned into an unsigned variable. Two defects:

1. `PIN_GPS_EN` is `-1` on boards with no enable pin. `if(-1)` passes, so
   `gpsResetPin` becomes **4294967295**. `start_gps()` and `stop_gps()` then call
   `pinMode()` and `digitalWrite()` with that on every `gps on` / `gps off`.
2. `0` is a legitimate pin number that the same test **rejects**, so a board wiring
   the enable line to pin 0 would silently never store it.

Case 1 is harmless today only because the core discards it
(`cores/nRF5/wiring_digital.c:28-32`):

```c
void pinMode( uint32_t ulPin, uint32_t ulMode ) {
  if (ulPin >= PINS_COUNT) { return; }
```

Resting on that is fragile, and it leaves `start_gps()`/`stop_gps()` silently
ineffective on RAK4631 with nothing to indicate why. A visible symptom: with the GPS
reported `deactivated`, the receiver keeps transmitting and bytes keep accumulating in
the `Serial1` buffer.

## Change

- `gpsResetPin` becomes `int`, initialised to `-1` meaning "no enable pin".
- The macro is tested with `>= 0`, excluding `-1` and accepting `0`.
- Both toggles are guarded, so the no-pin case is explicit and logged.

## Behaviour is unchanged on every board in tree

| Board family | `PIN_GPS_EN` | Before | After |
|---|---|---|---|
| `rak4631` | `-1` | out-of-range write, dropped by the core | explicit skip, logged |
| `gat562_*` | `33` | stores 33 | stores 33 |
| `rak3401` | n/a | excluded by existing `#ifndef RAK_3401` | unchanged |
| hypothetical pin 0 | `0` | never stored (bug) | stored correctly |

`rak4631` is the only variant defining `PIN_GPS_EN=-1`, and no variant uses pin 0, so
that half is a latent fix.

## Deliberately not changed, and why this is a draft

The suggestion on #3423 was to "always set `gpsResetPin` to the tested `ioPin`, and
only override it when `PIN_GPS_EN` is present and not `-1`".

I think that would be a regression. On RAK4631 the pin that wakes the receiver is
**`WB_IO2` (pin 34), the shared 3V3_S WisBlock slot rail**, confirmed on hardware:

```
DEBUG: Serial GPS: 63 byte(s) on Serial1 with IO pin 34 (not yet parsed)
```

`stop_gps()` drives `gpsResetPin` LOW, and it is called at boot on stock firmware via
the `#ifndef FORCE_GPS_ALIVE` path. Storing `ioPin` there would therefore pull the slot
rail low on every boot, powering down the RAK12002 RTC and anything else in the slots.

So there are two separable questions:

- **Is the pin handling wrong?** Yes, and this PR fixes that without changing power
  behaviour.
- **Should a GPS toggle drive a shared rail at all?** That is a design decision about
  RAK power control, and it is yours rather than mine. If the answer is yes, this PR is
  the wrong shape and the fix needs to distinguish a GPS-private enable pin from a
  shared rail.

## Testing

Hardware, RAK4631 + RAK12002:

```
clock     ->  20:37 - 16/9/2026 UTC
gps       ->  on, active, no fix, 0 sats
gps off   ->  DEBUG: GPS: no enable pin configured, power state unchanged
gps       ->  on, deactivated, no fix, 0 sats
gps on    ->  DEBUG: GPS: no enable pin configured, power state unchanged
clock     ->  20:38 - 16/9/2026 UTC
```

The clock still runs across both toggles, confirming the slot rail is untouched.

Builds: `RAK_4631_repeater` and `GAT562_Mesh_Tracker_Pro_companion_radio_usb`, the two
`RAK_WISBLOCK_GPS` families that differ in this code.

**Not verified:** the pin-0 path, since no variant uses it; a board with a real
GPS-private enable pin, since I do not have one; and any non-nRF52 core that could
compile `RAK_WISBLOCK_GPS`.

## Note

Textually close to #3423, which edits the log line a few lines above. No overlapping
hunks, but whichever merges second will want a trivial rebase.


ptr727:
Tracked by #3426, which states the defect and the open design question this draft deliberately leaves to maintainers.

PR #3428, opened 2026-09-17
Add Meshnology W10 board support

tgies:
Add support for Meshnology W10 board, including display, PMU, RTC, GPS, and sensor support.

Written with reference to Meshtastic's implementation for the same board.

Radio is controlled through an MCP23017 IO expander and polled IRQs. RadioLib HAL for this variant adds a wrapper that polls DIO1. This necessitates adding virtual isIRQPending() and isRecvIRQPending() hooks so that this board can work without changing the path taken on boards that use the latched ISR flag. So, despite touching the RadioLib wrapper stuff, this should be no runtime change on unrelated boards.

RTC is overridden so that the PCF85063A is not auto-detected as the slightly different and incompatible PCF8563. Also adds some paranoid validation of RTC state to steer clear of a memory safety bug in RTCLib encountered during development.

Adds a bunch of new tests around this IO expander topology, peripherals, etc.


tgies:
Disclaimers:
- Meshnology gave me the hardware free of charge to develop this. I cold-emailed them offering to attempt the MeshCore port if they would provide hardware, to which they agreed. No contract or payment was involved, and I do not have any preexisting or ongoing relationship with the vendor.
- LLMs touched this (especially generating boilerplate test code). I am a qualified and experienced software engineer since the before-times and I have rigorously reviewed anything LLMs touched.
- Meshtastic's W10 support was consulted for information about hardware configuration and known hardware quirks. Care has been taken not to copy substantive implementation code from GPLv3 Meshtastic to MIT MeshCore.

PR #3434, opened 2026-09-17
Pass `TELEM_RAK12500_ADDRESS` to the RAK12500 I2C probe 🤖🤖

ptr727:
## The bug

`src/helpers/sensors/EnvironmentSensorManager.cpp` defines the RAK12500's I2C address inside an
`#ifndef`, which is the file's convention for "a variant may override this":

```cpp
#ifndef TELEM_RAK12500_ADDRESS
#define TELEM_RAK12500_ADDRESS   0x42     //RAK12500 Ublox GPS via i2c
#endif
```

The probe then never passes it:

```cpp
if (ublox_GNSS.begin(Wire) == true){
```

`SFE_UBLOX_GNSS::begin` is declared
`bool begin(TwoWire &wirePort = Wire, uint8_t deviceAddress = 0x42, ...)`, so the address actually
probed is the library's own default, not the macro.

`grep -rn TELEM_RAK12500_ADDRESS src/ variants/ platformio.ini examples/` returns only the two
lines of the definition itself. It is the one `TELEM_*_ADDRESS` in this file that nothing consumes
— every other one reaches its probe through the sensor table (`TELEM_AHTX_ADDRESS`,
`TELEM_BME280_ADDRESS`, `TELEM_SHT4X_ADDRESS`, …) or a driver constructor (`TELEM_INA219_ADDRESS`,
`TELEM_INA226_ADDRESS`).

## Why it is latent, and why it is still worth fixing

No variant in the tree overrides the macro, and the library's default is also `0x42`, so **nothing
diverges today and this changes no behaviour in-tree**. The defect is that the knob exists and does
nothing: a variant that overrode it — which the `#ifndef` invites — would silently probe one
address while every other reference to the macro named another, and the symptom would be a GPS
that is simply not detected.

## The fix

```cpp
if (ublox_GNSS.begin(Wire, TELEM_RAK12500_ADDRESS) == true){
```

This is the only `ublox_GNSS.begin` call site in the tree.

## Verification

- Signature confirmed against the pinned library
  (`SparkFun_u-blox_GNSS_Arduino_Library.h:715`), not assumed.
- `RAK_4631_repeater` builds. The `RAK_WISBLOCK_GPS` block is compiled into that env
  (`ENV_INCLUDE_GPS && RAK_BOARD && !RAK_WISMESH_TAG`), confirmed by the `ublox_GNSS` symbol and
  the SparkFun translation unit in the resulting ELF, so the changed line is actually built.
- **Not verified on hardware, and it cannot be here.** The RAK4631 available to me carries a CASIC
  receiver on the UART rather than a u-blox on I2C, so it takes the serial fallback and never
  reaches this line. A board with an actual RAK12500 would be needed to exercise it — and since
  the address is unchanged in-tree, the observable behaviour would be identical either way.


Blake-Latchford:
> A board with an actual RAK12500 would be needed to exercise it — and since
the address is unchanged in-tree, the observable behaviour would be identical either way.

If you don't have hardware to run it on, why make the change in the first place? Have you verified this is actually a problem in reality?

PR #3436, opened 2026-09-18
WIO-E5 Fix: Pin Radiolab for stm32 and replace ltoa

### LINKED ISSUE #3334: wio-e5-mini_repeater fails to build: unpinned ststm32 floated to 20.0.0, where 'ltoa' is no longer declared
## Summary

`platformio.ini` leaves `ststm32` unpinned. On 2026-08-31 it floated to
**20.0.0**, whose newlib no longer declares `ltoa` — a non-standard GNU
extension — and `wio-e5-mini_repeater` stops compiling with no code change
responsible:

```
src/helpers/TxtDataHelpers.cpp:105:5: error: 'ltoa' was not declared in
this scope; did you mean 'utoa'?
  105 |     ltoa(int_part, p, 10);
```

## This is the second platform to hit the same call

PR #2025 (native Linux/Ardulinux repeater) ran into the identical problem
and works around it in `TxtDataHelpers` — but guarded by
`ARDULINUX_PLATFORM`, so the workaround does not reach any other target.
STM32 on the 20.x toolchain now needs the same thing for a different
reason.

Two independent platforms hitting one non-standard extension suggests the
guard-per-platform approach will keep accruing branches. `ltoa` here is
only converting a non-negative `int_part` to decimal, which `snprintf` does
portably:

```c
p += snprintf(p, <remaining>, "%ld", (long)int_part);
```

That would let PR #2025 drop its guard rather than add one.

## Isolation

Not assumed from the commit it first appeared on — that commit touched only
docs and unrelated environments:

1. Reproduced locally at that commit.
2. Reverted **only** that commit's `platformio.ini` change and rebuilt —
   failed identically.
3. Pinned `ststm32@19.7.1` and rebuilt — succeeded in 49s.

## Workaround

Pinning `platformio/ststm32@19.7.1`, the last of the 19.x line and what CI
was already passing with, restores the build. Reported here rather than
kept downstream since it is upstream's file and upstream's target — happy
to open a PR for the `snprintf` change if that is the direction you'd
prefer.

Blake-Latchford:
Fixes #3334

The major rev from 19.x to 20.x breaks Radiolab, so pin it for now. This should fix the failing wio-e5-mini_repeater builds.

ltoa is removed in STM32duino, so it was replaced with snprintf. The p[] buffer is always 16 bytes. Copy/paste for why claude thinks this is fine below, but also if this is broken it wouldn't be new.

>exp2 in [23, 30] (int_part = mantissa << (exp2 - 23)): this is where int_part can reach its true max, up to 2,147,483,520 — 10 digits. That's what the snprintf bound of 11 (10 digits + NUL) is sized for. But in this branch frac_part is never assigned anything other than its initial 0, so only *p++ = '0' runs for the fraction — 1 char, not 7.

> exp2 in [0, 22] (int_part = mantissa >> (23 - exp2)): this is the only branch where frac_part can also be nonzero and enter the 7-digit loop. Here int_part < 2^23 by construction, so it's capped at 7 digits, never 10.


fdlamotte:
just ran into the issue ...

will dig into it, thanks for the fix ...

PR #3437, opened 2026-09-18
Add missing "reboot" command to the CLI text dispatcher

shleepong:
## What

Adds a `reboot` branch to `handleCommand()` (the text dispatcher reached via `CMD_RUN_CLI_COMMAND`, used by BLE/WiFi/USB companion clients).

## Why

`handleCommand()` had no `reboot` branch, so sending `reboot` over this path replied "Unknown command". `CMD_REBOOT` (a separate binary opcode) and `checkCLIRescueCmd()`'s own special-cased `reboot` handling both bypass `handleCommand()` entirely, so this gap was never noticed on those paths.

## How

Matches the existing convention elsewhere (`checkCLIRescueCmd()`): calls `board.reboot()` directly, no reply attempted since it doesn't return.

Compile-verified against current `dev` (`heltec_v4_companion_radio_ble`).


liamcottle:
Hey there! Thanks for the PR! I'm opting to merge in my PR https://github.com/meshcore-dev/MeshCore/pull/3442 instead, to avoid all the AI generated comments, and also add the shutdown/poweroff commands.

PR #3438, opened 2026-09-18
Harmonize "get wifi.enabled" reply to on/off

shleepong:
## What

Changes `get wifi.enabled` to reply `on`/`off` instead of a bare `1`/`0`.

## Why

Every other boolean `get` in this codebase replies `on`/`off`: `radio.rxgain` (`CommonRadioPrefs.cpp`) and `radio.fem.rxgain` (`HeltecV4Board.cpp`). `get wifi.enabled` was the one outlier.

## A question before merging

This changes an existing CLI reply format. If any client (app, script, `meshcore-cli`, etc.) already parses the raw `1`/`0` value, this would be a breaking change for it. Flagging this explicitly rather than assuming it's risk-free -- happy to adjust the approach if there's a compatibility concern I'm not aware of.

Compile-verified against current `dev` (`heltec_v4_companion_radio_ble`).


Blake-Latchford:
You should probably remove the comments.

Blake-Latchford (approved):

Blake-Latchford:
Oh, have you actually executed the code and verified it does what you expect?

Blake-Latchford (changes requested):
Describe testing completed on hardware.

liamcottle:
This would then make the `set wifi.enabled 0/1` out of sync with it's own get command, since the `get wifi.enabled` no longer replies with the same value that you just set.

The other `get` handlers, that reply with `on/off` also accept `on/off` in the `set` command, rather than `0/1`.

So if this was to be changed for the getter, it should probably be changed for the setter as well.

shleepong:
@liamcottle Good catch on the get/set desync. Pushed a commit that makes the setter accept on/off too (0/1 still works).

Found one more spot testing on real hardware right after: the setter's own confirmation reply still printed the raw 1/0. Fixed to reply on/off there too, so get, set-input, and set-reply are all consistent now.

Tested on real hardware (Wireless Tracker V2, BLE). Screenshot below.

image

Blake-Latchford (approved):
LGTM

PR #3455, opened 2026-09-19
Recover T-Echo I2C bus before Wire initialization

J-PNW:
A T-Echo soft reset can leave an I2C slave holding SDA low, causing the first Wire transfer during RTC initialization to hang before radio initialization. Recover the bus before `Wire.begin()` by releasing SDA, pulsing SCL while SDA remains low, and generating a STOP.

The recovery sequence uses open-drain outputs and bounded delays. It is limited to the LilyGO T-Echo board implementation.

Validation:

- Built the exact PR head `ccebd3b8ab24c2b98220b5d4482ed3332e0a1d80` from a clean checkout on September 19, 2026 with PlatformIO: `LilyGo_T-Echo_repeater` and `LilyGo_T-Echo_room_server` both succeeded.
- The same board patch also builds in local repeater and room-server configurations based on v1.16.0. Both local devices repeatedly returned to a responsive CLI after DFU, including the latest local room/repeater updates.
- The exact current-upstream binaries have not been flashed to hardware. Battery-and-USB-disconnected cold boots and explicit stuck-SCL behavior remain unverified. Keep this as a draft pending those checks and review.

The separate local PCF8563 read-caching investigation is not included in this PR.


IoTThinks:
Is it happening in current version?
Does the issue make T-Echo unable to boot?

BTW, you should base the PR on dev branch.

variants/lilygo_techo/TechoBoard.cpp
```diff
@@ -5,9 +5,39 @@
...
+// nRF52 TWIM driver then busy-waits forever on the first transfer and the
+// firmware looks dead before it ever reaches radio_init(). Clock the bus by
+// hand until SDA is released, then issue a STOP, exactly as the I2C spec's
+// bus-clear procedure describes. Cheap, and a no-op on a healthy bus.
+static void i2cBusRecover() {
+  pinMode(PIN_WIRE_SCL, OUTPUT_S0D1);   // open-drain with pull-up
+  pinMode(PIN_WIRE_SDA, INPUT_PULLUP);
+  digitalWrite(PIN_WIRE_SCL, HIGH);
+  delayMicroseconds(10);
+  for (int i = 0; i < 18 && digitalRead(PIN_WIRE_SDA) == LOW; i++) {
```
IoTThinks, line 19: Why is there a for loop here?

PR #3462, opened 2026-09-20
Add multiple battery low voltage readings to prevent early shutdown (especially during TX spikes)

P4tches:
During power spikes, especially during TX, battery voltage sags below AUTO_SHUTDOWN_MILLIVOLTS. To help avoid this problem, 3 consecutive readings are taken before shutdown is invoked.

Are additional readings going to affect battery life?
No, this change continues to read only every 8 seconds, leading to ~24s of low battery before shutoff.

Is this dangerous running the battery below the autoshutdown?
No, the floor is 3300mV on LilyGo/Thinknode and 3400mV on Heltec/RAK, both safely above the 3000mV cutoff of the Lithium cells.


Blake-Latchford (approved):
Tested on my Heltec V3 and a bench top power supply. Confirmed that a 'transient' drop over 8s is detected, but only a 24s low voltage condition shuts the board down.

@P4tches note that since I'm nobody on this project my approval doesn't actually mean anything in the process. Hopefully a rando confirming your code works means something.

srl295 (approved):
LGTM FWIW BBQ

PR #3468, opened 2026-09-21
Thinknode m4 - don't touch crystal and OTA name

weebl2000:
I once made a build for someone that had the M4, I saw it was merged into dev now, these two things differed from my branch.


weebl2000:
@fdlamotte

fdlamotte:
you're right ...
I had this in my git for some time ... thought it was better in the dev branch than lost for anyone ;)

Issue #3409, opened 2026-09-14
BUG: set agc.reset.interval command parser does not accept full input range


cwichura:
agc_reset_interval is a uint8, and stored divided by four.  As such, it should accept a value up to 255*4=1000 seconds.

The CLI parser stores `interval = atoi(string) / 4` at https://github.com/meshcore-dev/MeshCore/blob/0679dbeffc504d562d2f09eb072fdc223f8ffc2a/src/helpers/CommonCLI.cpp#L475.  It appears the compiler sees the interval is a unit8 and invokes a version of `atoi()` that already returns a unit8.  The stored value becomes `(input mod 256) / 4`, making 252 the largest input accepted.  At 256, it rolls over to 0, and so on.

Blake-Latchford:
@cwichura do you want to verify the change fixed this bug?

cwichura:
@Blake-Latchford I don't have the environment to compile and test your PR, myself.  But reviewing the PR, it looks like it addresses it correctly.  And it looks like Scott has already merged it.

PR #3469, opened 2026-09-21
Constrain `agc.reset.interval` and allow intervals > 255s

Blake-Latchford:
getAgcResetInt(uint8_t) implicitly truncated `agc.reset.interval` interval. The division to uint8_t happens inside that function. Underflow and overflow rolled over in ways surprising to a user.

Tested in Heltec v3.
| Sent | Baseline | Fixed |
|---|---|---|
| `set … 252` | 252 | 252 |
| `set … 256` | **0** | 256 |
| `set … 1000` | **232** | 1000 |
| `set … 1020` | **252** | 1020 |
| `set … 5000` | **136** | 1020 (clamped) |
| `set … -5` | **248** | 0 (clamped) |
| `set … 17` | 16 | 16 |


Blake-Latchford:
https://github.com/meshcore-dev/MeshCore/issues/3409

Issue #940, opened 2025-10-14
(Bug) room server - allow.read.only should not allow an empty "guest" password


n2aws:
If you enable "allow.read.only" on a room server right now,  you can set a blank guest password.

In that scenario, a user who doesn't type in a password (who should normally be read-only) is granted read/write access.

When allow.read.only is enabled,  it should force a non-blank guest password, thus enabling "blank password" users to be read-only, instead of read/write

axhoff:
Opened #2995 with a focused fix and authorization-matrix regression tests. A blank configured guest password now disables authenticated read/write access, so blank logins receive the existing guest/read-only role when `allow.read.only` is enabled and are rejected when it is disabled.

Issue #864, opened 2025-09-29
Guest got admin rights


DerWossi:
Repeater Version 1.9.0-b92d9bd (Build: 28-Sep-2025)allows guests to manage the system...

I set new passwords for both roles and could replicate the issue even after reboot.

73, Peter

liamcottle:
Howdy!

If you logged in to a repeater with an admin password, you are added to the access control list.
So, the next time you try to login in with the same companion (with a blank password) it will remember your access level.

In v1.29.0 of the mobile app, you can tap the "Access Control" button in the repeater admin extra tools section to manage the list of users that have access without a password.

You will find if you remove yourself from the ACL, trying to log in with a blank password will indeed log you in as a guest.

From firmware v1.9.0+, logging in with a blank password will check if you are in the ACL, if not, checks if you provided the admin password, if not, checks if you provided the guest password.

DerWossi:
Whow! That was fast ;-)

I'll update right now.

THX

PR #3475, opened 2026-09-22
Fix room, repeater, and sensor login so an empty guest password cannot grant access.

KB3CMT:
An empty guest password matched strcmp("", "") and granted room read-write. Blank ACL logins also skipped replay checks and did not refresh last_activity, and a later password login could demote an admin.

Empty guest passwords no longer match. Existing roles are never lowered. Blank ACL logins update last_activity and reject replayed timestamps. Room READ_ONLY clients cannot post. Room REQ treats an equal timestamp as a replay.

Tested: pio run -e Heltec_v3_room_server, Heltec_v3_repeater, and Heltec_v3_sensor on current dev.


agessaman:
The repeater half of this change removes behavior that was added on purpose. It appears to me the room read-write grant is a separate bug.

Scott added `guest.password` in `42f2c1ec` (10 Feb 2025). The repeater check was `memcmp` of `strlen(guest_password)`. An unset password has length 0, and a zero-length compare always matches, so any login became a guest. The same commit gave the room server an explicit `len > 0` guard and left that guard off the repeater. The stats handler comment in that commit is "guests can also access this now."

`c4cc3dd1` (24 Feb 2025) switched the repeater to `strcmp` and still added no empty-password guard. After that, only a blank login matches an unset guest password. That is the behavior this pull request rejects.

Liam described that result as expected on #864. After explaining that a blank login keeps an ACL role, he wrote that removing yourself from the ACL and logging in with a blank password "will indeed log you in as a guest," in the order ACL, then admin password, then guest password. #2437 documents the same rule: "Guest password match (or blank pw when guest pw is unset) -> GUEST."

The room case is different. #940 is that a blank guest password plus `allow.read.only` grants read-write instead of guest. Stock room firmware sets `ROOM_PASSWORD` to `hello`, so this shows up when that password is cleared. Repeaters have no `ROOM_PASSWORD`, so an unset guest password is the default and currently means open guest login.

The other room and sensor changes still look like ACL bugs: `setperm` read-only (role 1) can post because the check is only `== PERM_ACL_GUEST`, a later login overwrites a stored higher role, and a blank ACL re-login skips the replay check that the password path already has.

KB3CMT:
Thanks Adam. I read the commits and the note on #864. The repeater side is doing what it was built to do with no guest password set, a blank login is a guest. I'll put that check back the way it was.

What I would like to remain in this request:

Room - A cleared guest password currently matches as read-write, so allow.read.only cannot assign a guest. That is #940. The stock room password is hello, so this shows up after that password is cleared. A read-only role can still post, because the post check only looks for guest. A later login can replace a stored admin role with a lower one. A blank re-login for someone already on the ACL skips the replay check and leaves last_activity at 0, so sync stays stuck after a reboot. 

Sensor - Same blank re-login path, same replay check. 

Repeater - besides putting the empty guest password back: I also stopped a login from demoting a higher stored role, and a blank re-login now does the replay check and refreshes last_activity. Say if you want those two out as well and this request limited to the room and the sensor.

PR #3476, opened 2026-09-22
Start the room and sensor boot splash from UITask::begin so a late init still shows

KB3CMT:
Room and sensor firmware compared millis() to BOOT_SCREEN_MILLIS, so the splash was already over if UITask::begin() ran more than 4 seconds after power-on. Repeater firmware already records _started_at in begin().

Room and sensor now do the same. The splash lasts 4 seconds from when the display task starts.

Tested: pio run -e Heltec_v3_room_server and pio run -e Heltec_v3_sensor on current dev.


Blake-Latchford (approved):
LGTM

PR #3477, opened 2026-09-22
Report identity save failures, reject truncated adverts, and require exact CLI on/off tokens.

KB3CMT:
IdentityStore::save() returned true after a failed write, so a node could reboot believing its identity was stored. AdvertDataParser copied lat/lon, feature fields, and the name without checking the remaining length, which can write past the name buffer. CLI setters for repeat, allow.read.only, loop.detect, and bridge.enabled matched a prefix, so values such as "offline" or "onn" were accepted.

save() now returns the write result. The advert parser stops when a flagged field does not fit and caps the name. Those CLI commands require the exact token and otherwise return an error.

Tested: pio run -e Heltec_v3_repeater and pio run -e Heltec_v3_room_server on current dev (69f3022a).


src/helpers/CommonCLI.cpp
```diff
@@ -560,13 +574,13 @@ void CommonCLI::handleSetCmd(uint32_t sender_timestamp, char* command, char* rep
   } else if (memcmp(config, "loop.detect ", 12) == 0) {
     config += 12;
     uint8_t mode;
-    if (memcmp(config, "off", 3) == 0) {
+    if (strcmp(config, "off") == 0) {
       mode = LOOP_DETECT_OFF;
-    } else if (memcmp(config, "minimal", 7) == 0) {
```
Blake-Latchford, line 565: Did you intend to remove the null terminator checking here?

src/helpers/AdvertDataHelpers.cpp
```diff
@@ -39,19 +39,26 @@ bool AdvertDataParser::isValidName(const char *n) {
   AdvertDataParser::AdvertDataParser(const uint8_t app_data[], uint8_t app_data_len) {
```
Blake-Latchford, line 39: Does it make sense to get a partial advert? I assume the application would throw it away with an invalid checksum or something, right?

Blake-Latchford (changes requested):
I'm not a maintainer, so you are free to ignore me. I'm just a rando with time to review this.

I've had better luck with smaller changes. You made some changes that I don't think were strictly necessary, and have multiple functional changes in one PR. 

Best wishes.

src/helpers/CommonCLI.cpp
KB3CMT: Yes. The old memcmp(config, "off", 3) compared the three letters and stopped. It never looked at the byte after them, so offline and off matched the same as off. strcmp(config, "off") compares through the null, so only the exact token matches. Same for minimal, moderate, and strict. That is the same fix as repeat and bridge.enabled, where a prefix such as offline or onn used to be accepted.

src/helpers/AdvertDataHelpers.cpp
KB3CMT: A short advert can arrive with a valid signature. Mesh::onRecvPacket verifies the signature over the public key, the timestamp, and the app_data bytes that are actually in the packet. It does not check that the flag bits have their fields. A sender can sign a one-byte app_data with the lat/lon bit set, and verify() still succeeds.

onAdvertRecv already returns when !parser.isValid(). Before this change the constructor still copied the missing lat/lon or feature bytes past app_data, then left _valid false. The early return stops that read. The packet is still dropped. The signature is not what drops it.

KB3CMT:
Thanks for reading it. This pull request does do three separate things: IdentityStore::save() returns the write result, the advert parser stops when a flagged field does not fit, and the CLI on/off setters require the whole token. They came out of the same review pass, so I put them in one request. I can split them into three if that is easier to review. I have recently gotten involved over here so I apologize if I missed some guidance, I just dug in the way I would at work. Not trying to cause any issues, just give back after spending 20 years avoiding getting involved with any major projects.

Blake-Latchford:
> I apologize if I missed some guidance

Now I'm wondering what that might look like. CONTIRIBUTING.md says some stuff about small PRs. I did the same as you, coming from a professional context this all seems totally reasonable. In this context the volunteer(?) maintainers have mountains of AI slop getting thrown at them every day. 

Practically I would leave it as one PR for a while. If nobody looks at it, close the PR and re-open as smaller ones.

src/helpers/CommonCLI.cpp
Blake-Latchford: I'm a bit less convinced this is worth fixing. It removes the risk of funny words working unexpectedly, at the expense of a missed null terminator (programmer error) resulting in a buffer overrun. Also SECURITY.md specifically calls out not being interested in "Physical access attacks", which I think includes anything over serial.

Redundant disclaimer is redundant: My opinion is my own and should not be assumed to reflect the MeshCore project maintainers' opinions.

KB3CMT: Yeah, fair. The buffer already has a null before this line runs. On the mesh path we write one at data[len] after decrypt, and the serial code writes one as each character comes in, so strcmp stops there.

The old memcmp only looked at the first 7 letters, so minimalxyz would still set the mode. That's the bit I wanted to stop.

SECURITY.md reads to me like JTAG or pulling keys off the UART. This same set command also comes in over the mesh when an admin sends it.

Happy to pull the exact-token checks back out and leave the identity save result and the advert length checks if that keeps the PR smaller or break in to 3 seperate.

PR #3480, opened 2026-09-22
Apply radio gain settings consistently

johnterickson:
## Summary

This PR centralizes application of the three radio gain settings used by MeshCore firmware roles:

1. Internal radio boosted RX gain (`radio.rxgain`)
2. External front-end-module RX gain / LNA (`radio.fem.rxgain`)
3. External front-end-module software TX gain (`radio.fem.txgain`)

The existing repeater behavior is used as the baseline. Before this PR, the repeater initialized and applied these defaults:

| Setting | Existing repeater default |
|---|---:|
| Internal boosted RX gain on SX1262/SX1268 | `SX126X_RX_BOOSTED_GAIN` when defined; otherwise `1` (on) |
| Internal boosted RX gain on other radios | `0` in `NodePrefs` |
| External FEM RX/LNA | `1` (on) |
| External FEM software TX gain | `0` (off) |

The new `helpers/RadioSettings.h` preserves that policy and lets all node types use the same code to apply it.

## Motivation

The node roles did not all initialize and apply gain settings in the same way:

- Repeater, room-server, and companion roles already initialized and applied all three persisted settings.
- Sensor persisted all three settings, but did not initialize or apply `radio.rxgain` through its role startup path.
- KISS modem and secure/terminal chat relied on radio-driver initialization for internal RX gain and did not explicitly apply external FEM defaults.

This made firmware behavior depend on which role was compiled for otherwise identical hardware. The difference was particularly visible on Heltec V4 hardware: its SX1262 boosted-gain build setting was already applied by the driver, while its controllable external FEM RX/LNA was not enabled by KISS or terminal-chat firmware.

## Implementation

A small shared helper now owns the default policy and application order:

```cpp
constexpr uint8_t defaultRxBoostedGain() {
#if defined(USE_SX1262) || defined(USE_SX1268)
#ifdef SX126X_RX_BOOSTED_GAIN
  return SX126X_RX_BOOSTED_GAIN;
#else
  return 1;
#endif
#else
  return 0;
#endif
}
```

This intentionally mirrors the original repeater/room/companion default logic as closely as possible to keep the behavioral change and review surface small.

`applyRadioGainSettings()` applies the three values together:

```cpp
radio.setRxBoostedGainMode(rx_boosted_gain);
board.setLoRaFemLnaEnabled(fem_rx_gain);
board.setLoRaFemPaGainEnabled(fem_tx_gain);
```

Roles with persisted preferences continue to pass their stored values. Roles without persisted gain preferences use `applyDefaultRadioGainSettings()`, which supplies the existing repeater defaults: internal RX according to the SX1262/SX1268 logic, FEM RX on, and FEM TX preference off.

## Defaults by node type

The table distinguishes role-level defaults from board/driver side effects. "Same as old repeater" means the new role default is exactly the policy that repeater firmware used before this PR.

| Node type | Old role defaults/application | New role defaults/application | Same as old repeater default? | Effective change |
|---|---|---|---|---|
| Repeater | SX1262/8 RX used `SX126X_RX_BOOSTED_GAIN`, otherwise on; FEM RX on; FEM TX off. All persisted values were applied. | Same values and persistence, applied through the shared helper. | **Yes, exactly. This is the baseline.** | Refactor only; no intended behavior change. |
| Room server | Same defaults and application as repeater. | Same values and persistence, applied through the shared helper. | **Yes, before and after.** | Refactor only; no intended behavior change. |
| Companion radio | Same defaults and application as repeater. | Same values and persistence, applied through the shared helper. | **Yes, before and after.** | Refactor only; no intended behavior change. |
| Sensor | FEM RX on and FEM TX off were initialized and applied. `rxgain` existed in persisted `NodePrefs`, but the sensor role did not initialize it from the repeater policy or apply it at startup. | Internal RX, FEM RX, and FEM TX use the same defaults as the old repeater. Persisted values are applied together. | **Yes. The new sensor defaults now match the old repeater defaults.** | `radio.rxgain` is now effective for sensor firmware. On Heltec V4, the default remains boosted RX on. |
| Secure/terminal chat | No role-level gain policy. Internal RX depended on radio-driver initialization. External FEM RX/TX preferences were not applied. | Uses the old repeater defaults directly: SX1262/8 RX build setting or on fallback; FEM RX on; FEM TX preference off. | **Yes. The new defaults are copied from the old repeater defaults.** | Controllable external FEM RX/LNA is now enabled by default. |
| KISS modem | No role-level gain policy. Internal RX depended on radio-driver initialization. External FEM RX/TX preferences were not applied. | Uses the old repeater defaults directly: SX1262/8 RX build setting or on fallback; FEM RX on; FEM TX preference off. | **Yes. The new defaults are copied from the old repeater defaults.** | Controllable external FEM RX/LNA is now enabled by default. |

## Heltec V4 behavior

Heltec V4 is the motivating hardware example.

| Setting | Old repeater | Old KISS | New repeater | New KISS |
|---|---:|---:|---:|---:|
| SX1262 boosted RX | On | On via SX1262 driver initialization | On | On |
| External FEM RX/LNA | On | Off/not explicitly applied | On | On |
| FEM software TX-gain preference | Off | Not explicitly applied | Off | Off |
| Board TX-mode PA switching | Automatic | Automatic | Automatic | Automatic |

The FEM TX preference does not turn off the fixed Heltec V4 transmit path. `HeltecV4Board::onBeforeTransmit()` still switches the external FEM into TX mode for every transmission, and `onAfterTransmit()` returns it to RX mode. This PR does not change those board hooks.

## Persisted settings

For repeater, room-server, companion, and sensor roles, saved configuration still overrides defaults:

- `radio.rxgain`
- `radio.fem.rxgain`
- `radio.fem.txgain`

The defaults are only used for a new configuration. Existing persisted values continue to be loaded before the shared helper applies them.

The sensor role now implements the existing `CommonCLICallbacks::setRxBoostedGain()` hook, making changes to `radio.rxgain` effective rather than merely serializable.

## Non-SX1262/SX1268 scope

The helper intentionally preserves the original managed-role logic:

- `USE_SX1262` or `USE_SX1268`: use `SX126X_RX_BOOSTED_GAIN` when provided; otherwise default on.
- Other radio families: role-level default is off.

For managed roles, this matches their existing zero-initialized `NodePrefs` behavior outside SX1262/SX1268 builds.

For KISS and secure/terminal-chat roles, the helper now explicitly applies the non-SX default. A target that separately enables boosted gain during `radio_init()` through another radio-family-specific mechanism could therefore be reset to off by the role helper. This PR deliberately does not reinterpret other macros such as `RX_BOOSTED_GAIN`, in order to remain as close as possible to the original repeater default logic. Reviewers may prefer either:

- keeping this exact repeater-derived policy for strict cross-role consistency, or
- limiting explicit internal-RX application to SX1262/SX1268 while still applying FEM defaults on all boards.

This tradeoff is called out explicitly because it is the only area where applying a formerly implicit role policy can affect a non-SX radio target.

## Deliberately unchanged

This PR does **not** change:

- TX power values or clamping
- Frequency, bandwidth, spreading factor, or coding rate
- Board-specific PA switching
- RadioLib register patches
- FEM hardware detection
- KISS protocol framing or commands
- Any Linux broker behavior

Unsupported board controls remain harmless through the existing `MainBoard` default methods, which return `false` and perform no operation.

## Validation

- `native_kiss_modem`: **8/8 passed**
- Native test suite: **40/40 passed**
  - Includes config serializer tests: **9/9 passed**
- Heltec V4 role builds:
  - `heltec_v4_repeater`: passed
  - `heltec_v4_room_server`: passed
  - `heltec_v4_companion_radio_usb`: passed
  - `heltec_v4_sensor`: passed
  - `heltec_v4_terminal_chat`: passed
  - `heltec_v4_kiss_modem`: passed
- `Station_G2_kiss_modem`: passed
  - Verifies explicit `SX126X_RX_BOOSTED_GAIN=0` remains disabled.
- `t1000e_kiss_modem`: passed
  - Compile coverage for a non-SX1262/SX1268 radio target.

## Review focus

1. Is the old repeater policy the desired common default for all roles?
2. Should KISS and terminal chat explicitly apply internal RX gain on non-SX1262/SX1268 radios, or leave those radio families to target initialization?
3. Is enabling controllable FEM RX/LNA by default for KISS and terminal chat desirable on all supporting boards?
4. Are the existing repeater defaults of FEM RX on and FEM TX preference off still the correct common baseline?


Blake-Latchford:
Have you flashed this code to a device to see if it works?

johnterickson:
Yes on Heltec v4 as repeater and KISS. I've left it in draft for the moment as I am still seeing some weirdly poor RX (dozens of dB) on the KISS modem. I think this code is right and it has something to do with AGC as that also has different code paths on KISS vs repeater, but I wanted to have it fully E2E verified at least on the v4 before publishing.  Apologies if draft mode was not as quiet/non-notifying as I had assumed.

Blake-Latchford:
> Apologies if draft mode was not as quiet/non-notifying as I had assumed.

I'm a nobody trying to help, not a core maintainer. Skimming your description I flagged it as AI written, and wasn't going to bother reading your PR unless A) it wasn't already abandoned and B) you had flashed it to a device.

src/helpers/RadioSettings.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include <MeshCore.h>
+
+namespace mesh {
+
+constexpr uint8_t defaultRxBoostedGain() {
+#if defined(USE_SX1262) || defined(USE_SX1268)
```
Blake-Latchford, line 8: nit: I think the nesting here is confusing. I think it would be clearer rewritten as if/elseif/else.

Blake-Latchford (commented):
Disclaimer: I'm not a MeshCore dev, I'm just a rando trying to help. You are welcome to ignore my feedback.

I think you're doing two things in this PR: Changing configuration behavior, and centralizing the logic. I think the bug fix you're trying to fix is quite simple, and on its own would be quite likely to get merged.

I've seen a lot of duplicated logic in MeshCore. It isn't clear to me if it is a deliberate optimization, or an artifact of the evolution of the code base. I feel way less comfortable commenting on that part of this change.

src/helpers/RadioSettings.h
johnterickson: Funny you mention that -  I originally had it rewritten that way, but changed it to this so it was a clean/lateral code move. I'll clean up.

johnterickson:
[keeping this in draft / standby until #3496  goes in per the discussion over there]

PR #3496, opened 2026-09-24
KISS modem: AGC, FEM gain and RX boosted gain control

agessaman:
## Summary

Adds SetHardware sub-commands so a KISS host can discover and configure RF settings that were previously fixed at build time. Hosts probe a capability bitmask instead of relying on the firmware version, since boards on the same firmware expose different controls. `KISS_FIRMWARE_VERSION` is now 2. Hosts that don't use the new commands see no change. Settings are not persisted; the host re-applies them after a reconnect.

| Command | Request | Reply |
|---|---|---|
| GetCapabilities | `0x1B` | `0x9B`: uint32 LE bits. 0 AGC, 1 FEM RX gain, 2 FEM TX gain, 3 radio RX boosted gain |
| Set/GetAgcResetInterval | `0x1C` + uint16 LE seconds / `0x1D` | `0x9D`: effective seconds (0 = off, max 1020, 4 s steps) |
| Set/GetFemState | `0x1E` + apply mask + value mask / `0x1F` | `0x9F`: capability mask + value mask |
| Set/GetRxBoostedGain | `0x20` + enable / `0x21` | `0xA1`: state |

New error `0x08` (Unsupported). Set commands reply with the Get response. Full details are in `docs/kiss_modem_protocol.md`.

## Details

- **AGC:** the fixed 30 s scheduler moves from `main.cpp` into `KissModem::loop()` and becomes configurable. The default is now 32 s, so a host can set it back exactly: values are rounded down to 4 s steps, the same as MeshCore's `agc.reset.interval`.
- **FEM gain:** reached through each board's existing `get/set radio.fem.*` commands via `handleCommand()`, so no new board interface is needed. The result of a set is checked by reading the state back. A change that arrives during a transmission is applied after TxDone. FEM replies go out in request order and are never dropped.
- **RX boosted gain:** the KISS equivalent of `radio.rxgain`. Radio drivers declare support through `supportsRxBoostedGain()`. While transmitting it replies TxBusy. Otherwise any received packet is delivered first, since LR2021 drops to standby to apply the change. The reply carries the state read back from the radio.

## Fixes to existing code

- **KISS build:** `setTxBufferSize()` is now called only on HWCDC builds. This fixes the Station G3 and Heltec Tracker V2 KISS builds.
- **FEM boards:** `set radio.fem.*` crashed on a NULL prefs store. The writes are now guarded, matching T-Beam 1W.
- **LR2021:** `get radio.rxgain` always reported off. The driver now tracks the setting per band and rolls back if a write fails.
- **LR1110 / LR1121:** a failed gain write was still recorded as applied.
- **`lr11x0ResetAGC`:** a gain set at runtime was lost on AGC reset unless the build defined `RX_BOOSTED_GAIN`.

## Testing

- Native KISS tests: 43/43 pass (`pio test -e native_kiss_modem`).
- Builds pass for KISS on the FEM boards and on SX1262, SX1276, LR1110, LR2021 and STM32WLx boards. The regular firmware builds pass for every FEM board and for SX1268 and LR1121.
- Tested successfully on a Heltec V4.3 OLED (SX1262 + KCT8103L FEM) with the matching openHop branches, `feat/kiss-agc-fem-control` in [openhop_core](https://github.com/agessaman/openhop_core/tree/feat/kiss-agc-fem-control) and [openhop_repeater](https://github.com/agessaman/openhop_repeater/tree/feat/kiss-agc-fem-control). Capabilities are reported correctly (AGC, FEM RX gain and RX boosted gain; FEM TX gain unsupported). AGC interval, FEM RX gain and RX boosted gain can each be set, and the replies show the state read back from the board.
- Still to check on hardware: FEM TX gain on Station G3; Wio Tracker L1 1W TX gain (whether its pin powers only the PA); RX boosted gain on an LR11x0 board.


agessaman:
@ViezeVingertjes thoughts?

src/helpers/radiolib/LR11x0Reset.h
```diff
@@ -15,7 +15,6 @@ inline void lr11x0ResetAGC(LR11x0* radio, float freqMHz, bool rx_boost_gain) {
   // Re-calibrate for the actual operating frequency (band=4MHz matches RadioLib default).
   radio->calibrateImageRejection(freqMHz - 4.0f, freqMHz + 4.0f);
 
-#ifdef RX_BOOSTED_GAIN
+  // always reapply: the gain may have been changed at runtime even when no build default is set
   radio->setRxBoostedGainMode(rx_boost_gain);
```
ViezeVingertjes: The same is done in `src/helpers/radiolib/SX126xReset.h` but it has a slightly different name `SX126X_RX_BOOSTED_GAIN`, probably needs to be applied there too?

ViezeVingertjes:
> @ViezeVingertjes thoughts?

LGTM; one comment to keep it consistent and it should be good to go.
I heard setting AGC causing issue's on i think `LR2021` before, so that probably is solved with this also. ^^

agessaman:
Great. I updated that. I'm going to mark this as ready.

ViezeVingertjes (approved):

johnterickson:
two FYI 
1. KISS radio dB are wrong by a factor of 4x https://github.com/meshcore-dev/MeshCore/pull/3498
2. I also have work in flight to make this normalized across node types https://github.com/meshcore-dev/MeshCore/pull/3480

agessaman:
Thanks for flagging these!

On the SNR scaling: the x4 in RxMeta is intentional. It's the same quarter-dB fixed-point encoding used for path SNRs and in the companion protocol, and `kiss_modem_protocol.md` documents it ("value x4"). Hosts divide by 4 to get dB. #3498 doesn't change the KISS output. If Scott takes both PRs, I'm happy to rebase around it in whatever order they land.

On #3480: I'm pretty sure the weak KISS RX you saw may be the FEM LNA being off at boot. On `dev`, `LoRaFEMControl` defaults it off, and only the roles with prefs (repeater, companion, etc.) turn it on via `attachDynamicPrefs()`, which KISS never calls. With this PR, a host can turn it on with SetFemState (`0x1E`). Also, in case it's useful: `dev` removed the `MainBoard` FEM virtuals (`setLoRaFemLnaEnabled()` etc.) when FEM control moved into variant code. That's why this PR reaches FEM through the boards' `radio.fem.*` commands via `handleCommand()`. #3480 might need something similar if it moves to `dev`, but that's for you and Scott to decide.

johnterickson:
Thanks - super helpful notes @agessaman!

#3498 So there was one place that it was wrong, so I decided to make a helper function to make it so all node/clients get it right.  I've now realized that the one place it was wrong was in the code I was adding. 😄

 #3480 I like the direction you are going here - it helps keep the KISS minimal. I will wait for this to go in and then I'll rebase on top of `dev` and see if there's anything still valuable.

